PR #5, opened 2017-01-25
treedataset, sstb, .gitignore

alrojo:
Minor comments (in blue) are in the code.

**_More general comments_**

**train/val/test specific arguments**

One more concern would be that for a given split we might want a given type of reaction. e.g. for training we want subtrees of trees. The most convenient way to do this would be to parse a boolean "subtrees=True", but the current interface for Dataset.split does not allow train/val/test specific arguments.

**vocab**

We should consider allowing the user to not have '"\<unk\>"' and '"\<pad\>"' as optional argument.

**TreeDataset hardcoded pipeline**

The TreeDataset is a wrapper for a hardcoded pipeline to sstb, perhaps this could be done more elegantly using the field's pipeline.


jekbradbury:
@alrojo made some changes, which involved breaking and then fixing things; can you give a second look and confirm that I fixed everything I broke?

PR #28, opened 2017-04-14
fix bug in preprocess field

fehiepsi:
Fix bug when `six.text_type` converts a list into a string.


jekbradbury:
Thanks! I'm sorry I didn't catch that I'd broken the SPINN example; we should really add unit tests here but I haven't had time :(

PR #29, opened 2017-04-18
utf-8 encoding

donglixp:
For python 2.*, if the dataset contains special tokens, there are encoding errors to read the files.

A better way to fix this problem is to add an argument to indicate the encoding of the input file.


soumith:
thanks!

jihunchoi:
This causes an error in Python 3, since a `str` object does not have the `decode` method.
I'll make a PR resolving this problem soon, if nobody is working on this now.

PR #31, opened 2017-04-25
Fix Python 3 compatibility

jihunchoi:
This PR fixes the bug referred by https://github.com/pytorch/text/pull/29#issuecomment-296878933.


soumith:
thank you!

PR #33, opened 2017-05-17
fix non-UTF8 error

kobikun:
in python 2.7, there is exception for decoding word into utf8.
So I added the type checking and converting UTF8 codes.

### DIFF
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -57,7 +57,7 @@ def load_word_vectors(root, wv_type, dim):
             if wv_size is None:
                 wv_size = len(entries)
             try:
-                if isinstance(word, six.string_types):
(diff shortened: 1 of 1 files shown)


jekbradbury:
I think this is wrong on Py3? What you probably want is `six.bytestype`.

kobikun:
I tried `six.bytestype`. but it is not work.
`six.string_types` is working on py2 and py3.

jekbradbury:
Sorry, I meant `six.binary_type`. In Py3, the lines being read are `bytes` objects, so the `if` wouldn't be triggered for `str` or `six.string_types`.

kobikun:
@jekbradbury  thanks~
I fix into `binary_type` and commited.

PR #38, opened 2017-06-20
Minor cleanup

nelson-liu:
Noticed a few unused imports, etc while reading through the package code.

Also some cosmetic fixes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,4 @@
 #!/usr/bin/env python
-import os
-import shutil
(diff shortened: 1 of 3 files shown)


nelson-liu:
Also the SpaCy setup instructions at https://github.com/pytorch/text/blob/master/torchtext/data.py#L61 are outdated, perhaps it'd be better to just link to their docs instead of enumerating the commands to run?

PR #40, opened 2017-06-26
Parameterize pad_token

Romainpaulus:
Allow fields to use a different pad token than `<pad>`, especially useful if the field is already numericalized or non vocabulary-based.

### DIFF
--- a/torchtext/data.py
+++ b/torchtext/data.py
@@ -116,13 +116,13 @@ class Field(object):
             eos_token=None, fix_length=None, tensor_type=torch.LongTensor,
             preprocessing=None, postprocessing=None, lower=False,
             tokenize=(lambda s: s.split()), include_lengths=False,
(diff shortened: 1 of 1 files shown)


jekbradbury:
There are a couple other places that `"<pad>"` shows up (in `vocab.py` and `datasets/snli.py`)

Romainpaulus:
They show up as default parameters for `Vocab.__init__` and `ParsedTextField.__init__`, so they are not hardcoded. The only time `Vocab.__init__` is called in `data.py` on line 203, and it uses `self.pad_token` in `__init__` (via `specials`): https://github.com/pytorch/text/blob/master/torchtext/data.py#L203

I think it's OK to leave it in `snli.py` since it's already following a specific format.

jekbradbury:
OK, this looks good then.

PR #47, opened 2017-07-03
Add tests for simple Field methods

nelson-liu:
This PR adds tests for the relatively simple `Field` methods  `Field.pad` and `Field.preprocess` -- the others are a bit more involved (you'd need to create `Dataset`s to properly test them), so I think I want to go "from the ground up" and write the tests for `Dataset` / other things that this uses before coming back to finish it.

This PR also adds some input validation to `get_tokenizer` and moves the get_tokenizer function into `data.utils` (though its still accessible from `data.get_tokenizer`, so nothing should break)

In general, let me know if there are datatypes that these methods should accept but I do not test; it's a bit hard to infer at times what the intended input is.

### DIFF
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -50,7 +50,10 @@ source activate testenv
 # Install requirements via pip in our conda environment
 pip install -r requirements.txt
 
-# Install PyTorch if we are running tests
+# Install the following only if running tests
 if [[ "$SKIP_TESTS" != "true" ]]; then
+    # SpaCy English models
+    python -m spacy download en
+    # PyTorch
     conda install --yes pytorch torchvision -c soumith
 fi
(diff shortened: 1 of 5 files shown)


jekbradbury:
My approach is basically duck typing I think? So for something like `preprocessing=` I'm currently supporting both `data.Pipeline(<some lambda>)` (the built-in constructor) and generic callables/custom subclasses of `Pipeline` that do things to the whole sentence at once (e.g. what's done [here](https://github.com/pytorch/text/blob/master/torchtext/datasets/snli.py) with a plain lambda--because the purpose of `Pipeline` is to provide a callable that automatically maps over tokens, if you _don't_ want that behavior you can just provide a callable). It's kind of messy, but the goal to make every part of the process hackable for weird custom datasets.

Issue #48, opened 2017-07-06
How to use pytorch text in the projects


mambuDL:
Hi, 

I've very limited python knowledge. I couldn't find how to integrate pytorch-text to my projects. When I type `from dataloaders.text.torchtext import data` to python reply after `import torch` I got the following error:

```
>>> from dataloaders.text.torchtext import data
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named dataloaders.text.torchtext

```

Then I decided, my system don't have the corrresponding package and I tried to install it like torch-vision package. (`pip install torchvision`) Unfortunatelly i couldn't make it since there isn't such a package. So how could I use this package ?

nelson-liu:
It's not on PyPI yet, so you can't pip install. With your terminal, cd to the location of the torchtext code on your disk (download or clone the repo first), then run: python setup.py install

mambuDL:
Thanks for the reply. It solved my problem. Unfortunatelly now I have another one. I tried to use the translator.py file inside the tests folder. To do I downloaded the same dataset and put a folder. Inside the code I only change the path at line-25 so that it points to the data where I've downloaded it. I got the following error: 

```
$ python translation.py 

    Warning: no model found for 'de'

    Only loading the 'de' tokenizer.


    Warning: no model found for 'en'

    Only loading the 'en' tokenizer.

Traceback (most recent call last):
  File "translation.py", line 27, in <module>
    fields=(DE, EN))
  File "build/bdist.linux-x86_64/egg/torchtext/data/dataset.py", line 56, in splits
  File "build/bdist.linux-x86_64/egg/torchtext/datasets/translation.py", line 35, in __init__
    EN.build_vocab(train.trg, max_size=50000)
  File "build/bdist.linux-x86_64/egg/torchtext/data/example.py", line 44, in fromlist
  File "build/bdist.linux-x86_64/egg/torchtext/data/field.py", line 83, in preprocess
  File "translation.py", line 14, in tokenize_de
    return [tok.text for tok in spacy_de.tokenizer(url.sub('@URL@', text))]
TypeError: Argument 'string' has incorrect type (expected unicode, got str)

```

nelson-liu:
[...]

PR #49, opened 2017-07-06
update preprocess to always convert to unicode

jekbradbury:
Mentioned in #48

### DIFF
--- a/torchtext/data/field.py
+++ b/torchtext/data/field.py
@@ -79,7 +79,7 @@ class Field(object):
 
     def preprocess(self, x):
         """Load a single example using this field, tokenizing if necessary.
-        
+
         If the input is a Python 2 `str`, it will be converted to Unicode
         first. If `sequential=True`, it will be tokenized. Then the input
         will be optionally lowercased and passed to the user-provided
--- a/torchtext/data/field.py
+++ b/torchtext/data/field.py
@@ -78,10 +78,15 @@ class Field(object):
         self.pad_token = pad_token if self.sequential else None
 
     def preprocess(self, x):
-        """Load a single example using this field, tokenizing if necessary."""
+        """Load a single example using this field, tokenizing if necessary.
+        
+        If the input is a Python 2 `str`, it will be converted to Unicode
+        first. If `sequential=True`, it will be tokenized. Then the input
+        will be optionally lowercased and passed to the user-provided
+        `preprocessing` Pipeline."""
         if self.sequential and isinstance(x, six.string_types):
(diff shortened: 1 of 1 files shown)


torchtext/data/field.py
```diff
@@ -81,10 +81,9 @@ def preprocess(self, x):
         """Load a single example using this field, tokenizing if necessary."""
         if self.sequential and isinstance(x, six.string_types):
             x = self.tokenize(x)
+        x = Pipeline(six.text_type)(x)
```
nelson-liu: shouldn't this before the tokenize call?

jekbradbury: Hmm. It wasn't before for the `lower=True` case but you're right that that probably makes more sense. Is there any reason a Py2 user would provide a tokenization function that only works on bytes? Maybe BPE? I guess in that case they should just wrap it in their own encode/decode step and have the torchtext public interface always be Unicode.

nelson-liu: > have the torchtext public interface always be Unicode.

Not sure if i understood what you mean, doesn't this change make the torchtext public interface agnostic to `unicode` or `str` (since we just convert to `unicode` under the hood)?

jekbradbury: yeah, I guess I just meant the interface for tokenizers.

torchtext/data/field.py
```diff
@@ -78,10 +78,15 @@ def __init__(
         self.pad_token = pad_token if self.sequential else None
 
     def preprocess(self, x):
-        """Load a single example using this field, tokenizing if necessary."""
+        """Load a single example using this field, tokenizing if necessary.
+        
+        If the input is a Python 2 `str`, it will be converted to Unicode
```
nelson-liu: maybe clearer to put `unicode` instead of Unicode, to make it explicit that we're referring to the [the python 2 unicode type?](https://docs.python.org/2/library/functions.html#unicode)

torchtext/data/field.py
```diff
@@ -78,13 +78,17 @@ def __init__(
...
+        first. If `sequential=True`, it will be tokenized. Then the input
+        will be optionally lowercased and passed to the user-provided
+        `preprocessing` Pipeline."""
         if self.sequential and isinstance(x, six.string_types):
+            x = Pipeline(six.text_type)(x)
```
nelson-liu: maybe i'm being dense, but why not just put this outside the `if`?

jekbradbury: it's also failing CI, whoops...

PR #50, opened 2017-07-06
Test Field converts str to unicode for SpaCy

nelson-liu:
Before #49 , I was converting the test string to unicode, but I believe that we should now handle this internally.


nelson-liu:
oh whoops, i was converting to specifically test the tokenizer, instead of the preprocess function --- ignore this.

PR #51, opened 2017-07-07
support field is none

alrojo:
Field is None to conveniently support data that does not require any preprocessing/padding/numericalization.


jekbradbury:
Can you switch the `if`s around so the conditions are `if field is None: [...] else: [...]` rather than `if field is not None...`?

PR #63, opened 2017-07-12
Savable iterator

Romainpaulus:
Implements `state_dict` and `load_state_dict` methods on `Iterator` instances in order to load/save checkpoints that maintain the shuffling order, including with `BucketIterator`. This required a more deterministic way to keep track of RNG states and reproduce them with the `RandomShuffler` class.

### DIFF
--- a/torchtext/data/iterator.py
+++ b/torchtext/data/iterator.py
@@ -24,9 +24,11 @@ class RandomShuffler(object):
         self.random_state = random.getstate()
         random.setstate(old_state)
 
-    def shuffle(self, data):
+    def __call__(self, data):
+        """Shuffle and return a new list."""
         with self.use_internal_state():
-            random.shuffle(data)
+            list_data = list(data)
+            return random.sample(list_data, len(list_data))
 
 class Iterator(object):
     """Defines an iterator that loads batches of data from a Dataset.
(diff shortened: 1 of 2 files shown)


torchtext/data/iterator.py
```diff
@@ -1,11 +1,34 @@
...
+        old_state = random.getstate()
+        random.setstate(self.random_state)
+        yield
+        self.random_state = random.getstate()
+        random.setstate(old_state)
+
+    def __call__(self, data):
+        """Shuffle and return a new list."""
+        with self.use_internal_state():
+            list_data = list(data)
```
jekbradbury: I think you can get away without this `list` call? Since `random.sample` doesn't modify its argument?

torchtext/data/iterator.py
```diff
@@ -81,7 +111,17 @@ def data(self):
...
         """Set up the batch generator for a new epoch."""
-        self.batches = batch(self.data(), self.batch_size, self.batch_size_fn)
+
+        if self._restored_from_state:
+            self.random_shuffler.random_state = self._random_state_this_epoch
+            self.batches = batch(self.data(), self.batch_size, self.batch_size_fn)
+            self._restored_from_state = False
+
+        else:
+            self._random_state_this_epoch = deepcopy(self.random_shuffler.random_state)
```
jekbradbury: I would avoid the deepcopy here by making `random_state` a `property` of the random shuffler whose "get" returns a copy of the state

torchtext/data/iterator.py
```diff
@@ -81,7 +111,17 @@ def data(self):
 
     def init_epoch(self):
         """Set up the batch generator for a new epoch."""
-        self.batches = batch(self.data(), self.batch_size, self.batch_size_fn)
+
+        if self._restored_from_state:
+            self.random_shuffler.random_state = self._random_state_this_epoch
+            self.batches = batch(self.data(), self.batch_size, self.batch_size_fn)
```
jekbradbury: I think you can move this below the if-else and avoid code duplication

torchtext/data/iterator.py
Romainpaulus: `len(data)` doesn't work if `data` is an iterator, which is the case in the `pool` function when calling `random_shuffler(batch(...))`. In that case, should I convert to list outside of the `RandomShuffler` class?

PR #90, opened 2017-08-09
Switch the order of tokenization and lower

Deepblue129:
Subword tokenization is dependant on if the sequence is lowercase or not.


jekbradbury:
Traditional text tokenizers might rely on the original tokenization of text, so I would probably make the custom field you're building a subclass of `Field` that overrides `preprocess` rather than switching the order for all users?

Deepblue129:
@jekbradbury You are right. My mistake.

One thing I am worried about when overriding `preprocess` is the amount of code I would need to copy. Could there be a clean way of defining the pipeline without restating the entire pipeline?

PR #92, opened 2017-08-10
[WIP] TREC dataset

bmccann:
```
import torch
from torchtext import data
from torchtext import datasets
inputs = data.Field(lower=True, include_lengths=True, batch_first=True)
answers = data.Field(sequential=False)
print('Generating train, dev, test splits')
train, test = datasets.TREC.splits(inputs, answers)
print(train[0].text, train[0].label)
```
yields

```
Generating train, dev, test splits
['how', 'did', 'serfdom', 'develop', 'in', 'and', 'then', 'leave', 'russia', '?'] DESC
```
When `fine_grained=True`:

```
Generating train, dev, test splits
['how', 'did', 'serfdom', 'develop', 'in', 'and', 'then', 'leave', 'russia', '?'] DESC:manner
```
Let me know if there is anything else you'd like to see. I'll leave the WIP until I hear 
back.

### DIFF
--- a/torchtext/datasets/trec.py
+++ b/torchtext/datasets/trec.py
@@ -4,6 +4,7 @@ from .. import data
 from six.moves import urllib
 from bs4 import UnicodeDammit
 
+import chardet
 
 class TREC(data.ZipDataset):
 
@@ -33,8 +34,8 @@ class TREC(data.ZipDataset):
         label_field.preprocessing = data.Pipeline(get_label_str)
 
         for line in open(os.path.expanduser(path), 'rb'):
-            line = UnicodeDammit(line).unicode_markup.split()
-            label, text = line[0], ' '.join(line[1:])
+            # there is one non-ASCII byte: sisterBADBYTEcity; replaced with space
+            label, _, text = line.replace(b'\xf0', b' ').decode().partition(' ')
             examples.append(data.Example.fromlist([text, label], fields))
 
         super(TREC, self).__init__(examples, fields, **kwargs)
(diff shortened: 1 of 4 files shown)


jekbradbury:
Does the dataset contain corrupted UTF-8? If not I’d rather just load it with Python rather than adding another dependency? (If it does then this is probably fine, although a built-in error strategy might also work)

torchtext/datasets/trec.py
```diff
@@ -0,0 +1,109 @@
...
+            fine_grained: Whether to use the fine-grained (50-class) version of TREC
+                or the coarse grained (6-class) version.
+            Remaining keyword arguments: Passed to the constructor of
+                data.Dataset.
+        """
+        fields = [('text', text_field), ('label', label_field)]
+        examples = []
+
+        def get_label_str(label):
+                return label.split(':')[0] if not fine_grained else label
```
jekbradbury: indentation is weird here?

torchtext/datasets/trec.py
```diff
@@ -0,0 +1,109 @@
...
+    @classmethod
+    def splits(cls, text_field, label_field, root='.',
+               train=train_filename, test=test_filename, **kwargs):
+        """Create dataset objects for splits of the TREC dataset.
+
+        Arguments:
+            text_field: The field that will be used for the sentence.
+            label_field: The field that will be used for label data.
+            root: The root directory that the dataset's zip archive will be
+                expanded into; therefore the directory in whose trees
```
jekbradbury: I think the bit about trees is specific to SST

torchtext/datasets/trec.py
```diff
@@ -0,0 +1,109 @@
+import os
+
+from .. import data
+from six.moves import urllib
+
+
+class TREC(data.ZipDataset):
```
jekbradbury: I don't think you're using any methods from `ZipDataset`? Would this work the same if it just inherited from `Dataset`?

PR #94, opened 2017-08-15
better word vectors + charngram vectors

bmccann:


torchtext/datasets/sentiment.py
```diff
@@ -89,17 +88,16 @@ def iters(cls, batch_size=32, device=0, root='.', wv_dir='.',
...
+                element one of the available pretrained vectors (see Vocab.load_vectors)
             Remaining keyword arguments: Passed to the splits method.
         """
         TEXT = data.Field()
         LABEL = data.Field(sequential=False)
 
         train, val, test = cls.splits(TEXT, LABEL, root=root, **kwargs)
 
-        TEXT.build_vocab(train, wv_dir=wv_dir, wv_type=wv_type, wv_dim=wv_dim)
+        TEXT.build_vocab(train, vectors=None)
```
jekbradbury: `vectors=vectors`

torchtext/vocab.py
```diff
@@ -139,33 +67,184 @@ def __init__(self, counter, max_size=None, min_freq=1, wv_dir=os.getcwd(),
...
+        """
+        if not isinstance(vectors, list):
+            vectors = [vectors]
+        vecs = []
+        tot_dim = 0
+        for v in vectors:
+            wv_type = v.split('.')[0]
+            wv_dim = int(v.split('.')[-1][:-1])
+            if wv_type == 'glove':
+                wv_name = '.'.join(v.split('.')[1:-1])
```
jekbradbury: seems a little roundabout? i think `partition` is relevant here and in line 102

torchtext/vocab.py
```diff
@@ -139,33 +67,184 @@ def __init__(self, counter, max_size=None, min_freq=1, wv_dir=os.getcwd(),
...
+            for i, v in enumerate(vectors):
+                end_dim = start_dim + vecs[i].dim
+                self.vectors[i][start_dim:end_dim] = vecs[i][token]
+                start_dim = end_dim
+            assert(start_dim == tot_dim)
+            start_dim = 0
+
+    def set_vectors(self, stoi, vectors, dim, unk_init='zero'):
+        self.vectors = torch.Tensor(len(self), dim)
+        self.vectors.normal_(0, 1) if unk_init == 'random' else self.vectors.zero_()
```
jekbradbury: would be nice to parameterize this but i'm not sure how -- pass a callback? a tuple?

torchtext/vocab.py
```diff
@@ -139,33 +67,184 @@ def __init__(self, counter, max_size=None, min_freq=1, wv_dir=os.getcwd(),
...
-            wv_index = wv_dict.get(token, None)
-            if wv_index is not None:
-                self.vectors[i] = wv_arr[wv_index]
+    def __getitem__(self, token):
+        if token in self.stoi:
+            return self.vectors[self.stoi[token]]
+        else:
+            vector = torch.Tensor(1, self.dim).zero_()
+            if self.unk_init == 'random':
+                vector.normal_(0, 1)
```
jekbradbury: what's the conceptual reason we need to have two places where unk initialization happens?

bmccann:
Woops, forgot to add my flake8 commits. Should be there now. Looking at other comments.

torchtext/vocab.py
bmccann: I'm not actually using this; just left it in case someone wanted to manually set the vectors. Not sure what you mean by parameterizing it though. Since I'm not using it, I'm very open to changing it around. Just let me know what you're thinking.

jekbradbury: parameterize meaning allow the user to supply a random init other than N~(0, 1). but that makes sense; leaving it in is probably a good call.

bmccann: Oh, I see. You were referring to how it is randomly initialized. Yeah, that would be worth a shot.

bmccann: Through a commit in there with zero_ as a default init function. Should make it possible to use any callback that returns a tensor of the same size as the input tensor.

jekbradbury:
I'm merging this, but we should be clear in 0.2 release notes about the two API changes this introduces: `wv_type` etc. are replaced by `vectors` (fairly straightforward) and the changes to word vector ordering and lookup, which won't affect most people but might be more subtle for anyone who relies on specifics of their current behavior.

PR #95, opened 2017-08-16
Batch should be sorted by decreasing size.

Deepblue129:
**Intended effect:**
- `rnn.pack_padded_sequence` requires that a minibatch be sorted by decreasing order.
- Curriculum learning requires that the batches are sorted in increasing order.

**Proposed Solution:**
Flip the sign of the `self.sort_key(x)` when creating the Batch.

### DIFF
--- a/torchtext/data/iterator.py
+++ b/torchtext/data/iterator.py
@@ -157,8 +157,10 @@ class Iterator(object):
                     continue
                 self.iterations += 1
                 self._iterations_this_epoch += 1
-                # NOTE: Find out more here for why we reverse: https://github.com/pytorch/text/pull/95
-                yield Batch(minibatch.reverse(), self.dataset, self.device,
+                # NOTE: `rnn.pack_padded_sequence` requires that a minibatch be sorted by decreasing order,
+                # which requires reversing relative to typical sort keys
+                minibatch.reverse()
+                yield Batch(minibatch, self.dataset, self.device,
                             self.train)
             if not self.repeat:
                 raise StopIteration
--- a/torchtext/data/iterator.py
+++ b/torchtext/data/iterator.py
@@ -157,8 +157,8 @@ class Iterator(object):
                     continue
                 self.iterations += 1
                 self._iterations_this_epoch += 1
-                minibatch_decreasing_size = sorted(minibatch, key=lambda x: -self.sort_key(x))
-                yield Batch(minibatch_decreasing_size, self.dataset, self.device,
+                # NOTE: Find out more here for why we reverse: https://github.com/pytorch/text/pull/95
+                yield Batch(minibatch.reverse(), self.dataset, self.device,
                             self.train)
             if not self.repeat:
                 raise StopIteration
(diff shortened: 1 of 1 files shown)


nelson-liu:
Not sure how i feel about this (and thus curriculum learning) being on by default.

jekbradbury:
I support the idea of intra-batch sorting being the opposite of inter-batch sorting, since the only reason for the former is to support packed sequences. It won’t turn curriculum learning on by default if you use a shuffled iterator like BucketIterator.

jekbradbury:
I believe that what should happen here is just a reverse() call though, since the sorts should always use the same key and just have opposite order.

nelson-liu:
> It won’t turn curriculum learning on by default if you use a shuffled iterator like BucketIterator.

good point, this was my main concern.

Deepblue129:
Thank you for your comments. Made the change you suggested. 

Added a comment because that `reverse` seems out of nowhere unless you have extra context.

jekbradbury:
You need to do `minibatch.reverse()` in a separate line, because `.reverse()` returns `None` (you could also use `reversed` but there's no reason not to do it in-place here).

Also, maybe the comment should just say something like "`pack_padded_sequence` requires that a minibatch be sorted by decreasing order, which requires reversing relative to typical sort keys" rather than asking the reader to copy and paste a GitHub URL?

torchtext/data/iterator.py
```diff
@@ -157,8 +157,8 @@ def __iter__(self):
                     continue
                 self.iterations += 1
                 self._iterations_this_epoch += 1
-                minibatch_decreasing_size = sorted(minibatch, key=lambda x: -self.sort_key(x))
-                yield Batch(minibatch_decreasing_size, self.dataset, self.device,
+                # NOTE: Find out more here for why we reverse: https://github.com/pytorch/text/pull/95
```
nelson-liu: why not just explain the rationale?

nelson-liu: whoops, just saw @jekbradbury 's comment on the PR --- +1 to that.

nelson-liu:
LGTM when CI passes.

Deepblue129:
@nelson-liu Did you know we both go to UW Comp Sci? And both worked @ Google?

jekbradbury:
Flake8 failed with `./torchtext/data/iterator.py:160:91: E501 line too long (107 > 90 characters)`

JianyuZhan:
Hi, @jekbradbury , I still stumbled on this issue even with this commit(a504b9). See https://github.com/OpenNMT/OpenNMT-py/issues/189

I think the `minibatch.reverse()` fix is not correct. In my case, it is using the `pool()` method to return batches for iterating over.  The `pool()` method would shuffle the `minibatch` examples, then later when  we `minibatch.reverse()` in the `__iter__` ,  it is still no satisfying the requirement of `rnn.pack_padded_sequence`.

I think this fix should be explicitly sort the `minibatch` by decreasing size before the `Batch` constructor.

I just tracked down the issue to the text code and found this issue.  Hope I don't misunderstand the code.

kyteague:
I don't think this is intuitive as it goes against what is expected when using a sort key in Python.

This also breaks a popular downstream library: https://github.com/IBM/pytorch-seq2seq
I made an issue about it there as well: https://github.com/IBM/pytorch-seq2seq/issues/77

PR #99, opened 2017-08-23
Fix loading pre-trained word vectors

### LINKED ISSUE #98: Loading pre-trained word vectors is broken
The logic to load the pre-trained word vectors seems to be broken.  I looked in `test/vocab.py` but didn't see any tests that covered correctness of loaded vectors.

Here's a snippet to reproduce the issue:

```python
from torchtext.vocab import Vocab 
from collections import Counter

the_vocab = Vocab(Counter(['the']), vectors='glove.6B.50d')

the_index = the_vocab.stoi['the']
print(the_index)
print(the_vocab.vectors.numpy()[the_index, :])
```

With the master branch this displays:

```
2
[  3.02532905e-17   1.40129846e-45   2.10194770e-44   0.00000000e+00
   1.00185045e-16   1.40129846e-45   1.40129846e-44   0.00000000e+00
   0.00000000e+00   0.00000000e+00   0.00000000e+00   0.00000000e+00
   0.00000000e+00   0.00000000e+00   0.00000000e+00   0.00000000e+00
   0.00000000e+00   0.00000000e+00   0.00000000e+00   0.00000000e+00
   4.84570079e-17   1.40129846e-45   1.54142831e-44   0.00000000e+00
   0.00000000e+00   0.00000000e+00   0.00000000e+00   0.00000000e+00
   0.00000000e+00  -2.00000000e+00   8.72027897e-07  -4.65774808e-10
   4.82658813e-26   1.40129846e-45   0.00000000e+00   0.00000000e+00
   0.00000000e+00   0.00000000e+00  -8.29709032e-27   3.23318939e+02
   9.88949980e-32   1.40129846e-45   1.58111748e-32   1.40129846e-45
  -8.15662628e+28  -9.19432299e+27   9.88943397e-32   1.40129846e-45
   7.55058708e-26   1.40129846e-45]
```

However, if I manually download and unpack the GloVe 50d file, this is the line with `the` token:

```
the 0.418 0.24968 -0.41242 0.1217 0.34527 -0.044457 -0.49688 -0.17862 -0.00066023 -0.6566 0.27843 -0.14767 -0.55677 0.14658 -0.0095095 0.011658 0.10204 -0.12792 -0.8443 -0.12181 -0.016801 -0.33279 -0.1552 -0.23131 -0.19181 -1.8823 -0.76746 0.099051 -0.42125 -0.19526 4.0071 -0.18594 -0.52287 -0.31681 0.00059213 0.0074449 0.17778 -0.15897 0.012041 -0.054223 -0.29871 -0.15749 -0.34758 -0.045637 -0.44251 0.18785 0.0027849 -0.18411 -0.11514 -0.78581
```

matt-peters:
Fixes #98 and adds a test for correctness.


nelson-liu:
Thanks for the PR and issue @matt-peters ! It'd be good to run this on CI -- could you rename the test file to test_vocab.py / put the test in a test case?

bmccann:
Confirmed!

matt-peters:
Addressed the PR comments. Now that the test is running in Travis, the build is failing for other reasons.  In the 2.X build, it's due to `os.makedirs` not supporting the `exist_ok` kwarg. The 3.X build was killed by Travis for an unknown reason, perhaps because it is running into a hard Travis limit like disk or memory usage or build time.

nelson-liu:
> In the 2.X build, it's due to os.makedirs not supporting the exist_ok kwarg. 

I've always made `os.makedirs` version invariant by using:
```
if not os.path.exists(path):
    os.makedirs(path)
```

> The 3.X build was killed by Travis for an unknown reason, perhaps because it is running into a hard Travis limit like disk or memory usage or build time.

I doubt it's disk or build time, as I've run longer builds involving downloads of larger datasets -- my bet would be on memory consumption when downloading or unzipping...

nelson-liu:
so after digging a bit deeper by looking through [verbose test logs from travis runs on my own fork](https://travis-ci.org/nelson-liu/text/builds/267454297), the issue seems to be [in memory consumption while loading the vectors from disk](https://github.com/pytorch/text/blob/a5049b9d70a699986ae839aca178c33376717cde/torchtext/vocab.py#L177). It's also quite surprising how slow that part of the code is (at least from looking at tqdm output on travis) -- is that due to the repeated decodes to utf-8 from binary / is that really necessary?

bmccann:
> In the 2.X build, it's due to os.makedirs not supporting the exist_ok kwarg.

Sorry about this. It should have been caught in the tests on the original PR, but I didn't realize that Travis didn't actually run test/vocab.py. I suppose this means that the other dataset test files were also not run outside of my local environment.

matt-peters:
The travis build is now passing.  I created and checked in a small test GloVe file used in the test to avoid downloading and processing the full file every time the build runs.  Longer term, it's better to host a smaller test file remotely to also exercise the download logic during the build.

jekbradbury:
Thanks all, this looks good (especially the minimal glove test). I believe the glove-loading logic is the fastest it can be in pure Python (utf-8 conversion has to happen somewhere), but I’d be happy to be proven wrong!

PR #104, opened 2017-09-03
Run linter on python 3 on CI

nelson-liu:
This PR adds Python 3 linting to the test matrix, and fixes an infraction (`unicode` type doesn't exist in python 3).


soumith:
thanks!

PR #107, opened 2017-09-04
Revtok integration: reversible fields and subwords

jekbradbury:
Integrates [revtok](https://github.com/jekbradbury/revtok), a simple fully-reversible tokenizer that optionally supports subwords. Adds the `ReversibleField.reverse` API to go from a padded batch of token IDs to a list of detokenized sentences. Speed is not great right now but I'm working on [an implementation in a fast language](https://github.com/jekbradbury/Revtok.jl) and will add an optional wrapper for that in the Python revtok when it's ready.

Added a first stab at a test (want to release revtok to pypi before I add it to requirements.txt and the test can actually run); will add more.

Usage example:
```python
>>> from torchtext import data, datasets
>>> TEXT = data.SubwordField()
>>> LABEL = data.Field(sequential=False)
>>> train, dev, test = datasets.SNLI.splits(TEXT, LABEL)
>>> LABEL.build_vocab(train)
>>> TEXT.build_vocab(train, max_size=2000) # currently takes about 8 minutes
>>> train[1].hypothesis
['\ue302 a ', ' person ', ' is ', ' at ', ' a ', ' diner ', ' ordering ', ' an ', ' omelette ', '. ']
>>> TEXT.segment(train, dev, test) # currently takes about 6 minutes
>>> train[1].hypothesis
['\ue302 a ', ' person ', ' is ', ' at ', ' a ', ' di', 'ner ', ', ', ' ', 'or', 'd', 'ering ', ' an ', ' o', 'm', 'el', 'e', 't', 'te ', '. ']
>>> train_iter, dev_iter, test_iter = data.BucketIterator.splits((train, dev, test), batch_size=16)
>>> b = next(iter(train))
>>> TEXT.reverse(b.premise.data)
['In a park a kid is chasing pigeons and two men are walking', 'A basketball player is hanging onto the rim while the ball is in the basket', 'An old man is speaking in a brown fedora and blue jacket.', ...]
```


test/test_subword.py
```diff
@@ -0,0 +1,24 @@
...
+        LABEL = data.Field(sequential=False)
+        RAW = data.Field(sequential=False, use_vocab=False)
+        raw, = TREC.splits(RAW, LABEL, train='TREC_10.label',
+                           validation=None, test=None)
+        cooked, = TREC.splits(TEXT, LABEL, train='TREC_10.label',
+                              validation=None, test=None)
+        LABEL.build_vocab(cooked)
+        TEXT.build_vocab(cooked, max_size=100)
+        TEXT.segment(cooked)
+        batch = next(iter(data.Iterator(cooked, 1)))
```
bmccann: Iterator needs shuffle=False

test/test_subword.py
```diff
@@ -0,0 +1,24 @@
...
+        self.assertEqual(TEXT.reverse(batch.text)[0], raw[0].text)
```
bmccann: breaking inside reverse -- batch.text is a variable and doesn't have t_(); can change to transpose that isn't in place and it works

torchtext/data/field.py
```diff
@@ -196,3 +199,54 @@ def numericalize(self, arr, device=None, train=True):
...
+        super(ReversibleField, self).__init__(**kwargs)
+
+    def reverse(self, batch):
+        try:
+            import revtok
+        except ImportError:
+            print("Please install revtok.")
+            raise
+        if not self.batch_first:
+            batch.t_()
```
bmccann: Using t() would allow this to work on variables and tensors, which seems to be desirable based on the TREC example you wrote for subwords

test/test_subword.py
```diff
@@ -0,0 +1,24 @@
...
+from torchtext.datasets import TREC
+
+
+class TestSubword(unittest.TestCase):
+    def test_subword(self):
+        TEXT = data.SubwordField()
+        LABEL = data.Field(sequential=False)
+        RAW = data.Field(sequential=False, use_vocab=False)
+        raw, = TREC.splits(RAW, LABEL, train='TREC_10.label',
+                           validation=None, test=None)
```
bmccann: There is no validation set available for TREC, but this would work: 

`raw, = TREC.splits(RAW, LABEL, train=TREC.test_filename, test=None)`

torchtext/data/field.py
bmccann: Decision was to not support variables in reverse: ac4f794 alters the test that broke on this line

test/test_subword.py
bmccann: ac4f794

test/test_subword.py
bmccann: ac4f794 -- decision was to preserve reverse as is and change the test

test/test_subword.py
```diff
@@ -0,0 +1,38 @@
...
+        cooked, = TREC.splits(TEXT, LABEL, train=TREC.test_filename, test=None)
+        LABEL.build_vocab(cooked)
+        TEXT.build_vocab(cooked, max_size=100)
+        TEXT.segment(cooked)
+        print(cooked[0].text)
+        batch = next(iter(data.Iterator(cooked, 1, shuffle=False)))
+        self.assertEqual(TEXT.reverse(batch.text.data)[0], raw[0].text)
+
+    def test_subword_snli(self):
+        TEXT = data.SubwordField()
```
jekbradbury: This is likely too slow for Travis

jekbradbury:
Notes from CI:
- [ ] declare encoding on field.py for Python 2
- [ ] fix subword TREC test to use CPU device
- [ ] fix Vocab tests that explicitly provide specials to also include `'<unk>'` (and note as a breaking change that also closes #121)

jekbradbury:
I think I messed up somewhere in a rebase, so the push build is failing on Python 2 due to a failure to round-trip the capitalization thing even though the PR build is succeeding and the branch claims to be up to date with master...

windweller:
I installed revtok by `pip install revtok`...maybe the version is not correct?
This error was thrown:

```
  File "text_classifier.py", line 298, in <module>
    TEXT = data.ReversibleField(sequential=True, tokenize=tokenizer, lower=True)
  File "/torchtext/data/field.py", line 324, in __init__
    super(ReversibleField, self).__init__(**kwargs)
  File "/torchtext/data/field.py", line 148, in __init__
    self.tokenize = get_tokenizer(tokenize)
  File "/torchtext/data/utils.py", line 32, in get_tokenizer
    import revtok
  File "/revtok/__init__.py", line 1, in <module>
    from vocab.vocab import Vocab, OutOfVocabularyException
ImportError: No module named vocab.vocab
```

Is the revtok on pip is not the correct version, how should we install it??

jekbradbury:
Currently you need to install from source github.com/jekbradbury/revtok

PR #114, opened 2017-09-08
Fix a bug of the set_vectors function

jihunchoi:
The current code has a small bug; thus it is semantically incorrect and also makes an error.
This PR fixes the issue.


nelson-liu:
@jihunchoi Thanks for the PR! i've been writing tests for the `Vocab` class at #102  --- do you mind if i cherry-pick your commit there and add a test for it?

jihunchoi:
Nice! Then I think this PR could be closed after yours is merged into master. :)

Issue #100, opened 2017-08-24
Loading Custom Word Embeddings


shuoyangd:
Hi All,

It seems to me that currently `Vocab.load_vectors` function only supports loading word embeddings from a pre-defined list of sources. I wonder if it is an option to add support for loading custom word embeddings from local or external sources?

Shuoyang

nelson-liu:
I think this is a good enhancement. @shuoyangd would it suffice to be able to load word embeddings from something like a Python dictionary of word to embedding?

nelson-liu:
i guess it'd also be a straightforward modification to just enable reading arbitrary vector files in w2v or glove format from disk...

nelson-liu:
This is fixed by #115, @jekbradbury close?

shuoyangd:
Fantastic. Please close if it's fixed.

jekbradbury:
This is fixed, but that PR breaks referring to specific GloVe types with strings, and I'd like to have that for backwards compatibility.

PR #115, opened 2017-09-09
Custom vectors and refactor

bmccann:
Depends on https://github.com/pytorch/text/pull/102.

Allows users to use custom embeddings (https://github.com/pytorch/text/issues/100).

Refactor to simplify Vectors.

### DIFF
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -86,11 +86,12 @@ class Vocab(object):
     def __len__(self):
         return len(self.itos)
 
-    def extend(self, v, sort=True):
+    def extend(self, v, sort=False):
         words = sorted(v.itos) if sort else v.itos
         for w in words:
-            self.itos.append(w)
-            self.stoi[w] = len(self.itos) - 1
+            if w not in self.stoi:
+                self.itos.append(w)
+                self.stoi[w] = len(self.itos) - 1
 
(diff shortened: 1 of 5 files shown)


torchtext/vocab.py
```diff
@@ -92,57 +85,28 @@ def __eq__(self, other):
...
-                       glove.6B.100d
-                       glove.6B.200d
-                       glove.6B.300d
-              unk_init (callback): by default, initialize out-of-vocabulary word vectors
-                  to zero vectors; can be any function that takes in a Tensor and
-                  returns a Tensor of the same size. Default: torch.Tensor.zero_
-              expand_vocab (bool): expand vocabulary to include all words for which
-                  the specified pretrained word vectors are available
+              vectors: one of or a list containing instantiations of the
+                  GloVe, CharNGram, or Vectors classes
```
jekbradbury: Not allowing these strings might be too breaking

bmccann: I don't think we should support both. The string wrapper was too brittle and forced everyone into GloVe format. It would also make it hard to deal with custom vectors defined by strings.

jekbradbury: All I mean is a dictionary that maps the existing strings to lambdas (thunks) that call the corresponding class constructors

PR #119, opened 2017-09-13
Finish tests for Field class

### LINKED ISSUE #78: Using a field representing real numbers with the iterator
I am trying to learn a regressor on text data and I use torchtext in all my other tasks but I see a problem in using it for this use case.

I define the field for targets as follows:
```
TARGETS = data.Field(
            sequential=False, tensor_type=torch.DoubleTensor, batch_first=True)
self.fields = [('targets', TARGETS), ('text', TEXT)]
self.train, self.val, self.test = data.TabularDataset.splits(
            path=self.path,
            train=self.train_suffix,
            validation=self.val_suffix,
            test=self.test_suffix,
            format=formatting,
            fields=self.fields)
TEXT.build_vocab(self.train)
```

I have a file that contains tab separate <values>\t<text>

When I make iterators out of it, 
```
train_iter, val_iter, test_iter = data.Iterator.splits(
                (self.train, self.val, self.test),
                batch_sizes=(self.batch_size, self.test_batch_size,
                             self.test_batch_size),
                sort_key=lambda x: len(x.text),
                shuffle=True)
print(next(iter(train_iter)))
```
it gives me an error when getting the next batch:

> AttributeError: 'Field' object has no attribute 'vocab'

I know this is because I didn't run .build_vocab for the TARGETS field. But why do I really need to do this? What if I just want to get real numbers and compute losses on them? 

Any workaround is appreciated. If I am doing something wrong, please let me know too.

### LINKED ISSUE #96: Does Field support numerical features?
Hi,

I am processing a text dataset. Beside the text feature which could be processed by ```Field``` in a subclass of ```Dataset```, there is some float numbers which I would like to use with the text. So how could I create a subclass of ```data.Dataset``` with appropriate ```Field```?

Thanks,

nelson-liu:
This PR is a continuation of #47 and finishes testing the various functions for the field class (`build_vocab` and `numericalize`).

I also wanted to fix #78 and add a test for it, but I'm unsure how to properly do this. One way would be to simply convert the numerical features to python ints dictated by the input `tensor_type` member (e.g. `int` for `LongTensor`, `float` for `FloatTensor`, etc). This feels pretty brittle, does anyone else have any other suggestions?

### DIFF
--- a/torchtext/data/field.py
+++ b/torchtext/data/field.py
@@ -79,7 +79,6 @@ class Field(object):
         torch.cuda.LongTensor: int
     }
 
-
     def __init__(
             self, sequential=True, use_vocab=True, init_token=None,
             eos_token=None, fix_length=None, tensor_type=torch.LongTensor,
--- a/torchtext/data/field.py
+++ b/torchtext/data/field.py
@@ -57,6 +57,29 @@ class Field(object):
         pad_token: The string token used as padding. Default: "<pad>".
     """
 
+    # Dictionary mapping PyTorch tensor types to the appropriate Python
+    # numeric type.
+    tensor_types = {
+        torch.FloatTensor: float,
+        torch.cuda.FloatTensor: float,
+        torch.DoubleTensor: float,
+        torch.cuda.DoubleTensor: float,
+        torch.HalfTensor: float,
+        torch.cuda.HalfTensor: float,
+
+        torch.ByteTensor: int,
+        torch.cuda.ByteTensor: int,
+        torch.CharTensor: int,
+        torch.cuda.CharTensor: int,
+        torch.ShortTensor: int,
+        torch.cuda.ShortTensor: int,
+        torch.IntTensor: int,
+        torch.cuda.IntTensor: int,
+        torch.LongTensor: int,
+        torch.cuda.LongTensor: int
+    }
+
+
     def __init__(
             self, sequential=True, use_vocab=True, init_token=None,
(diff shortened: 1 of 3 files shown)


jekbradbury:
I think that would be okay? I can't imagine any situation where someone has data in a string that looks like a float and is asking for a FloatTensor, but doesn't want the data converted from string to float. If it's already of the target datatype, this is a no-op anyway.

nelson-liu:
>If it's already of the target datatype, this is a no-op anyway.

Is this even possible with the standard data flow? I feel like everything
is converted to string.

jekbradbury:
If you write a custom Dataset that e.g. loads from HDF5, maybe?

torchtext/data/field.py
```diff
@@ -180,11 +213,23 @@ def numericalize(self, arr, device=None, train=True):
...
-            arr = self.postprocessing(arr, train)
+        else:
+            if self.tensor_type not in self.tensor_types:
+                raise ValueError(
+                    "Specified Field tensor_type {} can not be used with "
+                    "use_vocab=False because we do not know how to numericalize it. "
+                    "Please raise an issue at "
+                    "https://github.com/pytorch/text/issues".format(self.tensor_type))
+            numericalization_func = self.tensor_types[self.tensor_type]
+            # It doesn't make sense to explictly coerce to a numeric type if
```
nelson-liu: Not too sure how i feel about this part --- what's a use case for passing in sequential data that shouldn't use a vocabulary, and is this (don't do any explicit numericalization, but postprocess it) the proper way to handle it?

PR #120, opened 2017-09-13
Add aliases for pretrained Vectors for backwards compatibility

nelson-liu:
Fixes https://github.com/pytorch/text/issues/100, specifically https://github.com/pytorch/text/issues/100#issuecomment-329303657

Enables the `__init__` argument of `Vocab` to take one of the pretrained aliases we were using before, and it'll translate it into a `Vectors` object of the appropriate type.


jekbradbury:
Thanks Nelson!

PR #121, opened 2017-09-15
Adding is_label option in Field to remove unknown label

GregorySenay:
Hi all,

I add in the Field and Vocal class, the option "is_label", to be able to remove the default <unk> token when you are creating the LABEL dictionary.

This avoid to have an additional <unk> label in the class list.

### DIFF
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -60,7 +60,7 @@ class Vocab(object):
 
             counter.subtract({tok: counter[tok] for tok in ['<unk>'] + specials})
         else:
-            self.stoi = defaultdict(_default_unk_index) #list)
+            self.stoi = defaultdict(_default_unk_index)
 
             self.itos = []
 
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -60,7 +60,7 @@ class Vocab(object):
 
             counter.subtract({tok: counter[tok] for tok in ['<unk>'] + specials})
         else:
-            self.stoi = defaultdict(list)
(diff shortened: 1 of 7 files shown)


torchtext/vocab.py
```diff
@@ -40,21 +40,38 @@ def __init__(self, counter, max_size=None, min_freq=1, specials=['<pad>'],
...
+            self.stoi.update({tok: i for i, tok in
                           enumerate(['<unk>'] + specials)})
-        self.itos = ['<unk>'] + specials
+            self.itos = ['<unk>'] + specials
 
-        counter.subtract({tok: counter[tok] for tok in ['<unk>'] + specials})
+            counter.subtract({tok: counter[tok] for tok in ['<unk>'] + specials})
+        
+        else:
+            counter.update(specials)
```
nelson-liu: as it currently stands, this would add the `<pad>` token to your class list as well.

nelson-liu:
I like the idea behind this, it'd be nice to get a class list out of a Field. Not sure if this is the best way to do it, though --- will think a bit more about it.

torchtext/vocab.py
GregorySenay: Okay in that case, with the is_label option I can also remove the "specials"

jekbradbury:
Added a similar feature as a part of #107; let me know if that doesn't work for you.

GregorySenay:
Great! it works like this for me.
In case for others, you have to call the LABEL like this:
LABEL = data.Field(sequential=False, **unk_token=None**)

PR #126, opened 2017-09-21
iwslt

bmccann:


torchtext/datasets/translation.py
```diff
@@ -73,3 +75,70 @@ def splits(cls, exts, fields, root='.data',
             os.path.join(path, test), exts, fields, **kwargs)
         return tuple(d for d in (train_data, val_data, test_data)
                      if d is not None)
+
+
+class IWSLT(TranslationDataset, data.Dataset):
+    """Defines a dataset for the multi-modal IWSLT 2016 task"""
```
jekbradbury: Not multimodal

test/translation.py
```diff
@@ -18,6 +18,34 @@ def tokenize_en(text):
     return [tok.text for tok in spacy_en.tokenizer(url.sub('@URL@', text))]
 
 
+# Testing IWSLT
+DE = data.Field(tokenize=tokenize_de)
+EN = data.Field(tokenize=tokenize_en)
+
+train, val, test = datasets.IWSLT.splits(exts=('.de', '.en'), fields=(DE, EN))
```
jekbradbury: This will add a ton of time to tests if we made it a real TestCase; I'm not sure what the right thing to do here is but I might just leave it like this

nelson-liu: I was planning on doing the following for the datasets:

1. Write a temporary test file in the proper format
2. Read the temporary test file, and assert that the read data looks correct

Alternatively/In addition, if the time for processing the whole dataset is low (say less than 5 or 10 mins, we can cache the downloads or something) we can put it into the slow test cases.

torchtext/datasets/translation.py
```diff
@@ -73,3 +75,70 @@ def splits(cls, exts, fields, root='.data',
...
+
+    @classmethod
+    def splits(cls, exts, fields, root='.data',
+               train='train', val='IWSLT16.TED.tst2013',
+               test='IWSLT16.TED.tst2014', **kwargs):
+        """Create dataset objects for splits of the IWSLT dataset.
+
+        Arguments:
+
+            root: directory containing Multi30k data
```
nelson-liu: `s/Multi30k/IWSLT`?

PR #134, opened 2017-10-02
Update README.md to fix #133

### LINKED ISSUE #133: Update README.md with a working example on data.TabularDataset for json files
In README.md, the current example for data.TabularDataset with json file input looks like this:

sentiment = data.TabularDataset(
    path='data/sentiment/train.json', format='json',
    fields=[{'sentence_tokenized': ('text', data.Field(sequential=True)),
             'sentiment_gold': ('labels', data.Field(sequential=False))}])

where "fields" attribute is a list that contains a dictionary. This will lead to an AttributeError: 'list' object has no attribute 'items' because in torchtext/data/example.py we have:

    @classmethod
    def fromJSON(cls, data, **fields**):
        return cls.fromdict(json.loads(data), **fields**)

    @classmethod
    def fromdict(cls, data, **fields**):
        ex = cls()
        for key, vals in **fields.items()**:

where the field value (currently a list) is processed by function fromdict with a ".item()" operator.

So the example in the README.md file should be changed to:

sentiment = data.TabularDataset(
    path='data/sentiment/train.json', format='json',
    fields={'sentence_tokenized': ('text', data.Field(sequential=True)),
             'sentiment_gold': ('labels', data.Field(sequential=False))})

where "fields" attribute is fed with a dictionary directly.

czhang99:
Fixed #133

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@ pos = data.TabularDataset(
 
 sentiment = data.TabularDataset(
     path='data/sentiment/train.json', format='json',
-    fields=[{'sentence_tokenized': ('text', data.Field(sequential=True)),
-             'sentiment_gold': ('labels', data.Field(sequential=False))}])
+    fields={'sentence_tokenized': ('text', data.Field(sequential=True)),
+             'sentiment_gold': ('labels', data.Field(sequential=False))})
 ```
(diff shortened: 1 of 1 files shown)


jekbradbury:
Thanks for noticing this!

PR #138, opened 2017-10-09
Fix WT2, restructure splits, and add WMT14

jekbradbury:
Also adds nonoptional dependencies to setup.py so `pip install torchtext` will fetch them.

The only known possibly-breaking change (I'd call it a bugfix) is that `Dataset.splits` will now use `os.path.join` rather than concatenation to join provided paths.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,14 @@
-# NLP tools
-nltk
-spacy
-
 # Progress bars on iterators
 tqdm
 
 # Downloading data and other files
 requests
+
+# Optional NLP tools
+nltk
+spacy
+
+# Required for tests only:
 
 # Style-checking for PEP8
 flake8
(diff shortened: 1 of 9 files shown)


bmccann:
Nice; the added download functionality is great, especially for Google Drive. Is there any way to unify the splits for the other datasets even more?

jekbradbury:
Agreed, though the others are a lot more complicated and this one allows direct comparison with Google's GNMT/Transformer results

jekbradbury:
Also the confirm_token thing sits there for easily five minutes before actually downloading?? I don't know what's up with that, the logic is exactly the same as what Google uses (and it looks like they got theirs from StackOverflow anyway)

jekbradbury:
Anyone know what to do about the bogus flake8 error? I think I've seen it before but I don't remember what if anything makes it go away (other than `#NOQA`)

PR #150, opened 2017-10-18
bug fix: snli data loader

sivareddyg:
The current SNLI dataset class fails for the test case test/snli.py.  Even the SNLI example in [examples/snli](https://github.com/pytorch/examples/blob/master/snli/train.py#L22) fails. 
This should fix it.

### DIFF
--- a/torchtext/datasets/snli.py
+++ b/torchtext/datasets/snli.py
@@ -1,5 +1,3 @@
-import os
-
 from .. import data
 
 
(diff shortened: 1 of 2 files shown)


jekbradbury:
Accidentally pushed to master a few minutes ago -- can you remove the last two commits?

jekbradbury:
Also make sure to break line 40 in `snli.py` to satisfy the linter.

sivareddyg:
@jekbradbury thanks! I was planning to do this today but since you already did, is there something else I have to do for this PR?

sivareddyg:
@jekbradbury all checks passed :)

PR #157, opened 2017-10-21
Sequence Labeling Dataset

sivareddyg:
Very handy for working with sequences tagged with labels. Once this request is accepted, I will send a pull request of a [sequence labeling task](https://github.com/sivareddyg/examples/tree/master/sequence_labeling) to [pytorch/examples](https://github.com/pytorch/examples)


jekbradbury:
Thanks so much, this is really useful!

I believe you don't need to provide a custom `splits` method; the standard `Dataset.splits` should work fine here. Also, we're using `root` as the name for the directory like `.data` and `path` as the name for the full dataset path, so your keyword args should be updated to agree with that.

sivareddyg:
@jekbradbury thanks! I have incorporated your suggestion. Travis CL tests pass.

jekbradbury:
What I mean is that you can delete the entire `splits` method because the parent class already has exactly the same implementation.

sivareddyg:
That makes sense. I removed it. Thanks!

sivareddyg:
@jekbradbury do you want me to do anything else for this to be merged? Feel free. The only difference with other datasets is I introduced load_default_dataset function. I can get rid of it but I thought this function is more transparent than loading default dataset using splits.

I plan to push sequence labeling task to pytorch/examples and would appreciate if this can be merged. Thanks!

jekbradbury:
Sorry, one last thing: I think the clearest approach would be to make `SequenceLabelingDataset` (or maybe call it `TaggingDataset`) an abstract class that doesn't have a "default" dataset, and then add a subclass `UniversalDependenciesPOS` that provides URLs/dirname attributes and a `splits` method with the default filenames that uses `super` to call `Dataset.splits`, along the lines of what we do for `TranslationDataset`s. I can make these changes and merge tomorrow if you want.

sivareddyg:
No worries, I can do this. Thanks!

sivareddyg:
I would prefer SequenceTagging or SequenceLabeling instead of just Tagging. What do you suggest?

jekbradbury:
I like`SequenceTaggingDataset`

PR #166, opened 2017-11-05
update wordvector parameters for language modeling and add tests

keon:
I found that `wv_*` parameters are no longer used and cause errors.
So I updated the params and added tests to verify.


jekbradbury:
Thanks, this looks right. Can you fix the lint though? (Flake8 doesn't like bare `except`)

PR #170, opened 2017-11-07
changed special token indexing

siddsach:
Changed specials token indexing in field's numericalize function to make sure that pad_token is 0, so that torchtext works with torch.rnn.utils.pack_padded_sequence

### DIFF
--- a/torchtext/data/field.py
+++ b/torchtext/data/field.py
@@ -240,7 +240,7 @@ class Field(RawField):
                     x = [x]
                 counter.update(x)
         specials = list(OrderedDict.fromkeys(
-            tok for tok in [self.unk_token, self.pad_token, self.init_token,
+            tok for tok in [self.pad_token, self.unk_token, self.init_token,
                             self.eos_token]
             if tok is not None))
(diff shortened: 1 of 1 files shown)


jekbradbury:
Doesn't `pack_padded_sequence` accept any padding token because it uses the provided lengths rather than a particular masking value? Or is the issue actually `pad_packed_sequence`?
I think ultimately this is probably the right choice, but it's pretty breaking.

siddsach:
Yeah you're right. I don't understand why it would break it though, because all I did was change the order of the special tokens, and I don't see any assumptions made elsewhere about the order of the special tokens in the vocab object.

jekbradbury:
I don't mean break other parts of torchtext; I mean break user code that assumes padding is 1 or unks are 0 (I'm not sure how much of that there is) and break saved models that use dictionaries/embeddings with padding=1 and unk=0 (there's a lot of those).

jekbradbury:
Will close this for now, though I agree that if we were to start over it would probably be a good idea to switch the padding and unk indices.

PR #179, opened 2017-11-16
Add Documentation

keon:


docs/source/conf.py
```diff
@@ -0,0 +1,243 @@
...
+project = 'torchtext'
+copyright = '2017, Torch Contributors'
+author = 'Torch Contributors'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = '0.2.0'
```
nelson-liu: perhaps it's better to just import the package and use `package.__version__`? e.g. https://github.com/pytorch/pytorch/blob/master/docs/source/conf.py#L78

keon: In order to do this, I made some changes on setup.py
https://github.com/keon/text/blob/5003ffb56017768c1dbafae81587321c5652729b/setup.py

and saved version inside `__version__` in `torchtext/__init__.py`

nelson-liu: whoops, didn't realize that `__version__` was not yet set --- thanks for handling that. i would have used something like https://github.com/kennethreitz/setup.py/blob/master/setup.py#L38 and https://github.com/kennethreitz/setup.py/blob/master/mypackage/__version__.py, but this looks like it should work too.

keon: how about this? https://github.com/keon/text/commit/7c68bb1b4268b498fe2add77cdc6be5d154c1729

EntilZha:
I was about to open a PR to add documentation when I saw this one (should have checked earlier). My guess is that some of the changes are duplicate, and some are not. When might this be merged so I can reconcile and make a PR for the non-duplicate changes?

My changes are currently at https://github.com/EntilZha/text with working RTD docs at https://torch-text.readthedocs.io/en/latest/. If you are interested, we can also figure out how to transfer that over and/or have it track the main repo instead of my fork. Thanks!

jekbradbury:
I'm not super familiar with Sphinx/RTD, so I can't really judge the differences/overlaps between your approaches, but it sounds like even if @EntilZha's docs are working he'd rather I merge this PR first, so I'll go ahead and do that.

Issue #124, opened 2017-09-18
Field containing list of characters for each word


sriniiyer:
Hi,

For every example, I want to store a (#words, max_char_size) tensor, containing characters in each word in the sequence. Is there a way to do this currently using Field ? I think right now, numericalize expect a single list and doesn't allow a list of lists.

Thanks!

nelson-liu:
Yeah, I think you're right in that you can't do this with the current Field. Supporting more than basic tokenization shouldn't be too difficult though, and is definitely a good feature to have.

sriniiyer:
cool thanks. Gonna hack something up in Batch and Field to do it for now.

kmkurn:
There is a `SubwordField` implemented that might be able to achieve this but I am not sure. I also need this and currently I just make my own custom field. Maybe @jekbradbury can help?

jekbradbury:
SubwordField is for subword-splitting tokenizers that still result in a flat list of tokens (at the subword rather than word level) so I think you'd have to make your own custom field for now.

kmkurn:
@jekbradbury I see. Would you be interested in a PR for such a field?

jekbradbury:
Sure!

PR #197, opened 2017-12-30
Implement a nested field

kmkurn:
This PR mostly addresses #124 by introducing a nested field: a field that can contain another field. Please see `NestedField` class docstring for more info and `TestNestedField` class in `test/test_field.py` for examples. A complete list of what's introduced in this PR:

* Implement a `NestedField` class. This class implements a custom `preprocess`, `pad`, `build_vocab`, and `numericalize` method. The `pad` method is especially hacky since it uses monkeypatching. Unfortunately I can't think of any other idea. Suggestions for a more elegant approach are welcomed.
* Add `Sphinx` and `sphinx_rtd_theme` to `requirements.txt` to allow building docs locally.
* Put installation instructions in README, including the optional requirements (which is not optional if you want to run tests). Without this, contributors need to see `build_tools/install.sh` to figure those out.
* Update .gitignore to exclude coverage files which are produced when generating coverage reports locally.

Please let me know what you think @jekbradbury.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@ docs/build
 
 # Download folder
 .data
+
+# Coverage reports
+.coverage
+htmlcov/
--- a/README.md
+++ b/README.md
@@ -8,11 +8,36 @@ This repository consists of:
 - [torchtext.data](#data) : Generic data loaders, abstractions, and iterators for text (including vocabulary and word vectors)
 - [torchtext.datasets](#datasets) : Pre-built loaders for common NLP datasets
 
-# Data
+## Installation
+
+Make sure you have Python 2.7 or 3.5+ and PyTorch 0.2.0 or newer. You can then install torchtext using pip:
+
+```
+pip install torchtext
+```
+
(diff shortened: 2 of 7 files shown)


jekbradbury:
This is really great. The problem of padding in a nested field is pretty complicated, and your solution is better than anything I would have come up with. I also really like the other improvements in the PR; thanks for taking those on too.

kmkurn:
No problem. Thanks for accepting this. I guess #124 can be closed now right?

Anyway I notice a possible bug in `NestedField`. I originally intended that users can nest as many fields as they wish. However, `pad` does not accommodate this; it assumes that the nesting field is not another `NestedField`. It is probably possible to modify pad to allow that, but right now I'm not sure how. So, the best way for now I think is to disallow `NestedField` as a nesting field by raising error if that happens. What do you think?

jekbradbury:
I agree; it’s probably not necessary to support arbitrary nesting unless someone has an immediate use case.

Issue #198, opened 2018-01-03
Informative __str__, __repr__ of Batch


mttk:
It might be useful to have a more verbose / formatted output of `str()` for a `Batch` object (when compared to using `vars()`).

To this end, I wrote a short update that produces this:
```
[torchtext.data.batch.Batch of size 16 from IMDB]
	[.text]:[torch.LongTensor of size 143x16]
	[.label]:[torch.LongTensor of size 16]
```
This deviates a bit from the standard output of `_tensor_str`, but is short & good enough. Open points where I'd like input:

1. Handling empty Batches (data == None) -- what's the str() output? ex. 'Empty torchtext.data.batch.Batch instance'
2. Display tensor summary (as in `_tensor_str`) forall fields? I find that this makes the output cluttered and the indentation gets messed up easily. 
3. Does this warrant a PR?

jekbradbury:
I like this; a PR would be great! I agree with your proposal for empty batches in (1) and also agree with using only the type descriptors rather than the full tensor summaries.

mttk:
Cool, I'm closing this, will get a PR up in a day & reference this.

PR #211, opened 2018-01-31
Informative __str__, __repr__ of Batch

mttk:
From #198: short addition for printing out batches. Prints out short descriptions of all the fields present in the batch, along with the dims of the tensors.

The _short_str method might have a better home someplace else. It's largely copied (but reduced) from the _tensor_str's _str.


jekbradbury:
This may need a couple changes after the merger of `torch.Tensor` and `torch.autograd.Variable`, but torchtext master is targeting PyTorch 0.3 for now so this should work.

mttk:
sure, I'll keep track of the changes and update accordingly.

Issue #210, opened 2018-01-26
TabularDataset not working with new lines in strings


pmandera:
`TabularDataset` breaks if there are new lines in strings.

This is a direct consequence of how the data is read:
https://github.com/pytorch/text/blob/c5ce46051508a919eb830deca4187c749727151b/torchtext/data/dataset.py#L164

Perhaps using `csv` from python standard library would be a solution?

keitakurita:
I'm willing to file a pull request resolving this issue.

I have a question regarding [this line](https://github.com/pytorch/text/blob/db2e26c490d7d70db3edabf8ddc301275845d635/torchtext/data/example.py#L39) thought.
I don't quite understand what this comment means, as well as all the subsequent encoding and decoding. This doesn't appear in the `fromTSV` method either, so I'm not convinced that I need to replicate this functionality. 
Could someone tell me whether I can drop this, and if not, why?

jekbradbury:
Closed in #214

PR #214, opened 2018-02-03
Fix TabularDataset failing for files with newlines in the text (#210)

keitakurita:
I've made the TabularDataset class handle csv and tsv files using the standard csv library in Python. This should resolve the bugs surrounding newlines within the text (and probably some more hidden ones that are covered by the standard library).

I've added a unit test, run flake8 and confirmed the tests pass in both Python 2.7 and 3.6. If there's anything else I should do, please let me know:)


jekbradbury:
Thanks!

PR #216, opened 2018-02-03
Initialize vocab with counter object

flipsidetalk:
I added the ability for the build_vocab method in Field to take counter objects as an argument to initialize vocabulary frequency counts with. Sometimes, datasets are too big to be loaded all at once into memory for building the vocabulary. With this simple change, you can build vocab counts on individual folds, and then aggregate them by adding them as arguments into build_vocab for training.


jekbradbury:
This is good, but it looks like you also switch the order of `pad_token` and `unk_token`. That could be added as an optional keyword argument, but changing the default breaks saved models and has significant downstream impact on e.g. OpenNMT.

erip:
I think this can be closed. cc @Nayef211

PR #223, opened 2018-02-05
Error with the encoding on python 3

pouyana:
Using `pytorch` with `fast.ai` and not correctly configured Linux system leads to this kind of errors:

The code used

```
torchtext.datasets.IMDB.splits(TEXT, IMDB_LABEL, '/tmp/')
```

The error that is shown:

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 796: 
ordinal not in range(128)
```
Adding the correct encoding in the open statement solve this problem.

More information here:

http://forums.fast.ai/t/imdb-test-data-is-different-from-what-is-shown-in-the-class/8138/13

### DIFF
--- a/torchtext/datasets/imdb.py
+++ b/torchtext/datasets/imdb.py
@@ -29,7 +29,7 @@ class IMDB(data.Dataset):
 
         for label in ['pos', 'neg']:
             for fname in glob.iglob(os.path.join(path, label, '*.txt')):
-                with open(fname, 'r') as f:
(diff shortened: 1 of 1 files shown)


notnami:
This is a good idea in general because many Windows files still default to non-unicode encodings. I know PyTorch officially doesn't support Windows yet, but if or when it ever does, it would help to have the encoding specified explicitly.

keon:
I've had the same issue a while back.
I'll merge this since this is a minor change.

Issue #228, opened 2018-02-09
Encoding error when loading WikiText2 dataset


keitakurita:
On my machine, executing test/language_modeling.py raises an encoding error. 
The LanguageModelingDataset should have the option to specify the encoding of the file being read. Since this is a trivial feature to implement, I will send a pull request to solve this problem.

PR #229, opened 2018-02-09
Fixes and features around language modeling

keitakurita:
- Dealt with #228 (Encoding issue for LanguageModelingDataset)
- Added Penn Treebank dataset (commonly used dataset, so should be worth having)
- Added tests for built-in datasets


keitakurita:
I'm not quite sure why codecov/patch raises an error...

jekbradbury:
I merged this, but just realized that the PTB and WT2 tests (part of the optional Travis jobs because they take some time to download) are failing on Python 2 -- probably an encoding/unicode issue?

keitakurita:
Oh, sorry about that (I thought I ran the tests locally for python2 but must have made a mistake somewhere).
I'll look into it and try to file a PR that fixes the bug. Thanks for the notice!

Issue #160, opened 2017-10-26
Strongly suggest the ability to split the dataset


windweller:
So current the `.splits()` defined in `Dataset` seems to only take in pre-split files. This is very helpful for pre-defined, well-established datasets like WikiText, IMDB, etc. However, for people in other fields, it could be beneficial if the `Dataset` can provide a split function where I dump the whole data in, and it splits nicely into train, valid, test?

(Besides that, I love this library :) saved me so much trouble! Might move away from Tensorflow just because of this...

it's unfortunate that there's a lack of documentation, but I was still able to figure everything out within a few hours and connect to a PyTorch model)

jekbradbury:
Not a bad idea! Will try to add this if I get some time.

alejandrodumas:
I also think this library is great! Trying to use a seq2seq model with the
english to french data from: [http://www.manythings.org/anki/](http://www.manythings.org/anki/)
I arrived to the same problem, as you only have 1 big file. I want to help, but don't now
how we could handle the signature of the method, or do you have another method in mind @jekbradbury ? Don't know if adding another parameter for the train/val/test proportion is the best thing, but having another method would be strange

windweller:
Haha :) Another question/concern: can PyTorch save cached vocabulary list and numericalized files somewhere so it doesn't reparse/re-numericalize every time it runs?

jekbradbury:
You can definitely save the vocabulary using torch.save; however, we currently numericalize on the fly in order to not fill up GPU memory so you can't save that.

oja:
This should be closed, right?

PR #237, opened 2018-02-22
Add a split constructor for dataset

mttk:
Motivated by #160 

**Changes:**
- Added a split constructor for `Dataset` which allows for stratified / normal train-test-validation splits.
- Moved `RandomShuffler` from `iterator.py` to `data.utils.py` to avoid circular imports as the splitting relies on the Shuffler
- Added helper functions for splitting to `dataset.py`, namely: 
  * `check_split_ratio`
  * `stratify`
  * `rationed_split`

This is **far from a complete PR** as the (urgent) TODOs are:
- Add splitting option to TabularDataset constructor
- Write detailed tests
- Maybe refactor a bit

I have built and ran the changes myself, and in the limited tests I've done, everything seems to work. I'm open to a discussion on what the best way to implement the spliting mechanism is, and you can consider this PR just an initiative to start further discussion.

When / if the changes are green-lit I'll swiftly finish up the TODOs.


jekbradbury:
This is a much-needed enhancement! Thanks for the PR. In terms of design I think it's worth trying (at least for a little while) to use the same `splits` class method name as we've used with the pre-built dataset splits to denote a constructor that splits the data it's receiving as a list into train/test/etc. But another alternative is to always create the full `Dataset` first and then add a `split` _instance_ method that would return a tuple of `Dataset`s.

mttk:
Could you clarify on using the same `splits` class method? Simply renaming to "splits" would not work, so I assume it's not that.

I actually like the idea of constructing a Dataset first and then splitting, as I would assume this sort of thing to be often used along with `TabularDataset`, which already creates an instance when loading the data. The usage would look like this:

`
train, test, valid = TabularDataset(path, format, fields).split(ratio, stratified, field, rnd)
`
Which looks _passable_ for a one-liner.

It would be good to do some sort of automatic type inference for `format` if possible (to make it an optional argument) via the extension of 'path'. I can make the default stratify field 'label' to also reduce the necessary arguments.

jekbradbury:
OK, I think this API works well! Travis (flake8) is complaining about a couple indentations; if you add a couple tests this should be good to merge.

mttk:
Ok, I've fixed flake8, added tests and squashed all the changes into a single commit.

I also added an option to create and write a mock dataset (given `num_examples` and `num_labels`) to `TorchtextTestCase`, since I needed a dataset that's a bit larger to run tests on. Everything seems to work fine.

Unrelated -- I'm running some experiments on SQuAD and can probably clean up the script enough to add the dataset to torchtext, I saw that this is planned so let me know if you are already on it or have some pointers.

BangLiu:
@mttk Is the SQuAD dataset available? Also tested for 2.0?

PR #245, opened 2018-03-11
Enable passing args to Vectors constructor.

sidharthms:
This enables setting vectors cache location and init strategy for unk words.
Also fixing broken calls to load_vectors.


jekbradbury:
Can you try to make Travis happy? The `load_fasttext_vectors` test is breaking because you `os.path.join` with `None`.

sidharthms:
Changed Vector params to None instead of default values. The default values are now being set in the body. It seemed better to do this instead of repeating these defaults in the callers of Vector.

Issue #242, opened 2018-03-05
`pool` method always shuffles and sorts the data?


shuoyangd:
In the last few lines of `iterator.py`:

```
for p in batch(data, batch_size * 100, batch_size_fn):
        p_batch = batch(sorted(p, key=key), batch_size, batch_size_fn)
        for b in random_shuffler(list(p_batch)):
            yield b
```

... which shuffles the batch and sorts examples within the batch. Now this code is used in `BucketIterator`, which means that shuffle and sorting is always conducted regardless of the value of `self.sort_within_batch` and `self.shuffle`. This does not look like an intuitive behavior especially during debugging when reproducible behavior is desired. Am I missing something?

I have a version to do what I find intuitive. Let me know if I should submit a PR for review.

jekbradbury:
Feel free to submit a PR.

gongliym:
I think the original version is better. 
The current version is as follow:
```
    for p in batch(data, batch_size * 100, batch_size_fn):
        p_batch = batch(sorted(p, key=key), batch_size, batch_size_fn) \
            if sort_within_batch \
            else batch(p, batch_size, batch_size_fn)
```
The problem is that if we set the parameter ~~ sort_within_batch=False~~(default), then the batch is not optimized for minimum padding.
From my understanding, this part of code should:
1. takes ~~100*batch_size~~ examples
2. sort within these examples by length
3. generate batch iterator

shuoyangd:
I agree with your understanding. However,

```
The problem is that if we set the parameter "sort_within_batch=False"(default), then the batch is not optimized for minimum padding.
```

I'm not sure why that is a problem when you set "sort_within_batch=False". From what I recall, this is a configuration that you should only use for debugging?

PR #258, opened 2018-03-22
fix pool method

shuoyangd:
such that it does not shuffle/sort when iterator does not (#242)


shuoyangd:
seems like flake8 is happy now.

Issue #271, opened 2018-04-09
Error in BucketIterator


alumae:
This used to work, but fails in 0.2.2. Using Python 3.6.

Code:

```
import torchtext.data as data
import torchtext.datasets as datasets
TEXT = data.Field(lower=True, batch_first=True)
LABEL = data.Field(sequential=False)
train_data, test_data = datasets.IMDB.splits(TEXT, LABEL)
TEXT.build_vocab(train_data, max_size=10000)
LABEL.build_vocab(train_data)
train_iter, test_iter = data.BucketIterator.splits((train_data, test_data), batch_size=32, device=0, repeat=False)
batch = next(iter(train_iter))
```

The last line causes:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-19-2c227e277efb> in <module>()
----> 1 batch = next(iter(train_iter))
      2 print(batch.text)
      3 print(batch.label)

/usr/local/lib/python3.6/dist-packages/torchtext/data/iterator.py in __iter__(self)
    133     def __iter__(self):
    134         while True:
--> 135             self.init_epoch()
    136             for idx, minibatch in enumerate(self.batches):
    137                 # fast-forward if loaded from state

/usr/local/lib/python3.6/dist-packages/torchtext/data/iterator.py in init_epoch(self)
    109             self._random_state_this_epoch = self.random_shuffler.random_state
    110 
--> 111         self.create_batches()
    112 
    113         if self._restored_from_state:

/usr/local/lib/python3.6/dist-packages/torchtext/data/iterator.py in create_batches(self)
    240                                 random_shuffler=self.random_shuffler,
    241                                 shuffle=self.shuffle,
--> 242                                 sort=self.sort_within_batch)
    243 
    244 

TypeError: pool() got an unexpected keyword argument 'sort'
```

shuoyangd:
Sorry that's my bad. I've worked with this change on my local copy for a while, but I must have messed up something while going back and forth.

jekbradbury:
Closed by #272. Thanks for catching this.

PR #272, opened 2018-04-09
[Urgent] Fix argument to pool call

mttk:
Call to `pool` method failed due to a wrong argument name in #258 .

This should have failed tests and shouldn't have been merged as it breaks **every** BucketIterator called with `sort=False`, as reported in #271.


soumith:
thanks for the urgent fix.

jekbradbury:
Thanks everyone! Just pushed a fixed version to PyPI.

PR #273, opened 2018-04-09
Update iterator.py

caolusg:


soumith:
merged #272 (same PR)

PR #279, opened 2018-04-11
Copy the sort key while splitting datasets

### LINKED ISSUE #278: dataset.sort_key not retained after dataset.split()
Hi I was trying out the new `split()` functionality to split a test set into test and validation set. When the `BucketIterator` is being set up from a newly split dataset the sorting fails because `dataset.sort_key` is None. Looking at the `split()`  function I see that a new instance of type `Dataset` is created in which `sort_key` is None by default:
```
def split():
...
if not stratified:
            train_data, test_data, val_data = rationed_split(self.examples, train_ratio,
                                                             test_ratio, val_ratio, rnd)
            return tuple(Dataset(d, self.fields)
                         for d in (train_data, val_data, test_data) if d)
```

I guess one way would be to explicitly copy the `sort_key` into the new instance. A more generic way could be to make a copy of the current instance and replace only the `examples` attribute but I'm not sure if that is really needed.

Thanks!

mttk:
Fixes  #278, while kind of ugly.

### DIFF
--- a/torchtext/data/dataset.py
+++ b/torchtext/data/dataset.py
@@ -109,7 +109,7 @@ class Dataset(torch.utils.data.Dataset):
             train_data, test_data, val_data = rationed_split(self.examples, train_ratio,
                                                              test_ratio, val_ratio, rnd)
             splits = tuple(Dataset(d, self.fields)
-                         for d in (train_data, val_data, test_data) if d)
+                           for d in (train_data, val_data, test_data) if d)
 
             # In case the parent sort key isn't none
             if self.sort_key:
@@ -132,14 +132,13 @@ class Dataset(torch.utils.data.Dataset):
                 val_data += group_val
 
             splits = tuple(Dataset(d, self.fields)
-                         for d in (train_data, val_data, test_data) if d)
+                           for d in (train_data, val_data, test_data) if d)
 
             # In case the parent sort key isn't none
             if self.sort_key:
                 for subset in splits:
                     subset.sort_key = self.sort_key
             return splits
-
 
     def __getitem__(self, i):
         return self.examples[i]
--- a/torchtext/data/dataset.py
+++ b/torchtext/data/dataset.py
@@ -137,7 +137,7 @@ class Dataset(torch.utils.data.Dataset):
             # In case the parent sort key isn't none
             if self.sort_key:
                 for subset in splits:
-                    subset.sort_key = self.ort_key
+                    subset.sort_key = self.sort_key
             return splits
 
 
--- a/torchtext/data/dataset.py
+++ b/torchtext/data/dataset.py
@@ -108,8 +108,14 @@ class Dataset(torch.utils.data.Dataset):
         if not stratified:
             train_data, test_data, val_data = rationed_split(self.examples, train_ratio,
                                                              test_ratio, val_ratio, rnd)
-            return tuple(Dataset(d, self.fields)
+            splits = tuple(Dataset(d, self.fields)
                          for d in (train_data, val_data, test_data) if d)
+
+            # In case the parent sort key isn't none
+            if self.sort_key:
+                for subset in splits:
+                    subset.sort_key = self.sort_key
+            return splits
         else:
             if strata_field not in self.fields:
                 raise ValueError("Invalid field name for strata_field {}"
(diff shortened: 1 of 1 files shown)


mttk:
@jekbradbury thanks, must've been braindead while writing that part.

alumae:
I believe you also need to copy the sort_key and sort_within_batch fields.

mttk:
@alumae `sort_within_batch` is part of the Iterator (as well as `sort`), while the `sort_key`, which is being copied belongs to the dataset.

alumae:
Then it's probably a different issue, but it's similar to this one. 

Here's my code:

```
!wget https://phon.ioc.ee/~tanela/tmp/names_train.csv
!wget https://phon.ioc.ee/~tanela/tmp/names_test.csv

from torchtext import data
from torchtext import datasets
NAME = data.Field(tokenize=list, init_token="<bos>", eos_token="<eos>", include_lengths=True, batch_first=True)
LABEL = data.Field(sequential=False)
train_dataset, test_dataset = data.TabularDataset.splits(path=".", train='names_train.csv', test='names_test.csv', format='tsv', skip_header=True, fields=[('name', NAME), ('label', LABEL)])
NAME.build_vocab(train_dataset)
LABEL.build_vocab(train_dataset)
train_iter, test_iter = data.BucketIterator.splits((train_dataset, test_dataset), batch_size=32,  device=0, repeat=False)
print(next(iter(test_iter)))

```
Result:
```
TypeError                                 Traceback (most recent call last)
<ipython-input-21-1cd3949e427b> in <module>()
----> 1 print(next(iter(test_iter)))
      2 

/usr/local/lib/python3.6/dist-packages/torchtext/data/iterator.py in __iter__(self)
    133     def __iter__(self):
    134         while True:
--> 135             self.init_epoch()
    136             for idx, minibatch in enumerate(self.batches):
    137                 # fast-forward if loaded from state

/usr/local/lib/python3.6/dist-packages/torchtext/data/iterator.py in init_epoch(self)
    109             self._random_state_this_epoch = self.random_shuffler.random_state
    110 
--> 111         self.create_batches()
    112 
    113         if self._restored_from_state:

/usr/local/lib/python3.6/dist-packages/torchtext/data/iterator.py in create_batches(self)
    233     def create_batches(self):
    234         if self.sort:
--> 235             self.batches = batch(self.data(), self.batch_size,
    236                                  self.batch_size_fn)
    237         else:

/usr/local/lib/python3.6/dist-packages/torchtext/data/iterator.py in data(self)
     94         """Return the examples in the dataset in order, sorted, or shuffled."""
     95         if self.sort:
---> 96             xs = sorted(self.dataset, key=self.sort_key)
     97         elif self.shuffle:
     98             xs = [self.dataset[i] for i in self.random_shuffler(range(len(self.dataset)))]

TypeError: '<' not supported between instances of 'Example' and 'Example'
```

It works if I manually set:

```
test_iter.sort_within_batch = train_iter.sort_within_batch
test_iter.sort = train_iter.sort

print(next(iter(test_iter)))
```
Output:
```
[torchtext.data.batch.Batch of size 32]
	[.name]:('[torch.cuda.LongTensor of size 32x15 (GPU 0)]', '[torch.cuda.LongTensor of size 32 (GPU 0)]')
	[.label]:[torch.cuda.LongTensor of size 32 (GPU 0)]
```

mttk:
@alumae Yeah, this is (fortunately) unrelated to `split()`, but (unfortunately) related to default Dataset behavior with custom datasets. It should be clarified and documented better.

What happens is that the Dataset constructor doesn't know how to sort the `examples` in batches since your dataset can have arbitrarily named fields (in this case, `name` and `label`). I guess you want to sort over the length of `name`, but the dataset can't know that. Maybe you want to sort examples in the batch over `label`.

So - a fix (and I guess intended usage) would be to manually set the `sort_key` in iterator constructor. The `Iterator` only backs off to the `Dataset.sort_key` if you don't pass the `sort_key` argument to the iterator. An example:

```python
sort_key = lambda ex: len(ex.name)
# ...
train_iter, test_iter = data.BucketIterator.splits((train_dataset, test_dataset), batch_size=32,  device=-1, repeat=False, sort_key=sort_key)
```
This is done to minimize the padding within batches. If you don't want to do this, this is the alternative:

```python
train_iter, test_iter = data.BucketIterator.splits((train_dataset, test_dataset), batch_size=32,  device=-1, repeat=False, sort=False, sort_within_batch=False)

```

PR #282, opened 2018-04-11
[WIP] Dataset streaming

mttk:
Just to make keeping tabs on this easier:

- [x] Create streaming version of Dataset
- [x] Create streaming version of Field
- [x] Create streaming version of Iterator
  -  The dataset iterator breaks after buffer capacity overtakes whole dataset size
- [x] Make everything serializable
  - Doesn't work for SpaCy tokenization
- [x] Add a better way of buffering data (currently, it's just load full buffer, yield batches, load -- I'd like this to happen more smoothly)
  - This will be a TODO for later as I want to get an alpha version to work
- [ ] Sync to `pytorch/text` master
- [x] Minimize code duplication
- [ ] Add & run detailed tests

Right now, the code is quite ugly as I'm duplicating code wherever I can to make editing & debugging easier. This will change as some issues get cleaned up.


akurniawan:
hi @mttk, would love to use this once this is merged, may we know what the current status is?

mttk:
Hey, sorry for keeping radio silence wrt this. I'm honestly not happy with the current state of data streaming with respect to the state of torchtext (the main issue I have is that there is no simple way to make existing classes streamable, like adding a mixin, but all classes need to be rewritten anew).

I've been overwhelmed with other work, which has somewhat stabilized now. The plan is to first release a somewhat stable `0.3.0` to pypi (probably tomorrow). From then, I'll write some (minor) tests and then I'll add this branch to master. I've shied away from adding it earlier as I don't consider it stable, however having _some_ version of streaming and then ironing out the issues seems better than having none at all. 

Be aware that this will be an iterative process though, and that there will be errors :)
I'd expect to add this PR to master during next week.

ciphurus:
Hi guys - is this feature available now? Is there any simple example of how processed torchtext data and vocabulary can be saved and reloaded?

Froskekongen:
It would be great to have streaming data as part of torchtext! @mttk: Did you figure this one out?

siddsach:
@mttk would really appreciate this feature in torchtext!!! Most NLP datasets useful for deep learning are too big to have in memory.

ailurus1991:
any updates?

captainvera:
@mttk This would be a great feature! I'd really appreciate this in torchtext!

mridunarang:
any updates?

xdefeaterx:
ну шо там?

erip:
I think this can be safely closed. With the larger migration to datapipes which act as functional streams, I think this will be solved. cc @Nayef211

abhinavarora:
Closing this as obsolete. Created a label called obsolete, that we can add to such PRs and issues.

Issue #264, opened 2018-03-28
build_vocab on multi-dimension data?


shihan9:
line#250 in the `field.py` will throw error if `x` is a nested list while this is possible in some datasets. my suggestions is to replace 
```
counter.update(x)
```
with 
```
try:
    counter.update(x)
except TypeError:
    counter.update(flatten(x))
```
function `flatten` is defined as
```
def flatten(nested):
    for x in nested:
        if isinstance(x, (list,tuple)):
            yield from flatten(x)
        else:
            yield x
```

PR #286, opened 2018-04-13
added bAbI dataset

shihan9:
there are three things in this PR:

1. enhanced build_vocab on three dimensional data input, for example, the story in shape of [batch_size, memory_size, seq_len] of bAbI dataset. #264 

2. added Facebook bAbI dataset and its test file

3. changed urllib to requests for downloading bAbI dataset.

### DIFF
--- a/test/babi.py
+++ b/test/babi.py
@@ -1,0 +1,50 @@
+from torchtext import datasets
+
+# en-valid
+TRAIN_NUM = [0] + [900] * 16 + [904, 905, 900, 904]
+VAL_NUM = [0] + [100] * 16 + [96, 95, 100, 96]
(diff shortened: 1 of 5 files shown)


jekbradbury:
Is this bAbI-1k or bAbI-10k?

shihan9:
both. there is a bool argument called `tenK`.

shtoshni92:
Hi there,

I think you forgot to reverse the story for bAbI. See [here](https://github.com/pytorch/text/blob/28fc0556f56fb1e4e68302ec7f18d80698078fe6/torchtext/datasets/babi.py#L29) where the story is reversed for restricting to the latest sentences within memory_size but are never reversed back.

Issue #285, opened 2018-04-13
Feature: input, target = batch


keitakurita:
Thanks, as always, for the amazing work on this library!

I've noticed that various libraries (including fast.ai and torchsample) expect to be able to iterate over each batch generated from a DataLoader like
```python
for batch in dataloader:
    input, target = batch
```
where `input` is the input to a model and `target` is the target variable(s).

I thought it would be nice if the batches generated by the torchtext data loaders could more easily and intuitively interface with this kind of functionality. This would also make code using torchtext more independent of the dataset. 
The question is, how to distinguish between input and target variables.
Here's my idea: specify whether each field is a target field using a keyword argument `target_field` (or something to that effect) when initializing the `Field` object for each field. This will enable the batch object to know which fields to return as inputs and which to return as targets.

My question is, does this seem like a reasonable feature/API? If so, I would be happy to implement it.

mttk:
There's a `LabelField` implementation, however it's currently just a shallow wrapper (setting `unk_init` and `sequential` to None. It would be reasonable to extend it with a `target` flag.

What's your idea of combining multi-input datasets (e.g. SNLI)? `input` could be a tuple / namedtuple, or the batch could unpack into multiple variables.

Anyway, all in all seems reasonable to me.

keitakurita:
(If I'm understanding your question correctly) I think returning a tuple would be the most consistent with the expectations of the APIs I mentioned earlier and would be the easiest to deal with.
For a batch with input variables `x1`, `x2`, and `x3` and a target variable of `y`, I am thinking the batch should behave like
```python
(x1, x2, x3), y = b
```

The order of the variables will be the same as the order of the fields in the dataset.

mttk:
Closing as #285 has landed.

PR #288, opened 2018-04-15
Added batch iteration into input and target variables (#285)

keitakurita:
Now, the batch can be iterated into input and target variables like
```
>>> (x1, x2, x3), y = batch
```

where `x1` to `x3` are inputs to a downstream model and `y` is the label data for this batch.


torchtext/data/batch.py
```diff
@@ -58,6 +62,20 @@ def __str__(self):
     def __len__(self):
         return self.batch_size
 
+    def _get_vars(self, target):
```
mttk: Could this be changed to some enumeration and/or commented (ex `target="outputs"`, or just add a docstring)

torchtext/data/batch.py
```diff
@@ -58,6 +62,20 @@ def __str__(self):
...
+        fields = self.target_fields if target else self.input_fields
+        vars = tuple(getattr(self, f) for f in fields)
+        if len(vars) == 0:
```
mttk: Also brief inline comments on the cases here

keitakurita:
Hi, thanks for the review! I changed the code to reflect my intent more clearly, I think this should resolve the above problems, but if anything can be improved please tell me :)

mttk:
Awesome, thanks, everything seems good now!

SudoSharma:
Hey everyone, I ran into the same issue as @raheelqader in #422, with the "is_target" attribute being referenced for a "RawField" instance. I'd like to suggest we add an "is_target" attribute to "RawField" and set it to False by default. 

I'd like to contribute, but this is my first time. Should I just go ahead and do a PR once I've made the fix? Thanks! Looking forward to contributing.

mttk:
Hey, sure, feel free to submit a PR!

Issue #287, opened 2018-04-13
[Feature] Relax the memory requirements of vector loading


mttk:
When trying to load GloVe 840B, all the RAM (32G) on my server gets eaten up and the code breaks due to the whole set of vectors being loaded into memory and then indexed. 

I'd like to accommodate an option for loading only the subset of vectors which are present in the currently built vocab or union of vocabs (which is a common use-case). 

My idea is to extend the Vector class to accept an optional `vocab` argument which contains the words for which the vectors should be loaded.

PR #298, opened 2018-04-24
Fix vector loading memory requirements

mttk:
#287

- ~~Added two arguments to the `Vocab` class which allow the user to (1) retain only the pretrained vectors present in the dataset (`discard_unused`) or (2) keep vectors from a pre-defined vocabulary (`custom_vocab`).~~
- Memory usage when loading large pretrained vector files (Glove 840B / Glove 6B) ~~as well as the time it takes to load the vectors~~ has decreased **immensely**
- ~~Vectors default to loading only for tokens present in the dataset (`discard_unused=True`)~~

~~The vectors need to be initialized by passing the `vectors` argument to `build_vocab` as a **string** and not a `Vectors()` object for case (1).~~

~~Let me know if this seems sensible and I'll whip up a couple of tests. A documentation/tutorial review is also long overdue, but that'll come after the data streaming.~~

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,9 +18,12 @@ This repository consists of:
 Installation
 ============
 
-Make sure you have Python 2.7 or 3.5+ and PyTorch 0.2.0 or newer. You can then install torchtext using pip::
+
+Make sure you have Python 2.7 or 3.5+ and PyTorch 0.4.0 or newer. You can then install torchtext using pip::
 
     pip install torchtext
+    
+For PyTorch versions before 0.4.0, please use `pip install torchtext==0.2.3`.
 
 Optional requirements
 ---------------------
@@ -69,7 +72,7 @@ The data module provides the following:
   .. code-block:: python
 
       >>> # continuing from above
-      >>> mt_dev = data.TranslationDataset(
+      >>> mt_dev = datasets.TranslationDataset(
       ...     path='data/mt/newstest2014', exts=('.en', '.de'),
       ...     fields=(src, trg))
       >>> src.build_vocab(mt_train, max_size=80000)
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -59,5 +59,5 @@ if [[ "$SKIP_TESTS" != "true" ]]; then
     python -m nltk.downloader perluniprops nonbreaking_prefixes
 
     # PyTorch
-    conda install --yes pytorch torchvision -c soumith
+    conda install --yes pytorch torchvision -c pytorch
 fi
--- a/test/babi.py
+++ b/test/babi.py
@@ -1,0 +1,50 @@
+from torchtext import datasets
+
+# en-valid
+TRAIN_NUM = [0] + [900] * 16 + [904, 905, 900, 904]
+VAL_NUM = [0] + [100] * 16 + [96, 95, 100, 96]
+TEST_NUM = [0] + [1000] * 20
+
+# Testcase 1 (joint training)
+train_iter, val_iter, test_iter = datasets.BABI20.iters(task=1, joint=True)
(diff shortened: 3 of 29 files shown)


dancsi:
Hello, I have tried to use this pull request to load the default (english) FastText vectors, but unfortunately it seems that there are still some problems, as I ran out of RAM and swap (16 + 16GB) at some point after loading the vectors into the `array`, and before saving the `.pt` file...

mttk:
@dancsi the loading is still going to be limited by the vocab size, but this shouldn't happen.

I haven't yet documented the way to load the vectors with this method, but essentially you have to pass the string identifier of the vectors instead of the vector instance, e.g. instead of:

```
vectors = GloVe(name="42B", dim=300)
TEXT.build_vocab(train, test, val, vectors=vectors)
```
you should pass the string identifier of the vecors
```
TEXT.build_vocab(train, test, val, vectors="glove.42B.300d")
```
Since the constructors (FastText also) already load all of the vectors.

dancsi:
Sorry, I should have been more clear. I am loading the vectors like
```
v = torchtext.vocab.FastText()
```
just for testing purposes. Since there are 2,519,370 300-dimensional vectors, even storing them as double precision (which I doubt that it is necessary), that is only slightly above 6GB. Thus, I should comfortably be able to store 2 full copies in my RAM. This line, however, rapidly started filling my RAM after everything was loaded to `vectors`:
```
In [1]: import torchtext
In [2]: v = torchtext.vocab.FastText()
  0%|          | 0/2519371 [00:00<?, ?it/s]Skipping token 2519370 with 1-dimensional vector ['300']; likely a header
100%|██████████| 2519371/2519371 [04:27<00:00, 9412.75it/s]
# The RAM starts filling here :(
```

mttk:
Yeah, that's unrelated to this PR, but still an issue. I've managed to isolate it (it happens when copying vectors to torch.Tensor) and will update the PR tomorrow. After that, the usage is ~9G. Note that not only the vectors are stored, but also a bunch of strings, so that increases the memory usage.

dancsi:
Oh, I see. I don't know what exactly you fixed, but I worked around the issue by peeking in the first few lines of the input, to determine the dimension of the vectors, and then preallocate just one pytorch tensor of the appropriate size - so there is no intermediate `array` (and everything fits into <4GB)

mttk:
I did the same thing but with a numpy intermediate, which should've shared memory with the torch tensor, but obviously doesn't.

jekbradbury:
This is still WIP, right? Or should I merge it as is now? Thank you Martin for all the work you've been putting in recently!

dancsi:
I mean, I could submit a PR for fast vector loading. Pity that I can not push to this PR...

jekbradbury:
I believe you can push this branch (mttk/text) to your own GitHub account and submit a new PR that builds on it?

dancsi:
Well, this specific branch (mttk:vector_memreq) doesn't seem to exist anymore, so I would have to do some additional git trickery... Tomorrow...

mttk:
@jekbradbury yeah, still need to set up some tests, I'll do it over the weekend

@dancsi the branch exists, you should be able to see it here: https://github.com/mttk/text/tree/vector_memreq

mttk:
@jekbradbury in the end, it seems that the problem was caused by the conversion from `array.array` to `torch.tensor`. The memory footprint of loading the entire large dimensional vectors turns out to be manageable.

The filtering approach is problematic to implement due to caching of the vectors once loaded, which happens at the same time as filtering. This would mean that some name for the cached file should be generated based on the words in the vocabulary in order to differ between vocabs, which feels like unnecessary as the memory issues have been resolved.

So -- this can be merged as soon as Travis says OK.

mttk:
Closing in favor of #353

Issue #294, opened 2018-04-21
Make device arg take in torch.device as well


SsnL:

mttk:
could you elaborate a bit as `torch.device` is a ctx manager?

I'd expect the use-case to be that `torch.cuda.current_device()` is passed to the `Iterator`. Alternatively, if the device is unset I can fish out the current device in the same fashion. I'm not sure when the exact class would be passed as an arg.

PR #299, opened 2018-04-25
[Breaking] Pytorch 0.4 release compatibility fixes

### LINKED ISSUE #295: Update Field.numericalize volatile option for PyTorch 0.4.0

mttk:
#295
- Fixes volatile for pytorch 0.4 / 0.3 (with a version check)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,9 +18,12 @@ This repository consists of:
 Installation
 ============
 
-Make sure you have Python 2.7 or 3.5+ and PyTorch 0.2.0 or newer. You can then install torchtext using pip::
+
+Make sure you have Python 2.7 or 3.5+ and PyTorch 0.4.0 or newer. You can then install torchtext using pip::
 
     pip install torchtext
+    
+For PyTorch versions before 0.4.0, please use `pip install torchtext==0.2.3`.
 
 Optional requirements
 ---------------------
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -59,5 +59,5 @@ if [[ "$SKIP_TESTS" != "true" ]]; then
     python -m nltk.downloader perluniprops nonbreaking_prefixes
 
     # PyTorch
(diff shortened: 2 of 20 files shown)


SsnL:
Please fix the docstrings too.

SsnL:
It would be better if we can also raise an error/warning when the flag is set, so users can know that they need to update this.

mttk:
I also added weak support for `torch.device` now, as in #294. It's not ideal as I'm just unpacking the device index, but it will take some time to find all the spots where the code breaks if the variable can be either an `int` or a `torch.device` instance.

shihan9:
how about get rid of all the device arguments in torchtext? user could change the device using one-liner.

```python
input, target = input.to(device), target.to(device)
```

jekbradbury:
I'm open to either of these solutions, and I'm sorry I haven't been on top of 0.4 compatibility; let's get something merged soon so we have a working release

mttk:
We could dump the responsibility onto the user (I prefer this as it would make for some cleaner code), but that would kinda ruin backwards compatibility.

jekbradbury:
I'm OK with 0.4 support being a breaking release

mttk:
I went on a deleting spree:

- I removed every place that `train` was passed as an argument, except (1) in the constructor of `Iterator`, where it is used to set the repeat/shuffle/sort flags and (2) in `Batch.fromvars` since it was stored during serialization in previous versions, however there it's a no-op and is simply present so that previously stored models can be restored.
- I kept the `device` argument to be consistent with pytorch's `DataLoader` (which sets the device internally). However, now the argument has to either be a string or `torch.device` and is passed to the tensor constructor. In case the argument is an integer, a warning is logged.

Still waiting to see if this passes tests as my server is acting up.

mttk:
@jekbradbury the current build passes everything. When you get some time, please do a sanity check on the changes, but I don't see any obvious issues.

PR #302, opened 2018-04-28
[Bug Fix] numericalize doesn't create cuda tensor 

shihan9:
the fixes to pytorch 0.4 compatibility issues stop creating cuda tensor due to:
```python
self.tensor_type(arr, device=device)
```
tensor_type is torch.LongTensor by default which doesn't accept a `device` argument. as suggested by [migration guide](http://pytorch.org/2018/04/22/0_4_0-migration-guide.html), i change it to `torch.tensor` for numericalization. 

pytorch 0.4 also decouples type, device, layout from creation of tensors. i changed `tensor_type` to `dtype`, because the tensor_type like `torch.LongTensor` explicitly creates a CPU dense tensor of type long, which means more than just a type.

### DIFF
--- a/test/data/test_field.py
+++ b/test/data/test_field.py
@@ -336,7 +336,7 @@ class TestField(TorchtextTestCase):
         # Test basic usage
         int_field = data.Field(sequential=False, use_vocab=False)
         float_field = data.Field(sequential=False, use_vocab=False,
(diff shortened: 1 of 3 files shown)


shihan9:
#304

dolphin-zs:
Good fix! I also encounter the same problem. Wish to see the update.

PR #308, opened 2018-05-04
Fix vector dimension of OOV token (#275)

### LINKED ISSUE #275: wordvec of unk token
https://github.com/pytorch/text/blob/master/torchtext/vocab.py#L242

When the word is in the dictionary, the returned is a one-dim vector, while unk vector is a two-dim vector.

I assume this is not expected, right?

domaala:
This fixes #275 .


domaala:
The single test that was failing depended on Xavier initialization, which does not allow for 1-D tensors. Considering that we went for a 1-D vector of an OOV vector, I had to change it. 

However, this is kind of problematic for `unk_init `in general, so we should at least update the documentation to explain this 1-D constraint.

PR #345, opened 2018-06-29
Adding quality estimation based on the WMT 2018.

mtreviso:
Simple class for loading data of WMT 2018 quality estimation task. More info: [http://www.statmt.org/wmt18/quality-estimation-task.html](http://www.statmt.org/wmt18/quality-estimation-task.html)


mtreviso:
I will work more on this in the near future. For now I will close this PR.

PR #351, opened 2018-07-14
TabularDataSet format fix

cwarny:
Documentation says format can be passed in a case-insensitive way for tabular data sets, but this is not true. The if conditions compare unlowered format variable to lowercased 'csv' and 'tsv', etc.: `if format == 'csv':`. This PR fixes that.

### DIFF
--- a/torchtext/data/dataset.py
+++ b/torchtext/data/dataset.py
@@ -221,9 +221,10 @@ class TabularDataset(Dataset):
                 and also enables selecting a subset of columns to load.
             skip_header (bool): Whether to skip the first line of the input file.
         """
+        format = format.lower()
         make_example = {
             'json': Example.fromJSON, 'dict': Example.fromdict,
-            'tsv': Example.fromCSV, 'csv': Example.fromCSV}[format.lower()]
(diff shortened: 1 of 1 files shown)


WindChimeRan:
met same bug as well.

### data:

> so , unless that 's gon na be feasible , then ... \t so , if you can afford it , then ...
of course you did .	of course it is .
by now , singh 's probably been arrested . \t because he was probably just arrested at the moment by singh .

### upper

```python
TEXT = data.Field(sequential=True, tokenize=lambda x:x.split())
train_set = data.TabularDataset(path='./data/test.txt',format='TSV',
                                fields=[('Source', TEXT), ('Target', TEXT)])
print(train_set[0].Source)                       
```
`[u's']`      

### lower
```python
TEXT = data.Field(sequential=True, tokenize=lambda x:x.split())
train_set = data.TabularDataset(path='./data/test.txt',format='tsv',
                                fields=[('Source', TEXT), ('Target', TEXT)])
print(train_set[0].Source)         
```
`[u'so', u',', u'unless', u'that', u"'s", u'gon', u'na', u'be', u'feasible', u',', u'then', u'...']`

PR #353, opened 2018-07-21
Make vector loading more flexible.

notnami:
* Add support for passing a `max_vectors` argument, in situations where the entire set of pre-trained vectors can't fit in memory (most pre-trained vector sets are sorted in the descending order of word frequency, so if you pass, say, `max_vectors=100_000`, you will get pre-trained for 100_000 most common words)
* Add ability to read pre-trained vectors from gzipped files (without tar)
* Read pre-trained vectors by iterating over the vectors file object line by line, without reading the whole file into memory first
* Only load the file in binary mode to make parsing logic more compact
* Make vector loading more memory-efficient (based on #298)
* Minor refactoring to improve style

### DIFF
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -252,6 +252,10 @@ class Vectors(object):
         cache = '.vector_cache' if cache is None else cache
         self.unk_init = torch.Tensor.zero_ if unk_init is None else unk_init
         self.cache(name, cache, url=url, max_vectors=max_vectors)
+        self.itos = None
+        self.stoi = None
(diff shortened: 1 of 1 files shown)


notnami:
I fixed the issues with style, but now the Python 3.5 build has failed for an unclear reason. I believe that same build had passed before. I don't see an option to restart it, and not sure how to proceed.

mttk:
@notnami I restarted it, probably just timed out. Will merge when it passes.

mttk:
@notnami thanks a lot!

PR #361, opened 2018-08-07
migrate to sacremoses and add toktok tokenizer

### LINKED ISSUE #306: MosesTokenizer has been moved out of NLTK due to licensing issues
@jekbradbury great work here!

Due to https://github.com/nltk/nltk/issues/2000, we had to remove MosesTokenizer out of NLTK but now it's hosted on https://github.com/alvations/sacremoses 

```
pip install sacremoses
```

The silver lining is that the package comes with the data needed for tokenization so there's no need to keep the `nltk_data` directory =)

----

I would propose adding `sacremoses` on top of `nltk` because NLTK has another port of a nice tokenizer (by @jonsafari) that people overlook, https://github.com/nltk/nltk/blob/develop/nltk/tokenize/toktok.py (I think it's fast too)

keon:
Moses tokenizer has been deleted from nltk because of a license issue.
The current build is broken because of this change.
This fix resolves #306 
Also added toktok tokenizer based on the suggestion from @alvations

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ requests
 # Optional NLP tools
 nltk
 spacy
+sacremoses
 git+git://github.com/jekbradbury/revtok.git
 
 # Documentation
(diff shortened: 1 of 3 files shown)


alvations:
BTW, you can also use Toktok as a standalone now =) 

https://github.com/alvations/toktok

mttk:
@alvations thanks, will look into adding this soon :)

PR #362, opened 2018-08-07
Patch subword test failing in python 2.7

mttk:
[TestSubword](https://github.com/pytorch/text/blob/master/test/data/test_subword.py#L8) fails for python 2.7, making other tests not to run in recent PR's.

Since the problem seems to be external (revtok), this patch will skip the test for py2.7 so that the failing test does not stop patch and codecov.

### DIFF
--- a/test/data/test_subword.py
+++ b/test/data/test_subword.py
@@ -1,10 +1,14 @@
 import unittest
+import pytest
+import sys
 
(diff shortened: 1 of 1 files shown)


mttk:
Merging as the failing test is related to the tokenizer fixed in another pr.

PR #364, opened 2018-08-12
Check failing test in travis

mttk:


mttk:
@keon checked both here and locally with sacremoses 0.4 the same test still fails.

keon:
@mttk i think it is not the fault of sacremoses. I'll post a PR shortly

Issue #336, opened 2018-06-05
dataset.split()'s random_state argument causes TypeError


bentrevett:
Code:

``` python
from torchtext import data
from torchtext import datasets

TEXT = data.Field()
LABEL = data.LabelField()
train, test = datasets.IMDB.splits(TEXT, LABEL)
train, valid = train.split(random_state=1234)
```

Error:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-2b44beb35a3b> in <module>()
      6 LABEL = data.LabelField()
      7 train, test = datasets.IMDB.splits(TEXT, LABEL)
----> 8 train, valid = train.split(random_state=1)

~/.conda/envs/pytorch04/lib/python3.6/site-packages/torchtext/data/dataset.py in split(self, split_ratio, stratified, strata_field, random_state)
    108         if not stratified:
    109             train_data, test_data, val_data = rationed_split(self.examples, train_ratio,
--> 110                                                              test_ratio, val_ratio, rnd)
    111             return tuple(Dataset(d, self.fields)
    112                          for d in (train_data, val_data, test_data) if d)

~/.conda/envs/pytorch04/lib/python3.6/site-packages/torchtext/data/dataset.py in rationed_split(examples, train_ratio, test_ratio, val_ratio, rnd)
    291     # by ratio x length slices for each of the train/test/dev? splits
    292     N = len(examples)
--> 293     randperm = rnd(range(N))
    294     train_len = int(round(train_ratio * N))
    295 

~/.conda/envs/pytorch04/lib/python3.6/site-packages/torchtext/data/utils.py in __call__(self, data)
     96     def __call__(self, data):
     97         """Shuffle and return a new list."""
---> 98         with self.use_internal_state():
     99             return random.sample(data, len(data))

~/.conda/envs/pytorch04/lib/python3.6/contextlib.py in __enter__(self)
     79     def __enter__(self):
     80         try:
---> 81             return next(self.gen)
     82         except StopIteration:
     83             raise RuntimeError("generator didn't yield") from None

~/.conda/envs/pytorch04/lib/python3.6/site-packages/torchtext/data/utils.py in use_internal_state(self)
     81         """Use a specific RNG state."""
     82         old_state = random.getstate()
---> 83         random.setstate(self._random_state)
     84         yield
     85         self._random_state = random.getstate()

~/.conda/envs/pytorch04/lib/python3.6/random.py in setstate(self, state)
[...]

PR #383, opened 2018-09-14
Fix parameter's explanation in Dataset.split

nzw0301:
Current `random_state` takes a tuple returned by `random.getstate()`, not `int` (See #336 ). So I fix the docstring.


mttk:
Great, thanks! Will merge when Travis completes.

mttk:
@nzw0301 could you please just correct the line length that Flake is complaining about?

nzw0301:
@mttk Thank you for your comment. I fixed it.

PR #386, opened 2018-09-17
fixed ParseTextField postprocessing, added genre field to MultiNLI

putama:
- Fix ParsedTextField postprocessing that still accepts three args instead of two
- Added an option for ParsedTextField to whether to reverse the sentence as postprocessing
- Include parse trees for testing and reverse postprocessing test

### DIFF
--- a/test/nli.py
+++ b/test/nli.py
@@ -2,58 +2,65 @@ from torchtext import data
 from torchtext import datasets
 
 # Testing SNLI
-TEXT = data.Field()
-LABEL = data.Field(sequential=False)
+print("Run test on SNLI...")
+TEXT = datasets.nli.ParsedTextField()
+LABEL = data.LabelField()
+TREE = datasets.nli.ShiftReduceField()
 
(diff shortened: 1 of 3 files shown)


mttk:
@putama could you please fix the flake errors reported w.r.t. some line lengths and (just for my sake) elaborate on what was the error with the postprocessing fix?

putama:
Hi. I found that on field.py line 290, they changed the call to postprocessing function into only 2 arguments. This was the quickest fix I could think of.

mttk:
Ok, cool. Could you just fix the last flake complaint (IMO, just if/else the function call and leave the lambdas as arguments)

putama:
Done, thanks. Could you tell what the latest unsuccessful check was about?

mttk:
Functions are preferred to lambdas in PEP8 due to easier stack tracing (https://stackoverflow.com/questions/25010167/e731-do-not-assign-a-lambda-expression-use-a-def). 
They seem reasonable here as it's unlikely to use them again so the hack around that check is to put them as default args in function calls, where they're allowed.

Issue #354, opened 2018-07-22
FEATURE REQUEST: Remove undesired words from vocabulary


manuelsh:
When building the vocabulary, it would be great to be able to pass a list of undesired words (e.g. punctuation marks, stop words, etc) and they will not belong to the vocabulary (hence, will be '<unk>' when "numericalizing").

domaala:
Not sure if this helps you, but you could pass your own preprocessing/postprocessing pipeline to the `Field` constructor that would handle the removal of unwanted tokens.

manuelsh:
Yes, I actually did that, thanks!

However, being such a common thing to do in NLP, possibly makes sense to add it in the library?

domaala:
Possibly, yes!

mttk:
Added in #397

manuelsh:
Amazing, thanks!

PR #397, opened 2018-09-20
Feature to ignore tokens

keitakurita:
Feature implementation for #354 
Simply added option to pass stop words to a field during initialization and have it remove words during preprocessing. API is based on the [CountTokenizer from scikit-learn](http://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.CountVectorizer.html).
Stop words being an attribute of the field (not the vocab) seems natural when you consider the vocab as just a mapping from tokens to integers.

### DIFF
--- a/test/data/test_field.py
+++ b/test/data/test_field.py
@@ -345,15 +345,14 @@ class TestField(TorchtextTestCase):
         question_field.build_vocab(tsv_dataset)
 
         test_example_data = question_field.pad(
-                               [question_field.preprocess(x) for x in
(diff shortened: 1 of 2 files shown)


mttk:
I'd just add a typecheck for `stop_words` to be a set (since people might pass a list as well) for speed. If it fails, cast it to list.
Either that or add that the stopwords have to be a set in the function doc.

keitakurita:
Updated!

mttk:
Great, thanks!

manuelsh:
That's amazing! Thanks a lot!

PR #409, opened 2018-09-24
Make NestedField compatible with GPU

ppuliu:
Otherwise, will get this error: 
```
RuntimeError: legacy constructor for device type: cpu was passed device type: cuda, but device type must be: cpu
```


mttk:
Thanks!

Issue #355, opened 2018-07-22
Ignoring UNK words


manuelsh:
Cannot find the way to ignore UNK words when numericalising, i.e. instead by substituting them by a 0, it just ignore that word.

Is that implemented?

This is useful in classification problems, when you just want to remove 'UNK' words.

rumaak:
https://torchtext.readthedocs.io/en/latest/data.html#field 

> postprocessing – A Pipeline that will be applied to examples using this field after numericalizing but before the numbers are turned into a Tensor. The pipeline function takes the batch as a list, and the field’s Vocab. Default: None.

I tried this, but unsuccessfully. When used pipeline deleting 0 from sequence, I get this error:

`ValueError: expected sequence of length 12 at dim 1 (got 11)`

I don't think there is anything closer to your use case than `postprocessing`, yet that doesn't seem to work. If anyone could try it out, it would be great.

Minimal example I used:
```
import torchtext
from torchtext.data import *

def func(*x):
    y = x[0]
    return [z for z in y if z != 0]

pipe = Pipeline(func)
field = Field(postprocessing=pipe)

# 2 arbitrary datasets, ds2 must contain words ds1 doesn't contain
ds1 = TabularDataset(fname1,'tsv',[('text',field)])
ds2 = TabularDataset(fname2,'tsv',[('text',field)])

field.build_vocab(ds1)
itr = Iterator(ds2,4)
next(iter(itr)).text
```

mttk:
So, the lifecycle of a dataset goes as follows:

1. `preprocess` is called on example-level (row-by-row), but at this moment the vocabulary is not yet built
2. `build_vocab` is called on the pre-processed dataset, mapping the strings to indices
3. `process` is called (pad & numericalize) on an already constructed batch (by an `Iterator`)
4. (optional) 'postprocess' is called

You can't fit this part into 4, since the batches have to be fixed-size, and you can't fit it into 1 because you don't yet know the vocabulary. What I would suggest (for now) is manually placing that part in "2.5".

After you have built your vocab, you can access `stoi`. Then you can do something like this:

```python
def clean(example, vocab):
    return [word for word in example if word in vocab]
# train here is the `Dataset` instance obtained through .splits()
for example in train:
  # assuming field name is 'text'
  example.text = clean(example.text, TEXT.vocab.stoi)
```

mttk:
[...]

PR #413, opened 2018-09-25
Add's dataset feature - remove unkown words from dataset's examples

rumaak:
Response to issue #355. As @mttk pointed out this feature should be solved differently as to optimize for performance, so I implemented his idea for temporary fix. As this is something that will probably get removed in future, I haven't written any tests yet, but if needed, I am completely open to that option too.

I also changed the arguments @mttk suggested and I think this seems like reasonable approach, but I am open to change this too.

### DIFF
--- a/torchtext/data/dataset.py
+++ b/torchtext/data/dataset.py
@@ -197,6 +197,19 @@ class Dataset(torch.utils.data.Dataset):
                             shutil.copyfileobj(gz, uncompressed)
 
         return os.path.join(path, cls.dirname)
+ 
+
+    def filter_examples(self, field_names):
+        """Remove unknown words from dataset examples with respect to given field.
+
+        Arguments:
+            field_names (list(str)): Within example only the parts with field names in
+                field_names will have their unknown words deleted.
+        """
+        for i, example in enumerate(self.examples):
+            example_part = [word for word in getattr(example,field_name) if word in self.fields[field_name].vocab.stoi]
+            setattr(example, field_name, example_part)
+            self.examples[i] = example 
 
 
 class TabularDataset(Dataset):


torchtext/data/dataset.py
```diff
@@ -197,6 +197,19 @@ def download(cls, root, check=None):
...
+
+    def filter_examples(self, field_names):
+        """Remove unknown words from dataset examples with respect to given field.
+
+        Arguments:
+            field_names (list(str)): Within example only the parts with field names in
+                field_names will have their unknown words deleted.
+        """
+        for i, example in enumerate(self.examples):
+            example_part = [word for word in getattr(example,field_name) if word in self.fields[field_name].vocab.stoi]
```
nzw0301: Could you add space: `example,field_name` → `example, field_name`

rumaak: done, thanks

torchtext/data/dataset.py
```diff
@@ -197,6 +197,19 @@ def download(cls, root, check=None):
                             shutil.copyfileobj(gz, uncompressed)
 
         return os.path.join(path, cls.dirname)
+ 
+
+    def filter_examples(self, field_names):
```
mttk: The argument is `field_names`, while in the code you use `field_name` (it would make sense to allow for multiple fields to be filtered at the same time).

torchtext/data/dataset.py
```diff
@@ -197,6 +197,19 @@ def download(cls, root, check=None):
...
+        """Remove unknown words from dataset examples with respect to given field.
+
+        Arguments:
+            field_names (list(str)): Within example only the parts with field names in
+                field_names will have their unknown words deleted.
+        """
+        for i, example in enumerate(self.examples):
+            example_part = [word for word in getattr(example, field_name) if word in self.fields[field_name].vocab.stoi]
```
mttk: `stoi` is a list, so it would be good to cast it to a set (this can be done outside of the list comprehension otherwise this line is going to explode

torchtext/data/dataset.py
rumaak: Yeah I've looked at the tests and went on to fix them

torchtext/data/dataset.py
rumaak: Done

rumaak:
I will proceed to write the tests, they should be done within few hours

mttk:
Thanks, no rush!

test/data/test_dataset.py
```diff
@@ -342,6 +342,111 @@ def test_stratified_dataset_split(self):
...
+
+        assert dataset[1].text1 == ["nay", "answer", "me"]
+        assert dataset[1].text2 == ["stand", "unfold", "yourself"]
+        assert dataset[1].label == 0
+
+        assert dataset[2].text1 == ["Horatio"]
+        assert dataset[2].text2 == ["a", "piece", "of", "him"]
+        assert dataset[2].label == 0
+
+    def filter_init(self, ex_val1, ex_val2, ex_val3):
```
mttk: Could you move this method outside of the Test class (and add a comment that it is used in the filtering test)

mttk:
@rumaak the tests seem good, thanks! Just move the non-test method outside of the test class definition and this is good to merge.

test/data/test_dataset.py
rumaak: Sure

Issue #406, opened 2018-09-24
BucketIterator iterating more than the length


ksurya:
I have code like this:

```python
batch_size = 100

train_dataset, test_dataset = convos.split(split_ratio=0.7)

train_iterator = torchtext.data.BucketIterator(
    train_dataset,
    batch_size=3000,
    sort_key=lambda x: torchtext.data.interleave_keys(len(x.context), len(x.response)),
    device=device
)

print("Batch size: ", batch_size)
print("Train size: ", len(train_iterator))
```
```text
Batch size:  100
Train size:  1
```

Following this, I do

```python
for count, batch in enumerate(train_iterator):
    print(count)
```

And this program doesn't stop (at least I checked until count = 3000). I suppose it should iterate just once!

nzw0301:
Yes, your code does not stop. 

`Iteration` class's behavior may be confusing. If you want to iterate just once, you need to pass `False` to `repeat` argument like this.

```python
torchtext.data.BucketIterator(
    train_dataset,
    batch_size=3000,
    sort_key=lambda x: torchtext.data.interleave_keys(len(x.context), len(x.response)),
    device=device,
    repeat=False
)
```

`False` as the default value of `repeat` seems to be better than `None`.

ksurya:
Does this mean, the iterator keeps repeating and is there a way to fix the number of repeats?

nzw0301:
Yes, if you want to iterate just once in each epoch, my code snippet above can do that.

ksurya:
Thanks you.  I would suggest to set  default `repeat` to `True` in `BucketIterator` -- that may break existing codebases but I think typically iterators are not in infinite loop.

mttk:
Yeah, I believe it is more or less agreed that the default infinite loop in Iterators is confusing.

I'd say that the expected behavior is to iterate for one epoch (so, the default would be `repeat=False`), and leave the infinite loop as an option if users prefer it. This might break something backwards but it's manageable.

zhangguanheng66:
I will just close this issue since a PR is attached. Thanks all for the help.

PR #417, opened 2018-09-25
Make repeat argument False in iterator class as default

nzw0301:
See #406


mttk:
Had to go through the codebase and see if this affects anything else -- seems that it doesn't. Thanks!

nzw0301:
Thanks!

Issue #276, opened 2018-04-11
why torchtext does not support google word2vec?


caolusg:
Do we have plan to do this?

dasguptar:
Is there an easy way to download the pretrained Google News Word2Vec bin file? From what I remember, it used to be downloaded from a Google Drive link, which is probably not amenable to download from the command line.

PR #421, opened 2018-09-25
[WIP] Add word2vec class

nzw0301:
See #276.

To download skip-gram embeddings trained on google news, we can use [DL4J's url](https://s3.amazonaws.com/dl4j-distribution/GoogleNews-vectors-negative300.bin.gz).

I also checked whether this mirror model is the same as [the original model](https://code.google.com/archive/p/word2vec/).

```
$ diff GoogleNews-vectors-negative300.bin original-GoogleNews-vectors-negative300.bin
$ # same!
```

I will update `Vector`'s cache function to read the binary file of word2vec.

### DIFF
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -411,7 +411,8 @@ class GloVe(Vectors):
 
 class Word2Vec(Vectors):
     urls = {
-        'googlenews': 'https://s3.amazonaws.com/dl4j-distribution/GoogleNews-vectors-negative300.bin.gz'
+        'googlenews': 'https://s3.amazonaws.com/'
+                      'dl4j-distribution/GoogleNews-vectors-negative300.bin.gz'
     }
 
     def __init__(self, name='googlenews', **kwargs):
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -237,21 +237,20 @@ class SubwordVocab(Vocab):
 def _infer_shape(f):
     num_lines, vector_dim = 0, None
 
-    # Assuming word, [vector] format
-
-    # Read header
+    # Read the first line (it may be a header of word2vec/fasttext format)
     line = f.readline()
-    if isinstance(line, six.binary_type):
-        line = line.decode('utf-8')
-    row = line.rstrip().split(" ")
+    row = line.rstrip().split(b" ")
 
+    # Assuming word, [vector] format
     if len(row) == 2:
-        num_lines, vector_dim = list(map(int, row))
-    else:
-        vector = row[1:]
         # The header present in some (w2v) formats contains two elements.
+        num_lines, vector_dim = list(map(int, row))
+        num_lines += 1
+    else:
+        # glove format: no header line
+        vector = row[1:]
         vector_dim = len(vector)
-        num_lines += 1  # First element read
+        num_lines += 1
 
         for _ in f:
             num_lines += 1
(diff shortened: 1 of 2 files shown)


nzw0301:
I faced [this encoding error](https://github.com/RaRe-Technologies/gensim/wiki/Recipes-&-FAQ#q10-loading-a-word2vec-model-fails-with-unicodedecodeerror-utf-8-codec-cant-decode-bytes-in-position-). I guess that [a gensim`s function](https://radimrehurek.com/gensim/models/keyedvectors.html#gensim.models.keyedvectors.Word2VecKeyedVectors.load_word2vec_format) can avoid this problem easily, but this repo needs to depend on gensim.

mttk:
This would actually be a great addition. The encoding error can probably be handled without importing gensim (their code is on github, and the licence allows code reuse, so with proper attribution, this can be handled without introducing additional dependencies).

nzw0301:
Thank you for your comment! I will try it.

mttk:
I'll reopen this, let me know what you find out.

nzw0301:
For binary file of word2vec, this snippet used partially in gensim, can load word vectors.

```python
import numpy as np
from numpy import fromstring, dtype
import sys


if sys.version_info[0] >= 3:
    unicode = str


def any2unicode(text, encoding='utf8', errors='strict'):
    if isinstance(text, unicode):
        return text
    return unicode(text, encoding, errors=errors)


path = "GoogleNews-vectors-negative300.bin"
encoding = 'utf-8'
unicode_errors = 'strict'

with open(path, "rb") as f:
    # read header
    header = any2unicode(f.readline(), encoding=encoding)
    num_words, dim = list(map(int, header.split()))

    binary_len = dtype(np.float32).itemsize * dim
    for _ in range(num_words):
        # read word
        chars = []
        while True:
            ch = f.read(1)
            if ch == b' ':
                break
            if ch == b'':
                raise EOFError("unexpected end of input; is count incorrect or file otherwise damaged?")
            if ch != b'\n':  # ignore newlines in front of words (some binary files have)
                chars.append(ch)
        word = any2unicode(b''.join(chars), encoding=encoding, errors=unicode_errors)
        weights = fromstring(f.read(binary_len), dtype=np.float32).astype(np.float32)
        # do something to store word and word-vector
```

torchtext/vocab.py
```diff
@@ -442,6 +448,17 @@ def __init__(self, name='840B', dim=300, **kwargs):
         super(GloVe, self).__init__(name, url=url, **kwargs)
 
 
+class Word2Vec(Vectors):
```
zhangguanheng66: Please add some docs here (including examples) and unit tests. Thanks for the contribution.

nzw0301: Thank you for your comment. I will add examples to [`vocab.rst` file later.

nzw0301:
I haven't updated `Vector`'s `cache` function to load word2vec's binary format.

zhangguanheng66:
> I haven't updated `Vector`'s `cache` function to load word2vec's binary format yet.
> 
> Now, I consider two ways to add this word2vec's pre-trained model feature;
> 
> * change `cache` function in `Vector` class
> * add `cache` function in `Word2Vec` class

I'm wondering the backward compatibility if change the cache function in Vector. Another way to do so is to add cache function as a general API and call it within Word2Vec class. These are just my two cents.

nzw0301:
My apologies for not working on this PR. I'm afraid that I cannot work to continue this PR due to another project and my work. If some contributors are interested in this feature, please free free to take over this PR from either scratch or this branch.

best regards,

PR #425, opened 2018-09-27
Use progress bar for downloads

kklemon:
As downloading datasets may take quiet some time and torchtext already uses the [tqdm](https://github.com/tqdm/tqdm) library, it may be convenient to have a progress on datasets downloads too.


torchtext/utils.py
```diff
@@ -1,6 +1,8 @@
 import six
 import requests
 import csv
+import math
```
rumaak: Take a look at details from travis build. The log suggests this line is not used, so it should be deleted.

kklemon: I'm sorry, my fault. Forgot to remove this import statement.

Issue #349, opened 2018-07-10
Quotation mark at start of line corrupts data


KrisKorrel:
When I read in a `.tsv` file containing the following two lines, containing two fields (input and output)
```
a " c    b
" b c	 a
```
the first example will results in fields ['a', '"', 'c'], ['b']
While the second example will results in an input field ['b', 'c', 'a'] and no output field.
I think the desired behaviour would be that it would read as ['"', 'b', 'c'], ['a'], and be agnostic of the specific characters or tokens used.

I've seen this go wrong in multiple ways. It seems as if the character is not properly escaped.

Produced with torchtext 0.2.3 and python 3.6.5
The problem does not seem to appear with torchtext 0.2.1

AveryLiu:
From my experience with torchtext 0.3.0, I believe the cause of this problem stems from the use of ``unicode_csv_reader`` in the ``dataset.py``. 

The quotation marks are by default to declare a field, which means the text inside the quotation marks is deemed as a whole. You can change this behavior by setting the parameter ``quotechar`` in ``unicode_csv_reader`` to ``None`` to by-pass the problem. However, torchtext doesn't expose the interface, so it requires you to modify the codebase to fix the problem. 

Another workaround is simply to remove all punctuations in the text.

KrisKorrel:
I see, thanks. And I understand that for normal CSV use, it is necessary to use such a `quotechar` in order to handle newline characters, for example.
But for the use in torchtext, shouldn't it indeed be `None` by default? I think that all Unicode characters separated by commas (in the case of `.csv` or tabs in the case of `.tsv`) should be used agnostically / as such.

AveryLiu:
Yes, I agree with you on that point and I also suffered from that problem and fixed it with punctuation removal.

At the moment I don't see any problem with the ``quotechar`` to be set to ``None`` by default. Maybe the owners @jekbradbury @nelson-liu  have different consideration?

keitakurita:
I can imagine people wanting to read tsv/csv files that have text fields containing newlines by using quotations as well, so I'm not sure setting `quotechar` to `None` by default is a good idea. For me, it seems like a quotation mark being at the head of a field while not being used to contain text seems more like an edge case, though I'm open to being convinced otherwise.

A better workaround would seem to be to just have torchtext expose the `quotechar` parameter when reading csvs. If this seems like a reasonable solution, I'd be happy to implement it and file a PR.

KrisKorrel:
[...]

PR #432, opened 2018-09-29
Exposed parameters for csv reader to user through dataset

keitakurita:
Handles issue #349 and also exposes additional attributes of `csv_reader` to the user of `TabularDataset` so that they can modify the handling of csv files as they see fit.
Any keyword arguments can now be passed to csv_reader through the `csv_reader_params` keyword argument.


mttk:
Great, thanks a lot!

PR #450, opened 2018-10-14
Adjust `is_target` in LabelField and update a couple of docstrings

harkous:


mttk:
@harkous can you just fix the travis error (in field.py, the comment line 695 is too long so just break it into two lines)

mttk:
Thanks!

PR #453, opened 2018-10-19
Field serialization, tokenizers in different languages

mttk:
- Add the option to serialize fields with `torch.save` or `pickle.dump`
- Add basic tests
- Allows tokenizers in different languages for spacy (argument `tokenizer_language`)

All lambdas were removed from the pipeline and replaced with closures / partials or explicit functions. The responsibility of a custom tokenizer being serializable is on the user. 
An ugly hack overcomes the problem of `torch.[dtype]` not being serializable.


xelibrion:
Are you guys planning a release any time soon @mttk? It would be great to get this change from pypi.

PR #458, opened 2018-10-22
fix stop_words conversion to set

### LINKED ISSUE #457: Field.stop_words is overwritten and is no longer a set
https://github.com/pytorch/text/blob/758d356fd76ee6102de06a4cb5e414a3164c9e42/torchtext/data/field.py#L166-L173

I think this is an error and self.stop_words should not be overwritten after the condition.

artemisart:
fix #457

### DIFF
--- a/torchtext/data/field.py
+++ b/torchtext/data/field.py
@@ -163,14 +163,10 @@ class Field(RawField):
         self.pad_token = pad_token if self.sequential else None
         self.pad_first = pad_first
         self.truncate_first = truncate_first
-        if stop_words is not None:
-            try:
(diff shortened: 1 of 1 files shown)


mttk:
@artemisart thanks, can you just revert the change in line 215 (back to `self.stop_words is not None`)? Even though the current version only supports `set` / `None`, in which case the expressions are equivalent, I'd prefer the former version.

artemisart:
Ok, done !

mttk:
Great, will be merged as soon as Travis completes. Thanks a lot for spotting this!

PR #459, opened 2018-10-23
Set `is_target` to False by default in `RawField`

SudoSharma:
Fixing issue potentially caused by #288. You can now use RawField directly without any workarounds as mentioned in issue 422.


mttk:
Great, thanks!

Issue #454, opened 2018-10-22
[Feature request] Reading compressed (e.g. gzipped) datasets?


eugene-kharitonov:
Hello,
I wonder if there could be a support for reading zipped/gzipped torchtext.data.TabularDataset-s? 
In our case the dataset is small enough that we want to store its versions in the repo alongside with the code.
I understand it's not a mainstream use-case.  Happy to make a PR myself, if it makes sense.
A relevant snippet https://github.com/pytorch/text/blob/master/torchtext/data/dataset.py#L251

mttk:
This sounds as a reasonable addition. I believe the gist of the change would be changing `open` to some variant of `zipfile`, `gzipfile` etc., as I believe you intended with the referenced line.

A PR would be appreciated!

PR #466, opened 2018-11-02
[wip] TabularDatasets can be read from gzip/bzip2 files

eugene-kharitonov:
* This PR introduces a new parameter `compression` for the TabularDataset constructor. This param defaults to None. When set to 'gz' or 'bz2', we open the dataset file with the corresponding decompressor.
* UTs seem to be passing.

An attempt to address https://github.com/pytorch/text/issues/454

### DIFF
--- a/test/common/torchtext_test_case.py
+++ b/test/common/torchtext_test_case.py
@@ -6,6 +6,9 @@ import os
 import shutil
 import subprocess
 import tempfile
+import bz2
+import gzip
+import io
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


eugene-kharitonov:
I played around and it seems that things become uglier when the support for py2 is needed, somewhat beyond the 10-line-fix I had in mind :D 

@mttk - what is the policy around supporting py2?

vin-ivar:
Is this/something similar to this likely to be merged? Python 2 reaches EOL in 2020, is it really very critical to ensure support?

mttk:
Right now, it is pretty critical to ensure support as I believe a significant amount of users are using python 2. So, support won't be dropped in the near future.
@eugene-kharitonov could you elaborate on the interface differences?

Issue #447, opened 2018-10-12
Apparent bug for ``torchtext.data.vocab.Vocab`` causing collisions between OOV words and in vocab words.


nryant:
When using `torchtext` version 0.3.1, I notice a rather alarming bug affecting `torchtext.vocab.Vocab` that causes collisions between OOV words and non-OOV words. When a vocabulary is instantiated with `specials_first=True`, there is a collision between OOV words and the first special word:

 ```
from collections import Counter
import torchtext
word_freqs = Counter({'the' : 10, 'apple' : 5, 'cat' : 2})
word_vocab = torchtext.vocab.Vocab(word_freqs, specials=['<pad>'], specials_first=True)
for word in ['<pad>', 'dfsddfsdfds']:
    print(word, word_vocab.stoi[word])
```

which yields:

```
('<pad>', 0)
('dfsddfsdfds', 0)
```

Similarly, when `specials_first=False`, there is a collision between OOV words and the highest frequency word in the vocabulary:

```
word_vocab = torchtext.vocab.Vocab(word_freqs, specials=['<pad>'], specials_first=False)
for word in ['the', 'dfsddfsdfds']:
    print(word, word_vocab.stoi[word])
('the', 0)
('dfsddfsdfds', 0)
```

The culprit appears to be the following line of `vocab.py`:

`self.stoi.update({tok: i for i, tok in enumerate(self.itos)})`

I assume this behavior is unintended. If so, it's a relatively easy fix assuming that `_default_unk_index` sill never be changed to output anything other than 0.

mttk:
Yeah this is not the nicest bug. Assuming UNK=0 is ok since it's already hardcoded (so we can hardcode it further). I'll fix this when I get the time, if someone wants to take a stab at it, feel free.

PR #482, opened 2018-12-06
Fix collisions between oov words and in-vocab words (#447)

mtreviso:
Fix issue #447 by verifying if `<unk>` is not in specials list and then creates a defaultdict without default_factory.

### DIFF
--- a/test/test_vocab.py
+++ b/test/test_vocab.py
@@ -49,6 +49,38 @@ class TestVocab(TorchtextTestCase):
         self.assertEqual(v.itos, expected_itos)
         self.assertEqual(dict(v.stoi), expected_stoi)
 
+    def test_vocab_without_unk(self):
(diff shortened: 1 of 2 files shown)


mttk:
Thanks, this will work for now, but needs to be handled better.

raheelqader:
This change broke my code as my unknown token was <UNK> (in capital). I had to manually change <unk> to <UNK> vocab.py to fix it.

cpuhrsch:
The documentation wasn't updated that Vocab now by default doesn't prepend the unk token anymore ([as is being verified in this test](https://github.com/pytorch/text/commit/edafd9cb284010043d8aba7d8e57f91aaad77e99#diff-1b0eb69c4a8ebef932dcd9fd350c7be4R52)) unless passed explicitly. I fixed it [here](https://github.com/pytorch/text/pull/567/files#diff-7ac0ea3f4d05c510e01f73bf2491691bL46) and included <unk> in the default specials list to reflect the default behavior.

PR #495, opened 2019-01-28
fix parameter specials in Field.build_vocab

speedcell4:


mttk:
I believe that the specials argument was off-limits to users intentionally. Do you have any use-case not covered by pad, bos, eos or unk?

mttk:
Makes sense. AFAIK, the case #3 should be better handled while creating batches (because post-processing is used just once, and you would want to have different tokens masked between epochs). 
Could you fix the travis errors & write a test for one of those use-cases (ex. add a special token and then validate it's added)

speedcell4:
no, `postprocessing` will be used once in every batch

```python
sentence = 'I have an apple'


def postprocessing(batch, vocab):
    print(f'postprocessing is called')
    res_batch = []
    for ex in batch:
        res_ex = []
        for token in ex:
            if random.random() < 0.5:
                res_ex.append(vocab.stoi['<unk>'])
            else:
                res_ex.append(token)
        res_batch.append(res_ex)
    return res_batch


class Dummy(Dataset):
    def __init__(self, examples, fields):
        super(Dummy, self).__init__(examples, fields)

    @classmethod
    def iters(cls):
        WORD = Field(postprocessing=postprocessing)
        fields = [('word', WORD)]
        examples = [
            Example.fromlist([sentence], fields=fields),
        ]
        dataset = cls(examples, fields)
        WORD.build_vocab(dataset)
        return Iterator(dataset, batch_size=1)


if __name__ == '__main__':
    train = Dummy.iters()
    for _ in range(2):
        for batch in train:
            print(batch.word.tolist())

# postprocessing is called
# [[2], [0], [3], [4]]
# postprocessing is called
# [[2], [5], [0], [4]]
```

I will finish the unit tests soon

speedcell4:
this project should provide an edit configuration file, my PyCharm always fix the none-pep8 format automatically

PR #501, opened 2019-02-04
Add batch_size_multiple argument to torchtext.data.batch

guillaumekln:
FP16 acceleration on Tensor Cores requires each dimension to be a multiple of 8. I propose here to extend the `torchtext.data.batch` generator to possibly constrain the number of returned elements to be a multiple of a specified value.


guillaumekln:
Let me also update the `Iterator` class.

PR #506, opened 2019-02-19
Fix encoding for sequence tagging dataset

akurniawan:
This PR is to fix the following error that I have encountered. The data need to be encoded first before torchtext able to load it perfectly.

```
  File "/opt/conda/lib/python3.6/site-packages/torchtext/data/dataset.py", line 78, in splits
    os.path.join(path, train), **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/torchtext/datasets/sequence_tagging.py", line 29, in __init__
    for line in input_file:
  File "/opt/conda/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 301: ordinal not in range(128)
```


mttk:
Thanks!

PR #521, opened 2019-03-31
fix dead url for fasttext vectors

grll:
Previous URL has been returning 403 for quite a while now this new url seems to be working better.

### DIFF
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -407,7 +407,7 @@ class GloVe(Vectors):
 
 class FastText(Vectors):
 
-    url_base = "https://dl.fbaipublicfiles.com/fasttext/vectors-wiki/wiki.{}.vec"
+    url_base = 'https://dl.fbaipublicfiles.com/fasttext/vectors-wiki/wiki.{}.vec'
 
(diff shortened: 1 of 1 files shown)


torchtext/vocab.py
```diff
@@ -407,7 +407,7 @@ def __init__(self, name='840B', dim=300, **kwargs):
 
 class FastText(Vectors):
 
-    url_base = 'https://s3-us-west-1.amazonaws.com/fasttext-vectors/wiki.{}.vec'
```
nzw0301: Could you use `'` instead of `"` for consistency?

torchtext/vocab.py
```diff
@@ -407,7 +407,7 @@ def __init__(self, name='840B', dim=300, **kwargs):
...
+    url_base = "https://dl.fbaipublicfiles.com/fasttext/vectors-wiki/wiki.{}.vec"
```
nzw0301: Could you use ' instead of " for consistency?

grll: sure, I changed it.

mttk:
Thanks!

PR #530, opened 2019-04-26
Avoid installation of `test` folder

vfdev-5:
A small patch to avoid copying `test/common` and `test/data` folders when `torchtext` is installed.
Currently one can find them at
```
ls <python>/site-packages/test/
> common  data
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setup_info = dict(
     ],
 
     # Package info
-    packages=find_packages(exclude=('test',)),
+    packages=find_packages(exclude=('test','test.*')),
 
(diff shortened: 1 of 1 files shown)


vfdev-5:
Tested locally by installing with 
```
git clone https://github.com/vfdev-5/text.git -b patch-1 && text
pip install .
ls <python>/site-packages/test/
> no folder
```

mttk:
@vfdev-5 flake is complaining about missing whitespace, could you please fix that and then I'll merge

mttk:
Thanks!

PR #531, opened 2019-04-26
Fix the hard code for the unk_index

seanliu96:
Hi all,

I notice that the Vocab uses the defaultdict but the default number is always 0. 
If I set specials_first=False, then the unk_index must be wrong;
If I set the specials=['\<eos\>', '\<sos\>', '\<pad\>', '\<unk\>'], then the unk_index is also wrong.

The correct way is to generate the unk_index according to the specials.

Best regards,
Xin Liu

### DIFF
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -67,7 +67,6 @@ class Vocab(object):
             # only extend max size if specials are prepended
             max_size = None if max_size is None else max_size + len(specials)
 
-
         # frequencies of special tokens are not counted when building vocabulary
         # in frequency order
         for tok in specials:
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -63,8 +63,8 @@ class Vocab(object):
             # find the unk_index in specials
             if '<unk>' in specials:
                 self.unk_index = specials.index('<unk>')
-            elif '<pad>' in specials:
-                self.unk_index = specials.index('<pad>')
+            # elif '<pad>' in specials:
+                # self.unk_index = specials.index('<pad>')
 
         # frequencies of special tokens are not counted when building vocabulary
         # in frequency order
@@ -86,10 +86,10 @@ class Vocab(object):
             # find the unk_index in specials
             if '<unk>' in specials:
                 self.unk_index = specials.index('<unk>') + len(self.itos)
-            elif '<pad>' in specials:
-                self.unk_index = specials.index('<pad>') + len(self.itos)
+            # elif '<pad>' in specials:
+                # self.unk_index = specials.index('<pad>') + len(self.itos)
             self.itos.extend(list(specials))
-        
+
         if self.unk_index is None:
             self.stoi = defaultdict()
         else:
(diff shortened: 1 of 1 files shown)


seanliu96:
I paste the result of test_vocab_without_unk. It doesn't raise a KeyError because I set the unk_index = pad_index if '\<pad\>' is in specials but '\<unk\>' is not.
Of course, it is a optional method. I do this because the tensor of '\<pad\>' is zero and that of '\<unk\>' is also zero.

However, the new Vocab can not be serialized because I use a lambda function.

    def test_vocab_without_unk(self):
        c = Counter({'hello': 4, 'world': 3, 'ᑌᑎIᑕOᗪᕮ_Tᕮ᙭T': 5, 'freq_too_low': 2})
        oov_word = 'OOVWORD'
        self.assertNotIn(oov_word, c)
    
        # tests for specials_first=True
        v_first = vocab.Vocab(c, min_freq=3, specials=['<pad>'], specials_first=True)
        expected_itos_first = ['<pad>', 'ᑌᑎIᑕOᗪᕮ_Tᕮ᙭T', 'hello', 'world']
        expected_stoi_first = {x: index for index, x in enumerate(expected_itos_first)}
        self.assertEqual(v_first.itos, expected_itos_first)
        self.assertEqual(dict(v_first.stoi), expected_stoi_first)
        self.assertNotIn(oov_word, v_first.itos)
        self.assertNotIn(oov_word, v_first.stoi)
    
        # tests for specials_first=False
        v_last = vocab.Vocab(c, min_freq=3, specials=['<pad>'], specials_first=False)
        expected_itos_last = ['ᑌᑎIᑕOᗪᕮ_Tᕮ᙭T', 'hello', 'world', '<pad>']
        expected_stoi_last = {x: index for index, x in enumerate(expected_itos_last)}
        self.assertEqual(v_last.itos, expected_itos_last)
        self.assertEqual(dict(v_last.stoi), expected_stoi_last)
        self.assertNotIn(oov_word, v_last.itos)
        self.assertNotIn(oov_word, v_last.stoi)
    
        # check if pad is mapped to the first index
        self.assertEqual(v_first.stoi['<pad>'], 0)
        # check if pad is mapped to the last index
        self.assertEqual(v_last.stoi['<pad>'], max(v_last.stoi.values()))
    
        # check if an oovword is not in vocab and a default unk_id is not assigned to it
>       self.assertRaises(KeyError, v_first.stoi.__getitem__, oov_word)
E       AssertionError: KeyError not raised

mttk:
@seanliu96 I'll try to get around the pickling error and then merge this. Thanks for the PR!

seanliu96:
@mttk Thanks for your efforts!

francoishernandez:
@seanliu96 we have some issues related to this at [OpenNMT-py](https://github.com/OpenNMT/OpenNMT-py).
Loading previously trained models breaks : http://forum.opennmt.net/t/issue-loading-pretrained-models-for-machine-translation/2982.
Any idea of a quick workaround?
Thanks!

mttk:
@francoishernandez the dirtiest fix is to change the error creating L115 of `vocab.py` to `if state.get("unk_index", None) is None:`. I can create a PR soonish.

PR #538, opened 2019-05-19
Modify download_from_url to retrieve stopped download

belerico:
Added the possibility to retrieve download of vectors if it was for some reason stopped


belerico:
I've add the possibility to retrieve download of vectors if it was for some reason stopped

Issue #513, opened 2019-03-18
How can I filter out too high frequency words


sheepover96:
Hi!

Now I want to filter out too frequency words when i call build_vocab() method like a following manner.
```
TEXT = Field(sequential=True, lower=True)
TEXT.build_vocab(lang, max_freq=40)
```
Is there any method to do this or do I need to implement it by my self?

zhangguanheng66:
For now, please implement it as needed. If we continuously receive such feature request, we will add it via a PR.

Nayef211:
I don't believe we plan to add any features about high frequency words (only low frequency words are supported for the time being). Closing this for now but will reopen if we continue getting requests for this feature.

PR #541, opened 2019-05-30
add max_freq argument to Vocab: filter out high frequency tokens

sheepover96:
Add max_freq argument to Vocab to filter out high frequency tokens.
Based on following  Issue https://github.com/pytorch/text/issues/513


zhangguanheng66:
Thanks. Looks good to me. Could you also add the unit tests coving max_freq, like 0.0, 1.0, 100, float('int').

zhangguanheng66:
@sheepover96 Could you rebase and resolve the conflicts to land? Thanks.

torchtext/vocab.py
```diff
@@ -33,15 +33,18 @@ class Vocab(object):
...
+                 vectors_cache=None, specials_first=True):
         """Create a Vocab object from a collections.Counter.
 
         Arguments:
             counter: collections.Counter object holding the frequencies of
                 each value found in the data.
             max_size: The maximum size of the vocabulary, or None for no
                 maximum. Default: None.
+            max_freq: The max frequency of tokens in the vocabulary.
+                      Values less than 1 will be set to 1. Default: inf (no filter).
```
zhangguanheng66: This may not be important. Should we also check max_freq larger than min_freq?

sheepover96: Thank you for reviewing!
In the case that min_freq is larger than max_freq, should we return some error?

zhangguanheng66: Definitely. It makes sense to return an error message.

sheepover96:
CI test for python2 seems to failed by package(nltk_data) download error.
Am I able to fix it?
https://travis-ci.org/pytorch/text/jobs/571905378

zhangguanheng66:
> CI test for python2 seems to failed by package(nltk_data) download error.
> Am I able to fix it?
> https://travis-ci.org/pytorch/text/jobs/571905378

Thanks. Please fix it in a separate PR so it's easy for the record.

Issue #491, opened 2019-01-16
`input, target = batch` syntax is not working if the batch is created by BPTTIterator


ikorolev93:
Minimal reproduction code:
```
import torchtext.datasets as datasets

train, valid, test = datasets.WikiText2.iters()
input, target = next(iter(train))
```

Expected: no errors
Actual result:
```
Traceback (most recent call last):
  File ".../main.py", line 4, in <module>
    input, target = next(iter(train))
  File ".../venv/lib/python3.7/site-packages/torchtext/data/batch.py", line 78, in __iter__
    yield self._get_field_values(self.input_fields)
AttributeError: 'Batch' object has no attribute 'input_fields'
```
---
Looking at the code, `BPTTIterator` calls `Batch.fromvars` that creates a `Batch`, but never defines the `input_fields` and `target_fields` (unless it's passed in `**kwargs`, which is not the case). Copying the appropriate initialization code from constructor, it starts working.
So either `Batch.fromvars` should calculate `input_fields` and `target_fields` from `dataset`, or `BPTTIterator` should pass them in `**kwargs`.

mttk:
Thanks for reporting, will fix this ASAP.

zhangguanheng66:
Hi @mttk just wondering if there are any progress for this issue. Thanks.

mttk:
Hey @zhangguanheng66, I've (re)identified the problem, and I have a quick and dirty solution. I'll open a PR right away (#542), but the language modelling pipeline needs a large restructure.

Nayef211:
We have migrated all datasets to torchdata datapipes and should not be seeing this issue any longer.

PR #542, opened 2019-05-30
Add input and output field lists to fromvars init

mttk:
Dirty fix for #491


zhangguanheng66:
@mttk Thanks for the PR. If you feel the language model pipeline needs a large reconstruction, we could plan it via conference calls.

Nayef211:
Closing since iterator no longer exists within torchtext

PR #546, opened 2019-05-31
fix epoch prop

vincentzlt:
fix iterator's epoch property.


Nayef211:
Closing since `Iterator` class no longer exists in torchtext.

PR #550, opened 2019-06-27
Supervised learning datasets

zhangguanheng66:
Add a few supervised learning datasets.
Add a tutorial to use the new dataset for sentiment analysis.

### DIFF
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -1,6 +1,8 @@
 import os
 import torchtext.data as data
 from torchtext.datasets import WikiText2, PennTreebank
+from torchtext.datasets import TextClassificationDataset, \
+    AG_NEWS
 
 from ..common.test_markers import slow
 from ..common.torchtext_test_case import TorchtextTestCase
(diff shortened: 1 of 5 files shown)


zhangguanheng66:
@facebook-github-bot retest this please

docs/tutorials/text_sentiment_ngrams.ipynb
```diff
@@ -0,0 +1,375 @@
+{
```
cpuhrsch: Can you split this out into a separate PR just to make this one smaller?

zhangguanheng66: NP

test/data/test_builtin_datasets.py
```diff
@@ -47,3 +49,25 @@ def test_penntreebank(self):
...
+        txt_cls = TextClassificationDataset("ag_news", root=datadir, ngrams=2)
+        train_iter, test_iter, valid_iter = txt_cls.iters(batch_size=128,
+                                                          device="cpu")
+        ag_news_cls = AG_NEWS(root=datadir, ngrams=3)
+        train_iter, test_iter, valid_iter = ag_news_cls.iters(batch_size=128,
+                                                              device="cpu")
+
+        # Delete the dataset after we're done to save disk space on CI
+        if os.environ.get("TRAVIS") == "true":
+            datafile = os.path.join(self.project_root,
```
cpuhrsch: Instead of this you should be able to create a temporary directory via https://docs.python.org/3/library/tempfile.html , which has all of this built in and is therefore safer.

torchtext/datasets/text_classification.py
```diff
@@ -0,0 +1,484 @@
...
+import re
+import torch
+from ..utils import download_and_extract_archive, makedir_exist_ok
+import random
+from .. import data
+from tqdm import tqdm
+from ..data import dataset
+
+
+class TextClassificationDataset(data.Dataset):
```
cpuhrsch: Is it possible to avoid creating this and have each dataset (e.g. AG_NEWS ) use only https://pytorch.org/docs/stable/data.html instead?

torchtext dataset abstractions need a lot of work and we should avoid the proliferation of torchtext-style datasets until we did a wider review.

This should be fine even if it means that we first release AG_NEWS, etc. with less features.

PR #553, opened 2019-06-28
Issue templates

zhangguanheng66:
Create some issue templates, similar to PyTorch, and help our users to bring up issue.


zhangguanheng66:
> I'm assuming this is mostly a copy-paste from pytorch, which is fine.
> 
> It'd be nice to have the collect_env.py function exposed within pytorch and used from there so we don't miss out on subtle bug fixes etc. - Let's mark that as a follow-up task?

Personally, I would like to leave the collect_env.py file in pytorch without carrying it. I just need to have the information of torchtext version.

PR #556, opened 2019-07-09
[WIP] add new APIs to build dataset

zhangguanheng66:
With some very basic APIs (download_from_url, extract_archive, build_dictionary, rationed_split, generate_ngrams), we hope to build new datasets directly from torch.utils.data.dataset. The existing abstract dataset class in TorchText (i.e. torchtext.data.Dataset) is too complicated. We hope to gain flexibility with this simplified pattern. A few supervised learning dataset will be implemented based on this new pattern. As more necessary APIs are added, we hope the new APIs allow a user to do everything they want to do.

Steps to build datasets with the new pattern:
- The download_from_url and extract_archive functions are used to download and extract the online raw data. 
- Some text preprocessors/processors (e.g. spaCy) are required to clean the data and tokenize the sentence. It should be noted that the text preprocessors vary with the actual datasets so users are expected to prepare the preprocessors/processors. A generic preprocessor API is not pursued here.
- The build_vocab function is to build vocab object based on the tokens lists.
- Use rationed_split function to split the initial data into train/test/valid sets.
- The train/test/valid sets generated from last step are passed to the generate_iterators function. Here, data are batched. The output (train/test/valid iterators) can be used to train/test models.

### DIFF
--- a/torchtext/data/dataset.py
+++ b/torchtext/data/dataset.py
@@ -323,8 +323,22 @@ def stratify(examples, strata_field):
 
 
 def rationed_split(examples, train_ratio, test_ratio, val_ratio, rnd):
-    # Create a random permutation of examples, then split them
-    # by ratio x length slices for each of the train/test/dev? splits
+    """Create a random permutation of examples, then split them by ratios
+
+    Arguments:
+        examples: a list of data
+        train_ratio, test_ratio, val_ratio: split fractions.
+        rnd: a random shuffler
+
+    Examples:
+        >>> examples = []
+        >>> train_ratio, test_ratio, val_ratio = 0.7, 0.2, 0.1
+        >>> rnd = torchtext.data.dataset.RandomShuffler(None)
+        >>> train_examples, test_examples, valid_examples = \
+                torchtext.data.dataset.rationed_split(examples, train_ratio,
+                                                      test_ratio, val_ratio,
+                                                      rnd)
+    """
     N = len(examples)
     randperm = rnd(range(N))
     train_len = int(round(train_ratio * N))
--- a/torchtext/data/iterator.py
+++ b/torchtext/data/iterator.py
@@ -252,6 +252,17 @@ class BucketIterator(Iterator):
                                 sort_within_batch=self.sort_within_batch)
 
 
+def generate_iterators(datasets, batch_sizes, **kwargs):
+    """Create Iterator objects for multiple splits of a dataset.
+    """
+    ret = []
+    for i in range(len(datasets)):
+        train = i == 0
+        ret.append(Iterator(
+            datasets[i], batch_size=batch_sizes[i], train=train, **kwargs))
+    return tuple(ret)
+
+
 def batch(data, batch_size, batch_size_fn=None):
     """Yield elements from data in chunks of batch_size."""
     if batch_size_fn is None:
--- a/torchtext/data/iterator.py
+++ b/torchtext/data/iterator.py
@@ -252,28 +252,6 @@ class BucketIterator(Iterator):
                                 sort_within_batch=self.sort_within_batch)
 
 
-def generate_iterators(datasets, batch_sizes, **kwargs):
-    """Create Iterator objects for multiple splits of a dataset.
-
-    Arguments:
-        datasets: a tuple of Datasets
-        batch_sizes: a list of batch_size, corresponding to each dataset.
-        Remaining keyword arguments: Passed to the constructor of Iterator.
-
-    Examples:
-        >>> batch_size = 64
-        >>> train_iter, test_iter, valid_iter = generate_iterators(
-            (train_dataset, test_dataset, valid_dataset), [batch_size] * 3,
-            device="cpu")
-    """
-    ret = []
-    for i in range(len(datasets)):
-        train = i == 0
-        ret.append(Iterator(
-            datasets[i], batch_size=batch_sizes[i], train=train, **kwargs))
-    return tuple(ret)
-
-
 def batch(data, batch_size, batch_size_fn=None):
     """Yield elements from data in chunks of batch_size."""
     if batch_size_fn is None:
--- a/torchtext/data/iterator.py
+++ b/torchtext/data/iterator.py
@@ -254,6 +254,17 @@ class BucketIterator(Iterator):
 
 def generate_iterators(datasets, batch_sizes, **kwargs):
     """Create Iterator objects for multiple splits of a dataset.
+
+    Arguments:
+        datasets: a tuple of Datasets
+        batch_sizes: a list of batch_size, corresponding to each dataset.
+        Remaining keyword arguments: Passed to the constructor of Iterator.
+
+    Examples:
+        >>> batch_size = 64
+        >>> train_iter, test_iter, valid_iter = generate_iterators(
+            (train_dataset, test_dataset, valid_dataset), [batch_size] * 3,
+            device="cpu")
     """
     ret = []
     for i in range(len(datasets)):
--- a/torchtext/utils.py
+++ b/torchtext/utils.py
@@ -3,7 +3,6 @@ import requests
 import csv
 from tqdm import tqdm
 import os
-import errno
 import tarfile
 
 
@@ -28,7 +27,18 @@ def reporthook(t):
 
 
 def download_from_url(url, path):
-    """Download file, with logic (from tensor2tensor) for Google Drive"""
+    """Download file, with logic (from tensor2tensor) for Google Drive
+
+    Arguments:
+        url: the url for online Dataset
+        path: directory and filename for the downloaded dataset.
+
+    Examples:
+        >>> url = 'http://www.quest.dcs.shef.ac.uk/wmt16_files_mmt/validation.tar.gz'
+        >>> path = './validation.tar.gz'
+        >>> torchtext.utils.download_from_url(url, path)
+    """
+
     def process_response(r):
         chunk_size = 16 * 1024
         total_size = int(r.headers.get('Content-length', 0))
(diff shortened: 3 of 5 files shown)


torchtext/data/iterator.py
```diff
@@ -252,6 +252,29 @@ def create_batches(self):
                                 sort_within_batch=self.sort_within_batch)
 
 
+def generate_iterators(datasets, batch_sizes, **kwargs):
+    """Create Iterator objects for multiple splits of a dataset.
+
+    Arguments:
+        datasets: a tuple of Datasets
+        batch_sizes: a list of batch_size, corresponding to each dataset.
+        ngrams: the number of ngrams.
```
cpuhrsch: This is specific to text, but the function has potential for being generic by itself. Is there a way to split out this "ngrams" concept?

zhangguanheng66: It seems a typo. I don't think iterator function needs ngrams.

torchtext/utils.py
```diff
@@ -25,7 +27,18 @@ def inner(b=1, bsize=1, tsize=None):
 
 
 def download_from_url(url, path):
-    """Download file, with logic (from tensor2tensor) for Google Drive"""
+    """Download file, with logic (from tensor2tensor) for Google Drive
```
cpuhrsch: We could take this as a chance to move this stuff into pytorch core and then use it here.

zhangguanheng66: yeah. As long as they are not text specific (I think most of them are not), we can move them to the core library.

text_classification.py
```diff
@@ -0,0 +1,487 @@
...
+        self.fields = fields
+
+    def __getitem__(self, i):
+        return self.examples[i]
+
+    def __len__(self):
+        try:
+            return len(self.examples)
+        except TypeError:
+            return 2**32
```
cpuhrsch: Does this mean self.examples could be something else than a list? Also, why call it examples, isn't it the entire corpus?

text_classification.py
```diff
@@ -0,0 +1,487 @@
...
+                'float' token will be used.
+            ngrams: a contiguous sequence of n items from s string text.
+                Default: 1
+
+        Examples:
+            >>> txt_cls = TextClassificationDataset("ag_news", ngrams=2)
+
+        """
+        self.dataset_name = dataset_name
+        if dataset_name in supported_text_classification_datasets.keys():
```
cpuhrsch: You have the class instances below (e.g. AG_NEWS), so you use this to deduplicate some code I think.
 The main reason for that seems to be that you're using "load_text_classification_data" and have a predefined split.

text_classification.py
```diff
@@ -0,0 +1,487 @@
...
+            - valid_iter: a iterator based on
+                train.csv file with 1-split_ratio percent.
+
+        """
+
+        train, test, valid = self.split(split_ratio, random_state)
+        return generate_iterators(
+            (train, test, valid), [batch_size] * 3, device=device)
+
+    def download(self):
```
cpuhrsch: Can we move this out of TextClassificationDataset?

text_classification.py
```diff
@@ -0,0 +1,487 @@
...
+        os.makedirs(self.raw_folder)
+        filename = self.dataset_name + '_csv.tar.gz'
+        url = self.url
+        path = os.path.join(self.raw_folder, filename)
+        download_from_url(url, path)
+        extract_archive(path, self.raw_folder, remove_finished=True)
+
+        print('Dataset %s downloaded.' % self.dataset_name)
+
+    def preprocess(self, raw_folder=None, processed_folder=None):
```
cpuhrsch: This seems useful in its own right, i.e. it applies a preprocessing function to a set of text files. Can we move it out?

text_classification.py
```diff
@@ -0,0 +1,487 @@
...
+
+        src_filepath = os.path.join(raw_folder, 'test.csv')
+        if not os.path.isfile(src_filepath):
+            self.download()
+        tgt_filepath = os.path.join(processed_folder, self.dataset_name + '.test')
+        self.text_normalize(src_filepath, tgt_filepath)
+
+        print("Dataset %s preprocessed." % self.dataset_name)
+
+    def text_normalize(self, src_filepath, tgt_filepath):
```
cpuhrsch: Can we move this out of the class and make it a general function? This seems useful by itself. We just need to add a bit of an explanation I think (i.e. what kind of text this is useful for).

text_classification.py
```diff
@@ -0,0 +1,487 @@
...
+             - YelpReviewPolarity
+             - YelpReviewFull
+             - YahooAnswers
+             - AmazonReviewPolarity
+             - AmazonReviewFull
+
+    """
+
+    def __init__(self, dataset_name, root='.data',
+                 text_field=None, label_field=None, ngrams=1):
```
cpuhrsch: The number of ngrams is passed in as a parameter for load_text_classification_data, which in turn calls into label_text_processor and returns the preprocessed "examples" (corpus). build_vocab then takes an instance of self.  I agree that using the iterator approach is more efficient, but in this instance it is iterating over a corpus that's already entirely in memory. So, when build the vocabulary it should be enough to only pass in self.examples. Then build_vocab could stand entirely by itself (maybe even called "build_vocab_from_list") .

text_classification.py
```diff
@@ -0,0 +1,487 @@
...
+                                             'processed')
+
+        filepath = os.path.join(self.processed_folder, self.dataset_name + '.train')
+        self.train_examples = self.load_text_classification_data(filepath, ngrams)
+
+        filepath = os.path.join(self.processed_folder, self.dataset_name + '.test')
+        self.test_examples = self.load_text_classification_data(filepath, ngrams)
+        self.examples = self.train_examples + self.test_examples
+
+        self.fields['text'].vocab = build_vocab(self, self.fields['text'], 'text')
```
cpuhrsch: I think we can call into this using self.examples. Then build_vocab can be more generic . This however also requires us to get rid of "fields".

text_classification.py
zhangguanheng66: This is the data loaded from train/test files. It covers the entire corpus in this case.

text_classification.py
zhangguanheng66: This interface is old, and not generic. The major function components in this class should be "download", "preprocess", "load token list", "general iterator/batch". That's all.

text_classification.py
zhangguanheng66: I will try to improve the build_vocab API. For now, I think two separate components, data (e.g. examples) and token processor (e.g. field) make sense.

text_classification.py
zhangguanheng66: yeap, download function is now independent and can stay outside.

text_classification.py
zhangguanheng66: Similar to download, we don't need those wrappers inside the class.

text_classification.py
zhangguanheng66: Same as above

zhangguanheng66:
Split the supervised learning dataset into a new PR. https://github.com/pytorch/text/pull/557
Land this one first for the necessary APIs.

torchtext/vocab.py
```diff
@@ -216,6 +218,33 @@ def set_vectors(self, stoi, vectors, dim, unk_init=torch.Tensor.zero_):
...
+        if not field.sequential:
+            x = [x]
+        try:
+            counter.update(x)
+        except TypeError:
+            counter.update(chain.from_iterable(x.text))
+    specials = list(OrderedDict.fromkeys(
+        tok for tok in [field.unk_token, field.pad_token, field.init_token,
+                        field.eos_token] if tok is not None))
+    return Vocab(counter, specials=specials, **kwargs)
```
cpuhrsch: Vocab does a lot of things that seem unrelated to dictionaries, for example "load_vectors" and "set_vectors", "__eq__" etc.  - maybe we should audit it and see if we can remove a few things for 0.4?

Vocab is also undocumented and not referenced by anything, even though it seems pretty useful.

zhangguanheng66: Let me see if I can make some changes there. I cannot estimate the efforts to split vacab and vector etc.

torchtext/utils.py
```diff
@@ -75,3 +88,31 @@ def unicode_csv_reader(unicode_csv_data, **kwargs):
 def utf_8_encoder(unicode_csv_data):
     for line in unicode_csv_data:
         yield line.encode('utf-8')
+
+
+def extract_archive(from_path, to_path=None, remove_finished=False):
```
cpuhrsch: Add a flag like "format=tar" to indicate it's using tar so we can extend it later on. This is also very useful in general.

torchtext/data/iterator.py
```diff
@@ -252,6 +252,28 @@ def create_batches(self):
...
+    Examples:
+        >>> batch_size = 64
+        >>> train_iter, test_iter, valid_iter = generate_iterators(
+            (train_dataset, test_dataset, valid_dataset), [batch_size] * 3,
+            device="cpu")
+    """
+    ret = []
+    for i in range(len(datasets)):
+        train = i == 0
+        ret.append(Iterator(
```
cpuhrsch: It would be nice if we didn't need to use any existing legacy stuff since we're planning to remove / revamp it

torchtext/data/iterator.py
```diff
@@ -252,6 +252,28 @@ def create_batches(self):
                                 sort_within_batch=self.sort_within_batch)
 
 
+def generate_iterators(datasets, batch_sizes, **kwargs):
```
cpuhrsch: Let's write model without batching first, because batching is a perf. optimization and we haven't measured it yet.

torchtext/vocab.py
cpuhrsch: It's ok if we build a very minimal and advertised Dictionary instead of replicating all features of Vocab. That will also give us the chance to make it fast.

torchtext/utils.py
zhangguanheng66: We can simply extend it by adding the other branches between "if" and "else"

torchtext/data/iterator.py
zhangguanheng66: Then, I will move this to tutorial.

torchtext/vocab.py
```diff
@@ -216,6 +218,34 @@ def set_vectors(self, stoi, vectors, dim, unk_init=torch.Tensor.zero_):
                 self.vectors[i] = unk_init(self.vectors[i])
 
 
+def build_dictionary(dataset, field, data_name, **kwargs):
+    """Construct the Vocab object for the field from a dataset.
```
cpuhrsch: Still need to update the docs :)

torchtext/data/dataset.py
```diff
@@ -323,8 +323,22 @@ def stratify(examples, strata_field):
 
 
 def rationed_split(examples, train_ratio, test_ratio, val_ratio, rnd):
-    # Create a random permutation of examples, then split them
-    # by ratio x length slices for each of the train/test/dev? splits
+    """Create a random permutation of examples, then split them by ratios
```
fmassa: This looks a lot like [`torch.utils.data.random_split`](https://github.com/pytorch/pytorch/blob/f622ec8084c1778c6943474796efa52e9b9cf695/torch/utils/data/dataset.py#L262), but where the splits don't need necessarily to be random.

torchtext/data/dataset.py
```diff
@@ -323,8 +323,22 @@ def stratify(examples, strata_field):
...
+
+    Arguments:
+        examples: a list of data
+        train_ratio, test_ratio, val_ratio: split fractions.
+        rnd: a random shuffler
+
+    Examples:
+        >>> examples = []
+        >>> train_ratio, test_ratio, val_ratio = 0.7, 0.2, 0.1
+        >>> rnd = torchtext.data.dataset.RandomShuffler(None)
```
fmassa: Why do we re-create a `RandomShuffler` in `torchtext`, what is the use-case, instead of using `torch.utils.data.RandomShuffler`?

PR #557, opened 2019-07-11
[WIP] Add new datasets for text classification.

zhangguanheng66:
Add a few supervised learning datasets, including
             - AG_NEWS
             - SogouNews
             - DBpedia
             - YelpReviewPolarity
             - YelpReviewFull
             - YahooAnswers
             - AmazonReviewPolarity
             - AmazonReviewFull

### DIFF
--- a/torchtext/datasets/text_classification.py
+++ b/torchtext/datasets/text_classification.py
@@ -9,7 +9,7 @@ import random
 from tqdm import tqdm
 
 
-def download(url, raw_folder, dataset_name):
+def download_extract_archive(url, raw_folder, dataset_name):
     """Download the dataset if it doesn't exist in processed_folder already."""
 
     train_csv_path = os.path.join(raw_folder,
@@ -149,7 +149,7 @@ class TextClassificationDataset(torch.utils.data.Dataset):
                                              'processed')
         filepath = os.path.join(self.processed_folder, self.dataset_name + '.train')
         if not os.path.isfile(filepath):
-            download(self.url, self.raw_folder, self.dataset_name)
+            download_extract_archive(self.url, self.raw_folder, self.dataset_name)
             _preprocess(self.raw_folder, self.processed_folder, self.dataset_name)
         with open(filepath) as src_data:
             self.train_examples = _load_text_classification_data(src_data, self.fields, ngrams)
--- a/torchtext/datasets/text_classification.py
+++ b/torchtext/datasets/text_classification.py
@@ -96,7 +96,7 @@ def _load_text_classification_data(src_data, fields, ngrams=1):
 
 
 class TextClassificationDataset(torch.utils.data.Dataset):
-    """Defines text classification datasets.
+    """Defines an abstract text classification datasets.
         Currently, we only support the following datasets:
 
              - AG_NEWS
@@ -112,7 +112,7 @@ class TextClassificationDataset(torch.utils.data.Dataset):
 
     def __init__(self, url, root='.data',
                  text_field=None, label_field=None, ngrams=1):
-        """Create a text classification dataset instance.
+        """Initiate text-classification dataset.
 
         Arguments:
             url: url of the online raw data files.
@@ -123,11 +123,6 @@ class TextClassificationDataset(torch.utils.data.Dataset):
                 'float' token will be used.
             ngrams: a contiguous sequence of n items from s string text.
                 Default: 1
-
-        Examples:
-            >>> url = 'https://drive.google.com/uc?export=download&id=0Bz8a_Dbh9QhbUDNpeUdjb0wxRms'
-            >>> txt_cls = TextClassificationDataset(url, ngrams=2)
-
         """
 
         super(TextClassificationDataset, self).__init__()
@@ -190,8 +185,14 @@ class AG_NEWS(TextClassificationDataset):
     def __init__(self, root='.data', text_field=None, label_field=None, ngrams=1):
         """Create supervised learning dataset: AG_NEWS
 
-        Inputs:
-            See TextClassificationDataset() class
+        Arguments:
+            root: Directory where the dataset are saved. Default: ".data"
+            text_field: The field that will be used for the sentence. If not given,
+                'spacy' token will be used.
+            label_field: The field that will be used for the label. If not given,
+                'float' token will be used.
+            ngrams: a contiguous sequence of n items from s string text.
+                Default: 1
 
         Examples:
             >>> text_cls = torchtext.datasets.AG_NEWS(ngrams=3)
@@ -215,8 +216,14 @@ class SogouNews(TextClassificationDataset):
                  label_field=None, ngrams=1):
         """Create supervised learning dataset: SogouNews
 
-        Inputs:
-            See TextClassificationDataset() class
+        Arguments:
+            root: Directory where the dataset are saved. Default: ".data"
+            text_field: The field that will be used for the sentence. If not given,
+                'spacy' token will be used.
+            label_field: The field that will be used for the label. If not given,
+                'float' token will be used.
+            ngrams: a contiguous sequence of n items from s string text.
+                Default: 1
 
         Examples:
             >>> text_cls = torchtext.datasets.SogouNews(ngrams=3)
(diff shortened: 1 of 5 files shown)


torchtext/datasets/text_classification.py
```diff
@@ -0,0 +1,410 @@
...
+    download_from_url(url, path)
+    extract_archive(path, raw_folder, remove_finished=True)
+
+    print('Dataset %s downloaded.' % dataset_name)
+
+
+def preprocess(raw_folder, processed_folder, dataset_name):
+    """Preprocess the csv files."""
+
+    def text_normalize(src_filepath, tgt_filepath):
```
cpuhrsch: I think you can split this out as well. It seems useful in its own right. But instead of taking a filepath, it can take a string and return a string (or a list of strings).

torchtext/datasets/text_classification.py
```diff
@@ -0,0 +1,410 @@
...
+    test = TextDataset(test_examples, fields)
+    test.sort_key = sort_key
+    valid = TextDataset(valid_examples, fields)
+    valid.sort_key = sort_key
+
+    return generate_iterators(
+        (train, test, valid), [batch_size] * 3, device=device)
+
+
+class TextDataset(torch.utils.data.Dataset):
```
cpuhrsch: Do we still need this?

zhangguanheng66: I actually include it here as a new abstract class. Later on, I can simply merge it into TextClassification Dataset.

torchtext/datasets/text_classification.py
zhangguanheng66: Sure thing. Will do it.

torchtext/datasets/text_classification.py
```diff
@@ -0,0 +1,416 @@
...
+    line = line.lower()
+    label, text = line.split(",", 1)
+    label = "__label__" + re.sub(r'[^0-9\s]', '', label)
+    text = re.sub(r'[^a-zA-Z0-9\s]', ' ', text)
+    text = ' '.join(text.split())
+    line = label + ' , ' + text + ' \n'
+    return line
+
+
+def preprocess(raw_folder, processed_folder, dataset_name):
```
cpuhrsch:
```
for filename in os.listdir('dirname):
    with open(filename, 'r') as a, open(filename + '.processed', 'w') as b:
        b.write(text_normalize(a.readline())
```

cpuhrsch: I'd delete this and expect the user to know how to do this

cpuhrsch: Maybe make it an example / add it to the tutorial

torchtext/datasets/text_classification.py
```diff
@@ -0,0 +1,416 @@
...
+        with open(tgt_filepath, 'w') as new_data:
+            for line in src_data:
+                lines.append(text_normalize(line))
+            random.shuffle(lines)
+            new_data.writelines(lines)
+
+    print("Dataset %s preprocessed." % dataset_name)
+
+
+def load_text_classification_data(filepath, fields, ngrams=1):
```
cpuhrsch: I'd add an underscore ("_load_text[...]"), because this function is very specific to these datasets and as such a helper.

torchtext/datasets/text_classification.py
```diff
@@ -0,0 +1,416 @@
...
+    filename = dataset_name + '_csv.tar.gz'
+    url = url
+    path = os.path.join(raw_folder, filename)
+    download_from_url(url, path)
+    extract_archive(path, raw_folder, remove_finished=True)
+
+    print('Dataset %s downloaded.' % dataset_name)
+
+
+def text_normalize(line):
```
cpuhrsch: Maybe "basic_normalization_english" or such? Is there a reference here (maybe within a paper)? Can be immediately moved into c++. It's very useful in general and we can indicate that.

torchtext/datasets/text_classification.py
```diff
@@ -0,0 +1,416 @@
...
+from torchtext.data.utils import generate_ngrams
+from torchtext.vocab import build_vocab
+import random
+from torchtext import data
+from tqdm import tqdm
+from torchtext.data import dataset
+from torchtext.data.iterator import generate_iterators
+
+
+def download(url, raw_folder, dataset_name):
```
cpuhrsch:
```
extract_archive(
    download_from_url(
        torchtext.datasets.supervised.urls['AG_NEWS']) #returns a path in cache, 
        "path/to/AG_NEWS/folder"
    )
)
```

torchtext/datasets/text_classification.py
```diff
@@ -0,0 +1,416 @@
...
+        return
+
+    os.makedirs(raw_folder)
+    filename = dataset_name + '_csv.tar.gz'
+    url = url
+    path = os.path.join(raw_folder, filename)
+    download_from_url(url, path)
+    extract_archive(path, raw_folder, remove_finished=True)
+
+    print('Dataset %s downloaded.' % dataset_name)
```
cpuhrsch: Use the logging package for logging because it allows the user to set verbosity (this is an info line) https://docs.python.org/3/library/logging.html

torchtext/datasets/text_classification.py
cpuhrsch: I don't this necessary needs to be its own function

torchtext/datasets/text_classification.py
```diff
@@ -0,0 +1,416 @@
...
+        return ex
+
+    examples = []
+    with open(filepath) as src_data:
+        for line in tqdm(src_data):
+            examples.append(label_text_processor(line, fields, ngrams))
+    return examples
+
+
+def iters(train_examples, test_examples, fields, sort_key,
```
cpuhrsch: Let's add batching separately if we see that we need it.

torchtext/datasets/text_classification.py
```diff
@@ -0,0 +1,416 @@
...
+        self.root = root
+        self.raw_folder = os.path.join(self.root, self.__class__.__name__, 'raw')
+        self.processed_folder = os.path.join(self.root,
+                                             self.__class__.__name__,
+                                             'processed')
+        filepath = os.path.join(self.processed_folder, self.dataset_name + '.train')
+        if not os.path.isfile(filepath):
+            download(self.url, self.raw_folder, self.dataset_name)
+            preprocess(self.raw_folder, self.processed_folder, self.dataset_name)
+        self.train_examples = load_text_classification_data(filepath, self.fields, ngrams)
```
cpuhrsch: Since it's all being read into memory I'd expect the user to just pass in this already from memory. So this doesn't need to take a path, but can take two lists. We can then add support for paths later on.

torchtext/datasets/text_classification.py
```diff
@@ -0,0 +1,416 @@
...
+        """Create supervised learning dataset: AG_NEWS
+
+        Inputs:
+            See TextClassificationDataset() class
+
+        Examples:
+            >>> text_cls = torchtext.datasets.AG_NEWS(ngrams=3)
+
+        """
+        self.url = 'https://drive.google.com/uc?export=download&id=0Bz8a_Dbh9QhbUDNpeUdjb0wxRms'
```
cpuhrsch: Can you expose this as part of a static list? I think this is something that could be useful by itself. The user can then also use it as a reference for other programs via something like torchaudio.datasets.supervised.URLS['AG_NEWS']

zhangguanheng66: Sure. I haven't updated the APIs to address your comments above. Will do them all together.

torchtext/datasets/text_classification.py
zhangguanheng66: the supervised learning paper didn't mention how they did pre-processing. fastText just did this pre-processing in the bash script.

torchtext/datasets/text_classification.py
zhangguanheng66: I would like to support the one-line data loading for the supervised learning dataset. I think we can just leave this function here for these supervised learning datasets.

torchtext/datasets/text_classification.py
zhangguanheng66: Fixed.

torchtext/datasets/text_classification.py
zhangguanheng66: Fixed

torchtext/datasets/text_classification.py
zhangguanheng66: fixed

torchtext/datasets/text_classification.py
zhangguanheng66: minor change. Put download and extract_archive together.

PR #564, opened 2019-07-20
C++ Vocabulary

ShahriarSS:
Hi @zhangguanheng66 Here is the vocabulary.


torchtext/csrc/core/vocabulary.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <stdint.h>
+#include <string>
+#include <vector>
+
+namespace torchtext {
+namespace core {
+class Vocabulary {
+  static const size_t MAX_VOCAB_SIZE = 30000000;
+  static const size_t negative = static_cast<size_t>(-1);
```
cpuhrsch: Why is this necessary?

torchtext/csrc/core/vocabulary.h
```diff
@@ -0,0 +1,36 @@
+#ifndef VOCABULARY_H
+#define VOCABULARY_H
+
+#include <stdint.h>
```
cpuhrsch: Why not use the cpp compatibility header https://en.cppreference.com/w/cpp/header/cstdint

torchtext/csrc/core/vocabulary.cpp
```diff
@@ -0,0 +1,37 @@
+#include "vocabulary.h"
+#include <iostream>
+
+namespace torchtext {
+namespace core {
+size_t Vocabulary::word_hash(const std::string& word) {
+  return std::hash<std::string>()(word);
```
cpuhrsch: Do you know what hash function this uses? I'd prefer https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function

setup.py
```diff
@@ -25,6 +32,55 @@ def find_version(*file_paths):
 VERSION = find_version('torchtext', '__init__.py')
 long_description = read('README.rst')
 
+
+def get_extensions():
```
cpuhrsch: I assume this is copy paste from torchvision?

.clang-format
```diff
@@ -0,0 +1,88 @@
+---
```
cpuhrsch: I'd want to add this in a separate PR.

cpuhrsch:
Could you rename this from Vocabulary to Dictionary please :)

torchtext/csrc/core/vocabulary.h
ShahriarSS: Do you mean negative? I'm using ```size_t``` for indexes which is unsigned and indexes must initially be negative(-1). So this is represents negative in unsigned integer (std::string uses the same method for negative).

torchtext/csrc/core/vocabulary.cpp
ShahriarSS: No I don't but it can't be a bad one. Can it? I'll look into Fowler hash function.

setup.py
ShahriarSS: Yes. I'm not an expert in Python so I copied and changed to fit.

.clang-format
ShahriarSS: OK.

torchtext/csrc/core/vocabulary.cpp
ShahriarSS: @cpuhrsch I looked into it and apparently GCC is using MurmurHashUnaligned2 (see [this](https://stackoverflow.com/questions/19411742/what-is-the-default-hash-function-used-in-c-stdunordered-map)). Maybe it would be better if we implemented the hash function ourselves to be sure that it will perform the same everywhere.

cpuhrsch: Yes ideally we implement it ourselves, because this hash function also has particular properties for text.

ShahriarSS: I did implement it.

torchtext/csrc/core/dictionary.cpp
```diff
@@ -0,0 +1,45 @@
...
+  auto words_size = words.size();
+  auto index = word_hash(word) % words_size;
+
+  while (words[index].index != negative && words[index].word != word)
+    index = (index + 1) & words_size;
+
+  return index;
+}
+
+size_t Dictionary::add_word(const std::string& word) {
```
cpuhrsch: After adding words, at some point, the user will ideally want to sort the dataset based on the frequency. The most frequent word should be the index 0, the second most frequent index 1, etc. - This is a memory optimization. Words follow [Zipf's law](https://en.wikipedia.org/wiki/Zipf%27s_law). When you read in a cacheline of data from this lookup you're most likely going to read index 0. Since you always read 64 bytes, that'll result in certain number of the first few words to always be read. The probability of hitting an index close to 0 is very high due to Zipf's law, so on the second read you'll likely be able to use the data you just read into cache. If the indicies on the other hand are random you'll constantly trash cache for a single word and read much more than you need to.

On that note, using std::string is probably not a good idea, because you need tight control over the structure that's used to represent the word. We can however add these optimizations later on once you have a benchmark (please add that as well) and can look at the memory pressure of a typical read. A good benchmark is tokenizing and creating a dictionary based on Wikipedia. At some point we might also want to align an entry struct or use SmallVector to store the first few characters on the stack.

erip:
I think this is safe to close since we have a pybind and torchbind version of Vocab, cc @parmeet

PR #565, opened 2019-07-24
Refactor text_classification, improve extract_archive and add a small example

cpuhrsch:


torchtext/datasets/text_classification.py
```diff
@@ -110,73 +112,47 @@ class TextClassificationDataset(torch.utils.data.Dataset):
 
     """
 
-    def __init__(self, url, root='.data',
-                 text_field=None, label_field=None, ngrams=1):
+    def __init__(self, root, ngrams):
         """Initiate text-classification dataset.
 
         Arguments:
             url: url of the online raw data files.
```
zhangguanheng66: We don't need this url in the docs.

PR #568, opened 2019-07-26
Update text_normalize() function in text_classification.

zhangguanheng66:
As the title says.


torchtext/datasets/text_classification.py
```diff
@@ -27,27 +27,36 @@
...
+                   ' ']
 
 
 def text_normalize(line):
     """
     Basic normalization for a line of text.
     Normalization includes
     - lowercasing
-    - replacing all non-alphanumeric characters with whitespace
+    - complete some basic text normalization for English words.
```
cpuhrsch: I think it's a good idea to explicitly list the normalizations that are applied so that the user knows exactly what's going to happen without having to look at the code.

test/data/test_utils.py
```diff
@@ -45,3 +46,11 @@ def test_get_tokenizer_toktokt(self):
             data.get_tokenizer(1)
         with self.assertRaises(ValueError):
             data.get_tokenizer("some other string")
+
+    def test_text_nomalize_function(self):
```
cpuhrsch: You should make this more rigorous by implementing the bash function and running it across all of AG_NEWS. I think it's worth doing that, because it's easy to miss something. In the future we might update this test as the basic normalization function becomes more sophisticated.

torchtext/datasets/text_classification.py
```diff
@@ -27,26 +27,38 @@
...
 def text_normalize(line):
     """
     Basic normalization for a line of text.
     Normalization includes
     - lowercasing
-    - replacing all non-alphanumeric characters with whitespace
+    - complete some basic text normalization for English words as follows:
+        add spaces before and after '\'', '.', ',', '(', ')', '!', , '?',
+        remove '\"',
+        replace r'<br \/>', ';', ':', multiple spaces with single space
```
cpuhrsch: nit: I think there's an extra "r"

I also think it'd be more readable (albeit more verbose) if this was split up into separate lines

test/data/test_utils.py
```diff
@@ -45,3 +46,19 @@ def test_get_tokenizer_toktokt(self):
...
+        ref_lines = []
+        test_lines = []
+        with open('../asset/test.csv') as src_data:
+            for line in src_data:
+                test_lines.append(text_normalize(line))
+
+        with open('../asset/ag_news.test') as ref_data:
+            for line in ref_data:
+                line = line.split()
+                line[0] = line[0][9:]  # remove '__label__'
```
cpuhrsch: Can you add an assert that you indeed only remove '__label__'?

test/data/test_utils.py
```diff
@@ -45,3 +46,19 @@ def test_get_tokenizer_toktokt(self):
             data.get_tokenizer(1)
         with self.assertRaises(ValueError):
             data.get_tokenizer("some other string")
+
+    def test_text_nomalize_function(self):
+        # Test text_nomalize function in torchtext.datasets.text_classification
+        ref_lines = []
+        test_lines = []
+        with open('../asset/test.csv') as src_data:
```
cpuhrsch: can you name this something more expressive? Maybe "text_normalization_ag_news_test.csv" or such. Or move it into a different folder "reference" and the .test dataset into "data".

test/data/test_utils.py
```diff
@@ -45,3 +46,19 @@ def test_get_tokenizer_toktokt(self):
...
+            for line in src_data:
+                test_lines.append(text_normalize(line))
+
+        with open('../asset/ag_news.test') as ref_data:
```
cpuhrsch: Please use the csv reader abstraction for this, it'll be more robust and aligned with the rest of the code

```
    from torchtext.utils import unicode_csv_reader

    with io.open(data_path, encoding="utf8") as f:
        reader = unicode_csv_reader(f)
        for row in reader:
```

test/data/test_utils.py
zhangguanheng66: let me rename it.

torchtext/datasets/text_classification.py
```diff
@@ -27,26 +27,46 @@
...
+        add spaces before and after '\''
+        remove '\"',
+        add spaces before and after '.'
+        replace '<br \/>'with single space
+        add spaces before and after ','
+        add spaces before and after '('
+        add spaces before and after ')'
+        add spaces before and after '!'
+        add spaces before and after '?'
+        replace ';' multiple spaces with single space
```
cpuhrsch: What does this mean?

zhangguanheng66: typo. replace multiple spaces with single space

cpuhrsch: But doesn't that only happen in the end? It looks like the corresponding re only replaces ';' with a single space

test/data/test_utils.py
```diff
@@ -45,3 +47,22 @@ def test_get_tokenizer_toktokt(self):
...
+        ref_lines = []
+        test_lines = []
+
+        data_path = 'test/asset/text_normalization_ag_news_test.csv'
+        with io.open(data_path, encoding="utf8") as f:
+            reader = unicode_csv_reader(f)
+            for row in reader:
+                test_lines.append(text_normalize(' , '.join(row)))
+
+        with open('../asset/text_normalization_ag_news_ref_results.test') as ref_data:
```
cpuhrsch: Please use io.open here as well, because the reference is presumably encoded in utf-8

test/data/test_utils.py
```diff
@@ -45,3 +47,22 @@ def test_get_tokenizer_toktokt(self):
...
+            for line in ref_data:
+                line = line.split()
+                line[0] = line[0][9:]  # remove '__label__'
+                ref_lines.append(line)
+
+        assert ref_lines == test_lines
```
cpuhrsch: Please use self.assertEqual because it'll show up as an error in the tests instead of a separate assert

torchtext/datasets/text_classification.py
zhangguanheng66: it replaces ';' with a single space. At the very end, multiple spaces will be combined together.

PR #569, opened 2019-07-26
Improved download and extract functions plus download scripts

cpuhrsch:


torchtext/data/utils.py
```diff
@@ -13,11 +14,57 @@ def _spacy_tokenize(x, spacy):
...
+    """
+
+    line = line.lower()
+    for pattern_re, replaced_str in zip(_normalize_pattern_re, replaced_string):
+        line = pattern_re.sub(replaced_str, line)
+    return line.split()
+
+
 def get_tokenizer(tokenizer, language='en'):
     # default tokenizer is string.split(), added as a module function for serialization
```
zhangguanheng66: We need docs for this part (you can leave it to me) since this function will be called by users.

torchtext/datasets/text_classification.py
```diff
@@ -27,55 +28,13 @@
...
-    Returns a list of tokens after splitting on whitespace.
-    """
-
-    line = line.lower()
-    for pattern_re, replaced_str in zip(_normalize_pattern_re, replaced_string):
-        line = pattern_re.sub(replaced_str, line)
-    return line.split()
-
-
 def _csv_iterator(data_path, ngrams, yield_cls=False):
```
zhangguanheng66: There is another one in text_classification/vocab.py.

torchtext/utils.py
```diff
@@ -73,7 +91,7 @@ def process_response(r):
         url = url + "&confirm=" + confirm_token
         response = session.get(url, stream=True)
 
-    process_response(response)
+    return process_response(response, root, filename)
 
 
 def unicode_csv_reader(unicode_csv_data, **kwargs):
```
zhangguanheng66: We need an example/docs here. Happy to add.

examples/text_classification/download.py
```diff
@@ -0,0 +1,23 @@
+import logging
```
zhangguanheng66: Is this an example to show download function? I feel it is not relevant to text_classification things. Better to move to torch.utils docs. Or you could even put it as a separate example under examples directory

zhangguanheng66: Since download/extract_archive functions are more orthogonal now, we could add this file under examples directory. They are not so relevant to text_classification things.

examples/text_classification/vocab.py
```diff
@@ -0,0 +1,39 @@
+import logging
```
zhangguanheng66: same as above. this example is to show how to build vocab so it should be under examples directory.

torchtext/data/utils.py
```diff
@@ -13,11 +14,57 @@ def _spacy_tokenize(x, spacy):
...
+
+replaced_string = [' \'  ', '',
+                   ' . ', ' ',
+                   ' , ', ' ( ',
+                   ' ) ', ' ! ',
+                   ' ? ', ' ',
+                   ' ', ' ']
+
+
+def _basic_english_normalize(line):
```
zhangguanheng66: we need to update the unit test

torchtext/utils.py
zhangguanheng66: Since this API will be exposed to the outside, I would expect more docs and examples (I could add them).

torchtext/utils.py
```diff
@@ -27,39 +28,56 @@ def inner(b=1, bsize=1, tsize=None):
...
     Examples:
         >>> url = 'http://www.quest.dcs.shef.ac.uk/wmt16_files_mmt/validation.tar.gz'
-        >>> path = './validation.tar.gz'
-        >>> torchtext.utils.download_from_url(url, path)
+        >>> torchtext.utils.download_from_url(url)
+        >>> '.data/validation.tar.gz'
     """
 
-    def process_response(r):
+    def process_response(r, root, filename):
```
zhangguanheng66: So this changes the API. Make sure it doesn't break any existing applications.

examples/text_classification/download.py
cpuhrsch: The only reason it's specific to text classification is because it pulls the URL from the text_classification package. But we can make the URL an argument.

torchtext/utils.py
cpuhrsch: Ah, but it's a local function. We should prepend it with "_"

examples/text_classification/download.py
zhangguanheng66: I think use an URL argument makes more sense. It's just a download API and we want people to use it separately.

PR #570, opened 2019-07-29
Docs

zhangguanheng66:
Update docs.
Fix an error "root directory is not found"
Fix an error "field number exceeds the limit."


torchtext/datasets/text_classification.py
```diff
@@ -71,9 +72,16 @@ def __init__(self, vocab, data, labels):
...
-            ngrams: a contiguous sequence of n items from s string text.
+            vocab: Vocabulary object used for dataset.
+            data: a list of label/tokens tuple. tokens are a tensor after
+                numericalizing the string tokens. label is an integer.
+                [(label1, tokens1), (label2, tokens2), (label2, tokens3)]
+            label: a set of the labels.
+                {label1, label2}
+
+        Examples:
+            See _setup_datasets() function in torchtext.datasets.text_classification
```
cpuhrsch: Since this is internal (due to the leading underscore) we can't use it as an example :/

torchtext/datasets/text_classification.py
```diff
@@ -99,6 +107,9 @@ def get_vocab(self):
 
 
 def _setup_datasets(dataset_name, root='.data', ngrams=2, vocab=None):
+    if not os.path.exists(root):
```
cpuhrsch: I have an error that is thrown if the user doesn't do this. I don't think we should create a directory (and in particular many directories) if it doesn't exist. For example, a user might supply a very long, wrong directory by accident.

torchtext/utils.py
```diff
@@ -100,9 +101,34 @@ def _process_response(r, root, filename):
...
+
+    Examples:
+        >>> from torchtext.utils import unicode_csv_reader
+        >>> import io
+        >>> with io.open(data_path, encoding="utf8") as f:
+        >>>     reader = unicode_csv_reader(f)
+
+    """
+
+    # Fix field larger than field limit error
```
cpuhrsch: Nice!

torchtext/datasets/text_classification.py
zhangguanheng66: Sure. If the error message is desired, we don't need to create a directory here. I will remove it.

PR #575, opened 2019-07-30
build_iterable_dataset_from_iterator

cpuhrsch:


examples/text_classification/iterable_train.py
```diff
@@ -0,0 +1,157 @@
...
+from torchtext.vocab import build_vocab_from_iterator
+
+from tqdm import tqdm
+
+
+def generate_batch(batch):
+
+    def generate_offsets(data_batch):
+        offsets = [0]
+        for entry in data_batch:
```
mttk: This could be replaced by `np.cumsum`

examples/text_classification/iterable_train.py
```diff
@@ -0,0 +1,157 @@
...
+            offsets.append(offsets[-1] + len(entry))
+        offsets = torch.tensor(offsets[:-1])
+        return offsets
+
+    cls = torch.tensor([entry[0] for entry in batch])
+    text = [entry[1] for entry in batch]
+    offsets = generate_offsets(text)
+    text = torch.cat(text)
+    return text, offsets, cls
```
mttk: perhaps `label` instead of `cls`?

examples/text_classification/iterable_train.py
```diff
@@ -0,0 +1,157 @@
...
+        with torch.no_grad():
+            output = model(text, offsets)
+            accuracy = (output.argmax(1) == cls).float().mean().item()
+            total_accuracy.append(accuracy)
+    print("Test - Accuracy: {}".format(sum(total_accuracy) / len(total_accuracy)))
+
+def csv_iterator(data_path, ngrams, vocab, epochs=1, is_worker=False):
+    tokenizer = get_tokenizer("basic_english")
+    with io.open(data_path, encoding="utf8") as f:
+        f.seek(0, 2)
```
mttk: This part could be pretty confusing if not commented

examples/text_classification/iterable_train.py
cpuhrsch: We don't want a hard numpy dependency, but I agree that'd be prettier

examples/text_classification/iterable_train.py
cpuhrsch: Agreed, we'll work through all examples and add lots of comments.

examples/text_classification/iterable_train.py
cpuhrsch: Sure

PR #578, opened 2019-08-01
Update generate_batch function and add a run script.

zhangguanheng66:


examples/text_classification/train.py
```diff
@@ -129,6 +123,8 @@ def test(data_):
                         help='device (default=cpu)')
     parser.add_argument('--data', default='.data',
                         help='data directory (default=.data)')
+    parser.add_argument('--dictionary', default='NONE',
```
cpuhrsch: If you don't provide a default and the user doesn't use --dictionary, args.dictionary will be None. So if you want your default to be None, you don't need to pass in a value for the default keyword, you can just leave it out.

examples/text_classification/train.py
```diff
@@ -67,12 +62,18 @@ def train(lr_, num_epoch, data_):
...
-                      collate_fn=generate_batch, num_workers=args.num_workers)
     optimizer = torch.optim.SGD(model.parameters(), lr=lr_)
     scheduler = torch.optim.lr_scheduler.StepLR(optimizer, 1, gamma=args.lr_gamma)
-    num_lines = num_epochs * len(data)
+    num_lines = num_epochs * len(data_) / batch_size * split_ratio
     for epoch in range(num_epochs):
+
+        train_len = int(len(data_) * split_ratio)
+        sub_train_, sub_valid_ = \
+            random_split(data_, [train_len, len(data_) - train_len])
```
cpuhrsch: Usually this is done only once at the beginning of the script

examples/text_classification/train.py
```diff
@@ -88,6 +89,20 @@ def train(lr_, num_epoch, data_):
                         progress * 100, scheduler.get_lr()[0], loss))
         # Adjust the learning rate
         scheduler.step()
+
+        # Test the model on valid set
```
cpuhrsch: We have the test function for this

examples/text_classification/train.py
```diff
@@ -129,6 +146,8 @@ def test(data_):
                         help='device (default=cpu)')
     parser.add_argument('--data', default='.data',
                         help='data directory (default=.data)')
+    parser.add_argument('--dictionary',
+                        help='path to save vocab (default=NONE)')
```
cpuhrsch: You don't need to mention (default=None) here, it's implied.

examples/text_classification/train.py
```diff
@@ -34,29 +35,23 @@ def generate_batch(batch):
...
 torch.utils.data.DataLoader is recommended for PyTorch users to load data.
-We use DataLoader here to load datasets and send it to the train()
+We use DataLoader here to load datasets and send it to the train_and_valid()
 and text() functions.
 
 """
 
 
-def train(lr_, num_epoch, data_):
+def train_and_valid(lr_, num_epoch, data_):
```
cpuhrsch: Pass in train_data_ and validation_data_ (or so)

examples/text_classification/train.py
```diff
@@ -67,12 +62,15 @@ def train(lr_, num_epoch, data_):
         data_: the data used to train the model
     """
 
-    data = DataLoader(data_, batch_size=batch_size, shuffle=True,
-                      collate_fn=generate_batch, num_workers=args.num_workers)
     optimizer = torch.optim.SGD(model.parameters(), lr=lr_)
     scheduler = torch.optim.lr_scheduler.StepLR(optimizer, 1, gamma=args.lr_gamma)
-    num_lines = num_epochs * len(data)
+    num_lines = num_epochs * len(sub_train_) / batch_size * split_ratio
```
cpuhrsch: why do you need to multiply this by the split_ratio and divide by the batch_size? The only thing that should have changed is the length of the dataset. Also, you're instantiating the DataLoader on each epoch, which can be a costly operation.

cpuhrsch: Now that we've moved over to using StepLR I think we don't need this number anymore anyway, right?

examples/text_classification/train.py
```diff
@@ -154,9 +160,17 @@ def test(data_):
                           embed_dim, len(train_dataset.get_labels())).to(device)
     criterion = torch.nn.CrossEntropyLoss().to(device)
 
-    train(lr, num_epochs, train_dataset)
-    test(test_dataset)
+    # split train_dataset into train and valid
+    train_len = int(len(train_dataset) * split_ratio)
+    sub_train_, sub_valid_ = \
```
cpuhrsch: I'm not sure this works if split_ratio is 1 - have you tried running it with '1'?

PR #582, opened 2019-08-04
Added the code

ShahriarSS:
@cpuhrsch ```basic_english_normalize```. The same structure as ```C++ Vocab```. Also this is only the C++ implementation of Python code but I think this can be optimized further. I'll try to make it faster. @cpuhrsch Please give your feedback on optimization.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,14 @@
 import os
 import io
 import re
+import sys
+import torch
 from setuptools import setup, find_packages
+from torch.utils.cpp_extension import CppExtension
+
+import glob
+import shutil
+import distutils.command.clean
 
 
 def read(*names, **kwargs):
@@ -25,6 +32,55 @@ def find_version(*file_paths):
 VERSION = find_version('torchtext', '__init__.py')
 long_description = read('README.rst')
 
+
+def get_extensions():
+    this_dir = os.path.dirname(os.path.abspath(__file__))
+    extensions_dir = os.path.join(this_dir, 'torchtext', 'csrc')
+
+    main_file = glob.glob(os.path.join(extensions_dir, 'text_extension.cpp'))
+    source_core = glob.glob(os.path.join(extensions_dir, 'core', '*.cpp'))
+
+    sources = main_file + source_core
+    extension = CppExtension
+
+    define_macros = []
+    extra_compile_args = {}
+
+    if sys.platform == 'win32':
+        define_macros += [('torchtext_EXPORTS', None)]
+
+    sources = [os.path.join(extensions_dir, s) for s in sources]
+
+    include_dirs = [extensions_dir]
+
+    ext_modules = [
+        extension(
+            'torchtext._C',
+            sources,
+            include_dirs=include_dirs,
+            define_macros=define_macros,
+            extra_compile_args=extra_compile_args,
+        )
+    ]
+
+    return ext_modules
+
+
+class clean(distutils.command.clean.clean):
+    def run(self):
+        with open('.gitignore', 'r') as f:
+            ignores = f.read()
+            for wildcard in filter(None, ignores.split('\n')):
+                for filename in glob.glob(wildcard):
+                    try:
+                        os.remove(filename)
+                    except OSError:
+                        shutil.rmtree(filename, ignore_errors=True)
+
+        # It's an old-style class in Python 2.7...
+        distutils.command.clean.clean.run(self)
+
+
 setup_info = dict(
     # Metadata
     name='torchtext',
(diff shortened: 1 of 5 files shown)


cpuhrsch:
Thanks @ShahriarSS!

A few comments: 1) Could you hook up the extension code to replace the Python version of basic_english_normalize? That'll also add more test coverage. 2) Could you add some benchmarks (google benchmark is good tool to add here) and compare the performance with the Python version?

setup.py
```diff
@@ -25,6 +32,55 @@ def find_version(*file_paths):
...
+    if sys.platform == 'win32':
+        define_macros += [('torchtext_EXPORTS', None)]
+
+    sources = [os.path.join(extensions_dir, s) for s in sources]
+
+    include_dirs = [extensions_dir]
+
+    ext_modules = [
+        extension(
+            'torchtext._C',
```
jamarshon: this should just be '_C'. See https://github.com/pytorch/audio/blob/master/setup.py#L109
```
    ext_modules=[
        CppExtension(
            '_torch_sox',
```

ShahriarSS: @jamarshon The problem is still there.

jamarshon: you have to replace all occurences of torchtext._C with _C in the code

jamarshon:
@ShahriarSS You can use https://github.com/pytorch/text/pull/586 as rough suggestion if you want
You need a relatively high version of gcc/g++ for regex features as the default one 4.8.4 in trusty is not sufficient and trying 4.9 will give std::bad_alloc
```
before_install:
  - export CXX="g++-6" CC="gcc-6"
addons:
  apt:
    sources:
      - ubuntu-toolchain-r-test
    packages:
      - gcc-6
      - g++-6 
```

CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+cmake_minimum_required(VERSION 2.8)
```
jamarshon: I don't think we need this file as we are using setup.py to compile

setup.py
```diff
@@ -25,6 +32,55 @@ def find_version(*file_paths):
...
+
+def get_extensions():
+    this_dir = os.path.dirname(os.path.abspath(__file__))
+    extensions_dir = os.path.join(this_dir, 'torchtext', 'csrc')
+
+    main_file = glob.glob(os.path.join(extensions_dir, 'text_extension.cpp'))
+    source_core = glob.glob(os.path.join(extensions_dir, 'core', '*.cpp'))
+
+    sources = main_file + source_core
+    extension = CppExtension
```
jamarshon: can we just use CppExtension instead of creating a variable for it?

torchtext/csrc/core/utils.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+std::vector<std::string> basic_english_normalize(std::string line) {
+  std::transform(line.begin(), line.end(), line.begin(), tolower);
+  for (const auto& P : _patterns)
+    line = std::regex_replace(line, P.first, P.second);
+  return split(line, ' ');
+}
+
+std::vector<std::string> split(const std::string& text, char splitter) {
+  std::string string;
```
jamarshon: is this the same as using stringstream? e.g.
```c++
  std::string token;
  std::istringstream ss(text);
  std::vector<std::string> res;
  while (getline(ss, token, splitter)) 
    res.push_back(token);
  return res;
```

torchtext/csrc/core/utils.h
```diff
@@ -0,0 +1,21 @@
+#ifndef UTILS_H
+#define UTILS_H
+
+#include <torch/torch.h>
```
jamarshon: #include <torch/extension.h>

setup.py
```diff
@@ -25,6 +32,55 @@ def find_version(*file_paths):
...
+            include_dirs=include_dirs,
+            define_macros=define_macros,
+            extra_compile_args=extra_compile_args,
+        )
+    ]
+
+    return ext_modules
+
+
+class clean(distutils.command.clean.clean):
```
jamarshon: this function removes all .gitignore files which include requirements.txt, seems not intended

CMakeLists.txt
ShahriarSS: If we're going to have C++ API we do.

setup.py
ShahriarSS: ```extension``` can later be set to ```CUDAExtension``` but right now we have a function that has nothing to do with torch so I think that we should leave it this way.

ShahriarSS:
Thanks to @jamarshon 's test PR I was able to Finish the code for extension but I think that performance of C++ code is worse than Python equivalent here. @cpuhrsch What do you think?

setup.py
jamarshon: I believe the function will still remove requirements.txt because the first line of the .gitignore is `*.txt` which will remove all text files?

PR #584, opened 2019-08-09
Fix the overflow error by changing maxInt

rouzbeh-afrasiabi:
csv.field_size_limit input was changed from sys.maxsize to maxInt. This Will resolve the "OverflowError: Python int too large to convert to C"  error.


zhangguanheng66:
run "flake8 /torchtext/utils.py" to reproduce the errors.

PR #585, opened 2019-08-09
Adding installation step to Travis

ShahriarSS:


zhangguanheng66:
I think you could add "if [[ "$SKIP_TESTS" != "true" ]]; then" before "python setup.py install".

build_tools/travis/install.sh
```diff
@@ -60,4 +60,7 @@ if [[ "$SKIP_TESTS" != "true" ]]; then
 
     # PyTorch
     conda install --yes pytorch torchvision -c pytorch
+
+    # Installation
+    python setup.py install
```
zhangguanheng66: I think you could add "if [[ "$SKIP_TESTS" != "true" ]] here.

ShahriarSS: But @zhangguanheng66 the code is inside the ```if``` statement. Is this nessesary?

ShahriarSS:
@zhangguanheng66 All the tests are passing now.

PR #591, opened 2019-08-20
Fix backwards compatibility of vocab serialization

mttk:
(Hopefully) fixes the issue mentioned in #531, after the PR was merged.

Previous instances of "Vocab" didn't have the "unk_index" attribute, which the deserialization wrongly assumes will be present. This is causing issues when loading models stored with previous versions of torchtext. I added a bandaid fix as well as a test checking whether loading works on a mocked older version of the Vocab instance.

@francoishernandez I would appreciate if you can test the patch once it's merged.


francoishernandez:
Just installed from your branch, seems to work fine, thanks!

PR #597, opened 2019-09-09
Add Sentencepiece binding to torchtext, plus example to build torchtext dataset with sentencepiece

zhangguanheng66:
Add sentencepiece binding to torchtext.
Show an example to build torchtext dataset with sentencepiece. Reproduce fastTest's results in the example with sentencepiece binding.


examples/sentencepiece/example_bpm_dataset.py
```diff
@@ -0,0 +1,115 @@
+import logging
+import torch
+import io
+from torchtext.utils import download_from_url, extract_archive, unicode_csv_reader
+from torchtext.data.utils import generate_sp_tokenizer
+from os import path
+
+URLS = {
```
cpuhrsch: You can get this from https://github.com/pytorch/text/blob/master/torchtext/datasets/text_classification.py#L11

examples/sentencepiece/train.py
```diff
@@ -0,0 +1,162 @@
...
+    label = torch.tensor([entry[0] for entry in batch])
+    text = [entry[1] for entry in batch]
+    offsets = [0] + [len(entry) for entry in text]
+
+    offsets = torch.tensor(offsets[:-1]).cumsum(dim=0)
+    text = torch.cat(text)
+    return text, offsets, label
+
+
+from torch.utils.data import DataLoader
```
cpuhrsch: Imports should always be at the top of the file. Install a formatter such as https://github.com/hhatto/autopep8 - you can then apply it via something like https://github.com/google/vim-codefmt

examples/sentencepiece/train.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description='Apply SentencePiece to text classification dataset.')
+    parser.add_argument('--dataset', default='YelpReviewFull',
+                        help='dataset name')
+    parser.add_argument('--num-epochs', type=int, default=10,
+                        help='num epochs (default=10)')
+#    parser.add_argument('--embed-dim', type=int, default=32,
```
cpuhrsch: Should this be deleted?

examples/sentencepiece/train.py
```diff
@@ -0,0 +1,162 @@
...
+    if not os.path.exists(args.data_directory):
+        print("Creating directory {}".format(args.data_directory))
+        os.mkdir(args.data_directory)
+
+    import example_bpm_dataset as text_classification
+    train_data, test_data = text_classification.setup_datasets(args.dataset,
+                                                               root='.data',
+                                                               vocab_size=args.vocab_size)
+    from torchtext.datasets.text_classification import LABELS
+    NUN_CLASS = len(LABELS[args.dataset])
```
cpuhrsch: nit: NUM?

examples/sentencepiece/train.py
```diff
@@ -0,0 +1,162 @@
...
+        mins = secs / 60
+        secs = secs % 60
+
+        print('Epoch: %d' % (epoch + 1),
+              ' | time in %d minutes, %d seconds' % (mins, secs))
+        print('\tLoss:\t %.4f(train)\t|\tAcc:\t %.1f'
+              % (train_loss, train_acc * 100), '%')
+        print('\tLoss:\t %.4f(valid)\t|\tAcc:\t %.1f'
+              % (valid_loss, valid_acc * 100), '%')
+#        print(f'\tLoss: {train_loss:.4f}(train)\t|\tAcc: {train_acc * 100:.1f}%(train)')
```
cpuhrsch: Should this be deleted?

test/data/test_utils.py
```diff
@@ -70,3 +70,60 @@ def test_text_nomalize_function(self):
...
+    def test_get_tokenizer_sentencepiece(self):
+        # Test SentencePiece option, and verify it properly handles punctuation.
+        test_sample = 'SentencePiece is an unsupervised text tokenizer and detokenizer'
+        model_path = 'test/asset/spm_example.model'
+        tokenizer = data.get_tokenizer("sentencepiece", spm_name=model_path)
+
+        import sys
+        # Handle byte string in Python2 and Unicode string in Python3, respectively
+        if sys.version_info < (3, 0):
+            ref_results = ['\xe2\x96\x81Sent', 'ence', 'P', 'ie', 'ce', '\xe2\x96\x81is',
```
cpuhrsch: How do you create the references?

torchtext/data/utils.py
```diff
@@ -71,7 +71,7 @@ def _basic_english_normalize(line):
     return line.split()
 
 
-def get_tokenizer(tokenizer, language='en'):
+def get_tokenizer(tokenizer, language='en', spm_name=None):
```
cpuhrsch: This flag seems unrelated and too general for a geneic "get_tokenizer" function. If spm_name is not None for a tokenizer that's not "sentencepiece" it has no effect or meaning.

torchtext/data/utils.py
```diff
@@ -155,6 +155,15 @@ def get_tokenizer(tokenizer, language='en'):
         except ImportError:
             print("Please install revtok.")
             raise
+    elif tokenizer == "sentencepiece" and spm_name is not None:
```
cpuhrsch: This tokenizer is special in that there's a file it uses to be created. That makes it different from all the other ones. Maybe it should't be here?

torchtext/data/utils.py
```diff
@@ -224,6 +233,74 @@ def _get_ngrams(n):
             yield ' '.join(x)
 
 
+def generate_sp_tokenizer(filename, vocab_size=20000,
```
cpuhrsch: I think it makes sense to move functions like this into their own file (maybe)?

torchtext/data/utils.py
```diff
@@ -224,6 +233,74 @@ def _get_ngrams(n):
...
+        >>> iterator = ['Generic data loaders, abstractions, and iterators',
+        >>>             'Pre-built loaders for common NLP datasets']
+        >>> spm_data_generator('spm_user.model', iterator)
+        >>> [[3046, 7706, 319, 14246, 174, 5, 4838, 4, 18995, 4, 5, 14, 41, 86, 6106],
+        >>>  [3493, 15, 14528, 14246, 174, 18, 3054, 3397, 264, 319, 17493]]
+    """
+
+    import sentencepiece as spm
+    sp_user = spm.SentencePieceProcessor()
+    sp_user.load(spm_name)
```
cpuhrsch: This means that every time I want to use this, I have to load the model again instead of loading it once and then applying the processor.

torchtext/data/utils.py
```diff
@@ -224,6 +233,74 @@ def _get_ngrams(n):
...
+
+    data = []
+    for corpus in iterator:
+        token_ids = sp_user.encode_as_ids(corpus)
+        data.append(token_ids)
```
cpuhrsch: This consumes an iterator, yet returns a list which materializes all data. Maybe this should be an iterator

examples/sentencepiece/run_script.sh
```diff
@@ -0,0 +1,5 @@
+if [ ! -d ".data" ]; then
```
vincentqb: This is the entry point into running the example, right? do you have a README or instructions to indicate how to use the example?

vincentqb:
By the way, one check is failing. What would have to be done to pass it?

examples/sentencepiece/train.py
vincentqb: I personally like [black](https://pypi.org/project/flake8-black/), since it autoformats the code for you, and [isort](https://github.com/timothycrosley/isort).

examples/sentencepiece/example_bpm_dataset.py
vincentqb: This file is a similar version of `text_classification`, should some structures be reused, or copied (as done here) appropriate?

zhangguanheng66:
> By the way, one check is failing. What would have to be done to pass it?

That one is fine to fail. For torchtext, we just need to monitor travis.

torchtext/data/utils.py
zhangguanheng66: Very good point. I will move those func to a `Transform` class.

examples/sentencepiece/run_script.sh
zhangguanheng66: Add a README file in the example folder.

examples/sentencepiece/train.py
```diff
@@ -0,0 +1,162 @@
...
+            output = test_model(text, offsets)
+            loss = criterion(output, cls)
+            loss += loss.item()
+            acc += (output.argmax(1) == cls).sum().item()
+
+    return loss / len(data_), acc / len(data_)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
```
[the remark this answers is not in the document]
zhangguanheng66: Not really. Just import `argparse` at the beginning of the script.

examples/sentencepiece/train.py
```diff
@@ -0,0 +1,156 @@
+import torch
```
cpuhrsch: Instead of copy-pasting train.py here you could modify the other examples to use the sentencepiece abstraction you implemented in the main library. That probably makes more sense to do after you migrate current tokenizers to the transform / functional split.

torchtext/data/functional.py
```diff
@@ -0,0 +1,37 @@
...
+    Examples:
+        >>> from torchtext.data.utils import generate_sp_tokenizer
+        >>> generate_sp_tokenizer('test.csv', vocab_size=23456,
+        >>>                       model_prefix='spm_user')
+        >>> import sentencepiece as spm
+        >>> sp_user = spm.SentencePieceProcessor()
+        >>> sp_user.load('spm_user.model')
+    """
+
+    import sentencepiece as spm
```
vincentqb: nit: should this be at the beginning?

examples/sentencepiece/README.md
```diff
@@ -0,0 +1,13 @@
+# This is an example to create a dataset with SentencePiece binding.
```
vincentqb: nit: title should probably just be "# Create a dataset with SentencePiece binding" since the text already refers to the content as an example, and its in the example folder :)

test/data/test_transforms.py
```diff
@@ -0,0 +1,38 @@
+from ..common.torchtext_test_case import TorchtextTestCase
+
+
+class TestUtils(TorchtextTestCase):
+    def test_sentencepiece_encode_as_ids(self):
+        from torchtext.data.transforms import sentencepiece_encode_as_ids
```
vincentqb: nit: same about import

test/data/test_transforms.py
```diff
@@ -0,0 +1,38 @@
...
+        spm_generator = sentencepiece_encode_as_ids(model_path)
+
+        ref_results = [15340, 4286, 981, 1207, 1681, 17, 84, 684, 8896, 5366,
+                       144, 3689, 9, 5602, 12114, 6, 560, 649, 5602, 12114]
+
+        self.assertEqual(list(spm_generator([test_sample]))[0],
+                         ref_results)
+
+    def test_sentencepiece_encode_as_pieces(self):
+        import sys
```
vincentqb: nit: same about imports

torchtext/data/transforms.py
```diff
@@ -0,0 +1,67 @@
...
+        output: a generator over the ids based on SentencePiece model.
+
+    Examples:
+        >>> from torchtext.data.transforms import sentencepiece_encode_as_pieces
+        >>> sp_tokens_generator = sentencepiece_encode_as_pieces("m_user.model")
+        >>> list_a = ["sentencepiece encode as pieces", "examples to   try!"]
+        >>> list(sp_tokens_generator(list_a))
+            [['_sentence', 'piece', '_en', 'co', 'de', '_as', '_pieces'],
+             ['_example', 's', '_to', '_try', '!']]
+
```
vincentqb: nit: extra space?

torchtext/data/transforms.py
```diff
@@ -0,0 +1,67 @@
...
+        output: a generator over the ids based on SentencePiece model.
+
+    Examples:
+        >>> from torchtext.data.transforms import sentencepiece_encode_as_ids
+        >>> sp_id_generator = sentencepiece_encode_as_ids("m_user.model")
+        >>> list_a = ["sentencepiece encode as pieces", "examples to   try!"]
+        >>> list(sp_id_generator(list_a))
+            [[9858, 9249, 1629, 1305, 1809, 53, 842],
+             [2347, 13, 9, 150, 37]]
+
```
vincentqb: nit: extra space?

torchtext/data/transforms.py
```diff
@@ -0,0 +1,67 @@
...
+    Outputs:
+        output: a generator over the ids based on SentencePiece model.
+
+    Examples:
+        >>> from torchtext.data.transforms import sentencepiece_encode_as_pieces
+        >>> sp_tokens_generator = sentencepiece_encode_as_pieces("m_user.model")
+        >>> list_a = ["sentencepiece encode as pieces", "examples to   try!"]
+        >>> list(sp_tokens_generator(list_a))
+            [['_sentence', 'piece', '_en', 'co', 'de', '_as', '_pieces'],
```
vincentqb: Out of curiosity: did anything changed in the tokens in the last few commits? They come out as changed :)

zhangguanheng66: I removed those Non-ASCII character '\xe2' in the example.

zhangguanheng66:
> I'm noticing the dataset class does not use iterator. Is this meant to represent a consensus we agreed on?

Not to the dataset level. At current stage, I would like to implement a few transforms (two in this case) in the format of generators, apply them as tokenizers. As we have more usage cases, we can sit together and make a final decision across all the domains. :)

examples/text_classification/train.py
```diff
@@ -137,6 +137,10 @@ def test(data_):
                         help='device (default=cpu)')
     parser.add_argument('--data', default='.data',
                         help='data directory (default=.data)')
+    parser.add_argument('--use-sp-tokenizer', type=bool, default=False,
+                        help='use sentencepiece tokenizer (default=False)')
+    parser.add_argument('--vocab-size', type=int, default=20000,
```
cpuhrsch: I'd change this to say "--sp-vocab-size" just so it's less confusing

torchtext/data/functional.py
```diff
@@ -0,0 +1,44 @@
...
+
+    spm_training_string = "--input={} \
+                           --vocab_size={} \
+                           --model_prefix={} \
+                           --model_type={}".format(filename,
+                                                   vocab_size,
+                                                   model_prefix,
+                                                   model_type)
+    spm.SentencePieceTrainer.train(spm_training_string)
+    return None
```
cpuhrsch: Why return None?

cpuhrsch: Ah because the straining also involves writing to disk - is it possible to just return the object and have the user save it herself (if necessary). That way we don't add any constraints for where this is supposed to be saved or whether it's supposed to be saved at all.

torchtext/data/transforms.py
```diff
@@ -0,0 +1,71 @@
...
+        >>> from torchtext.data.transforms import sentencepiece_encode_as_ids
+        >>> sp_id_generator = sentencepiece_encode_as_ids("m_user.model")
+        >>> list_a = ["sentencepiece encode as pieces", "examples to   try!"]
+        >>> list(sp_id_generator(list_a))
+            [[9858, 9249, 1629, 1305, 1809, 53, 842],
+             [2347, 13, 9, 150, 37]]
+    """
+
+    sp_model = spm.SentencePieceProcessor()
+    sp_model.Load(spm_path)
```
cpuhrsch: Instead of loading the model here, can this just accept a sentence piece model and nothing else? That way we don't add any constraints on where the model is supposed to be loaded from. It could however be good to have a "load_sentencepiece_model_from_path" or something that wraps the first two lines (just because static constructors are more idiomatic than a method that overwrites an entire state).

torchtext/data/transforms.py
```diff
@@ -0,0 +1,71 @@
...
+        >>> from torchtext.data.transforms import sentencepiece_encode_as_pieces
+        >>> sp_tokens_generator = sentencepiece_encode_as_pieces("m_user.model")
+        >>> list_a = ["sentencepiece encode as pieces", "examples to   try!"]
+        >>> list(sp_tokens_generator(list_a))
+            [['_sentence', 'piece', '_en', 'co', 'de', '_as', '_pieces'],
+             ['_example', 's', '_to', '_try', '!']]
+    """
+
+    sp_model = spm.SentencePieceProcessor()
+    sp_model.Load(spm_path)
```
cpuhrsch: Same thing as above. Here the "load...from_path" function can be used to pass into this. Also, if you can load it independently you can run this multiple times.

Further, this doesn't need to return a function that then yields, it can yield itself.

torchtext/data/transforms.py
```diff
@@ -0,0 +1,71 @@
...
+    return _internal_func
+
+
+def sentencepiece_encode_as_pieces(spm_path):
+    """A sentencepiece model for tokenizing a text sentence into
+       a generator over the tokens.
+
+    Arguments:
+        spm_path: the file path saving the sentencepiece model.
+        txt_iter: input sentence text generator.
```
cpuhrsch: This doesn't accept a "txt_iter" argument.

torchtext/data/transforms.py
```diff
@@ -0,0 +1,71 @@
...
+    sp_model = spm.SentencePieceProcessor()
+    sp_model.Load(spm_path)
+
+    def _internal_func(txt_iter):
+        for line in txt_iter:
+            yield sp_model.EncodeAsIds(line)
+    return _internal_func
+
+
+def sentencepiece_encode_as_pieces(spm_path):
```
cpuhrsch: If "pieces" means "tokens" in our language I suggest we use tokens instead of pieces here because we need to standardize on a language within our project even if other projects might use a different language.

torchtext/data/functional.py
zhangguanheng66: I have checked the API in sentencepiece, and it seems they don't have the option to return an object.

PR #598, opened 2019-09-11
Fix example argument help strings

ribeirompl:
- Corrected help string in argument parser for default epochs
- Added help string in argument parser for --save-model-path (same help string as in iterable_train.py in examples/text_classification)

### DIFF
--- a/examples/text_classification/train.py
+++ b/examples/text_classification/train.py
@@ -118,7 +118,7 @@ if __name__ == "__main__":
         description='Train a text classification model on text classification datasets.')
     parser.add_argument('dataset', choices=text_classification.DATASETS)
     parser.add_argument('--num-epochs', type=int, default=5,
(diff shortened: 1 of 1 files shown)


examples/text_classification/train.py
```diff
@@ -118,7 +118,7 @@ def test(data_):
         description='Train a text classification model on text classification datasets.')
     parser.add_argument('dataset', choices=text_classification.DATASETS)
     parser.add_argument('--num-epochs', type=int, default=5,
-                        help='num epochs (default=3)')
+                        help='num epochs (default=5)')
```
zhangguanheng66: Any reason to change the number of epochs? If not, I would suggest to keep the original one.

ribeirompl: The actual default is 5 (line above), however in the help string it says it's 3.

PR #600, opened 2019-09-18
added xnli dataset

bentrevett:
This is initial work on adding the [XNLI dataset](http://www.nyu.edu/projects/bowman/xnli/) to TorchText. 

Weirdly, the XNLI dataset only has a "dev" and a "test" set, no training set. This means I had to make some decisions that I'm not sure were for the best. 

I've either removed or hard coded the `train` argument to be `None`, which means the `XNLI` dataset works fine with `.splits`, but not with `.iters`. due to [these](https://github.com/pytorch/text/blob/master/torchtext/datasets/nli.py#L120) lines being hardcoded.

The XNLI dataset is used for "cross-language sentence encoding", so I have also added a `language_field`.

### DIFF
--- a/test/nli.py
+++ b/test/nli.py
@@ -64,3 +64,30 @@ batch = next(iter(train_iter))
 print("Numericalize premises:\n", batch.premise)
 print("Numericalize hypotheses:\n", batch.hypothesis)
 print("Entailment labels:\n", batch.label)
+
+# Testing XNLI
+print("Run test on XNLI...")
+TEXT = data.Field()
+LABEL = data.LabelField()
+GENRE = data.Field()
+LANGUAGE = data.Field()
+
+val, test = datasets.XNLI.splits(TEXT, LABEL, GENRE, LANGUAGE)
+
+print("Fields:", val.fields)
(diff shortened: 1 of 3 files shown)


zhangguanheng66:
@bentrevett Thanks for your contribution. We are right now re-designing torchtext dataset and hopefully make it more compatible with `torch.utils.data`. One of the ideas is to switch to `DataLoader` and `Sampler` in `torch.utils.data`. You can find the new pattern as text_classification dataset [here](https://github.com/pytorch/text/blob/master/torchtext/datasets/text_classification.py)

zhangguanheng66:
@bentrevett If you still want to add the XNLI dataset, we are happy to accept it even under current structure.

bentrevett:
@zhangguanheng66 Sure. I deleted the branch with the changes though so will have to close this and make a new pull request.

zhangguanheng66:
@bentrevett We are always welcome some new datasets for torchtext. Once we have concluded the new design structure, we could mitigate the existing datasets into the new pattern. Appreciate your contributions.

PR #602, opened 2019-09-19
additions to download_from_url and extract_archive

bentrevett:
`download_from_url` would always fail to download from a non-google drive link due to a `KeyError: 'content-disposition'` from [this](https://github.com/pytorch/text/blob/master/torchtext/utils.py#L53) line. This was fixed by explicitly getting the filename from the url instead of the url header.

This fixed caused some issues where google drive links thought end of the url was the filename, so this is handled by setting the filename back to `None` after checking if the url is or isn't a google drive link.

I also added support for .zip files in `extract_archive`. This involved removing the `archive` argument and having the function get it from the end of the filename. Had a quick look but this didn't seem to break any functionality.

Also, the `download_extract.py` file did not seem to work in the first place. It would try and do an `os.path.split(path)` on `validation.tar.gz` and then `root` would just be an empty string which would cause an error, changing the default to `./validation.tar.gz` seems to fix that, but I am not 100% sure this is the intended functionality.


zhangguanheng66:
You can use flake8 and run the lint error tests locally.

test/test_utils.py
```diff
@@ -0,0 +1,110 @@
...
+
+        # download file and ensure is in correct location
+        archive_path = utils.download_from_url(url)
+        assert target_archive_path == archive_path
+
+        # assert that non-valid file (not an archive) raises error
+        with self.assertRaises(NotImplementedError):
+            utils.extract_archive(archive_path)
+
+        # remove file
```
zhangguanheng66: Are there any extracted files which need to be removed after the unit test?

zhangguanheng66:
@bentrevett Thanks for the contributions. Nice jobs. I just update the PR with the master branch. Will merge it after CI tests pass.

PR #610, opened 2019-10-02
Add a new dataset - Enwik9

zhangguanheng66:
enwik9: compressed size of first 109 bytes of enwiki-20060303-pages-articles.xml.

It's part of Large Text Compression Benchmark project ([here](http://mattmahoney.net/dc/text.html)).

Benchmark results:
389.595s download_from_url: downloading
7.571s extract_archive
504.140s preprocess_raw_enwik9: remove html, images.
35.772s generate_offsets
0.128s get_vocab
0.012s read_lines_from_iterator

So the most time consuming tasks are downloading and pre-processing (a.k.a. clean up the raw data). Therefore, the processed data are saved to disk for re-use.


torchtext/data/functional.py
```diff
@@ -16,7 +16,7 @@
 def generate_sp_model(filename, vocab_size=20000,
                       model_type="unigram",
                       model_prefix='m_user'):
-    """Train a SentencePiece tokenizer.
+    r"""Train a SentencePiece tokenizer.
```
vincentqb: nit: This string change could have gone in a separate PR :)

torchtext/data/functional.py
```diff
@@ -114,3 +114,40 @@ def _internal_func(txt_iter):
         for line in txt_iter:
             yield sp_model.EncodeAsPieces(line)
     return _internal_func
+
+
+def custom_replace(replace_pattern):
+    r"""A transform to convert text string
```
vincentqb: nit: should there be a period at the end of the sentence?

torchtext/datasets/unsupervised_learning.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def read_lines_from_iterator(data_path, offsets, begin_line, num_lines):
+    with open(data_path) as f:
+        f.seek(offsets[begin_line])
+        for i in range(num_lines):
+            yield f.readline()
+
+
+def normalized_raw_enwik9(input_filename, output_filename):
```
vincentqb: nit: maybe it would be a good idea to define what normalize mean in this case?

torchtext/datasets/unsupervised_learning.py
```diff
@@ -0,0 +1,138 @@
...
+            while True:
+                line = f1.readline()
+                if not line:
+                    break
+                line = list(enwik9_norm_transform([line]))[0]
+                if line != ' ' and line != '':
+                    if line[0] == ' ':
+                        line = line[1:]
+                    f2.writelines(line + '\n')
+    return
```
vincentqb: nit: is it needed to have a blank return?

torchtext/datasets/unsupervised_learning.py
```diff
@@ -0,0 +1,138 @@
...
+            begin_line: the number of beginning line. Default: 0
+            num_lines: the number of lines to be loaded. Default: 6348957
+            root: Directory where the datasets are saved. Default: ".data"
+            data: a list of label/tokens tuple. tokens are a tensor after
+
+        Examples:
+            >>> from torchtext.datasets import EnWik9
+            >>> enwik9 = EnWik9(num_lines=20000)
+            >>> vocab = enwik9.get_vocab()
+
```
vincentqb: nit: blank line :)

torchtext/datasets/unsupervised_learning.py
```diff
@@ -0,0 +1,138 @@
...
+        processed_file = os.path.join(root, 'norm_enwik9')
+        if not os.path.exists(processed_file):
+            url = 'http://mattmahoney.net/dc/enwik9.zip'
+            dataset_zip = download_from_url(url,
+                                            path=os.path.join(root, 'enwik9.zip'),
+                                            root=root)
+            extracted_file = extract_archive(dataset_zip)
+            raw_file = os.path.join(root, extracted_file[0])
+            normalized_raw_enwik9(raw_file, processed_file)
```
vincentqb: nit: could people not want to normalize the dataset?

torchtext/datasets/unsupervised_learning.py
vincentqb: nit: is `normalized`? or `normalize`?

torchtext/datasets/unsupervised_learning.py
```diff
@@ -0,0 +1,138 @@
...
+
+        super(EnWik9, self).__init__()
+
+        processed_file = os.path.join(root, 'norm_enwik9')
+        if not os.path.exists(processed_file):
+            url = 'http://mattmahoney.net/dc/enwik9.zip'
+            dataset_zip = download_from_url(url,
+                                            path=os.path.join(root, 'enwik9.zip'),
+                                            root=root)
+            extracted_file = extract_archive(dataset_zip)
```
vincentqb: if the data is already downloaded or extracted, could we want to skip these steps?

torchtext/data/functional.py
```diff
@@ -114,3 +114,40 @@ def _internal_func(txt_iter):
...
+
+    Examples:
+        >>> from torchtext.data.functional import custom_replace
+        >>> custom_replace_transform = custom_replace([(r'S', 's'), (r'\s+', ' ')])
+        >>> list_a = ["Sentencepiece encode  aS  pieces", "exampleS to   try!"]
+        >>> list(custom_replace_transform(list_a))
+            ['sentencepiece encode as pieces', 'examples to try!']
+    """
+
+    import re
```
vincentqb: nit: should the import move to the beginning?

torchtext/datasets/__init__.py
```diff
@@ -10,7 +10,7 @@
     AG_NEWS, SogouNews, DBpedia, YelpReviewPolarity, \
     YelpReviewFull, YahooAnswers, \
     AmazonReviewPolarity, AmazonReviewFull
-
+from .unsupervised_learning import EnWik9
```
vincentqb: nit: same about import

torchtext/datasets/unsupervised_learning.py
cpuhrsch: I don't think this "read -> write" function is necessary. It's a simple operation

```
with open(input_filename, 'r') as f1, open(output_filename, 'w') as f2:
    for line in f1:
        line = line.strip() # Removes spaces to the left and right
        if len(line) > 0:
            f2.write(list(enwik9_norm_transform([line]))[0].lstrip())
````

This whole indexing, stripping etc. is strange. Is that really necessary?

File: torchtext/datasets/unsupervised_learning.py (hunk 4 above)
Comment (cpuhrsch) [reply torchtext/datasets/unsupervised_learning.py]: Python functions without a return value evaluate to None when called by default, so this isn't necessary.

Comment (vincentqb) [reply torchtext/datasets/unsupervised_learning.py]: Yup

**Comment by cpuhrsch** [comment]
@vincentqb - we consider datasets such as IMDB to be legacy at this point. The supervised learning datasets are more in line with our simplified philosophy.

**Comment by vincentqb** [comment]
> @vincentqb - we consider datasets such as IMDB to be legacy at this point. The supervised learning datasets are more in line with our simplified philosophy.

Thanks for pointing this out. It'd be nice if the PR would say which ones are the new format, and which ones are the old.

**Comment by cpuhrsch** [comment]
@vincentqb - since we're still in the process of more rigorously defining data processing and can't reference an RFC or wider discussion that kind of notice will mostly about noting that this does not comply to the current standard, but nothing more.

File: torchtext/datasets/unsupervised_learning.py (hunk 3 above)
Comment (zhangguanheng66) [reply torchtext/datasets/unsupervised_learning.py]: It's pretty time consuming to normalize the text. The idea is to re-use the normalized text file in the future. Of course, for a smaller text file, we don't need to dump the normalized text into a file and streaming will be a better option.

**Comment by zhangguanheng66** [comment]
> Still needs some work. Some very basic benchmarks using `-m cProfile -s tottime` would also be appreciated as a sanity check.

Sure. I will post the performance benchmark results in the PR later.

File: torchtext/datasets/unsupervised_learning.py (hunk 6 above)
Comment (zhangguanheng66) [reply torchtext/datasets/unsupervised_learning.py]: The original file contains html, image, etc and only texts are applicable for the word vector model.

File: torchtext/datasets/unsupervised_learning.py (hunk 7 above)
Comment (zhangguanheng66) [reply torchtext/datasets/unsupervised_learning.py]: the download function will automatically skip if the downloaded file is detected.

File: torchtext/datasets/__init__.py (hunk 9 above)
Comment (zhangguanheng66) [reply torchtext/datasets/__init__.py]: confused. This is grouped together with other `import`.

File: torchtext/data/functional.py (hunk 1 above)
Comment (zhangguanheng66) [reply torchtext/data/functional.py]: I feel it's a very small change so I just made it here :).

File: torchtext/data/functional.py [hunk 10]
```diff
@@ -114,3 +115,39 @@ def _internal_func(txt_iter):
...
+        >>> custom_replace_transform = custom_replace([(r'S', 's'), (r'\s+', ' ')])
+        >>> list_a = ["Sentencepiece encode  aS  pieces", "exampleS to   try!"]
+        >>> list(custom_replace_transform(list_a))
+            ['sentencepiece encode as pieces', 'examples to try!']
+    """
+
+    _patterns = list((re.compile(p), r)
+                     for (p, r) in replace_pattern)
+
+    def _internal_func(txt_iter):
```
vincentqb: What's the goal of wrapping the yield inside a function? Could you remove the `def` and the `return _internal_func`?

torchtext/datasets/unsupervised_learning.py
vincentqb: If the file was downloaded then extracted, and then the user deletes the downloaded file, it will be downloaded again even though it's already extracted, right?

torchtext/datasets/unsupervised_learning.py
```diff
@@ -0,0 +1,136 @@
...
+        return self._data[i]
+
+    def __len__(self):
+        return len(self._data)
+
+    def __iter__(self):
+        for x in self._data:
+            yield x
+
+    def get_vocab(self):
```
vincentqb: Will the vocab be part of the dataset? We had a discussion internally [here](https://fb.quip.com/LncwAsC1cUZt) about this.

torchtext/datasets/unsupervised_learning.py
```diff
@@ -0,0 +1,136 @@
...
+
+        self._vocab = None
+
+    def __getitem__(self, i):
+        return self._data[i]
+
+    def __len__(self):
+        return len(self._data)
+
+    def __iter__(self):
```
vincentqb: Since you have `__getitem__` and `__len__`, `__iter__` is implied.

torchtext/datasets/unsupervised_learning.py
```diff
@@ -0,0 +1,136 @@
...
+            >>> from torchtext.datasets import EnWik9
+            >>> enwik9 = EnWik9(num_lines=20000)
+            >>> vocab = enwik9.get_vocab()
+        """
+
+        super(EnWik9, self).__init__()
+
+        processed_file = os.path.join(root, 'norm_enwik9')
+        if not os.path.exists(processed_file):
+            url = 'http://mattmahoney.net/dc/enwik9.zip'
```
vincentqb: The user may want to customize the URL, etc.

torchtext/datasets/unsupervised_learning.py
```diff
@@ -0,0 +1,136 @@
...
+            extracted_file = extract_archive(dataset_zip)
+            raw_file = os.path.join(root, extracted_file[0])
+            preprocess_raw_enwik9(raw_file, processed_file)
+
+        # Meta information
+        offsets = generate_offsets(processed_file)
+        read_lines = read_lines_from_iterator(processed_file,
+                                              offsets, begin_line, num_lines)
+
+        self._data = []
```
vincentqb: The data is all loaded in memory and kept there. Is that desired?

torchtext/data/functional.py
zhangguanheng66: The `custom_replace` func returns `_internal_func`, which accepts `txt_iter` for replacement.  I could make those two as a single func but then, I have to re.compile the replace pattern every time.

torchtext/datasets/unsupervised_learning.py
zhangguanheng66: For this dataset, no. But we could make a more general case for similar datasets.

torchtext/datasets/unsupervised_learning.py
zhangguanheng66: Yes. If downloaded file is deleted, then it will be downloaded again. But I have a `if` check above to see if the processed_file exists. And in this case, downloading is much more expensive than extracting.

torchtext/datasets/unsupervised_learning.py
zhangguanheng66: Not all the data. Only partial data are loaded to the memory depending on `begin_line` and `num_lines`

torchtext/datasets/unsupervised_learning.py
zhangguanheng66: Yes. I don't need to have __iter__() here if we decide to use the map pattern.

torchtext/datasets/unsupervised_learning.py
zhangguanheng66: There are some arguments for this. For the old torchtext, `vocab` stay with `Field` and we don't like that. However, I also don't want a `vocab` apart completely from datasets. We still need to find a way to carry `vocab` somewhere.

PR #613, opened 2019-10-07
[WIP] added xnli dataset

bentrevett:
This is initial work on adding the [XNLI dataset](http://www.nyu.edu/projects/bowman/xnli/) to TorchText.

Weirdly, the XNLI dataset only has a "dev" and a "test" set, no training set. This means I had to make some decisions that I'm not sure were for the best.

I've either removed or hard coded the train argument to be None, which means the XNLI dataset works fine with .splits, but not with .iters. due to [these lines](https://github.com/pytorch/text/blob/master/torchtext/datasets/nli.py#L120) being hardcoded. My solution was to simply raise a `NotImplementedError` when trying to do `XNLI.iters`. Hopefully this is sufficient.

The XNLI dataset is used for "cross-language sentence encoding", so I have also added a language_field.


zhangguanheng66:
Change the title of the PR. Once you have done and need a review, please let us know. Thanks for your contributions.

bentrevett:
@zhangguanheng66 What did you want it changed to?

zhangguanheng66:
> @zhangguanheng66 What did you want it changed to?

Just mark it as WIP. Once you have done, we could start the reviewing.

test/nli.py
```diff
@@ -64,3 +64,31 @@
...
+
+TEXT.build_vocab(val)
+LABEL.build_vocab(val)
+GENRE.build_vocab(val, test)
+LANGUAGE.build_vocab(val, test)
+
+val_iter, test_iter = data.Iterator.splits((val, test), batch_size=3)
+
+batch = next(iter(val_iter))
+print("Numericalize premises:\n", batch.premise)
```
zhangguanheng66: In general, we don't have printout for the unit tests. You could use `Assert` to check the values.

torchtext/datasets/nli.py
```diff
@@ -161,3 +161,29 @@ def splits(cls, text_field, label_field, parse_field=None, genre_field=None,
...
+            extra_fields["language"] = ("language", language_field)
+
+        return super(XNLI, cls).splits(text_field, label_field,
+                                       extra_fields=extra_fields,
+                                       root=root, train=None,
+                                       validation=validation, test=test)
+
+    @classmethod
+    def iters(cls, *args, **kwargs):
+        raise NotImplementedError('XNLI dataset does not support iters')
```
zhangguanheng66: So why XNLI dataset doesn't support iters?

test/nli.py
bentrevett: All of the other dataset tests, i.e. test/imdb.py, test/sequence_tagging.py, all do it this way. I can re-write the whole of the test/nli.py to use unit tests though?

torchtext/datasets/nli.py
bentrevett: Because of [this](https://github.com/pytorch/text/blob/master/torchtext/datasets/nli.py#L120) line onward. `NLIDataset` always assumes there is a training set, which is not the case for the XNLI dataset. I can edit the `NLIDataset` class to check if `train` is `None` and act accordingly?

test/nli.py
bentrevett: unit tests have been added to the whole of test/nli.py now.

zhangguanheng66:
@bentrevett Do you need a review for this PR?

bentrevett:
@zhangguanheng66 Yes please.

zhangguanheng66:
Rebase to the master branch. Will merge after all the unit tests pass. Thanks for the contributions @bentrevett

bentrevett:
@zhangguanheng66 Sorry, still new to git. It looks like you've rebased to the master for me, is there anything I need to do on my end?

zhangguanheng66:
> @zhangguanheng66 Sorry, still new to git. It looks like you've rebased to the master for me, is there anything I need to do on my end?

No. I will just merge the PR. Thanks for the contributions.

PR #614, opened 2019-10-09
[DO NOT MERGE]fix Dataset.split() method

taufique74:
When **Dataset.split()** method is called with **split_ratio** as a list of **[train, test, val]**,
it always switches the **test** and **val** while returning.

### DIFF
--- a/torchtext/data/dataset.py
+++ b/torchtext/data/dataset.py
@@ -127,7 +127,7 @@ class Dataset(torch.utils.data.Dataset):
                 val_data += group_val
 
         splits = tuple(Dataset(d, self.fields)
-                       for d in (train_data, val_data, test_data) if d)
+                       for d in (train_data, test_data, val_data) if d)
 
         # In case the parent sort key isn't none
         if self.sort_key:
--- a/torchtext/data/dataset.py
+++ b/torchtext/data/dataset.py
@@ -125,7 +125,8 @@ class Dataset(torch.utils.data.Dataset):
                 train_data += group_train
                 test_data += group_test
                 val_data += group_val
-
+        
+        test_data, val_data = val_data, test_data
         splits = tuple(Dataset(d, self.fields)
                        for d in (train_data, test_data, val_data) if d)
 
--- a/torchtext/data/dataset.py
+++ b/torchtext/data/dataset.py
@@ -125,7 +125,7 @@ class Dataset(torch.utils.data.Dataset):
                 train_data += group_train
                 test_data += group_test
                 val_data += group_val
-        
+
         test_data, val_data = val_data, test_data
         splits = tuple(Dataset(d, self.fields)
                        for d in (train_data, test_data, val_data) if d)


zhangguanheng66:
This one will significantly break the current usage of torchtext. Instead of changing the code, we should change the docs, if necessary. @mttk

taufique74:
Then the doc shuld be modified as this is an unwanted behavior.

taufique74:
Swapping ```test_data``` and ```val_data``` before returning seems like did the job.
Can you please confirm whether it still breaks torchtext? @zhangguanheng66 
Thanks.

zhangguanheng66:
If my understanding is correct, you just need to update the docs.

taufique74:
Oh, I see.
Whenever I called 
```python
train, test, val = Dataset.split(examples, split_ratio=[0.75, 0.2, 0.05], random_state)
```
I got 37500, 2500, 10000 respectively for train, test and val instead of getting 37500, 10000, 2500 when examples size is 50000. Although switching the second and third ratio value got the job done.
Lets see what @mttk thinks, I just wish to see this issue resolved as it's kind of a pain to switch the ratios everytime I call split().
Thanks.

split

mttk:
I'm on mobile only for a bit, so only a brief comment:
- the order of split ratios in the arguments is such as we allow for valid to be empty (so, if one number is missing it's the valid ratio)
- you shouldn't swap test and valid (L129) because we allow for valid to be empty (if the number of examples is too small, valid size can be zero). This is an extreme edge case but I like to have it covered.
- I agree that the current return order isn't logical, it felt more 'natural' to return the standard [train, test, val] order of splits. We can change this to return [train, val, test] in order to be consistent with the ratio argument. IMO, only the change in L130 should be enough for this. This will break backward compat though, so we should label it as such.

I'm fine with keeping it as is, as well as changing it. Aesthetically, I like changing the order of returned splits to match the ratio arguments more.

taufique74:
I went over the code again, val_ratio could be zero because of rounding, and it will return an empty list as you said. Therefore, swapping test and val was such a bad idea. What I actually had in mind is passing the test.

In my first [commit](https://github.com/pytorch/text/pull/614/commits/368de8d73950af2ca3b06327bd98837e7853451c), I did exactly what you said, changing the order in L130, although it didn't pass some of the tests.

To be honest, I don't actually mind since I read the code and I know what's happening under the hood. However, as a huge fan of PyTorch's readability and pythonicness, this feels aesthetically unpleasing to me too. 

In any case, please let me know if you decide on changing the order of the return statement or changing the doc, I'd be more than happy to contribute.

mttk:
Yeah, the failing test is based on the old return order and should also be changed.
I agree that it's more aesthetically pleasing with the change so I'd vote for it.
@zhangguanheng66 is this change going to be relevant for the new dataset patterns or will it be obsolete? If it's the latter, I'd rather just update the docs.

zhangguanheng66:
We will encourage to use `random_split` in `torch.utils.data.dataset`. So I think we could just update the docs.

taufique74:
That would be a safe option I guess :)

zhangguanheng66:
> That would be the safe option I guess :)

Let me know if you have completed the PR and need a review.

taufique74:
As you said, we could just use `random_split` since `torchtext.data.Dataset` inherits from `torch.utils.data.Dataset`, although `random_split` lacks some functionalities like it doesn't have a stratified sampling option.

I'm not quite sure how the doc should be modified, or what are the things that should be mentioned in this particular case. 
You can guide me and provide some hints, or you can also modify yourself, either way is fine. Let me know if I should close the request.

Nayef211:
Closing since this is no longer relevant. We no longer use Dataset.split() and instead use a new functional API.

Issue #612, opened 2019-10-06
TabularDataset fails if lines start with "


neychev:
## 🐛 Bug
**Describe the bug**
`torchtext.data.TabularDataset` generates corrupted dataset from TSV file if one of the lines starts with quotation mark `"` 

**Example script to reproduce the bug is attached below**.

**To Reproduce**
1. Create file with one of the lines starting with `"` mark.
2. Create `TabularDataset` from this file.

**Expected behavior**
Dataset is generated correctly.

**Environment**

 - PyTorch Version: 1.2
 - OS: MacOS and Linux Ubuntu 18.04
 - How you installed PyTorch: pip
 - Build command you used (if compiling from source):
 - Python version: 3.6.7 and 3.7.3 respectively
 - CUDA/cuDNN version: None and CUDA 10.1 respectively
 - GPU models and configuration: None
 - Any other relevant information: torchtext version 0.4.0

**Example code to reproduce**:
```
#!/usr/bin/env python
# coding: utf-8

# In[1]:
import torchtext
from torchtext.datasets import TranslationDataset
from torchtext.data import Field, BucketIterator

from nltk.tokenize import WordPunctTokenizer

# In[2]:
broken_lines = [
    'First\tfirst\n',
    '"Second\tsecond\n',
    'Third "sentence"\tthird\n',
    'Fourth\tfourth\n',
    'Fifth\tfifth\n',
    'Sixth\tsixth\n'
]

broken_lines_another_location = [
    'First\tfirst\n',
    '"Second\tsecond\n',
    'Third\tthird\n',
    'Fourth\tfourth\n',
    'Fifth"sentence"\tfifth\n',
    'Sixth\tsixth\n'
]

fixed_lines = [
    'First\tfirst\n',
    'Second\tsecond\n',
    'Third "sentence"\tthird\n',
    'Fourth\tfourth\n',
    'Fifth\tfifth\n',
    'Sixth\tsixth\n'
]

with open('broken_lines.txt', 'w') as iofile:
    iofile.writelines(broken_lines)
with open('broken_lines_another_location.txt', 'w') as iofile:
    iofile.writelines(broken_lines_another_location)
with open('fixed_lines.txt', 'w') as iofile:
    iofile.writelines(fixed_lines)

# In[3]:
tokenizer_W = WordPunctTokenizer()
def tokenize(x, tokenizer=tokenizer_W):
    return tokenizer.tokenize(x.lower())

# In[4]:
[...]

PR #616, opened 2019-10-10
Make TabularDataset ignore quoting

hawkeoni:
This comes from https://github.com/pytorch/text/issues/612?_pjax=%23js-repo-pjax-container

Default behaviour while reading csv is to ignore delimeters between quotes, thus in cases with imbalanced quoting TabularDataset reads the data in a wrong way. Due to nature of our domain I believe it is important to ignore quoting in the dataset as a default parameter.

### DIFF
--- a/torchtext/data/dataset.py
+++ b/torchtext/data/dataset.py
@@ -1,5 +1,6 @@
 import io
 import os
+import csv
 import zipfile
 import tarfile
 import gzip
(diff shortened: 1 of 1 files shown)


hawkeoni:
The problem is in tests, which test how TabularDataset deals with newlines in quoting. In default mode it ignores newlines inside quotes, in my PR it ignores quoting. I'm not so sure about this PR being correct default behavior.

PR #621, opened 2019-10-21
Reduced data.dataset.stratify() to one iteration

ghost:
old:
```python
def stratify(examples, strata_field):
    unique_strata = set(getattr(example, strata_field))
    strata_maps = {s: [] for s in unique_strata}	
    for example in examples:
        strata_maps[getattr(example, strata_field)].append(example)
    return list(strata_maps.values())
```

new:
```python
def stratify(examples, strata_field):	
  strata_maps = defaultdict(list)
  for example in examples:	   
  	strata_maps[getattr(example, strata_field)].append(example)
  return list(strata_maps.values())
```

**In fact, in some cases, the original method is equivalent to traversing the `examples` three times.**

@mttk @zhangguanheng66


JahoJiang:
Actually, for some cases, the original method will traverse `examples` three times.

mttk:
This is fine, but due to the new direction that torchtext is taking, I'm not sure if it's necessary. @zhangguanheng66

Nayef211:
Closing since `Dataset` class no longer exists in torchtext.

PR #624, opened 2019-10-23
Re-write language_modeling datasets (PennTreebank, WikiText103, WikiText2)

zhangguanheng66:
Re-write the language modeling datasets with a new pattern which was first applied for text classification datasets in v.0.4.0 release.

## Motivation
The motivation for the new pattern is to simplify data processing in Torchtext and grant users more flexibility to build the pipeline. There are three major issues we want to solve:

- Remove the dependency of `Field` class, which couples tokenizer, vocabulary, split, batching and sampling, padding, and numericalization together. It's like a "black box" and confuses users about what's going on inside it. Instead, with the new pattern, the components mentioned above should be some basic building blocks. Users can build the data processing pipeline with the orthogonal components.
- Incompatible with `DataLoader` and `Sampler` in `torch.utils.data`. Some duplicate functionals (e.g. `Iterator`, `Batch`, `splits` in `torchtext`) should be replaced by the corresponding functionals in `torch.utils.data` to reduce the maintaince efforts.
- Unnecessary data structure. For example, `Example` class adds no structure, and should be replaced with tuple/dict or namedtuple.

## API for new language modeling datasets
To support "one-command" data loading, we have built a pipeline here to support PennTreebank, WikiText103, WikiText2. Users are welcome to build their ones if they follow the pattern. To load the new datasets, simple call the dataset API, as follow:
```
from torchtext.datasets import WikiText2
train_dataset, valid_dataset, test_dataset = WikiText2()
```
If you want to use a specific tokenizer, 
```
tokenizer = get_tokenizer("spacy")
train_dataset, valid_dataset, test_dataset = WikiText2(tokenizer=tokenizer)
```
If you just want the valid set:
```
vocab = train_dataset.get_vocab()
valid_dataset, = WikiText2(tokenizer=tokenizer, vocab=vocab, data_select='valid')
```
## Legacy code
We decide to move the old language modeling datasets (a.k.a. PennTreebank, WikiText103, WikiText2) to a legacy folder `torchtext.legacy.datasets`. In the past, you may use those datasets as follow:
```
import torchtext.data as data
from torchtext.legacy.datasets import WikiText2
TEXT = data.Field(lower=True, batch_first=True)
train_dataset, valid_dataset, test_dataset = WikiText2.splits(TEXT)
```

You can still use the legacy datasets, as follow:
```
import torchtext.data as data
from torchtext.legacy.datasets import WikiText2
TEXT = data.Field(lower=True, batch_first=True)
train_dataset, valid_dataset, test_dataset = WikiText2.splits(TEXT)
``` 

## Difference

- With the old pattern, users have to create a `Field` object including a specific tokenizer. In the new dataset API, user can pass the tokenizer directly to the dataset constructor

```
# Old pattern
TEXT = torchtext.data.Field(tokenize=get_tokenizer("basic_english"))
```
```
# New pattern
train_dataset, valid_dataset, test_dataset = WikiText2(tokenizer=get_tokenizer("spacy"))
```

- In the old dataset, `vocab` object is associated with `Field` class, and there is no way to apply a pre-trained `vocab` object. In the new dataset, `vocab` object can be obtained by
```
vocab = train_dataset.get_vocab()
```
and apply to generate new datasets
```
train_dataset, valid_dataset, test_dataset = WikiText103(vocab=vocab)
```

- The datasets with the new pattern return a tensor of token IDs, instead of tokens in the old pattern. If users would like to check the tokens, simply use the following command:
```
train_vocab = train_dataset.get_vocab()
tokens = [train_vocab.itos[id] for id in train_dataset]
```

- Unlike the old pattern using `BucketIterator.splits`, users are encouraged to use `torch.utils.data.DataLoader` to generate batches of data. You can specify how exactly the samples need to be batched using `collate_fn`.
```
# Generate 8x8 batches
import torch
from torch.utils.data import DataLoader
num_row = 8
def generate_rows(data):
	data = torch.tensor(data).view(num_row, -1).t().contiguous()
	return data
dataloader = DataLoader(train_dataset, batch_size=64, num_workers=4, collate_fn=generate_rows)
for batch in dataloader:
    # Send batch to the model.
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -129,6 +129,16 @@ Others are planned or a work in progress:
 
 See the ``test`` directory for examples of dataset usage.
 
+Legacy Code
+===========
+
+We are currently retiring several datasets as legacy code ```torchtext.legacy```:
+
+* Sentiment analysis: IMDb
+* Language modeling: abstract class + WikiText-2, WikiText103, PennTreebank
+
+These datasets are re-written with a new pattern that is introduced in `Release v0.5.0 <https://github.com/pytorch/text/releases>`_.
+
 Disclaimer on Datasets
 ======================
 
--- a/README.rst
+++ b/README.rst
@@ -132,7 +132,7 @@ See the ``test`` directory for examples of dataset usage.
 Legacy Code
 ===========
 
-We are currently retiring several datasets as legacy code ```torchtext.legacy```:
+We have currently retired several datasets and moved them under ```torchtext.legacy```:
 
 * Sentiment analysis: IMDb
 * Language modeling: abstract class + WikiText-2, WikiText103, PennTreebank
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -37,9 +37,9 @@ class TestDataset(TorchtextTestCase):
         from torchtext.datasets import WikiText2
         # smoke test to ensure wikitext2 works properly
         train_dataset, test_dataset, valid_dataset = WikiText2()
-        self.assertEqual(len(train_dataset), 1947375)
-        self.assertEqual(len(test_dataset), 230357)
-        self.assertEqual(len(valid_dataset), 203947)
+        self.assertEqual(len(train_dataset), 2049990)
+        self.assertEqual(len(test_dataset), 241859)
+        self.assertEqual(len(valid_dataset), 214417)
 
         vocab = train_dataset.get_vocab()
         tokens_ids = [vocab[token] for token in 'the player characters rest'.split()]
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -43,7 +43,7 @@ class TestDataset(TorchtextTestCase):
 
         vocab = train_dataset.get_vocab()
         tokens_ids = [vocab[token] for token in 'the player characters rest'.split()]
-        self.assertEqual(tokens_ids, [2, 285, 502, 699])
+        self.assertEqual(tokens_ids, [[2, 286, 503, 700]])
 
         # Delete the dataset after we're done to save disk space on CI
         datafile = os.path.join(self.project_root, ".data", "wikitext-2")
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -1,4 +1,5 @@
 import os
+import shutil 
 import torchtext.data as data
 from torchtext.datasets import AG_NEWS
 
@@ -9,6 +10,8 @@ from ..common.torchtext_test_case import TorchtextTestCase
 def conditional_remove(f):
     if os.path.isfile(f):
         os.remove(f)
+    elif os.path.isdir(f):
+        shutil.rmtree(f)
 
 
 class TestDataset(TorchtextTestCase):
@@ -46,6 +49,8 @@ class TestDataset(TorchtextTestCase):
         # Delete the dataset after we're done to save disk space on CI
         if os.environ.get("TRAVIS") == "true":
             datafile = os.path.join(self.project_root, ".data", "wikitext-2")
+            conditional_remove(datafile)
+            datafile = os.path.join(self.project_root, ".data", "wikitext-2-v1.zip")
             conditional_remove(datafile)
 
     @slow
@@ -100,5 +105,8 @@ class TestDataset(TorchtextTestCase):
 
         # Delete the dataset after we're done to save disk space on CI
         if os.environ.get("TRAVIS") == "true":
-            datafile = os.path.join(self.project_root, ".data", "AG_NEWS")
+            datafile = os.path.join(self.project_root, ".data", "ag_news_csv")
+            conditional_remove(datafile)
+
+            datafile = os.path.join(self.project_root, ".data", "ag_news_csv.tar.gz")
             conditional_remove(datafile)
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -30,9 +30,8 @@ class TestDataset(TorchtextTestCase):
                                                      bptt_len=30)
 
         # Delete the dataset after we're done to save disk space on CI
-        if os.environ.get("TRAVIS") == "true":
-            datafile = os.path.join(self.project_root, ".data", "wikitext-2")
-            conditional_remove(datafile)
+        datafile = os.path.join(self.project_root, ".data", "wikitext-2")
+        conditional_remove(datafile)
 
     def test_wikitext2(self):
         from torchtext.datasets import WikiText2
@@ -47,11 +46,10 @@ class TestDataset(TorchtextTestCase):
         self.assertEqual(tokens_ids, [2, 285, 502, 699])
 
         # Delete the dataset after we're done to save disk space on CI
-        if os.environ.get("TRAVIS") == "true":
-            datafile = os.path.join(self.project_root, ".data", "wikitext-2")
-            conditional_remove(datafile)
-            datafile = os.path.join(self.project_root, ".data", "wikitext-2-v1.zip")
-            conditional_remove(datafile)
+        datafile = os.path.join(self.project_root, ".data", "wikitext-2")
+        conditional_remove(datafile)
+        datafile = os.path.join(self.project_root, ".data", "wikitext-2-v1.zip")
+        conditional_remove(datafile)
 
     @slow
     def test_penntreebank_legacy(self):
@@ -68,9 +66,8 @@ class TestDataset(TorchtextTestCase):
                                                      bptt_len=30)
 
         # Delete the dataset after we're done to save disk space on CI
-        if os.environ.get("TRAVIS") == "true":
-            datafile = os.path.join(self.project_root, ".data", "penn-treebank")
-            conditional_remove(datafile)
+        datafile = os.path.join(self.project_root, ".data", "penn-treebank")
+        conditional_remove(datafile)
 
     def test_penntreebank(self):
         from torchtext.datasets import PennTreebank
@@ -85,13 +82,12 @@ class TestDataset(TorchtextTestCase):
         self.assertEqual(tokens_ids, [2, 2550, 3344, 1125])
 
         # Delete the dataset after we're done to save disk space on CI
-        if os.environ.get("TRAVIS") == "true":
-            datafile = os.path.join(self.project_root, ".data", 'ptb.train.txt')
-            conditional_remove(datafile)
-            datafile = os.path.join(self.project_root, ".data", 'ptb.test.txt')
-            conditional_remove(datafile)
-            datafile = os.path.join(self.project_root, ".data", 'ptb.valid.txt')
-            conditional_remove(datafile)
+        datafile = os.path.join(self.project_root, ".data", 'ptb.train.txt')
+        conditional_remove(datafile)
+        datafile = os.path.join(self.project_root, ".data", 'ptb.test.txt')
+        conditional_remove(datafile)
+        datafile = os.path.join(self.project_root, ".data", 'ptb.valid.txt')
+        conditional_remove(datafile)
 
     def test_text_classification(self):
         # smoke test to ensure ag_news dataset works properly
@@ -104,8 +100,7 @@ class TestDataset(TorchtextTestCase):
         self.assertEqual(len(ag_news_test), 7600)
 
         # Delete the dataset after we're done to save disk space on CI
-        if os.environ.get("TRAVIS") == "true":
-            datafile = os.path.join(self.project_root, ".data", "ag_news_csv")
-            conditional_remove(datafile)
-            datafile = os.path.join(self.project_root, ".data", "ag_news_csv.tar.gz")
-            conditional_remove(datafile)
+        datafile = os.path.join(self.project_root, ".data", "ag_news_csv")
+        conditional_remove(datafile)
+        datafile = os.path.join(self.project_root, ".data", "ag_news_csv.tar.gz")
+        conditional_remove(datafile)
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -1,6 +1,5 @@
 import os
 import torchtext.data as data
-from torchtext.datasets import WikiText2, PennTreebank
 from torchtext.datasets import AG_NEWS
 
 from ..common.test_markers import slow
@@ -14,7 +13,8 @@ def conditional_remove(f):
 
 class TestDataset(TorchtextTestCase):
     @slow
-    def test_wikitext2(self):
+    def test_wikitext2_legacy(self):
+        from torchtext.legacy.datasets import WikiText2
         # smoke test to ensure wikitext2 works properly
         ds = WikiText2
         TEXT = data.Field(lower=True, batch_first=True)
@@ -31,8 +31,26 @@ class TestDataset(TorchtextTestCase):
             datafile = os.path.join(self.project_root, ".data", "wikitext-2")
             conditional_remove(datafile)
 
+    def test_wikitext2(self):
+        from torchtext.datasets import WikiText2
+        # smoke test to ensure wikitext2 works properly
+        train_dataset, test_dataset, valid_dataset = WikiText2()
+        self.assertEqual(len(train_dataset), 1947375)
+        self.assertEqual(len(test_dataset), 230357)
+        self.assertEqual(len(valid_dataset), 203947)
+
+        vocab = train_dataset.get_vocab()
+        tokens_ids = [vocab[token] for token in 'the player characters rest'.split()]
+        self.assertEqual(tokens_ids, [2, 285, 502, 699])
+
+        # Delete the dataset after we're done to save disk space on CI
+        if os.environ.get("TRAVIS") == "true":
+            datafile = os.path.join(self.project_root, ".data", "wikitext-2")
+            conditional_remove(datafile)
+
     @slow
-    def test_penntreebank(self):
+    def test_penntreebank_legacy(self):
+        from torchtext.legacy.datasets import PennTreebank
         # smoke test to ensure penn treebank works properly
         TEXT = data.Field(lower=True, batch_first=True)
         ds = PennTreebank
@@ -47,6 +65,27 @@ class TestDataset(TorchtextTestCase):
         # Delete the dataset after we're done to save disk space on CI
         if os.environ.get("TRAVIS") == "true":
             datafile = os.path.join(self.project_root, ".data", "penn-treebank")
+            conditional_remove(datafile)
+
+    def test_penntreebank(self):
+        from torchtext.datasets import PennTreebank
+        # smoke test to ensure wikitext2 works properly
+        train_dataset, test_dataset, valid_dataset = PennTreebank()
+        self.assertEqual(len(train_dataset), 924412)
+        self.assertEqual(len(test_dataset), 82114)
+        self.assertEqual(len(valid_dataset), 73339)
+
+        vocab = train_dataset.get_vocab()
+        tokens_ids = [vocab[token] for token in 'the player characters rest'.split()]
+        self.assertEqual(tokens_ids, [2, 2550, 3344, 1125])
+
+        # Delete the dataset after we're done to save disk space on CI
+        if os.environ.get("TRAVIS") == "true":
+            datafile = os.path.join(self.project_root, ".data", 'ptb.train.txt')
+            conditional_remove(datafile)
+            datafile = os.path.join(self.project_root, ".data", 'ptb.ttest.txt')
+            conditional_remove(datafile)
+            datafile = os.path.join(self.project_root, ".data", 'ptb.valid.txt')
             conditional_remove(datafile)
 
     def test_text_classification(self):
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -88,7 +88,7 @@ class TestDataset(TorchtextTestCase):
         if os.environ.get("TRAVIS") == "true":
             datafile = os.path.join(self.project_root, ".data", 'ptb.train.txt')
             conditional_remove(datafile)
-            datafile = os.path.join(self.project_root, ".data", 'ptb.ttest.txt')
+            datafile = os.path.join(self.project_root, ".data", 'ptb.test.txt')
             conditional_remove(datafile)
             datafile = os.path.join(self.project_root, ".data", 'ptb.valid.txt')
             conditional_remove(datafile)
@@ -107,6 +107,5 @@ class TestDataset(TorchtextTestCase):
         if os.environ.get("TRAVIS") == "true":
             datafile = os.path.join(self.project_root, ".data", "ag_news_csv")
             conditional_remove(datafile)
-
             datafile = os.path.join(self.project_root, ".data", "ag_news_csv.tar.gz")
             conditional_remove(datafile)
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -43,7 +43,7 @@ class TestDataset(TorchtextTestCase):
 
         vocab = train_dataset.get_vocab()
         tokens_ids = [vocab[token] for token in 'the player characters rest'.split()]
-        self.assertEqual(tokens_ids, [[2, 286, 503, 700]])
+        self.assertEqual(tokens_ids, [2, 286, 503, 700])
 
         # Delete the dataset after we're done to save disk space on CI
         datafile = os.path.join(self.project_root, ".data", "wikitext-2")
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -1,5 +1,5 @@
 import os
-import shutil 
+import shutil
 import torchtext.data as data
 from torchtext.datasets import AG_NEWS
 
--- a/torchtext/data/functional.py
+++ b/torchtext/data/functional.py
@@ -188,20 +188,18 @@ def create_data_from_iterator(vocab, iterator, removed_tokens=None):
         >>> from torchtext.data.functional import simple_space_split
         >>> from torchtext.data.functional import create_data_from_iterator
         >>> vocab = {'Sentencepiece' : 0, 'encode' : 1, 'as' : 2, 'pieces' : 3}
-        >>> list(create_data_from_iterator(vocab,
-        >>>                                simple_space_split(["Sentencepiece as pieces",
+        >>> ids_iter = create_data_from_iterator(vocab,
+        >>>                               simple_space_split(["Sentencepiece as pieces",
         >>>                                                   "as pieces"]))
-        >>> [[0, 2, 3], [2, 3]]
+        >>> for ids in ids_iter:
+        >>>     print([num for num in ids])
+        >>> [0, 2, 3]
+        >>> [2, 3]
     """
 
     for tokens in iterator:
         if removed_tokens is None:
-            tokens = [vocab[token] for token in tokens]
+            yield iter(vocab[token] for token in tokens)
         else:
-            token_ids = list(filter(lambda x:
-                                    x not in [vocab[token] for token in removed_tokens],
-                                    [vocab[token] for token in tokens]))
-            tokens = token_ids
-        if len(tokens) == 0:
-            logging.info('Row contains no tokens.')
-        yield tokens
+            tokens = list(filter(lambda x: x not in removed_tokens, tokens))
+            yield iter(vocab[token] for token in tokens)
--- a/torchtext/data/functional.py
+++ b/torchtext/data/functional.py
@@ -195,9 +195,9 @@ def create_data_from_iterator(vocab, iterator, removed_tokens=None):
         >>> [0, 2, 3]
         >>> [2, 3]
     """
-    print("new iterator data")
     for tokens in iterator:
         if removed_tokens is None:
             yield iter(vocab[token] for token in tokens)
         else:
-            yield iter(map(vocab, filter(lambda x: x not in removed_tokens, tokens)))
+            yield iter(map(lambda x: vocab[x],
+                       filter(lambda x: x not in removed_tokens, tokens)))
--- a/torchtext/data/functional.py
+++ b/torchtext/data/functional.py
@@ -1,10 +1,8 @@
-import torch
 import sentencepiece as spm
 import re
 from torchtext.utils import unicode_csv_reader
 import logging
 import io
-from torchtext.vocab import Vocab
 
 __all__ = [
     "generate_sp_model", "load_sp_model",
@@ -179,7 +177,7 @@ def read_text_iterator(path, tokenizer):
 
 
 def create_data_from_iterator(vocab, iterator, removed_tokens=None):
-    r"""Create data from an token iterator.
+    r"""Yield a list of ids from an token iterator with a vocab.
 
     Arguments:
         vocab: the vocabulary convert token into id.
@@ -190,13 +188,12 @@ def create_data_from_iterator(vocab, iterator, removed_tokens=None):
         >>> from torchtext.data.functional import simple_space_split
         >>> from torchtext.data.functional import create_data_from_iterator
         >>> vocab = {'Sentencepiece' : 0, 'encode' : 1, 'as' : 2, 'pieces' : 3}
-        >>> create_data_from_iterator(vocab,
-        >>>                           simple_space_split(["Sentencepiece as pieces"]),
-        >>>                           removed_tokens=['<unk>'])
-        >>> tensor([0, 2, 3])
+        >>> list(create_data_from_iterator(vocab,
+        >>>                                simple_space_split(["Sentencepiece as pieces",
+        >>>                                                   "as pieces"]))
+        >>> [[0, 2, 3], [2, 3]]
     """
 
-    _data = []
     for tokens in iterator:
         if removed_tokens is None:
             tokens = [vocab[token] for token in tokens]
@@ -206,5 +203,4 @@ def create_data_from_iterator(vocab, iterator, removed_tokens=None):
             tokens = token_ids
         if len(tokens) == 0:
             logging.info('Row contains no tokens.')
-        _data += tokens
-    return torch.Tensor(_data).long()
+        yield tokens
--- a/torchtext/data/functional.py
+++ b/torchtext/data/functional.py
@@ -198,8 +198,9 @@ def create_data_from_iterator(vocab, iterator, removed_tokens=None):
         if removed_tokens is None:
             tokens = [vocab[token] for token in tokens]
         else:
-            token_ids = list(filter(lambda x: x not in removed_tokens, [vocab[token]
-                                    for token in tokens]))
+            token_ids = list(filter(lambda x:
+                                    x not in [vocab[token] for token in removed_tokens],
+                                    [vocab[token] for token in tokens]))
             tokens = token_ids
         if len(tokens) == 0:
             logging.info('Row contains no tokens.')
--- a/torchtext/data/functional.py
+++ b/torchtext/data/functional.py
@@ -201,5 +201,5 @@ def create_data_from_iterator(vocab, iterator, removed_tokens=None):
         if removed_tokens is None:
             yield iter(vocab[token] for token in tokens)
         else:
-            tokens = list(filter(lambda x: x not in removed_tokens, tokens))
-            yield iter(vocab[token] for token in tokens)
+            yield iter(vocab[token] for token in
+                       filter(lambda x: x not in removed_tokens, tokens))
--- a/torchtext/data/functional.py
+++ b/torchtext/data/functional.py
@@ -1,6 +1,10 @@
+import torch
 import sentencepiece as spm
 import re
-
+from torchtext.utils import unicode_csv_reader
+import logging
+import io
+from torchtext.vocab import Vocab
 
 __all__ = [
     "generate_sp_model", "load_sp_model",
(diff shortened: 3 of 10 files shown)


torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,213 @@
...
+    'WikiText103':
+        'https://s3.amazonaws.com/research.metamind.io/wikitext/wikitext-103-v1.zip',
+    'PennTreebank':
+        ['https://raw.githubusercontent.com/wojzaremba/lstm/master/data/ptb.train.txt',
+         'https://raw.githubusercontent.com/wojzaremba/lstm/master/data/ptb.test.txt',
+         'https://raw.githubusercontent.com/wojzaremba/lstm/master/data/ptb.valid.txt']
+}
+
+
+def _read_text_iterator(data_path, tokenizer):
```
cpuhrsch: These two functions look similar to what's in the classification datasets - is there a way to share this code via some simple abstractions?

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,213 @@
...
+    with io.open(data_path, encoding="utf8") as f:
+        reader = unicode_csv_reader(f)
+        for row in reader:
+            tokens = tokenizer(' '.join(row))
+            yield tokens
+
+
+def _create_data_from_iterator(vocab, iterator, include_unk):
+    _data = []
+    with tqdm(unit_scale=0, unit='lines') as t:
```
cpuhrsch: Is this noisy by default? Printing progress bars can cause a lot of noise.

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,213 @@
...
+    logging.info('Creating training data')
+    train_data = _create_data_from_iterator(
+        vocab, _read_text_iterator(train_path, tokenizer), include_unk)
+    logging.info('Creating testing data')
+    test_data = _create_data_from_iterator(
+        vocab, _read_text_iterator(test_path, tokenizer), include_unk)
+    logging.info('Creating valid data')
+    valid_data = _create_data_from_iterator(
+        vocab, _read_text_iterator(valid_path, tokenizer), include_unk)
+    return (LanguageModelingDataset(train_data, vocab),
```
cpuhrsch: This might be a good candidate for [namedtuples](https://docs.python.org/2/library/collections.html#collections.namedtuple).

cpuhrsch: ```datasets = namedtuple('Datasets', ['train', 'test', 'valid'], verbose=True)```

cpuhrsch:
Generally I think this looks good. In general, we need to write clear documentation that describes how to migrate to this new Dataset in comparison to the old one.

zhangguanheng66:
> Generally I think this looks good. In general, we need to write clear documentation that describes how to migrate to this new Dataset in comparison to the old one.

Yes. I could write some instructions after IMDB datasets.

torchtext/datasets/language_modeling.py
cpuhrsch: We talked about this offline and let's stay away from namedtuples until further discussion.

torchtext/data/functional.py
```diff
@@ -178,33 +176,31 @@ def read_text_iterator(path, tokenizer):
             yield tokens
 
 
-def create_data_from_iterator(vocab, iterator, include_unk):
-    r"""Create data from an token iterator.
+def create_data_from_iterator(vocab, iterator, removed_tokens=None):
```
cpuhrsch: Make sure to keep track of things that break BC for the release notes. As far as I can tell this does.

test/data/test_builtin_datasets.py
```diff
@@ -31,8 +34,28 @@ def test_wikitext2(self):
...
+        self.assertEqual(len(train_dataset), 1947375)
+        self.assertEqual(len(test_dataset), 230357)
+        self.assertEqual(len(valid_dataset), 203947)
+
+        vocab = train_dataset.get_vocab()
+        tokens_ids = [vocab[token] for token in 'the player characters rest'.split()]
+        self.assertEqual(tokens_ids, [2, 285, 502, 699])
+
+        # Delete the dataset after we're done to save disk space on CI
+        if os.environ.get("TRAVIS") == "true":
```
cpuhrsch: This seems like a generic cleanup pattern. You should use the capabilities the test framework brings along for this.

torchtext/data/functional.py
zhangguanheng66: It's the first time we add `create_data_from_iterator` in `torch.data.functional`.

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,199 @@
...
+    logging.info('Creating valid data')
+    valid_iter = create_data_from_iterator(
+        vocab, read_text_iterator(valid_path, tokenizer), removed_tokens)
+    valid_data = []
+    for tokens in valid_iter:
+        valid_data += tokens
+
+    return (LanguageModelingDataset(torch.Tensor(train_data).long(), vocab),
+            LanguageModelingDataset(torch.Tensor(test_data).long(), vocab),
+            LanguageModelingDataset(torch.Tensor(valid_data).long(), vocab))
```
vincentqb: A "dataset" is therefore a list of 3 datasets corresponding to train/test/validation? I assume the division is dictated by the source itself, right?

Could the user ever want to just get one of train/test/validation? What do you think of letting the user get one at the time when building a dataset?

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,199 @@
...
+    test_data = []
+    for tokens in test_iter:
+        test_data += tokens
+
+    logging.info('Creating valid data')
+    valid_iter = create_data_from_iterator(
+        vocab, read_text_iterator(valid_path, tokenizer), removed_tokens)
+    valid_data = []
+    for tokens in valid_iter:
+        valid_data += tokens
```
vincentqb: nit: I don't know that it's valuable to simplify this block, but it'd be more readable by just adding a space between line 91 and 92 :)

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,199 @@
...
+        for x in self._data:
+            yield x
+
+    def get_vocab(self):
+        return self._vocab
+
+
+def _setup_datasets(dataset_name, tokenizer=get_tokenizer("basic_english"),
+                    root='.data', vocab=None, removed_tokens=['<unk>']):
+    if dataset_name == 'PennTreebank':
```
vincentqb: If we use that function for future datasets, is there a more general strategy than just making `PennTreebank` a special case?

torchtext/datasets/language_modeling.py
vincentqb: For reference: [librispeech](https://github.com/pytorch/audio/blob/38d1a9b6325d8b69e9b26ce28b1f1a8e0040c51b/torchaudio/datasets/librispeech.py#L63) in audio simply lets the user chose which one to download.

zhangguanheng66: That could be an option with an extra input for that.
But it's kind of convention to generate three datasets together (if there are) and most users will need all the them for training and inference.

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,199 @@
...
+        vocab = build_vocab_from_iterator(read_text_iterator(train_path, tokenizer))
+    else:
+        if not isinstance(vocab, Vocab):
+            raise TypeError("Passed vocabulary is not of type Vocab")
+    logging.info('Vocab has {} entries'.format(len(vocab)))
+    logging.info('Creating training data')
+    train_iter = create_data_from_iterator(
+        vocab, read_text_iterator(train_path, tokenizer), removed_tokens)
+    train_data = []
+    for tokens in train_iter:
```
vincentqb: All those datasets load the data in memory, right? Would it be useful or more general to keep it out of memory with something like `train_data = itertools.chain.from_iterable(*train_iter)`?

zhangguanheng66: It's possible. For a very large dataset, we could iterate without caching.

torchtext/datasets/language_modeling.py
vincentqb: Would it be more flexible to have a function returning each separately, and then wrap all three with the API here?

Something like 
```
train = _setup_datasets_part(train)
test = _setup_datasets_part(test)
valid = _setup_datasets_part(valid)
return train, test, valid
```

cpuhrsch: the flag is probably a good idea. Someone might only want to run validation or testing in a separate process. either accept a string or a tuple of strings i'd say.

cpuhrsch: Is this consistent with the supervised datasets? If not, maybe we should add that option.

zhangguanheng66: Supervised datasets return train/test datasets all together. I feel that we should keep consistent among DAPIs or at least within torchtext. Of course, an extra input flag to select the dataset adds more flexibility :)

One usage case is concerning me: say you want valid dataset only (as proposed), then you have to call the API to generate train dataset again because the vocab object is generated based on train data. Instead, if we generate test/train/valid all together, then we re-use the vocab object for valid/test datasets.

cpuhrsch: I'd say, by default, we generate all three, but then we allow the user to create a single dataset. For that it's still up to the user whether they want to pass a custom (pretrained) Vocab, but we can mention in the docs that they probably want in most cases.

I'd expect a common usecase to be that a user generated all three in a training script and then only the test data in a separate evaluation script / context and wants to pass a Vocab. Someone might also only want to load the test dataset as yet another evaluation opportunity after having trained on another dataset (e.g. train on wikitext2, evaluate on penntreebank using a sentencepiece vocab).

On that note, if I want to inject sentencepiece as a vocab here, does that work?

vincentqb: > Instead, if we generate test/train/valid all together, then we re-use the vocab object for valid/test datasets.

What I would expect here is to have a function that produce one of the datasets given a vocab, similar to `_setup_datasets` but for only one of the datasets. We can then produce the vocab once, and pass it along when we create each datasets. Thoughts?

This can then be wrapped in a function like `_setup_datasets` which then returns all three together. This preserves consistency in torchtext, but also allows only one of the three to be generated, say as an evaluation script as mentioned by @cpuhrsch. 

> On that note, if I want to inject sentencepiece as a vocab here, does that work?

The API suggests, so: `LanguageModelingDataset(torch.Tensor(train_data).long(), vocab)` -- @zhangguanheng66 ?

zhangguanheng66: > I'd say, by default, we generate all three, but then we allow the user to create a single dataset. For that it's still up to the user whether they want to pass a custom (pretrained) Vocab, but we can mention in the docs that they probably want in most cases.
> 
> I'd expect a common usecase to be that a user generated all three in a training script and then only the test data in a separate evaluation script / context and wants to pass a Vocab. Someone might also only want to load the test dataset as yet another evaluation opportunity after having trained on another dataset (e.g. train on wikitext2, evaluate on penntreebank using a sentencepiece vocab).
> 
> On that note, if I want to inject sentencepiece as a vocab here, does that work?

Yes. As long as the `vocab` object is a callable function to convert tokens into ids.

In general, an option to generate a single dataset is not hard. I will think about the API.

zhangguanheng66: > > Instead, if we generate test/train/valid all together, then we re-use the vocab object for valid/test datasets.
> 
> What I would expect here is to have a function that produce one of the datasets given a vocab, similar to `_setup_datasets` but for only one of the datasets. We can then produce the vocab once, and pass it along when we create each datasets. Thoughts?
> 
> This can then be wrapped in a function like `_setup_datasets` which then returns all three together. This preserves consistency in torchtext, but also allows only one of the three to be generated, say as an evaluation script as mentioned by @cpuhrsch.
> 
> > On that note, if I want to inject sentencepiece as a vocab here, does that work?
> 
> The API suggests, so: `LanguageModelingDataset(torch.Tensor(train_data).long(), vocab)` -- @zhangguanheng66 ?

It's kind of the current API for `LanguageModelingDataset`. I'm thinking about some changes in `WikiText2`.

zhangguanheng66: I'm thinking about adding three input flags `train_filename`, `valid_filename`, `test_filename` in `WikiText2`. By default, those three flags have the corresponding file names.

Those input flags are necessary for other datasets, like translation datasets. For example, under German to English category, there are several training raw data files which represents different years (WMT_2009, WMT_2010, WMT_2011, etc). In this case, users have to choose which one they want to load.

Same thing here. By default, users will have train/valid/test datasets. However, if they turn any of `train_filename`, `valid_filename`, `test_filename` flags to `None`, the code won't generate the corresponding datasets. 
@cpuhrsch @vincentqb

zhangguanheng66: We talked about this offline. The decision is to add a keyword argument (a tuple of strings) to identify which datasets will be generated. The default value is a tuple of ('train', 'test'). Users have the flexibility to choose a single dataset with a proper `vocab` object.

vincentqb: Ok. 

In [CommonVoice](https://github.com/pytorch/audio/blob/38d1a9b6325d8b69e9b26ce28b1f1a8e0040c51b/torchaudio/datasets/commonvoice.py#L8), the implementation doesn't know about the meaning of train/test/etc. Instead, the zip file contains many tsv files (e.g. for train, test, ...), and the user simply specifies which file to use. 

An advantage is that the user could decide to create its own custom tsv file (say for training), and load that one.

zhangguanheng66: > Ok.
> 
> In [CommonVoice](https://github.com/pytorch/audio/blob/38d1a9b6325d8b69e9b26ce28b1f1a8e0040c51b/torchaudio/datasets/commonvoice.py#L8), the implementation doesn't know about the meaning of train/test/etc. Instead, the zip file contains many tsv files (e.g. for train, test, ...), and the user simply specifies which file to use.
> 
> An advantage is that the user could decide to create its own custom tsv file (say for training), and load that one.

Yes, that's kind of the same case for translation dataset so we need to provide an option to choose the file.
However, for most text dataset, the raw data files have been properly tagged with `train`, `test`, and `valid`.

vincentqb: Some datasets in audio even have "train_100", "train_500" etc built-in, see [this](https://github.com/pytorch/audio/blob/38d1a9b6325d8b69e9b26ce28b1f1a8e0040c51b/torchaudio/datasets/librispeech.py#L67). With only "train", "test" "valid", how would you like to deal with that here?

zhangguanheng66: In my early response, I propose to add three more flags `train_filename`, `valid_filename`, `test_filename`, which give you the flexibility to choose a specific file for training. I think either way should way and we just need to keep consistent across datasets within domain.

zhangguanheng66:
To conclude the discussions, a new flag `returned_datasets` is added to the API. By default, it is set to ('train', 'test','valid') so all the three subset are returned. Users could also choose any one or two of them, for example ('train', 'test'). If 'train' is not in the tuple, an vocab object should be provided which will be used to process valid and/or test data.

torchtext/data/functional.py
```diff
@@ -151,3 +153,54 @@ def simple_space_split(iterator):
...
+
+    Examples:
+        >>> from torchtext.data.functional import read_text_iterator
+        >>> tokenizer = get_tokenizer("basic_english")
+        >>> list((read_text_iterator('.data/ptb.train.txt', tokenizer)))
+            [['Sentencepiece', 'encode', 'as', 'pieces'], ['example', 'to', 'try!']]
+    """
+
+    with io.open(path, encoding="utf8") as f:
+        reader = unicode_csv_reader(f)
```
cpuhrsch: What if it's not a csv?

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,240 @@
...
+            test_data += tokens
+
+    valid_data = []
+    if valid_path is not None:
+        logging.info('Creating valid data')
+        valid_iter = create_data_from_iterator(
+            vocab, read_text_iterator(valid_path, tokenizer), removed_tokens)
+        for tokens in valid_iter:
+            valid_data += tokens
+    print("len(train_data): ", len(train_data))
```
cpuhrsch: Did you mean to leave this?

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,240 @@
...
+    def __iter__(self):
+        for x in self._data:
+            yield x
+
+    def get_vocab(self):
+        return self._vocab
+
+
+def _setup_datasets(dataset_name, tokenizer=get_tokenizer("basic_english"),
+                    root='.data', vocab=None, removed_tokens=['<unk>'],
```
cpuhrsch: Is removed_tokens text classification datasets do as well?

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,240 @@
...
+                    returned_datasets=('train', 'test', 'valid')):
```
cpuhrsch: "returned_datasets" is a bit of a strange name in my opinion. Maybe "splits" or something? cc @vincentqb

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,240 @@
...
+    train_path = None
+    test_path = None
+    valid_path = None
+    if dataset_name == 'PennTreebank':
+        if 'train' in returned_datasets:
```
cpuhrsch: I think it's worthwhile also supporting the case where `returned_datasets` is just a single string `train`.

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,240 @@
...
+    if valid_path is not None:
+        logging.info('Creating valid data')
+        valid_iter = create_data_from_iterator(
+            vocab, read_text_iterator(valid_path, tokenizer), removed_tokens)
+        for tokens in valid_iter:
+            valid_data += tokens
+    print("len(train_data): ", len(train_data))
+    print("len(test_data): ", len(test_data))
+    print("len(valid_data): ", len(valid_data))
+    return tuple(LanguageModelingDataset(torch.Tensor(d).long(), vocab)
```
cpuhrsch: torch.Tensor is the constructor for the Tensor class, but torch.tensor is the actual factory function (like torch.ones, etc.). We deprecated using the constructor, so we should use torch.tensor(d) here instead, so you can do the cast at the same time (even though it should be an int Tensor no matter what, since d contains ints).

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,240 @@
...
+    test_data = []
+    if test_path is not None:
+        logging.info('Creating testing data')
+        test_iter = create_data_from_iterator(
+            vocab, read_text_iterator(test_path, tokenizer), removed_tokens)
+        for tokens in test_iter:
+            test_data += tokens
+
+    valid_data = []
+    if valid_path is not None:
```
cpuhrsch: I think you could save a bit of space by writing something like

```
    paths = {}
    for fname in extracted_files:
        for k in ['train', 'valid', 'test']:
            if k in fname:
                paths[k] = os.path.join(root, fname)

    def gen_data(path_key, paths):
        path = paths[path_key]
        logging.info('Creating {} data'.format(path_key))
        return list(tokens for tokens in create_data_from_iterator(
            vocab, read_text_iterator(test_path, tokenizer), removed_tokens))

    return tuple(LanguageModelingDataset(torch.tensor(gen_data(k), vocab) for k in returned_datasets)
```

List expressions also tend to be faster than explicit for-loops.

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,240 @@
...
+        logging.info('Creating valid data')
+        valid_iter = create_data_from_iterator(
+            vocab, read_text_iterator(valid_path, tokenizer), removed_tokens)
+        for tokens in valid_iter:
+            valid_data += tokens
+    print("len(train_data): ", len(train_data))
+    print("len(test_data): ", len(test_data))
+    print("len(valid_data): ", len(valid_data))
+    return tuple(LanguageModelingDataset(torch.Tensor(d).long(), vocab)
+                 for d in (train_data, test_data, valid_data) if d != [])
```
cpuhrsch: This ignores the order in which the user passed the tuple to returned_datasets. It'll always return a subsequence of train_data, test_data, valid_data, but it won't return the correct order if the user specifies something like `(test, train, valid)`. It'd return `(train, valid, test)` instead.

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,240 @@
...
+                    root='.data', vocab=None, removed_tokens=['<unk>'],
+                    returned_datasets=('train', 'test', 'valid')):
+    train_path = None
+    test_path = None
+    valid_path = None
+    if dataset_name == 'PennTreebank':
+        if 'train' in returned_datasets:
+            train_path = download_from_url(URLS['PennTreebank'][0], root=root)
+        if 'test' in returned_datasets:
+            test_path = download_from_url(URLS['PennTreebank'][1], root=root)
```
cpuhrsch: The main difference here is that the `extracted_files` variable in the branch below can be pieced together by 

```
set = ...
extracted_files = [download_from_url(URLS['PennTreebank'][i], root=root)) for i in set]
```
Where set is chosen based on `returned_datasets`.

The subsequent "if 'train' in fname" logic then also applies to the paths here.

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,240 @@
...
+        if 'valid' in returned_datasets:
+            valid_path = download_from_url(URLS['PennTreebank'][2], root=root)
+    else:
+        dataset_tar = download_from_url(URLS[dataset_name], root=root)
+        extracted_files = extract_archive(dataset_tar)
+
+        for fname in extracted_files:
```
cpuhrsch: It's probably easier to write a search here instead of a full enumeration. That is

```
def get_data_path(key):
    for fname in extracted_files:
        if key in fname:
            return os.path.join(root, fname)
```

and then use that in context later on when you map from each entry of `returned_datasets` to a LanguageModelingDataset instance

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,240 @@
...
+                train_path = os.path.join(root, fname)
+            elif 'test' in returned_datasets and 'test' in fname:
+                test_path = os.path.join(root, fname)
+            elif 'valid' in returned_datasets and 'valid' in fname:
+                valid_path = os.path.join(root, fname)
+
+    if vocab is None:
+        logging.info('Building Vocab based on {}'.format(train_path))
+        if train_path is None:
+            raise TypeError("Train file is not defined correctly to generate vocabulary")
```
cpuhrsch: I think we can raise a more instructive message here, e.g.

```Must pass a vocab if train is not selected.```

**Comment by cpuhrsch** [comment]
I think the PR summary needs to be expanded since this does more than just moving the datasets to legacy. In particular I think it could be interesting to see answers to some of the following questions in that summary.

a) Why are we moving the current datasets to legacy instead of deleting them?
b) What is a summary of the new API?
c) How does this new API compare to the old one?
d) Are the new datasets less or more capable than the old ones?
e) Given an example usage of the old dataset, how do I adopt the new one?

File: torchtext/datasets/language_modeling.py (hunk 13 above)
Comment (vincentqb) [reply torchtext/datasets/language_modeling.py]: `return_datasets` would be more active voice. `concatenate` is not correct. `download`, `fetch`, `pick`, `select`, `split_into`? The keywork should hopefully imply it's the dataset-defined split. In that sense `fetch` might be appropriate. `split_into` sounds right but may imply there's also a way to select how to split.

`select` is probably good enough even though it's a little generic? Thoughts?

File: torchtext/datasets/language_modeling.py (hunk 12 above)
Comment (vincentqb) [reply torchtext/datasets/language_modeling.py]: Is that legacy option? Could it be a natural fit to belong within tokenizers to remove some?

File: torchtext/datasets/language_modeling.py (hunk 19 above)
Comment (vincentqb) [reply torchtext/datasets/language_modeling.py]: Same for above block of ifs.

File: torchtext/data/functional.py (hunk 10 above)
Comment (zhangguanheng66) [reply torchtext/data/functional.py]: Following this line, it calls `join` func before sending to `tokenizer`. Therefore, it works fine for non-csv files.
```
tokens = tokenizer(' '.join(row))
```

File: torchtext/datasets/language_modeling.py (hunk 12 above)
Comment (zhangguanheng66) [reply torchtext/datasets/language_modeling.py]: I don't think there is '<unk>' token in language modeling dataset, so I will just remove it.

File: torchtext/datasets/language_modeling.py (hunk 13 above)
Comment (zhangguanheng66) [reply torchtext/datasets/language_modeling.py]: I'm thinking about `data_select`.  `splits` may be confused with the one in legacy code.

File: torchtext/datasets/language_modeling.py (hunk 14 above)
Comment (zhangguanheng66) [reply torchtext/datasets/language_modeling.py]: Done

File: torchtext/datasets/language_modeling.py (hunk 15 above)
Comment (zhangguanheng66) [reply torchtext/datasets/language_modeling.py]: Fixed

File: torchtext/datasets/language_modeling.py (hunk 16 above)
Comment (zhangguanheng66) [reply torchtext/datasets/language_modeling.py]: Fixed!

File: torchtext/datasets/language_modeling.py (hunk 17 above)
Comment (zhangguanheng66) [reply torchtext/datasets/language_modeling.py]: Fixed!

File: torchtext/datasets/language_modeling.py (hunk 18 above)
Comment (zhangguanheng66) [reply torchtext/datasets/language_modeling.py]: fixed.

File: torchtext/datasets/language_modeling.py (hunk 20 above)
Comment (zhangguanheng66) [reply torchtext/datasets/language_modeling.py]: Fixed

File: torchtext/datasets/language_modeling.py (hunk 19 above)
Comment (zhangguanheng66) [reply torchtext/datasets/language_modeling.py]: Done

File: torchtext/data/functional.py (hunk 10 above)
Comment (cpuhrsch) [reply torchtext/data/functional.py]: For non-CSV files we might care about the "," as an explicit datapointer. So, for example, in a pure-text setting the comma might have a meaning, but in this case it gets replaced with a space without the user knowing about it, because the csv reader will interpret it as a boundary.

File: torchtext/data/functional.py [hunk 21]
```diff
@@ -151,3 +153,55 @@ def simple_space_split(iterator):
...
+        >>> from torchtext.data.functional import simple_space_split
+        >>> from torchtext.data.functional import create_data_from_iterator
+        >>> vocab = {'Sentencepiece' : 0, 'encode' : 1, 'as' : 2, 'pieces' : 3}
+        >>> list(create_data_from_iterator(vocab,
+        >>>                                simple_space_split(["Sentencepiece as pieces",
+        >>>                                                   "as pieces"]))
+        >>> [[0, 2, 3], [2, 3]]
+    """
+
+    for tokens in iterator:
```
cpuhrsch: One optimization here could be to yield an iterator instead of a list. This way we don't have to materialize the numbers per sentence which could be pretty large (and lists can be very slow).

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,236 @@
...
+        if 'train' not in _path.keys():
+            raise TypeError("Must pass a vocab if train is not selected.")
+        logging.info('Building Vocab based on {}'.format(_path['train']))
+        vocab = build_vocab_from_iterator(read_text_iterator(_path['train'], tokenizer))
+        logging.info('Vocab has {} entries'.format(len(vocab)))
+    else:
+        if not isinstance(vocab, Vocab):
+            raise TypeError("Passed vocabulary is not of type Vocab")
+
+    _data = {}
```
cpuhrsch: I don't think the leading underscore is necessary at all times. You typically prefix something with an underscore if it's supposed to be a private member function or member variable of a class, but not for every local variable.

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,236 @@
...
+    for item in _path.keys():
+        _data[item] = []
+        logging.info('Creating {} data'.format(item))
+        _iter = create_data_from_iterator(
+            vocab, read_text_iterator(_path[item], tokenizer), removed_tokens)
+        for tokens in _iter:
```
cpuhrsch: List expressions here could help performance

```data[item] = list(tokens for tokens in iter)```.

But actually, there is an opportunity for a performance optimization if we support returning an IterableDataset. So instead of materializing the entire dataset we just initialize it with an iterator. I think this is worth thinking about because it could help us make large datasets easily available (especially using the background iterator). However, should we include this here at this time? It could be a simple flag such as "return_iterable_datasets".

cc @vincentqb

File: torchtext/datasets/language_modeling.py [hunk 24]
```diff
@@ -1,217 +1,236 @@
...
+
+    if isinstance(data_select, str):
+        data_select = [data_select]
+    for item in data_select:
+        if item not in ('train', 'test', 'valid'):
+            raise TypeError('{} in data_select is not supported!'.format(item))
+
+    if dataset_name == 'PennTreebank':
+        extracted_files = []
+        if 'train' in data_select:
```
cpuhrsch: I think you can shorten this by defining ```select_to_index = {'train': 0, 'test': 1, 'valid': 2}```
and then

```extracted_files = [download_from_url(URLS['PennTreebank'][select_to_index[data_i]], root=root) for data_i in data_select]```

File: torchtext/datasets/language_modeling.py [hunk 25]
```diff
@@ -1,217 +1,236 @@
...
+            return fname
+
+
+def _setup_datasets(dataset_name, tokenizer=get_tokenizer("basic_english"),
+                    root='.data', vocab=None, removed_tokens=[],
+                    data_select=('train', 'test', 'valid')):
+
+    if isinstance(data_select, str):
+        data_select = [data_select]
+    for item in data_select:
```
cpuhrsch: You could do ```assert set(data_select) ==  set(('train', 'test', 'valid'))```.

torchtext/data/functional.py
zhangguanheng66: Sure. That's doable. Then, we materialize the token id outside the function.

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,236 @@
...
+    for item in _path.keys():
+        _data[item] = []
+        logging.info('Creating {} data'.format(item))
+        _iter = create_data_from_iterator(
+            vocab, read_text_iterator(_path[item], tokenizer), removed_tokens)
+        for tokens in _iter:
+            _data[item] += tokens
+
+    return tuple(LanguageModelingDataset(torch.tensor(_data[d]).long(), vocab)
+                 for d in data_select if _data[d] != [])
```
cpuhrsch: I think if ```_data[d]``` is empty we actually want to raise an error because the data the user requested is empty OR we simply return a dataset that loads nothing.

If the data is empty we might have downloaded something corrupted.

torchtext/datasets/language_modeling.py
zhangguanheng66: I think for now, we should return the materialized token ids for consistency.

torchtext/datasets/language_modeling.py
```diff
@@ -1,217 +1,236 @@
...
+    if dataset_name == 'PennTreebank':
+        extracted_files = []
+        if 'train' in data_select:
+            extracted_files.append(download_from_url(URLS['PennTreebank'][0], root=root))
+        if 'test' in data_select:
+            extracted_files.append(download_from_url(URLS['PennTreebank'][1], root=root))
+        if 'valid' in data_select:
+            extracted_files.append(download_from_url(URLS['PennTreebank'][2], root=root))
+    else:
+        dataset_tar = download_from_url(URLS[dataset_name], root=root)
```
cpuhrsch: Another thing we could do with these download functions is compare the result to a checksum. We added those capabilities in torchaudio. I think that might actually be worth it, if this checksum calculating is fast. If it's not fast we might need to make this a flag. This could save a lot of grief and is something rather common. Should we do this now? cc @vincentqb

vincentqb: Having a checksum is a good idea. However, since the goal is to soonish put the download function in sync, we should avoid investing time optimizing this particular version. I'd recommend using the one we worked on recently in [torchaudio](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/utils.py#L106) instead.

torchtext/datasets/language_modeling.py
vincentqb: This does seem like a good opportunity for optimization, and to start moving datasets in that direction. The return_iterable_datasets flag could default to "False" with a deprecation warning.

Do we know the length of `_iter`? If so, it's easy to return an iterable class that also offers a `__getitem__` that materialize lazyly.

The reason I'd see not to do this is if `_setup_datasets` is meant for users that would only experiment at small scale, or if we are moving away from this interface all together. Neither seems to be the case. I agree with consistency, but that would mean moving as many datasets as possible in that direction.

Thoughts?

cpuhrsch: IterableDatasets are a pretty new addition [about 5 months ago](https://github.com/pytorch/pytorch/pull/19228). This could be an opportunity to embrace that. 

The length will be hard to know for large datasets. We could call into stat and use the filesize, but that won't necessarily get us the number of lines of text.

Materializing lazily is only possible if we can read at an offset, but is an interesting idea. A diskcache iterator could allow people to write out the token ids onto disk as it's being consumed, but I think that's something to throw on top (should be easy given that IterableDataset behaves like an iterator).

torchtext/datasets/language_modeling.py
cpuhrsch: Yes we should copy this one into here as well. Do you want to create a PR?

cpuhrsch:
Thanks for updating the description! it's much clearer. 

Could you also add a paragraph that contrasts the old and new datasets? As in, do we give up on any capabilities, how much does the API differ at each step etc.

torchtext/datasets/language_modeling.py
zhangguanheng66: @vincentqb @cpuhrsch I'm fine to put one `download` function across domains if it doesn't break any current download activity.

torchtext/datasets/language_modeling.py
zhangguanheng66: I could write the iterabledataset and return the materialized one as default. Coming soon.

torchtext/data/functional.py
zhangguanheng66: Done!

torchtext/datasets/language_modeling.py
zhangguanheng66: Done!

torchtext/datasets/language_modeling.py
zhangguanheng66: Done!

torchtext/datasets/language_modeling.py
zhangguanheng66: Done!

torchtext/datasets/language_modeling.py
zhangguanheng66: Done!

torchtext/datasets/language_modeling.py
cpuhrsch: I think it's worth testing this, since it's our goal down the road :)

torchtext/data/functional.py
```diff
@@ -151,3 +153,53 @@ def simple_space_split(iterator):
...
+        >>>     print([num for num in ids])
+        >>> [0, 2, 3]
+        >>> [2, 3]
+    """
+
+    for tokens in iterator:
+        if removed_tokens is None:
+            yield iter(vocab[token] for token in tokens)
+        else:
+            tokens = list(filter(lambda x: x not in removed_tokens, tokens))
```
cpuhrsch: I think you can do this on the fly :). filter([...]) should be able to feed into iter and simply skip undesired items.

zhangguanheng66: Fixed. Thanks.

zhangguanheng66:
> Thanks for updating the description! it's much clearer.
> 
> Could you also add a paragraph that contrasts the old and new datasets? As in, do we give up on any capabilities, how much does the API differ at each step etc.

Add a session to compare the difference at the very end. @cpuhrsch

torchtext/data/functional.py
cpuhrsch: This doesn't work.

torchtext/data/functional.py
```diff
@@ -151,3 +152,53 @@ def simple_space_split(iterator):
...
+        >>>     print([num for num in ids])
+        >>> [0, 2, 3]
+        >>> [2, 3]
+    """
+
+    for tokens in iterator:
+        if removed_tokens is None:
+            yield iter(vocab[token] for token in tokens)
+        else:
+            yield iter(vocab[token] for token in
```
cpuhrsch: You can write this more compactly as ```iter(filter(lambda x: x is not None, [3, 4, None]))```

torchtext/data/functional.py
```diff
@@ -151,3 +152,52 @@ def simple_space_split(iterator):
...
+    """
+
+    with io.open(path, encoding="utf8") as f:
+        reader = unicode_csv_reader(f)
+        for row in reader:
+            tokens = tokenizer(' '.join(row))
+            yield tokens
+
+
+def create_data_from_iterator(vocab, iterator, removed_tokens=None):
```
vincentqb: Is the name of the function clear? `convert_tokens_to_ids`?

torchtext/legacy/__init__.py
```diff
@@ -0,0 +1,5 @@
+from . import datasets
+
+__version__ = '0.4.0'
```
vincentqb: nit: What does the version mean here?

README.rst
```diff
@@ -129,6 +129,16 @@ Others are planned or a work in progress:
 
 See the ``test`` directory for examples of dataset usage.
 
+Legacy Code
+===========
+
+We are currently retiring several datasets as legacy code ```torchtext.legacy```:
```
cpuhrsch: nit: We have currently retired several datasets and move them under ```torch.legacy```.

torchtext/data/functional.py
cpuhrsch: I don't think there is much value in this function as an abstraction. It wraps a one-liner

```iter(tokenizer(row) for row in io.open(path, encoding="utf8")``` once the incorrect usage of the unicode_csv reader has been removed.

PR #626, opened 2019-10-24
Remove unk hardcode

### LINKED ISSUE #618: arr = [[self.vocab.stoi[x] for x in ex] for ex in arr] KeyError: None
## 🐛 Bug
**Describe the bug**
I came across this error when using **data.Field**. It only happen when I define my own **unk_token** and set **min_freq** >1 at the same time.

**To Reproduce**
the code I use:
`SRC = data.Field(lower=True, unk_token="my_unk_token")
TGT = data.Field(lower=True)

train, val, test = datasets.IWSLT.splits(exts=('.de', '.en'), fields=(SRC, TGT))

SRC.build_vocab(train, min_freq=10)

train_iter = data.BucketIterator(dataset=train, batch_size=64,
sort_key=lambda x: data.interleave_keys(len(x.src), len(x.trg)))

batch = next(iter(train_iter))`

mttk:
Delegate the `unk_token` to arguments when constructing the vocabulary.

Fixes #618 , relatively major issue.

### DIFF
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -61,7 +61,7 @@ class Vocab(object):
         self.itos = list()
 
         self.unk_index = None
-        self.unk_token = unk_token # might be none
+        self.unk_token = unk_token  # might be none
         if specials_first:
             self.itos = list(specials)
             # only extend max size if specials are prepended
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -80,7 +80,7 @@ class Vocab(object):
                 break
             self.itos.append(word)
 
-        if unk_token is not None:
+        if unk_token is not None and unk_token in specials:
             unk_index = specials.index(unk_token)  # position in list
             # account for ordering of specials, set variable
             self.unk_index = unk_index if specials_first else len(self.itos) + unk_index
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -30,8 +30,9 @@ class Vocab(object):
         itos: A list of token strings indexed by their numerical identifiers.
     """
 
-    def __init__(self, counter, max_size=None, min_freq=1, unk_token='<unk>', specials=['<pad>'],
-                 vectors=None, unk_init=None, vectors_cache=None, specials_first=True):
+    def __init__(self, counter, max_size=None, min_freq=1, specials=['<unk>', '<pad>'],
+                 unk_token='<unk>', vectors=None, unk_init=None, vectors_cache=None,
+                 specials_first=True):
         """Create a Vocab object from a collections.Counter.
 
         Arguments:
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -109,8 +109,8 @@ class Vocab(object):
 
         if token_index is None:
             # An unknown word was encountered and unk_token is not used
-            raise KeyError(f'Token {token} not found in the vocabulary '
-                             'and the unknown token is not used.')
+            raise KeyError('Token {} not found in the vocabulary '
+                           'and the unknown token is not used.'.format(token))
 
         return token_index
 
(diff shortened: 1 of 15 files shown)


torchtext/vocab.py
```diff
@@ -30,10 +30,7 @@ class Vocab(object):
         itos: A list of token strings indexed by their numerical identifiers.
     """
 
-    # TODO (@mttk): Populate classs with default values of special symbols
-    UNK = '<unk>'
-
-    def __init__(self, counter, max_size=None, min_freq=1, specials=['<unk>', '<pad>'],
+    def __init__(self, counter, max_size=None, min_freq=1, unk_token='<unk>', specials=['<pad>'],
```
zhangguanheng66: Should `unk_token` argument be placed at the end? Not sure if this will introduce a BC here.

mttk: The order of kwargs is irrelevant so no (they are essentially treated as a dict).

zhangguanheng66: Not very end but behind `specials`?

mttk: I can move it, sure.

mttk:
This is enough to fix the issue. I would suggest we follow the example of huggingface/transformers (https://github.com/huggingface/transformers/blob/master/transformers/tokenization_utils.py#L52-L68) and having something akin to these special symbols predefined as properties in the vocab, which should prove to be easier in the long run than using a single list.

zhangguanheng66:
> This is enough to fix the issue. I would suggest we follow the example of huggingface/transformers (https://github.com/huggingface/transformers/blob/master/transformers/tokenization_utils.py#L52-L68) and having something akin to these special symbols predefined as properties in the vocab, which should prove to be easier in the long run than using a single list.
> @zhangguanheng66

I agree. It's more straightforward.

torchtext/vocab.py
```diff
@@ -106,7 +105,7 @@ def _default_unk_index(self):
         return self.unk_index
 
     def __getitem__(self, token):
-        return self.stoi.get(token, self.stoi.get(Vocab.UNK))
+        return self.stoi.get(token, self.unk_index)
```
zhangguanheng66: Say ```specials=['<pad>']``` and ```unk_token='<unk>' ```, self.unk_index = None (initiated).

Then the __getitem__ func will return `None` for unknown tokens. Not sure if this is correct.

mttk: You are correct, it should raise an error. Fixing.

torchtext/vocab.py
```diff
@@ -30,11 +30,9 @@ class Vocab(object):
         itos: A list of token strings indexed by their numerical identifiers.
     """
 
-    # TODO (@mttk): Populate classs with default values of special symbols
-    UNK = '<unk>'
```
zhangguanheng66: I think this will break the text_classification dataset here.
https://github.com/pytorch/text/blob/9afcfcdfdcb668780aba59fa1d06411645d54dd7/torchtext/datasets/text_classification.py#L52

mttk: I changed that line already :)

zhangguanheng66:
LGTM. Please let me know if you need a final review to merge the PR. Thanks.

mttk:
Sure, it's ready for merging from my side, let me know if you see anything problematic.

cpuhrsch:
Is this backwards compatible? Do we need to include these changes in our release notes? Please also have a look at https://github.com/pytorch/text/releases/tag/0.4.0 which includes notes relevant to the UNK token.

mttk:
You are right, it won't be backwards compatible as loading previously stored models will error out due to the missing attribute. 
I'll add a fix later this week, I'm swamped right now. This change should be mentioned in the release notes in case someone was explicitly using `Vocab.UNK` in their code.

vandykai (approved):

PR #627, opened 2019-10-24
Add BLEU score metric

sluks:
A quick review of MT quality scores:

**[BLEU]( https://www.aclweb.org/anthology/P02-1040.pdf) (Bilingual Evaluation Understudy)**, also called N-Gram Co-Occurence, was developed by IBM in 2001. It is a measure of similarity between a candidate translation and a set of reference translations, where each reference translation is equally valid. The essence of it is the following: we check how many n-grams the candidate and reference translation have in common (up to 4-grams). The idea is to focus on *precision*: how much of my candidate translation has correct words? However, if we only relied on precision, the machine could only output one word and precision would be super high but the translation (and recall) very bad. That’s why BLEU adds a “brevity penalty” to penalize for too short translations and improve recall.

* Pros: 
    * easy and fast to compute
    * reasonable estimate of Machine Translation (MT) quality, although limited
    * way more scalable than having humans do it
* Cons: 
    * doesn’t take meaning or grammar into account: if the references are not exhaustive (lack of synonyms) then the score is not accurate
    * no focus on sentence structure or order
    * score can be biased at sentence-level - BLEU is only meant to be used at corpus-level

**[NIST](http://www.mt-archive.info/HLT-2002-Doddington.pdf)** is an extension of BLEU, where we weigh the penalties of mis-matched n-grams: if we mis-match “rare” n-grams, the penalty will be higher than mis-matching common n-grams. The idea is to not give weight to usual, “stop word”-type of n-grams.

* Pros: 
    * less weight put on unimportant words compared to BLEU, leading to potentially more accurate scores
* Cons: 
    * still doesn’t take meaning into account, and this is potentially worse than BLEU: if the candidate translation has a valid synonym that’s not in the reference translations, then the penalty will be artificially higher

**[ROUGE](https://www.aclweb.org/anthology/W04-1013.pdf) (Recall-Oriented Understudy for Gisting Evaluation)** has the same idea as BLEU but focuses on recall instead of precision: we look at how many n-grams in the reference translations are in the candidate translation (instead of the reverse with BLEU). There are 5 different ROUGE variants: ROUGE-N, ROUGE-L, ROUGE-W, ROUGE-S, ROUGE-SU.

**[METEOR](https://www.cs.cmu.edu/~alavie/papers/BanerjeeLavie2005-final.pdf)**
Similar to BLEU, but considering synonyms and stemming. This is heavier to compute.

### DIFF
--- a/docs/source/data/metrics.rst
+++ b/docs/source/data/metrics.rst
@@ -6,3 +6,8 @@ torchtext.data.metrics
 
 .. automodule:: torchtext.data.metrics
 .. currentmodule:: torchtext.data.metrics
+
+:hidden:`bleu_score `
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. autofunction:: bleu_score
--- a/docs/source/data/metrics.rst
+++ b/docs/source/data/metrics.rst
@@ -1,0 +1,8 @@
+.. role:: hidden
+    :class: hidden-section
+
+torchtext.data.metrics
+===========================
+
+.. automodule:: torchtext.data.metrics
+.. currentmodule:: torchtext.data.metrics
--- a/test/data/test_metrics.py
+++ b/test/data/test_metrics.py
@@ -1,4 +1,5 @@
 from torchtext.data import metrics
+from torch.testing import assert_allclose
 from ..common.torchtext_test_case import TorchtextTestCase
 
 
@@ -18,19 +19,19 @@ class TestUtils(TorchtextTestCase):
         # Partial match
         candidate = [['My', 'full', 'pytorch', 'test']]
         refs = [[['My', 'full', 'pytorch', 'test', '!'], ['Different']]]
-        assert round(metrics.bleu_score(candidate, refs), 4) == 0.7788
+        assert_allclose(bleu_score(candidate, refs), 0.7788)
 
         # Bigrams and unigrams only
         candidate = [['My', 'pytorch', 'test']]
         refs = [[['My', 'full', 'pytorch', 'test'], ['Different']]]
-        assert round(metrics.bleu_score(candidate, refs, max_n=2,
-                     weights=[0.5, 0.5]), 4) == 0.5067
+        assert_allclose(bleu_score(candidate, refs, max_n=2,
+                                   weights=[0.5, 0.5]), 0.5067)
 
         # Multi-sentence corpus
         candidate = [['My', 'full', 'pytorch', 'test'], ['Another', 'Sentence']]
         refs = [[['My', 'full', 'pytorch', 'test'], ['Completely', 'Different']],
                 [['No', 'Match']]]
-        assert round(metrics.bleu_score(candidate, refs), 4) == 0.8409
+        assert_allclose(bleu_score(candidate, refs), 0.8409)
 
         # Empty input
         candidate = [[]]
@@ -49,10 +50,10 @@ class TestUtils(TorchtextTestCase):
                 [['I', 'have', 'made', 'a', 'terrible', 'mistake'], ['Big', 'mistake']]]
 
         # Value computed using nltk.translate.bleu_score.corpus_bleu(refs, candidate)
-        assert round(metrics.bleu_score(candidate, refs), 4) == 0.4573
-        assert round(metrics.bleu_score(candidate, refs, 3,
-                     weights=[0.33, 0.33, 0.33]), 4) == 0.4901
-        assert round(metrics.bleu_score(candidate, refs, 2,
-                     weights=[0.5, 0.5]), 4) == 0.5120
-        assert round(metrics.bleu_score(candidate, refs, 1,
-                     weights=[1]), 4) == 0.5516
+        assert_allclose(bleu_score(candidate, refs), 0.4573)
+        assert_allclose(bleu_score(candidate, refs, 3,
+                     weights=[0.33, 0.33, 0.33]), 0.4901)
+        assert_allclose(bleu_score(candidate, refs, 2,
+                     weights=[0.5, 0.5]), 0.5120)
+        assert_allclose(bleu_score(candidate, refs, 1,
+                     weights=[1]), 0.5516)
--- a/test/data/test_metrics.py
+++ b/test/data/test_metrics.py
@@ -49,11 +49,15 @@ class TestUtils(TorchtextTestCase):
                 ['She', 'loves', 'all', 'her', 'children', 'equally']],
                 [['I', 'have', 'made', 'a', 'terrible', 'mistake'], ['Big', 'mistake']]]
 
-        # Value computed using nltk.translate.bleu_score.corpus_bleu(refs, candidate)
+        # The comments below give the code used to get each hardcoded bleu score
+        # nltk.translate.bleu_score.corpus_bleu(refs, candidate)
         assert_allclose(bleu_score(candidate, refs), 0.4573199)
+        # nltk.translate.bleu_score.corpus_bleu(refs, candidate, weights=[0.33]*3)
         assert_allclose(bleu_score(candidate, refs, 3,
                      weights=[0.33, 0.33, 0.33]), 0.4901113)
+        # nltk.translate.bleu_score.corpus_bleu(refs, candidate, weights=[0.5]*2)
         assert_allclose(bleu_score(candidate, refs, 2,
                      weights=[0.5, 0.5]), 0.5119535)
+        # nltk.translate.bleu_score.corpus_bleu(refs, candidate, weights=[1])
         assert_allclose(bleu_score(candidate, refs, 1,
                      weights=[1]), 0.5515605)
--- a/test/data/test_metrics.py
+++ b/test/data/test_metrics.py
@@ -19,19 +19,19 @@ class TestUtils(TorchtextTestCase):
         # Partial match
         candidate = [['My', 'full', 'pytorch', 'test']]
         refs = [[['My', 'full', 'pytorch', 'test', '!'], ['Different']]]
-        assert_allclose(bleu_score(candidate, refs), 0.7788)
+        assert_allclose(bleu_score(candidate, refs), 0.7788007)
 
         # Bigrams and unigrams only
         candidate = [['My', 'pytorch', 'test']]
         refs = [[['My', 'full', 'pytorch', 'test'], ['Different']]]
         assert_allclose(bleu_score(candidate, refs, max_n=2,
-                                   weights=[0.5, 0.5]), 0.5067)
+                                   weights=[0.5, 0.5]), 0.5066641)
 
         # Multi-sentence corpus
         candidate = [['My', 'full', 'pytorch', 'test'], ['Another', 'Sentence']]
         refs = [[['My', 'full', 'pytorch', 'test'], ['Completely', 'Different']],
                 [['No', 'Match']]]
-        assert_allclose(bleu_score(candidate, refs), 0.8409)
+        assert_allclose(bleu_score(candidate, refs), 0.8408964)
 
         # Empty input
         candidate = [[]]
@@ -39,7 +39,7 @@ class TestUtils(TorchtextTestCase):
         assert metrics.bleu_score(candidate, refs) == 0
 
         # Long input, compared to NLTK implementation score
-        candidate = [['Lucille', 'Bluth', 'has', '3', 'sons'],
+        candidate = [['Lucille', 'B', 'has', '3', 'sons'],
                      ['She', 'loves', 'all', 'her', 'children', 'equally'],
                      ['No', 'match', 'here', 'at', 'all']]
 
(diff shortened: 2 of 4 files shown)


zhangguanheng66:
You can use flake8 to cover the lint error.

torchtext/data/functional.py
```diff
@@ -114,3 +115,72 @@ def _internal_func(txt_iter):
         for line in txt_iter:
             yield sp_model.EncodeAsPieces(line)
     return _internal_func
+
+
+def compute_ngrams(tokens, n):
```
zhangguanheng66: There is `ngram_iterator` [here](https://github.com/pytorch/text/blob/1f2bd140a700311b3c0d9f8e4a832953896981cd/torchtext/data/utils.py#L204)

torchtext/data/functional.py
```diff
@@ -114,3 +115,72 @@ def _internal_func(txt_iter):
...
+
+     Outputs:
+        output: a float between 0 and 1
+
+    Examples:
+        >>> from torchtext.data.functional import brevity_penalty
+        >>> brevity_penalty(9, 14)
+            0.5737534207374327
+    """
+    return math.exp(min(1 - refs_len/candidate_len, 0))
```
zhangguanheng66: Since this is just a single line, we don't need to have a func.

torchtext/data/functional.py
```diff
@@ -114,3 +115,72 @@ def _internal_func(txt_iter):
...
+        >>> from torchtext.data.functional import compute_ngrams
+        >>> tokens = ['This', 'is', 'definitely', 'me']
+        >>> compute_ngrams(tokens, 2)
+            [('This', 'is'), ('is', 'definitely'), ('definitely',
+            'me')]]
+    """
+    return list(zip(*[tokens[j:] for j in range(n)]))
+
+
+def compute_ngram_counter(tokens, max_n):
```
zhangguanheng66: And all the functions need a docs and unit test so it's better to minimize the number of functions.

docs/source/data/metrics.rst
```diff
@@ -0,0 +1,8 @@
+.. role:: hidden
+    :class: hidden-section
+
+torchtext.data.metrics
+===========================
+
+.. automodule:: torchtext.data.metrics
+.. currentmodule:: torchtext.data.metrics
```
zhangguanheng66: BLEU score should be added.
```
:hidden:`bleu_score `
~~~~~~~~~~~~~~~~~~~~~~~~~~~

.. autofunction:: bleu_score
```

test/data/test_metrics.py
```diff
@@ -0,0 +1,33 @@
+from torchtext.data import metrics
+from .common.torchtext_test_case import TorchtextTestCase
```
zhangguanheng66: from `..common ...`

torchtext/data/__init__.py
```diff
@@ -4,8 +4,9 @@
 from .field import RawField, Field, ReversibleField, SubwordField, NestedField, LabelField
 from .iterator import (batch, BucketIterator, Iterator, BPTTIterator,
                        pool)
+from .metrics import bleu_score
 from .pipeline import Pipeline
-from .utils import get_tokenizer, interleave_keys
+from .utils import get_tokenizer, interleave_keys, ngrams_iterator
```
sluks: @zhangguanheng66 do I actually need to import `ngrams_iterator` here ?

torchtext/data/metrics.py
```diff
@@ -0,0 +1,93 @@
...
+                Counter({('me',): 2,
+                 ('you',): 1,
+                 ('me', 'me'): 1,
+                 ('me', 'you'): 1,
+                 ('me', 'me', 'you'): 1})
+        """
+        assert max_n > 0
+
+        ngrams = [tuple(x.split(' ')) for x in ngrams_iterator(tokens, max_n)]
+        ngrams_counter = collections.Counter(ngrams)
```
cpuhrsch: Counter accepts iterators, so you can use ```collections.Counter(tuple(x.split(' ')) for x in ngrams_iterator(tokens, max_n))```. The advantage is that you don't need to materialize the entire string, so you'll save on memory roundtrips.

test/data/test_metrics.py
```diff
@@ -0,0 +1,33 @@
...
+
+        # No 4-gram
+        candidate = [['My', 'full', 'pytorch']]
+        refs = [[['My', 'full', 'pytorch', 'test'], ['Completely', 'Different']]]
+        assert metrics.bleu_score(candidate, refs) == 0
+
+        # Partial match
+        candidate = [['My', 'full', 'pytorch', 'test']]
+        refs = [[['My', 'full', 'pytorch', 'test', '!'], ['Different']]]
+        assert round(metrics.bleu_score(candidate, refs), 4) == 0.7788
```
vincentqb: What's the source for these scores? Or how were they computed?

torchtext/data/metrics.py
```diff
@@ -0,0 +1,93 @@
+import math
+import collections
+from torchtext.data.utils import ngrams_iterator
+
+
+def bleu_score(candidate_corpus, references_corpus, max_n=4, weights=[0.25] * 4):
+    """Computes the BLEU score between a candidate translation corpus and a references
+    translation corpus. Based on https://www.aclweb.org/anthology/P02-1040.pdf
+
+    Arguments:
```
cpuhrsch: As far as I can tell you don't use the assumption that candidate_corpus and reference_corpus are lists (you don't query their length and only iterate over their constituents) so this could be extended to "iterables".

torchtext/data/metrics.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Examples:
+        >>> candidate_corpus = [['I', 'ate', 'the', 'apple'], ['I', 'did']]
+        >>> references_corpus = [[['I', 'ate', 'it'], ['I', 'ate', 'apples']],
+                [['I', 'did']]]
+        >>> bleu_score(candidate_corpus, references_corpus)
+        >>> 0.7598356856515925
+    """
+
+    def _compute_ngram_counter(tokens, max_n):
```
vincentqb: nit: Any thoughts on putting this function outside of the `bleu_score` function?

cpuhrsch: Given on your (very clear) exposition of the different BLUE alternatives this sounds like something that'll become generally useful for the other metrics as well.

So it could be at the top-level (but still marked as private). Of course we can cross that bridge once we get there.

vincentqb: I second that.

torchtext/data/metrics.py
```diff
@@ -0,0 +1,93 @@
...
+        for ngram in clipped_counter:
+            clipped_counts[len(ngram) - 1] += clipped_counter[ngram]
+
+        for ngram in candidate_counter:  # TODO: no need to loop through the whole counter
+            total_counts[len(ngram) - 1] += candidate_counter[ngram]
+
+    if min(clipped_counts) == 0:
+        return 0.0
+    else:
+        pn = [clipped_counts[i] / total_counts[i] for i in range(max_n)]
```
cpuhrsch: A lot of the math here can be done using torch.Tensors - since this is part of the DAPI ecosystem it's ok to hard a hard dependency on that. Numpy on the other hand is something we can't have a hard dependency on, because pytorch/pytorch doesn't either.

torchtext/data/metrics.py
```diff
@@ -0,0 +1,93 @@
...
+        assert max_n > 0
+
+        ngrams = [tuple(x.split(' ')) for x in ngrams_iterator(tokens, max_n)]
+        ngrams_counter = collections.Counter(ngrams)
+
+        return ngrams_counter
+
+    assert max_n == len(weights), 'Length of the "weights" list has be equal to max_n'
+
+    clipped_counts = [0.0] * max_n
```
cpuhrsch: ```clipped_counts = torch.zeros(max_n)```

test/data/test_metrics.py
sluks: I computed them "by hand" by applying the math mentioned in the paper. Should I make this more explicit?

torchtext/data/metrics.py
```diff
@@ -0,0 +1,93 @@
...
+    for (candidate, refs) in zip(candidate_corpus, references_corpus):
+        candidate_len += len(candidate)
+
+        # Get the length of the reference that's closest in length to the candidate
+        refs_len_list = [float(len(ref)) for ref in refs]
+        refs_len = min(refs_len_list, key=lambda x: abs(len(candidate) - x))
+
+        reference_counters = collections.Counter()
+        for ref in refs:
+            reference_counters = reference_counters | _compute_ngram_counter(ref, max_n)
```
cpuhrsch: We carry around the count for tokens we will discard as soon as we see the candidate_counter. I wonder if candidate_counter could be used as an online filter when accumulating reference_counters, so we don't need to maintain that memory. It's also regrettable that we have to create a new collection for every reference. I'm sure most of the time that's fine, because many datasets will only have one target reference. It could be worth writing around that special case via
```
reference_counters = _compute_ngram_counter(refs[0], max_n)
for ref in refs[1:]:
   ...
```

torchtext/data/metrics.py
sluks: That's true, I'll change that right now it's an easy change

torchtext/data/metrics.py
```diff
@@ -0,0 +1,93 @@
...
+
+    candidate_len = 0.0
+    refs_len = 0.0
+
+    for (candidate, refs) in zip(candidate_corpus, references_corpus):
+        candidate_len += len(candidate)
+
+        # Get the length of the reference that's closest in length to the candidate
+        refs_len_list = [float(len(ref)) for ref in refs]
+        refs_len = min(refs_len_list, key=lambda x: abs(len(candidate) - x))
```
cpuhrsch: If I'm not missing something refs_len is being overwritten in each iteration. So in the end we only use the refs_len of the last (candidate, refs) pair?

cpuhrsch:
Great job! Thanks for writing this :) Most of my comments are around computational efficiency.

torchtext/data/metrics.py
zhangguanheng66: I think this is a good idea. In general, we only materialize the string if really necessary. Otherwise, we just send them to next generator pipeline.

torchtext/data/__init__.py
zhangguanheng66: Yes, you could.

torchtext/data/metrics.py
sluks: good point, I'll look into it

test/data/test_metrics.py
vincentqb: That's fine with me. Do you have more than 4 digits in your calculations? 

By the way, you could use "close to 0.7788" instead of round and equal.

torchtext/data/metrics.py
sluks: That's a great catch !! It should be `+=`. My tests didn't catch it because I only tested for very small corpuses. 
Related to this, I was wondering how I should test that this works for a large corpus. Should I check that I get the same result as the `nltk` implementation for a random corpus hardcoded in the tests file?

torchtext/data/metrics.py
```diff
@@ -0,0 +1,93 @@
...
+        max_n: the maximum order of n-gram wanted
+
+    Outputs:
+        output: a collections.Counter object with the unique n-grams and their
+            associated count
+
+    Examples:
+        >>> from torchtext.data.metrics import _compute_ngram_counter
+        >>> tokens = ['me', 'me', 'you']
+        >>> _compute_ngram_counter(tokens, 2)
```
zhangguanheng66: Do you want people to call `_compute_ngram_counter` directly? If so, you should keep the func name publicly.

test/data/test_metrics.py
zhangguanheng66: or `torch.testing.assert_allclose`

torchtext/data/metrics.py
```diff
@@ -0,0 +1,93 @@
+import math
+import collections
+from torchtext.data.utils import ngrams_iterator
+
+
+def _compute_ngram_counter(tokens, max_n):
```
zhangguanheng66: It seems that this func has only one line. You may not need a func here.

torchtext/data/metrics.py
cpuhrsch: Yes, using a well-established reference implementation isn't a bad idea. However, the values to compare against are better stored as a static asset, so that even though the reference implement might change, you still have the correct test data (we don't want a bug on their end influence us).

So, use ntlk to write out some test data, verify it's what you want and then use that as a reference within the test.

torchtext/data/metrics.py
cpuhrsch: The idea is that this function can be shared amongst all the other metrics (ROUGE, etc.), since they are all very similar in nature. Once those are implemented and depending on the added complexity we might end up deciding that this function either needs to be public or isn't necessary (since it's indeed only one line).

cpuhrsch:
If you want you could, for fun, benchmark this algorithm on some fixed dataset and see how some of the changes impact your runtime. You can then also compare this to known implementations  such as https://github.com/mjpost/sacreBLEU

torchtext/data/metrics.py
sluks: The assumption I took to keep this as a function was:
- Right now I use this function in these 2 places: https://github.com/pytorch/text/blob/c1fc7b0944e3b37d7cf05e1cf8ea3369842948b2/torchtext/data/metrics.py#L73-L75
- We might want to use it in other metrics in the future (it's a long shot as of now though)

torchtext/data/metrics.py
sluks: I replaced the math part with tensor operations in 4fc19dc8524bff9d9fe8b314645eda8bf547e9eb

Makes it more elegant without the list comprehensions!

cpuhrsch:
Generally this looks great to me! I'd be happy to merge that in its current form. There's still plenty of room for performance optimizations, but the overall setup seems fine and we have tests :D - Great job!

sluks:
@cpuhrsch @zhangguanheng66 In terms of github etiquette, should I rebase all these commits into 3 (one for metrics, one for the tests and one for docs) or into 1 only? For now I kept all the changes compared to the original PR in separate commits to keep track of each change

cpuhrsch:
@sluks - when merging we're choosing to squash and merge this PR into one. So, no need to worry about that.

sluks:
All right the PR passed all the tests and I added a few comments to clarify the test, I'm comfortable merging this from my side

PR #651, opened 2019-11-26
Re-write IMDB dataset in torchtext.experimental.datasets

zhangguanheng66:
Re-write IMDB dataset. See https://github.com/pytorch/text/pull/624 for motivation and API change. To load the new IMDB data with `torch.utils.data.DataLoader`
```
### Load a single dataset
from torchtext.experimental.datasets import IMDB
train1, = IMDB(data_select='train')

# Generate 8x8 batches
import torch
from torch.utils.data import DataLoader
def generate_rows(data):
	label_list = []
	txt_list = []
	for label, txt in data:
		label_list.append(label)
		txt_list.append(txt)
	return txt_list, label_list
	
dataloader = DataLoader(train1, batch_size=64, num_workers=4, collate_fn=generate_rows)
for txt, label in dataloader:
    # Send batch to the model.
```


zhangguanheng66:
Fix https://github.com/pytorch/text/issues/655

torchtext/datasets/text_classification.py
```diff
@@ -44,19 +45,17 @@ def _csv_iterator(data_path, ngrams, yield_cls=False):
...
-            else:
-                token_ids = list(filter(lambda x: x is not Vocab.UNK, [vocab[token]
-                                        for token in tokens]))
-                tokens = torch.tensor(token_ids)
-            if len(tokens) == 0:
-                logging.info('Row contains no tokens.')
-            data.append((cls, tokens))
-            labels.append(cls)
-            t.update(1)
+    for cls, tokens in iterator:
```
cpuhrsch: @vincentqb - we don't use tqdm here to track creating the Vocab. I'll start writing up some of these conventions for DAPI libraries so we can discuss them.

torchtext/__init__.py
```diff
@@ -4,7 +4,7 @@
 from . import vocab
 from . import legacy
 
-__version__ = '0.4.0'
```
vincentqb: What is the purpose of this PR? The title hints that it's about IMDB, right?

torchtext/datasets/text_classification.py
```diff
@@ -113,7 +112,8 @@ def get_vocab(self):
         return self._vocab
 
 
-def _setup_datasets(dataset_name, root='.data', ngrams=2, vocab=None, include_unk=False):
+def _setup_datasets(dataset_name, tokenizer=get_tokenizer("basic_english"),
```
cpuhrsch: A keyword argment such as 'root' can always be used positionally. So, if you're adding tokenizer a call such as ```AG_NEWS("/some/path", 2)``` is now invalid. To solve this you'll need need to move tokenizer to the very end.

Seeing this function call we should spend some time on making sure they align between this set of datasets and the unsupervised datasets we just added. I see that [over there](https://github.com/pytorch/text/blob/master/torchtext/datasets/language_modeling.py#L75) tokenizer comes first. So I'd suggest to move it at the end for unsupervised datasets as well.

cpuhrsch: Further there is also a different set of options. It might make sense to break BC to replace ```include_unk``` with the more generic ```removed_tokens=[]```.

torchtext/datasets/text_classification.py
vincentqb: Sure, I wouldn't want `tqdm` here indeed.

torchtext/datasets/text_classification.py
```diff
@@ -44,19 +45,17 @@ def _csv_iterator(data_path, ngrams, yield_cls=False):
...
+        if include_unk:
+            tokens = torch.tensor([vocab[token] for token in tokens])
+        else:
+            token_ids = list(filter(lambda x: x is not Vocab.UNK, [vocab[token]
+                                    for token in tokens]))
+            tokens = torch.tensor(token_ids)
+        if len(tokens) == 0:
+            logging.info('Row contains no tokens.')
+        data.append((cls, tokens))
+        labels.append(cls)
```
vincentqb: Just mean to confirm that there was a decision to materialize the the list, right?

torchtext/datasets/text_classification.py
```diff
@@ -373,6 +408,122 @@ def AmazonReviewFull(*args, **kwargs):
...
+                _iter = numericalize_tokens_from_iterator(vocab,
+                                                          txt_iter,
+                                                          removed_tokens)
+                for tokens in _iter:
+                    _data.append((0,
+                                  torch.tensor([token_id for token_id in tokens]).long()))
+    return _data
+
+
+def IMDB(tokenizer=get_tokenizer("basic_english"),
```
cpuhrsch: Why don't you use a setup similar to the other datasets here, i.e. forward to the common function? It looks like it should be similar enough or we can work on generalizing _setup_datasets to support it.

torchtext/datasets/text_classification.py
```diff
@@ -44,19 +45,17 @@ def _csv_iterator(data_path, ngrams, yield_cls=False):
 def _create_data_from_iterator(vocab, iterator, include_unk):
```
vincentqb: Same question as in [last PR](https://github.com/pytorch/text/pull/624#discussion_r350420442), is the name clear? Is this meant to follow the conventions established in the last PR?

torchtext/datasets/text_classification.py
```diff
@@ -155,6 +158,10 @@ def AG_NEWS(*args, **kwargs):
     Separately returns the training and test dataset
 
     Arguments:
+        tokenizer: the tokenizer used to preprocess raw text data.
```
vincentqb: Is this part of IMDB? Could you clarify the scope of this PR?

torchtext/datasets/text_classification.py
```diff
@@ -373,6 +408,122 @@ def AmazonReviewFull(*args, **kwargs):
...
+                                                          removed_tokens)
+                for tokens in _iter:
+                    _data.append((0,
+                                  torch.tensor([token_id for token_id in tokens]).long()))
+    return _data
+
+
+def IMDB(tokenizer=get_tokenizer("basic_english"),
+         root='.data', vocab=None, removed_tokens=[],
+         data_select=('train', 'test')):
```
vincentqb: This one doesn't have a validation set, right?

torchtext/datasets/text_classification.py
vincentqb: Using a function call in a function header usually raises a flake8 error since all the call to the function will use the same object reference. The function is [here](https://github.com/pytorch/text/blob/master/torchtext/data/utils.py#L74), but it's not clear to me that there won't be issue with that here.

torchtext/datasets/text_classification.py
vincentqb: Same about function call in header.

torchtext/datasets/text_classification.py
cpuhrsch: We can switch this over to the generator style. Probably a worthwhile optimizing and more consistent. @zhangguanheng66 - can you look into this?

torchtext/datasets/text_classification.py
```diff
@@ -373,6 +408,122 @@ def AmazonReviewFull(*args, **kwargs):
     return _setup_datasets(*(("AmazonReviewFull",) + args), **kwargs)
 
 
+def _generate_imdb_data(key, extracted_files, vocab, tokenizer, removed_tokens):
```
cpuhrsch: This function is very complex for what it does.

torchtext/datasets/text_classification.py
```diff
@@ -373,6 +408,122 @@ def AmazonReviewFull(*args, **kwargs):
...
+    _data = []
+    for fname in extracted_files:
+        if 'urls' in fname:
+            continue
+        elif key in fname:
+            if 'pos' in fname:
+                txt_iter = iter(tokenizer(row) for row in io.open(fname,
+                                                                  encoding="utf8"))
+                _iter = numericalize_tokens_from_iterator(vocab,
```
cpuhrsch: Please stop prefixing variables that aren't private member variables

torchtext/datasets/text_classification.py
cpuhrsch: Indeed the typical convention is to use `None` as default and then assign a value.

torchtext/datasets/text_classification.py
```diff
@@ -373,6 +408,122 @@ def AmazonReviewFull(*args, **kwargs):
...
+        >>> tokenizer = ngrams_tokenizer_wrapper(2)
+        >>> train_dataset, test_dataset = IMDB(tokenizer=tokenizer)
+    """
+
+    dataset_tar = download_from_url(URLS['IMDB'], root=root)
+    extracted_files = extract_archive(dataset_tar)
+
+    if isinstance(data_select, str):
+        data_select = [data_select]
+    for key in data_select:
```
cpuhrsch: Does this mechanism exist for the other functions? Looks like an extension that is only supported for IMDB.

torchtext/__init__.py
```diff
@@ -4,7 +4,7 @@
...
+__version__ = '0.5.0'
```
cpuhrsch: Please never bump the version randomly without doing a release or outside of a release branch.

zhangguanheng66: I thought after each release, we bump the version up for next release.

torchtext/datasets/text_classification.py
zhangguanheng66: Yes. I could fix it. Users won't see the change.

torchtext/datasets/text_classification.py
zhangguanheng66: We change the API to keep those text classification datasets consistent.

torchtext/datasets/text_classification.py
zhangguanheng66: no, there is not.

torchtext/datasets/text_classification.py
zhangguanheng66: I switch to the same API used by other text classification datasets.

torchtext/datasets/text_classification.py
zhangguanheng66: OK, I remove tqdm from `build_vocab`

torchtext/datasets/text_classification.py
zhangguanheng66: Fixed.

torchtext/datasets/text_classification.py
zhangguanheng66: IMDB calls `_setup_datasets` now.

torchtext/__init__.py
zhangguanheng66: Move the version back to '0.4.0'

torchtext/datasets/text_classification.py
zhangguanheng66: Fixed

torchtext/datasets/text_classification.py
zhangguanheng66: Since we add `tokenizer` to `_setup_datasets`, it affects all the other datasets here.

torchtext/datasets/text_classification.py
zhangguanheng66: Fixed. `tokenizer=None`.

torchtext/datasets/text_classification.py
zhangguanheng66: Merge it with `_setup_datasets`.

torchtext/__init__.py
zhangguanheng66: Fixed.

torchtext/datasets/text_classification.py
```diff
@@ -124,18 +136,47 @@ def _setup_datasets(dataset_name, root='.data', ngrams=2, vocab=None, include_un
...
+    else:
+        data_iterator = _csv_iterator(train_csv_path, tokenizer, ngrams, yield_cls=True)
+    data_iter = _create_data_from_iterator(vocab, data_iterator, removed_tokens)
+    train_data = []
+    train_labels = []
+    for cls, tokens in data_iter:
+        train_data.append((torch.tensor(cls),
+                           torch.tensor([token_id for token_id in tokens])))
+        train_labels.append(cls)
+    train_labels = set(train_labels)
```
vincentqb: Can you give an example of what is `train_labels` and `test_labels`? I was expecting them to be ordered.

torchtext/datasets/text_classification.py
```diff
@@ -124,18 +136,47 @@ def _setup_datasets(dataset_name, root='.data', ngrams=2, vocab=None, include_un
...
+    else:
+        data_iterator = _csv_iterator(test_csv_path, tokenizer, ngrams, yield_cls=True)
+    data_iter = _create_data_from_iterator(vocab, data_iterator, removed_tokens)
+    test_data = []
+    test_labels = []
+    for cls, tokens in data_iter:
+        test_data.append((torch.tensor(cls),
+                          torch.tensor([token_id for token_id in tokens])))
+        test_labels.append(cls)
+    test_labels = set(test_labels)
```
vincentqb: Same

torchtext/datasets/text_classification.py
```diff
@@ -373,6 +446,39 @@ def AmazonReviewFull(*args, **kwargs):
...
+
+    Examples:
+        >>> from torchtext.datasets import IMDB
+        >>> from torchtext.data.utils import get_tokenizer
+        >>> train_dataset, test_dataset = IMDB(ngrams=3)
+        >>> tokenizer = get_tokenizer("spacy")
+        >>> train_dataset, test_dataset = IMDB(tokenizer=tokenizer)
+    """
+
+    return _setup_datasets(*(('IMDB',) + args), **kwargs)
```
vincentqb: Would `_setup_datasets('IMDB', *args, **kwargs)` do the same?

torchtext/datasets/text_classification.py
zhangguanheng66: For classification problem, the `class` is saved in `train_data` as `cls` and used as target. Each data point has a `cls` and labels is actually a set of those `cls`

torchtext/datasets/text_classification.py
zhangguanheng66: yes. We decide to let all the dataset call the same function `_setup_datasets` but with a different function `_imdb_iterator` to read data from file.

torchtext/datasets/text_classification.py
```diff
@@ -41,23 +41,29 @@ def _csv_iterator(data_path, ngrams, yield_cls=False):
...
+        yield cls, iter(map(lambda x: vocab[x],
+                        filter(lambda x: x not in removed_tokens, tokens)))
+
+
+def _imdb_iterator(key, extracted_files, tokenizer, ngrams, yield_cls=False):
+    for fname in extracted_files:
+        if 'urls' in fname:
+            continue
+        elif key in fname:
+            if 'pos' in fname:
```
cpuhrsch: If 'pos' is not in fname and 'neg' is not in fname and 'urls' is not in fname, what's left?

If you expect there is no other choice it could be useful to add an assert.

torchtext/datasets/text_classification.py
```diff
@@ -41,23 +41,29 @@ def _csv_iterator(data_path, ngrams, yield_cls=False):
...
+                with io.open(fname, encoding="utf8") as f:
```
cpuhrsch: You could collapse this into one by saying

```label = 0 if 'pos' in fname else 1```
And then do 
```
 with io.open(fname, encoding="utf8") as f:
     token_iterator = ngrams_iterator(tokenizer(f.read()), ngrams)
     if yield_cls:
         yield int(label), token_iteratator
     else:
         yield token_iterator
```

Are you sure the explicit cast via ```int``` is necessary here?

File: torchtext/datasets/text_classification.py [hunk 18]
```diff
@@ -113,33 +119,61 @@ def get_vocab(self):
...
+def _setup_datasets(dataset_name, root='.data', ngrams=2, vocab=None,
+                    removed_tokens=[], tokenizer=None,
+                    data_select=('train', 'test')):
+    if not tokenizer:
+        tokenizer = get_tokenizer("basic_english")
+
+    if isinstance(data_select, str):
+        data_select = [data_select]
+    if not set(data_select).issubset(set(('train', 'test'))):
+        raise TypeError('data_select is not supported!')
```
cpuhrsch: I'd say something like ```TypeError('Given data selection {} is not supported!'.format(data_select))```. Then the user knows what she did wrong.

torchtext/datasets/text_classification.py
```diff
@@ -113,33 +119,61 @@ def get_vocab(self):
...
+            path['test'] = fname
 
     if vocab is None:
-        logging.info('Building Vocab based on {}'.format(train_csv_path))
-        vocab = build_vocab_from_iterator(_csv_iterator(train_csv_path, ngrams))
+        if 'train' not in data_select:
+            raise TypeError("Must pass a vocab if train is not selected.")
+        logging.info('Building Vocab based on train data')
+        if dataset_name == 'IMDB':
+            data_iterator = _imdb_iterator('train', extracted_files, tokenizer, ngrams)
```
cpuhrsch: Instead of passing this here you can change _setup_datasets to accept an iterator and create this in the classes that defer to the function. That'll save you this awkward branch and create a more general function that is agnostic to the specific name of the dataset.

torchtext/datasets/text_classification.py
```diff
@@ -113,33 +119,61 @@ def get_vocab(self):
...
-            TextClassificationDataset(vocab, test_data, test_labels))
+
+    data = {}
+    for item in data_select:
+        data[item] = {}
+        data[item]['data'] = []
+        data[item]['labels'] = []
+        logging.info('Creating {} data'.format(item))
+        if dataset_name == 'IMDB':
+            data_iterator = _imdb_iterator(item, extracted_files,
```
cpuhrsch: I understand that you need to generate the iterator twice, so we could also look into passing a function that is a generator factory to this setup function that'll allow you to generate an iterator over a particular dataset. That should keep this function shorter.

torchtext/datasets/text_classification.py
```diff
@@ -113,33 +119,61 @@ def get_vocab(self):
...
+                                           tokenizer, ngrams, yield_cls=True)
+        else:
+            data_iterator = _csv_iterator(path[item], tokenizer, ngrams, yield_cls=True)
+        data_iter = _create_data_from_iterator(vocab, data_iterator, removed_tokens)
+        for cls, tokens in data_iter:
+            data[item]['data'].append((torch.tensor(cls),
```
cpuhrsch: I wonder if it's better to create and pass a single Tensor or a list of Tensors. Should be interesting to benchmark. I'd imagine torch.Tensor indexing to be more efficient over a single block of memory than over a list of small Tensors each with their own, separate memory. This should be worth a quick benchmark.

So, more concretely, instead of append a torch.tensor here you create a list of torch.tensor(cls) and torch.tensor([token...]) and then stack them together to create one, large Tensor which is then used for construction of TextClassificationDataset.

torchtext/datasets/text_classification.py
```diff
@@ -227,7 +283,18 @@ def DBpedia(*args, **kwargs):
             Default: 1
         vocab: Vocabulary used for dataset. If None, it will generate a new
             vocabulary based on the train data set.
-        include_unk: include unknown token in the data (Default: False)
+        removed_tokens: removed tokens from output dataset (Default: [])
+        tokenizer: the tokenizer used to preprocess raw text data.
+            The default one is basic_english tokenizer in fastText. spacy tokenizer
+            is supported as well. A custom tokenizer is callable
+            function with input of a string and output of a token list.
+        data_select: a string or tupel for the returned datasets
```
cpuhrsch: nit: "tuple"

torchtext/datasets/text_classification.py
cpuhrsch: @vincentqb - as far as I remember that syntax isn't supported by Python2.

torchtext/datasets/text_classification.py
zhangguanheng66: Fixed

torchtext/datasets/text_classification.py
zhangguanheng66: Yes. I extended it to all the other datasets.

torchtext/datasets/text_classification.py
zhangguanheng66: If `fname` is not in these three cases, it just passes the loop and will not be used to extract data.

torchtext/datasets/text_classification.py
zhangguanheng66: Label is actually the index. When applying the dataset to the model, labels are used as the target so probably a floating value is more proper.

torchtext/datasets/text_classification.py
zhangguanheng66: Sure. I can do some benchmark for this issue.
Something to figure out is that each text tensor has different length. That's where `NestedTensor` could fit.

cpuhrsch: Yes, you're right, how did I miss that.... At least with a 2D Tensor we can translate an index to a range. Could still be interesting.

torchtext/datasets/text_classification.py
cpuhrsch: Yes, but can such a case ever come up with this dataset in particular?

torchtext/datasets/text_classification.py
```diff
@@ -113,33 +122,61 @@ def get_vocab(self):
...
+        if fname.endswith('test.csv') and 'test' in data_select:
+            path['test'] = fname
 
     if vocab is None:
-        logging.info('Building Vocab based on {}'.format(train_csv_path))
-        vocab = build_vocab_from_iterator(_csv_iterator(train_csv_path, ngrams))
+        if 'train' not in data_select:
+            raise TypeError("Must pass a vocab if train is not selected.")
+        logging.info('Building Vocab based on train data')
+        if dataset_name == 'IMDB':
```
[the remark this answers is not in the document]
cpuhrsch: I think that'll also be a great example of writing preprocessing on a generic source where the data might be backed by multiple or a single file in various formats.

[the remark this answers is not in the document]
zhangguanheng66: Yes. I think I need to write a demo to show how to build a pipeline for torchtext data processing, in case that users would like to load their own data.

torchtext/datasets/text_classification.py
zhangguanheng66: Did some benchmark:
```
import time
import torch
tensor1 = torch.rand((10, 10))
t0 = time.time()
for i in range(0, 10000):
	result = torch.mm(tensor1, tensor1)
print("Benchmark case 1 time: ", time.time() - t0)

list_tensor = [torch.rand((10))] * 10
tensor2 = torch.stack(list_tensor)
t0 = time.time()
for i in range(0, 10000):
	result = torch.mm(tensor2, tensor2)
print("Benchmark case 2 time: ", time.time() - t0)

Benchmark case 1 time:  0.1210787296295166
Benchmark case 2 time:  0.11946344375610352
```

torchtext/datasets/text_classification.py
vincentqb: Glad we deprecate python 2 soon :)

zhangguanheng66:
Fix #657

torchtext/vocab.py
```diff
@@ -553,9 +553,7 @@ def build_vocab_from_iterator(iterator):
     """
 
     counter = Counter()
-    with tqdm(unit_scale=0, unit='lines') as t:
```
cpuhrsch: Can we move this into a separate PR? tqdm was used within the dataset building file as well, but this is a wide-ranging change for the vocab which is used in many places.

zhangguanheng66: Moved it back.

torchtext/datasets/text_classification.py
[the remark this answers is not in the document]
vincentqb: On that note, can you also add to this PR an example code on how to load `train1` in the example you already provide?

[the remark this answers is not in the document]
zhangguanheng66: Sure. Added it.

torchtext/datasets/text_classification.py
```diff
@@ -113,36 +113,88 @@ def get_vocab(self):
         return self._vocab
 
 
-def _setup_datasets(dataset_name, root='.data', ngrams=2, vocab=None, include_unk=False):
+def generate_data_iterators(dataset_name, root, ngrams, tokenizer, data_select):
```
cpuhrsch: If this is an internal function, it is [customary](https://docs.python.org/3/tutorial/classes.html#private-variables) to use a leading underscore.

torchtext/datasets/text_classification.py
```diff
@@ -113,36 +113,88 @@ def get_vocab(self):
...
+            data[item]['data'].append((torch.tensor(cls),
+                                       torch.tensor([token_id for token_id in tokens])))
+            data[item]['labels'].append(cls)
+        data[item]['labels'] = set(data[item]['labels'])
+
+    return tuple(TextClassificationDataset(vocab, data[item]['data'],
+                                           data[item]['labels']) for item in data_select)
+
+
+def AG_NEWS(root='.data', ngrams=2, vocab=None, removed_tokens=[],
```
cpuhrsch: To avoid typos as might come along with wide-ranging changes or simply maintaining something as this long list of arguments and default values I suggest to create a new function that abstracts the function call. 

```
    return _setup_datasets(generate_data_iterators('AG_NEWS', root, ngrams,
                                                   tokenizer, data_select),
                           vocab, removed_tokens, data_select)
```

torchtext/datasets/text_classification.py
```diff
@@ -157,20 +209,34 @@ def AG_NEWS(*args, **kwargs):
     Arguments:
         root: Directory where the datasets are saved. Default: ".data"
         ngrams: a contiguous sequence of n items from s string text.
-            Default: 1
+            Default: 2
```
cpuhrsch: I'm not a fan of returning bigrams by default because:
a) the default vocab will be much larger
b) it's more intuitive in my opinion to get lines of tokens instead of tokens + bigram tokens
c) we can expect some users to maybe want to combine this with ngram iterator or similar abstractions, which can lead to nasty bugs
d) It's strongly BC-breaking. I understand this comes from the unsupervised dataset, I'd suggest we switch it over to "ngrams 1" instead of this to "2" by default.

torchtext/datasets/text_classification.py
```diff
@@ -113,36 +113,88 @@ def get_vocab(self):
         return self._vocab
 
 
-def _setup_datasets(dataset_name, root='.data', ngrams=2, vocab=None, include_unk=False):
```
cpuhrsch: Setting ngrams to 2 might have been a mistake on my end.

I'm not a fan of returning bigrams by default because:
a) the default vocab will be much larger
b) it's more intuitive in my opinion to get lines of tokens instead of tokens + bigram tokens
c) we can expect some users to maybe want to combine this with ngram iterator or similar abstractions, which can lead to nasty bugs

torchtext/datasets/text_classification.py
zhangguanheng66: Fixed

torchtext/datasets/text_classification.py
zhangguanheng66: Add a function `_initiate_datasets` to handle the two branches.

torchtext/datasets/text_classification.py
zhangguanheng66: Set `ngrams` default value to 1

torchtext/prototype/datasets/text_classification.py
```diff
@@ -0,0 +1,628 @@
...
+            yield x
+
+    def get_labels(self):
+        return self._labels
+
+    def get_vocab(self):
+        return self._vocab
+
+
+def _generate_data_iterators(dataset_name, root, ngrams, tokenizer, data_select):
```
cpuhrsch: I think this can be combined with _generate_imdb_data_iterators and abstracted further by allowing to pass an iters_group as an argument.

torchtext/prototype/datasets/text_classification.py
```diff
@@ -0,0 +1,628 @@
...
+                                           data[item]['labels']) for item in data_select)
+
+
+def _initiate_datasets(dataset_name, root='.data', ngrams=1, vocab=None,
+                       removed_tokens=[], tokenizer=None,
+                       data_select=('train', 'test')):
+    if isinstance(data_select, str):
+        data_select = [data_select]
+    if dataset_name == 'IMDB':
+        return _setup_datasets(_generate_imdb_data_iterators(dataset_name, root, ngrams,
```
cpuhrsch: The only difference here is _generate_imdb_data_iterators and _generate_data_iterators. If we create a more generic version of these two functions that can accept an iters_group we should be able to write less code.

examples/text_classification/README.md
```diff
@@ -1,7 +1,7 @@
 # This is an example to create a text classification dataset and train a sentiment model. 
 
 In the basic case, users can train the sentiment model in model.py with 
-AG_NEWS dataset in torchtext.datasets.text_classification. The dataset is
```
vincentqb: Is this meant to indicate that the interface is in pre-release and will evolve?

zhangguanheng66: Good catch. We need to fix it in this release as all the text classification datasets are moved to `prototype` folder.

cpuhrsch: @vincentqb - we talked a bit offline and thought it should be a good idea to move these new pieces into a preview or prototype folder first (even though we think they're much better) since the changes are quite drastic. What do you think?

vincentqb: Are the previous datasets still being deprecated? Is the new interface finalized? 

If yes and yes, then why delay? :)

cpuhrsch: The current datasets won't be deprecated during this release, instead we'll introduce a new prototype section.

The new interface is not finalized and we notice various inconsistencies bubbling up.

torchtext/prototype/datasets/text_classification.py
zhangguanheng66: fixed.

torchtext/prototype/datasets/text_classification.py
zhangguanheng66: fixed

examples/text_classification/README.md
zhangguanheng66: `prototype` folder allows us to evolve the interface without a lot of pressure from BC breaking. We could improve the datasets API while receiving feedbacks from OSS users. Once they are matured, we could move them to the main folder.

torchtext/prototype/datasets/text_classification.py
```diff
@@ -0,0 +1,613 @@
...
+    path = {}
+    if dataset_name != 'IMDB':
+        for fname in extracted_files:
+            if fname.endswith('train.csv') and 'train' in data_select:
+                path['train'] = fname
+            if fname.endswith('test.csv') and 'test' in data_select:
+                path['test'] = fname
+
+    iters_group = {}
+    if 'train' in data_select:
```
cpuhrsch: Both "_imdb_iterator" and "_csv_iterator" have a similar interface, except that _csv_iterator is passed a path. Instead of accepting a `path['train']` it could simply accept `'train'` like _imdb_iterator does. Then you only need to create an iterator via something like `_gen_imdb_iterator` and all of this branching etc. can disappear.

torchtext/datasets/text_classification.py
cpuhrsch: Your benchmark measures the amount of time a matmul takes which seems unrelated. What I suggested here is that we look at the performance of indexing into a list and retrieving a Tensor vs. indexing into a dense tensor. You're using a dense Tensor due to torch.stack in both cases, but you're not doing any indexing.

torchtext/datasets/text_classification.py
```diff
@@ -160,10 +178,21 @@ def AG_NEWS(*args, **kwargs):
             Default: 1
         vocab: Vocabulary used for dataset. If None, it will generate a new
             vocabulary based on the train data set.
-        include_unk: include unknown token in the data (Default: False)
+        removed_tokens: removed tokens from output dataset (Default: [])
```
cpuhrsch: Are these changes BC breaking since they'll live outside the experimental folder after all?

torchtext/experimental/datasets/text_classification.py
```diff
@@ -0,0 +1,142 @@
...
+        elif key in fname and ('pos' in fname or 'neg' in fname):
+            with io.open(fname, encoding="utf8") as f:
+                label = 1 if 'pos' in fname else 0
+                if yield_cls:
+                    yield label, ngrams_iterator(tokenizer(f.read()), ngrams)
+                else:
+                    yield ngrams_iterator(tokenizer(f.read()), ngrams)
+
+
+def _generate_data_iterators(dataset_name, root, ngrams, tokenizer, data_select):
```
cpuhrsch: I don't think you need to repeat all of this, but you can use it from the torchtext/datasets/text_classification file

torchtext/datasets/text_classification.py
cpuhrsch: Could you break out these BC breaking changes, which are unrelated to IMDB, into a separate PR?

torchtext/__init__.py
```diff
@@ -3,11 +3,13 @@
...
+from . import experimental
 
 __version__ = '0.4.0'
 
 __all__ = ['data',
            'datasets',
            'utils',
            'vocab',
-           'legacy']
+           'legacy',
```
cpuhrsch: You probably need to delete this

PR #684, opened 2020-01-23
Consider min frequency when create new vocab

lepoeme20:
User can not adjust min frequency to create new vocab when load a data from TextClassificationDataset and LanguageModelingDataset such as DBpedia, YelpReviewPolarity, AmazonReviewFull, WikiText103, PennTreebank, etc. 

I propose add 'min frequency' parameter so that users can adjust min frequency when they calling those datasets.

Thanks.

### DIFF
--- a/torchtext/datasets/text_classification.py
+++ b/torchtext/datasets/text_classification.py
@@ -113,7 +113,8 @@ class TextClassificationDataset(torch.utils.data.Dataset):
         return self._vocab
 
 
-def _setup_datasets(dataset_name, root='.data', ngrams=1, vocab=None, include_unk=False, min_freq=1):
+def _setup_datasets(dataset_name, root='.data', ngrams=1, vocab=None,
+                    include_unk=False, min_freq=1):
     dataset_tar = download_from_url(URLS[dataset_name], root=root)
     extracted_files = extract_archive(dataset_tar)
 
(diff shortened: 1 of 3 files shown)


zhangguanheng66:
Solution provided by @bentrevett. You could create a custom vocab object with `min_freq` as follow:
```
import torchtext
from torchtext.experimental.datasets import WikiText2
train_data, test_data, valid_data = WikiText2()
old_vocab = train_data.get_vocab()
new_vocab = torchtext.vocab.Vocab(counter=old_vocab.freqs, min_freq=10)
new_train_data, new_test_data, new_valid_data = WikiText2(vocab = new_vocab)
```

lepoeme20:
I agree with your mention "We would like to avoid a complicate API for the dataset."

However for users whose computer resources are not so good, the fact that call dataset which has very large vocab size without adjusting min frequency is very burdensome. This is the reason why I suggested a little bit complicated dataset API.

Tanks to your reply :)

mttk:
There could be a solution to this that doesn't overload parameters when constructing datasets, by either passing functions which can modify the frequency list before the vocabulary is constructed or using **kwargs (which might be cleaner).
1. 
```python
def build_vocab_from_iterator(iterator, pre_vocab_hooks=[]):
  # ...
  for hook in pre_vocab_hooks:
    counter = hook(counter) # ex. filter words with freq < 1, etc.
  word_vocab = Vocab(counter, **kwargs)
```

2. 
```python
def build_vocab_from_iterator(iterator, **kwargs):
  # ...
  word_vocab = Vocab(counter, **kwargs)
```

Otherwise, common things like frequency or stop word filtering become a bit too complicated.

cpuhrsch:
Please see for a discussion related to this: https://github.com/pytorch/text/issues/691

@lepoeme20 - What do you think about, for now, adding a flag that creates an IterableDataset which yields lines of raw UTF-8? You could then pass this dataset into build_vocab_from_iterator, since it has all the characteristics of an iterator, and still do tokenization + Vocab on the fly without having to fully read all text into memory.

More concretely, you'll then use build_vocab_from_iterator by itself, i.e.

```
vocab = build_vocab_from_iterator(<create IterableDataset>, min_freq=XYZ)
```

and can pass whatever flags you want :)

And then you can do

```
dataset = <create Dataset>(vocab = vocab)
```

To get a numericalized dataset that lives in memory.

erip:
I think this can be closed. vocab factory functions support `min_freq` now. cc @Nayef211

Nayef211:
Closing this as obsolete

PR #688, opened 2020-02-03
Added WMT News Crawl Dataset for language modeling

### LINKED ISSUE #645: Adding Google 1 Billion Benchmark Dataset to PyTorch dataset
Hello,

Currently, for language modelling, PyTorch has 3 built-in datasets (WikiText103, WikiText2, Penn Treebank). Would it be possible to add the Google 1 Billion Benchmark dataset as one of the PyTorch language modelling built-in data?

The link to the Google 1 Billion Benchmark dataset is below:

http://www.statmt.org/lm-benchmark/

Thanks,

anmolsjoshi:
Fixes #645 

- Added WMT News Crawl dataset for language modeling


anmolsjoshi:
@zhangguanheng66 I added the 2011 dataset per your instructions - was unable to find the validation and test sets. What are your thoughts on this?

anmolsjoshi:
@zhangguanheng66 I noticed an issue in writing this PR. 

zip and tar files are handled differently.

Assuming both .zip and .tar files are stored to .data, the output of filenames from extract_archive is different. 

.zip output files such as - "en-us/train.txt"

.tar output files such as - ".data/en-us/train.txt"

This causes inconsistency in the _setup_datasets function as the root is being joined to the outputted filenames. 

Is this be design? In my opinion, this should be corrected

anmolsjoshi:
@zhangguanheng66 as a note, this PR is able to download files correctly and setup the dataset just fine. But, it takes a very long time to create the dataset given that there are over 2 million rows in the News Crawl dataset - it takes a while to tokenize the entire dataset.

zhangguanheng66:
> @zhangguanheng66 as a note, this PR is able to download files correctly and setup the dataset just fine. But, it takes a very long time to create the dataset given that there are over 2 million rows in the News Crawl dataset - it takes a while to tokenize the entire dataset.

Thanks for the contribution. I understand. It's a huge dataset. 

How do you feel about the new abstract, compared with the old pattern? Any feedbacks?

anmolsjoshi:
@zhangguanheng66 this PR should be good to go. Let me know if you have any comments!

anmolsjoshi:
@zhangguanheng66 any thoughts on overloading the __iter__ method for language modeling?

zhangguanheng66:
> @zhangguanheng66 any thoughts on overloading the **iter** method for language modeling?

Ideally, the `iter` method should be handled by `DataLoader`, rather than torchtext. We want to eventually retire those legacy code, including `batch`, `split`.

torchtext/utils.py
```diff
@@ -178,7 +178,7 @@ def extract_archive(from_path, to_path=None, overwrite=False):
             for file_ in tar:
                 file_path = os.path.join(to_path, file_.name)
                 if file_.isfile():
-                    files.append(file_path)
+                    files.append(file_.name)
```
zhangguanheng66: Will this change introduce any BC breaking?

test/test_utils.py
```diff
@@ -1,118 +1,120 @@
...
-        archive_path = utils.download_from_url(url)
-        assert target_archive_path == archive_path
-
-        # assert that non-valid file (not an archive) raises error
-        with self.assertRaises(NotImplementedError):
-            utils.extract_archive(archive_path)
-
-        # remove file
-        conditional_remove(archive_path)
+import os
```
cpuhrsch: Why are there so many changes to this file?

torchtext/datasets/text_classification.py
```diff
@@ -119,9 +120,9 @@ def _setup_datasets(dataset_name, root='.data', ngrams=1, vocab=None, include_un
 
     for fname in extracted_files:
         if fname.endswith('train.csv'):
-            train_csv_path = fname
+            train_csv_path = os.path.join(root, fname)
```
cpuhrsch: This is BC breaking

torchtext/experimental/datasets/language_modeling.py
```diff
@@ -90,6 +87,12 @@ def _setup_datasets(dataset_name, tokenizer=get_tokenizer("basic_english"),
         dataset_tar = download_from_url(URLS[dataset_name], root=root)
         extracted_files = [os.path.join(root, d) for d in extract_archive(dataset_tar)]
 
+    if dataset_name == "WMTNewsCrawl":
+        data_select = ('train',)
```
cpuhrsch: Why is there a hardcoded data_select?

torchtext/experimental/datasets/text_classification.py
```diff
@@ -42,6 +43,7 @@ def _generate_data_iterators(dataset_name, root, ngrams, tokenizer, data_select)
 
     dataset_tar = download_from_url(URLS[dataset_name], root=root)
     extracted_files = extract_archive(dataset_tar)
+    extracted_files = [os.path.join(root, f) for f in extracted_files]
```
cpuhrsch: This is BC breaking

torchtext/utils.py
anmolsjoshi: Yes it will, but currently this function has different outputs for zip and tar files. WMT dataset is tar.gz given how _setup_datasets work, it appends the root to the extracted files, this will add the root twices for tar extracted files. 

I think it’s best to correct the function rather than include an if/else for non zip files

test/test_utils.py
anmolsjoshi: I’m not sure. I only changed a few lines, I’ll try pushing again

torchtext/datasets/text_classification.py
anmolsjoshi: It is. See comment above about .zip and .tar files being handled by extract_archive

torchtext/experimental/datasets/language_modeling.py
anmolsjoshi: There is no validation and test set provided. Would you prefer I added it as a default argument?

torchtext/experimental/datasets/text_classification.py
anmolsjoshi: See above!

anmolsjoshi:
Main reason for changes is due to the code below. 

I can revert back all my code to the original and include a if/else in _setup_datasets function for non zip files

```python
import os
from torchtext.utils import extract_archive, donwload_from_url

os.mkdir('.data')
download_from_url('http://www.quest.dcs.shef.ac.uk/wmt16_files_mmt/validation.tar.gz')
# '.data/validation.tar.gz'

download_from_url('https://s3.amazonaws.com/research.metamind.io/wikitext/wikitext-2-v1.zip')
# '.data/wikitext-2-v1.zip'

extract_archive('.data/validation.tar.gz')
# ['.data/val.de', '.data/val.en']

extract_archive('.data/wikitext-2-v1.zip')
# ['wikitext-2/', 'wikitext-2/wiki.test.tokens', 'wikitext-2/wiki.valid.tokens', 'wikitext-2/wiki.train.tokens']
```

@cpuhrsch @zhangguanheng66

anmolsjoshi:
Main reason for changes is due to the code below. 

I can revert back all my code to the original and include a if/else in _setup_datasets function for non zip files

```python
import os
from torchtext.utils import extract_archive, donwload_from_url

os.mkdir('.data')
download_from_url('http://www.quest.dcs.shef.ac.uk/wmt16_files_mmt/validation.tar.gz')
# '.data/validation.tar.gz'

download_from_url('https://s3.amazonaws.com/research.metamind.io/wikitext/wikitext-2-v1.zip')
# '.data/wikitext-2-v1.zip'

extract_archive('.data/validation.tar.gz')
# ['.data/val.de', '.data/val.en']

extract_archive('.data/wikitext-2-v1.zip')
# ['wikitext-2/', 'wikitext-2/wiki.test.tokens', 'wikitext-2/wiki.valid.tokens', 'wikitext-2/wiki.train.tokens']
```

@cpuhrsch @zhangguanheng66

torchtext/experimental/datasets/language_modeling.py
cpuhrsch: If this is the case you should error out and say it doesn't exist and have the user explicitly pass in this option. There are a few arguments to be made in favor of this:
- the number of expected return values depends on this option.
- the user might not be aware of the fact that there is no validation and no test dataset.

anmolsjoshi: Agreed! I will account for that and push changes later today

torchtext/experimental/datasets/text_classification.py
zhangguanheng66: I would prefer a separate PR to address this issue. It makes this PR a little bit overwhelming. I agree that both download/extract functions should handle zip/tar files consistently. @cpuhrsch how do you feel about it?

cpuhrsch: Agreed. It's unrelated to this and we might want to explicitly communicate this change.

cpuhrsch:
@anmolsjoshi - I think it makes sense to fix this separately and before merging this.

anmolsjoshi:
@zhangguanheng66 @cpuhrsch - I'll push a branch up later today fixing this tar/zip issue. And we can move forward with the WMT dataset after. 

Thanks for your review!

anmolsjoshi:
@zhangguanheng66 @cpuhrsch - quick question before I proceed - are the filenames returned from zip or tar correct i.e. should root folder be prepended to the path? 

I think the filenames from zip (root prepended) might be better as it is consistent with download_url function

cpuhrsch:
@anmolsjoshi - I think we should always return the full path, because it gives more flexibility and the user knows exactly what's going on. So, if prepending root gives us that, that's what we should do.

anmolsjoshi:
Closing in favor of #700

torchtext/utils.py
anmolsjoshi: Resolved by #692

torchtext/datasets/text_classification.py
anmolsjoshi: Resolved by #692

torchtext/experimental/datasets/text_classification.py
anmolsjoshi: Resolved by #692

test/test_utils.py
anmolsjoshi: Fixed by #692

anmolsjoshi:
@zhangguanheng66 @cpuhrsch I have incorporated changes requested in an earlier review and made some additional changes. Here is a summary:
- Removed any code related to extract_archive fix and moved it to #692 
- Added a check for data_select for WMT News Crawl, it can only be train as other datasets are not provided
- Added tests to load the dataset and check if errors are raised for incorrect data_select
- Made corrections to spelling errors, mainly correcting tupel to tuple

zhangguanheng66:
It seems that there is an even larger Wikitext dataset like this one,
https://dl.fbaipublicfiles.com/fairseq/data/wikipedia.en_filtered.gz
Any thought?

anmolsjoshi:
Reading from the [website](http://www.statmt.org/wmt11/translation-task.html#download), 2009 is the largest dataset.
> - From Europarl (403MB) md5 sha1
> - From the News Commentary corpus (41MB) md5 sha1
> - From the News Crawl corpus (2007 only) (1.1 GB) md5 
> - From the News Crawl corpus (2008 only) (3.4 GB) md5 
> - From the News Crawl corpus (2009 only) (3.7 GB) md5 
> - From the News Crawl corpus (2010 only) (1.4 GB) md5 
> - From the News Crawl corpus (2011 only) (229 MB) md5 sha1 

The reason I picked 2011 was due to your [comment](https://github.com/pytorch/text/issues/645#issuecomment-555568763) on #645

zhangguanheng66:
> Reading from the [website](http://www.statmt.org/wmt11/translation-task.html#download), 2009 is the largest dataset.
> 
> > * From Europarl (403MB) md5 sha1
> > * From the News Commentary corpus (41MB) md5 sha1
> > * From the News Crawl corpus (2007 only) (1.1 GB) md5
> > * From the News Crawl corpus (2008 only) (3.4 GB) md5
> > * From the News Crawl corpus (2009 only) (3.7 GB) md5
> > * From the News Crawl corpus (2010 only) (1.4 GB) md5
> > * From the News Crawl corpus (2011 only) (229 MB) md5 sha1
> 
> The reason I picked 2011 was due to your [comment](https://github.com/pytorch/text/issues/645#issuecomment-555568763) on #645

I think it could be interesting to include at least one more larger Crawl corpus so users have the option to try both.

anmolsjoshi:
Is the idea to have multiple functions for different years' datasets or provide an argument for the year?

zhangguanheng66:
> Is the idea to have multiple functions for different years' datasets or provide an argument for the year?

Correct me if I'm wrong @anmolsjoshi @cpuhrsch , I don't think there is a significant differences between years. We could just provide different Crawl corpus with different size. Users may want to train the models with different datasets depending on their memory.

anmolsjoshi:
Should I update the current dataset to 2009?

Which other datasets would you want to provide?

zhangguanheng66:
Maybe we could add one more argument (as you did for language) so user can explicitly choose the one they like. And in the docs, we clearly mark the number of tokens and the memory side for the corresponding dataset.

I'm using 2010 English one with 1.4G in total for my BERT model. The pipeline you built is flexible enough to switch between years/languages. Thanks a lot.

anmolsjoshi:
@zhangguanheng66 thanks for the comments. 

I've added an option where users can pass the year and a table in the docstrings with details about the news crawl datasets by year. 

Let me know what you think.

anmolsjoshi:
@zhangguanheng66 I saw discussion in #691 and #690, code in #696 - Is there value in decoupling vocab and LanguageModelingDataset as well?

zhangguanheng66:
@anmolsjoshi We want to de-couple the vocab object from dataset but are not very sure the design. I will work on some cases and pull you guys for a look.

anmolsjoshi:
Thanks! Let me know if any other changes are needed on this PR!

torchtext/experimental/datasets/language_modeling.py
```diff
@@ -73,7 +74,7 @@ def _get_datafile_path(key, extracted_files):
 
 def _setup_datasets(dataset_name, tokenizer=get_tokenizer("basic_english"),
                     root='.data', vocab=None, removed_tokens=[],
-                    data_select=('train', 'test', 'valid')):
+                    data_select=('train', 'test', 'valid'), **kwargs):
```
cpuhrsch: I'm not a fan of adding a generic **kwargs to this function, especially since it seems to serve the function of expanding data_select, which is already a well defined parameter. This runs the chance of making the APIs inconsistent. We should think about how to extend data_select to support cases like this.

The obvious choice is to expand it by using the cross-product, but that'll yield a lot of potential arguments. The other idea could be to have a NamedTuple object as a sort of argument object. Also, this seems to only create the training dataset. This means there is no predefined set, so the distinction between training, validation and test is arbitrary / not defined, so we might as well drop it. We could create an WMTNewsCrawlOptions namedtuple that is constructed by giving it a Year and a Language and passed to data_select. I'm sure there are some other options here as well.

PR #692, opened 2020-02-06
[Bug Fixing][BC Breaking] Unify tar and zip handling with extract_archive

anmolsjoshi:
Corrected behavior described in this [PR](https://github.com/pytorch/text/pull/688#issuecomment-582543743) 

Before this PR:
```python
import os
from torchtext.utils import extract_archive, donwload_from_url

os.mkdir('.data')
download_from_url('https://s3.amazonaws.com/research.metamind.io/wikitext/wikitext-2-v1.zip')
# '.data/wikitext-2-v1.zip'

extract_archive('.data/wikitext-2-v1.zip')
# ['wikitext-2/', 'wikitext-2/wiki.test.tokens', 'wikitext-2/wiki.valid.tokens', 'wikitext-2/wiki.train.tokens']
```

After this PR
```python
import os
from torchtext.utils import extract_archive, donwload_from_url

os.mkdir('.data')

download_from_url('https://s3.amazonaws.com/research.metamind.io/wikitext/wikitext-2-v1.zip')
# '.data/wikitext-2-v1.zip'

extract_archive('.data/wikitext-2-v1.zip')
# ['.data/wikitext-2/wiki.test.tokens', '.data/wikitext-2/wiki.valid.tokens', '.data/wikitext-2/wiki.train.tokens']
```

**NOTE**: This PR is BC Breaking, if you are using this function to build custom datasets, there is no need to write code to join the root and extracted filenames. The outputted filenames have the root folder prepended to the extracted filenames.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -30,6 +30,11 @@ class TestUtils(TorchtextTestCase):
         assert files == [os.path.join(root, 'val.de'),
                          os.path.join(root, 'val.en')]
 
+        # extract files with overwrite option True
+        files = utils.extract_archive(archive_path, overwrite=True)
+        assert files == [os.path.join(root, 'val.de'),
+                         os.path.join(root, 'val.en')]
+
         # remove files and archive
         for f in files:
             conditional_remove(f)
@@ -57,6 +62,10 @@ class TestUtils(TorchtextTestCase):
                          'en-ud-v2/README.txt']
         # extract files and ensure they are correct
         files = utils.extract_archive(archive_path)
+        assert files == [os.path.join(root, f) for f in correct_files]
+
+        # extract files with overwrite option True
+        files = utils.extract_archive(archive_path, overwrite=True)
         assert files == [os.path.join(root, f) for f in correct_files]
 
         # remove files and archive
@@ -87,6 +96,11 @@ class TestUtils(TorchtextTestCase):
 
         # extract files and ensure they are in the to_path directory
         files = utils.extract_archive(archive_path, to_path)
+        assert files == [os.path.join(to_path, 'val.de'),
+                         os.path.join(to_path, 'val.en')]
+
+        # extract files with overwrite option True
+        files = utils.extract_archive(archive_path, to_path, overwrite=True)
         assert files == [os.path.join(to_path, 'val.de'),
                          os.path.join(to_path, 'val.en')]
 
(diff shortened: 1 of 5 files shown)


anmolsjoshi:
@zhangguanheng66 I updated the title and description of this PR

anmolsjoshi:
I suspect the reason Github is oddly displaying tests/test_utils.py because of CRLF/LF. Although it does highlight what changed

anmolsjoshi:
TODO: will be adding tests for overwrite option

zhangguanheng66:
@cpuhrsch LGTM. I could run a smoke test to make sure the datasets work properly.

cpuhrsch:
Let's sit on this a little bit to make sure it doesn't break a lot of stuff. @zhangguanheng66 - I think it's worth running more tests on your end.

zhangguanheng66:
I have ran the smoke test for the release on the new branch and major datasets seem fine. The only difference, I notice, is that the directory path (e.g. 'wikitext-2/') is not included in the outputs of `extract_archive` function. This is different from the behaviors of the master branch.

anmolsjoshi:
> I notice, is that the directory path (e.g. 'wikitext-2/') is not included in the outputs of extract_archive function. This is different from the behaviors of the master branch.

@zhangguanheng66 I can update the code so wikitext-2/ is returned. I personally think that only filenames should be returned from this function

zhangguanheng66:
Did some offline research about the history of archive function. Long time ago (back to 2017), the download/archive functions were handled within `torchtext.dataset`. Then, `extract_archive` function was introduced with the support of tar file only. In October 2019, `extract_archive` function was extended to support zip file. But unfortunately, the `extract_archive` function has slightly different behaviors between tar and zip files (a.k.a. the root folder prepended to the path described by this PR). It's a bug fix and, IMP, the zip file should follow the behavior of the tar file. Thanks @anmolsjoshi for fixing it.

zhangguanheng66:
And please address the fix in the unit test.

zhangguanheng66:
@cpuhrsch there are two unit tests for `tar` and `zip`, respectively. In the tests, the extracted files have the paths with joint root and file path. see [here](https://github.com/pytorch/text/pull/692/commits/a4cdc72e0d9f157f423eb2bb2db611970aca3f26)

anmolsjoshi:
@cpuhrsch following up about this PR

zhangguanheng66:
Fix a bug introduced by https://github.com/pytorch/text/pull/602 where `extract_archive` func shows different behaviors for tar and zip files.

Issue #690, opened 2020-02-05
Adding validation splits to (experimental) text_classification datasets that do not have vocabulary built over them


bentrevett:
## 🚀 Feature
The experimental text_classification datasets should have a way to build a validation set from them, without the vocabulary being built over the validation set. 

**Motivation**

In ML, you should always have a test, validation and training set. In NLP, your vocabulary should be built from the training set only, and not from the test/validation. 

The current experimental text classification (IMDB) dataset does not have a validation set and automatically builds the vocabulary whilst loading the train/test sets. After loading the train and test sets, we would need to construct a validation set with `torch.utils.data.random_split`. The issue here is that our vocabulary has already been built over the validation set we are about to create. There is currently no way to solve this issue.

**Pitch**

`'valid'` should be accepted as a `data_select` argument and should create a validation set before the vocabulary has been created over the training set. As the IMDB dataset does not have a standardized validation split, we can do something like taking the last 20% of the training set. 

I am proposing something like the following after the `iters_group` is created [here](https://github.com/pytorch/text/blob/master/torchtext/experimental/datasets/text_classification.py#L63):

```python
from itertools import islice, tee

if 'valid' in iters_group.keys():
    train_iter_a, train_iter_b, train_iter_c = tee(iters_group['train'], 3)
    len_train = int(sum(1 for _ in train_iter_a) * 0.8)
    iters_group['valid'] = islice(train_iter_b, len_train, None)
    iters_group['train'] = islice(train_iter_c, 0, len_train)
    iters_group['vocab'] = islice(iters_group['vocab'] 0, len_train)
```

`tee` duplicates generators and `islice` slices into generators. We need to duplicate the training data iterator as we will be using it three times. We use the first iterator to get the length of the training set so we know what size the validation set will be (the last 20% of the examples from the training set). We then use `islice` to get the last 20% of the training examples to form the validation set, the first 80% of the training examples to use as the new training set, and the first 80% examples of the "vocab" set as this needs to match the training set as it is what we want to build our vocab from.

We can now correctly load a train, valid and test set with vocabulary built only over the training set:
```python
from torchtext.experimental import datasets

train_data, valid_data, test_data = datasets.IMDB(data_select=('train', 'valid', 'test'))
```

Can also load a custom vocabulary built from the original vocabulary like so (note that 'valid' needs to be in the `data_select` when building the original vocabulary):
```python
from torchtext import vocab
from torchtext.experimental import datasets

[...]

PR #696, opened 2020-02-19
Return raw tokens for experimental IMDB dataset

zhangguanheng66:
Return raw text data.

### Return tokens
```
import torchtext
from torchtext.experimental.datasets import IMDB
vocab = lambda x: x
train_dataset, test_dataset = IMDB(vocab=vocab)
```

### Return raw text
```
import torchtext
from torchtext.experimental.datasets import IMDB
vocab = lambda x: x
tokenizer = get_tokenizer('empty_tokenizer')
train_dataset, test_dataset = IMDB(vocab=vocab, tokenizer=tokenizer)
```

### DIFF
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -128,15 +128,12 @@ class TestDataset(TorchtextTestCase):
         conditional_remove(datafile)
 
     @slow
-    def test_iterabledataset_imdb(self):
+    def test_raw_imdb(self):
         from torchtext.experimental.datasets import IMDB
         # smoke test to ensure imdb works properly
-        train_dataset, = IMDB(data_select='train', raw_text=True)
-        cached_dataset = list(train_dataset)
-        self.assertEqual(len(cached_dataset), 25000)
-        test_dataset, = IMDB(data_select='test', raw_text=True)
-        cached_dataset = list(test_dataset)
-        self.assertEqual(len(cached_dataset), 25000)
+        train_dataset, test_dataset = IMDB(raw_text=True)
+        self.assertEqual(len(train_dataset), 25000)
+        self.assertEqual(len(test_dataset), 25000)
 
         # Delete the dataset after we're done to save disk space on CI
         datafile = os.path.join(self.project_root, ".data", "imdb")
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -129,9 +129,11 @@ class TestDataset(TorchtextTestCase):
 
     @slow
     def test_raw_imdb(self):
+        def return_same(x):
+            x
         from torchtext.experimental.datasets import IMDB
         # smoke test to ensure imdb works properly
-        train_dataset, test_dataset = IMDB(raw_text=True)
+        train_dataset, test_dataset = IMDB(vocab=return_same)
         self.assertEqual(len(train_dataset), 25000)
         self.assertEqual(len(test_dataset), 25000)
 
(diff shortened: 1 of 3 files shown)


test/data/test_builtin_datasets.py
```diff
@@ -126,3 +126,22 @@ def test_imdb(self):
         conditional_remove(datafile)
         datafile = os.path.join(self.project_root, ".data", "aclImdb_v1.tar.gz")
         conditional_remove(datafile)
+
+    @slow
+    def test_iterabledataset_imdb(self):
+        from torchtext.experimental.datasets import IMDB
+        # smoke test to ensure imdb works properly
+        train_dataset, = IMDB(data_select='train', raw_text=True)
+        cached_dataset = list(train_dataset)
```
cpuhrsch: If you're materializing the entire dataset you could also compare it to a sort-of hacky version of IMDB. We know the layout of this particular dataset and can simply read it in via read(<path>).read().split('\n'). Then we could do an actual comparison to an underlying reference (for now).

torchtext/experimental/datasets/text_classification.py
```diff
@@ -33,6 +44,19 @@ def _imdb_iterator(key, extracted_files, tokenizer, ngrams, yield_cls=False):
                     yield ngrams_iterator(tokenizer(f.read()), ngrams)
 
 
+def setup_iterable_datasets(dataset_name, root, ngrams, tokenizer, data_select):
```
cpuhrsch: We don't need ngrams and tokenizer for this right?

zhangguanheng66:
fix https://github.com/pytorch/text/issues/690

torchtext/experimental/datasets/text_classification.py
```diff
@@ -16,8 +16,15 @@
...
     for cls, tokens in iterator:
-        yield cls, iter(map(lambda x: vocab[x],
-                        filter(lambda x: x not in removed_tokens, tokens)))
+        if vocab is None:
+            yield cls, tokens
+        else:
+            if callable(vocab):
+                yield cls, iter(map(lambda x: vocab(x),
+                                filter(lambda x: x not in removed_tokens, tokens)))
+            else:
```
cpuhrsch: The assumption here is that if vocab is not a callable it'll be an instance of Vocab? We could add a type assertion here at the beginning.

We could also create a version that accepts a callable and then pass in the callable ```lambda x: vocab[x]```.

torchtext/experimental/datasets/text_classification.py
```diff
@@ -81,8 +88,13 @@ def _setup_datasets(dataset_name, root='.data', ngrams=1, vocab=None,
         logging.info('Creating {} data'.format(item))
         data_iter = _create_data_from_iterator(vocab, iters_group[item], removed_tokens)
         for cls, tokens in data_iter:
-            data[item]['data'].append((torch.tensor(cls),
-                                       torch.tensor([token_id for token_id in tokens])))
+            tok_list = [tok_id for tok_id in tokens]
+            try:
+                data[item]['data'].append((torch.tensor(cls),
```
cpuhrsch: In which cases will this conversion not succeed? Using try except is a pretty big hammer. ValueError could be raised by many functions.

bentrevett: This will fail when `tokens` is a list of string(s), e.g. the new `empty_tokenizer` is used for the case when you don't even want to tokenize your data.

zhangguanheng66:
merged in https://github.com/pytorch/text/pull/701

PR #697, opened 2020-02-21
remove duplicate document in dataset

srayuws:
Language Modeling part is duplicated at the start and between Entailment and Machine Translation

### DIFF
--- a/docs/source/datasets.rst
+++ b/docs/source/datasets.rst
@@ -36,6 +36,37 @@ The following datasets are available:
 .. contents:: Datasets
     :local:
 
+
+Language Modeling
+^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


srayuws:
In the documentation page https://pytorch.org/text/datasets.html here, there are two identical sections of `Language Modeling` :

https://pytorch.org/text/datasets.html#language-modeling, and
https://pytorch.org/text/datasets.html#id1

I think these two sections are duplicate? Or we need some more description about their difference?

zhangguanheng66:
Got it. I would rather remove the second one. Thanks.

srayuws:
Updated. Also added some empty newlines to make each section is separated by two lines as in other parts.

PR #701, opened 2020-02-28
Text classification datasets with new torchtext dataset abstraction

zhangguanheng66:
A new dataset abstraction that decouples data and vocab/tokenizer.

To load raw text dataset:
```
from torchtext.experimental.datasets import RawAG_NEWS
train, test = RawAG_NEWS()

# Process text data
from torchtext.experimental.datasets.text_classification import build_vocab
from torchtext.experimental.transforms import TokenizerTransform, VocabTransform, ToTensor
from torch.nn import Sequential

vocab = build_vocab(train, TokenizerTransform(get_tokenizer('basic_english')))
text_transform = Sequential(TokenizerTransform(get_tokenizer('basic_english')), VocabTransform(vocab), ToTensor())
label_transform = ToTensor()
for (label, txt) in train[:10]:
    print(label_transform(label), text_transform(txt))
```
Or wrap up everything above and load processed dataset with one-command:
```
from torchtext.experimental.datasets import AG_NEWS
train, test = AG_NEWS()
```


zhangguanheng66:
@fmassa @cpuhrsch @vincentqb

zhangguanheng66:
Need to have `VocabTransform` and `TokenizerTransform` scriptable.

```
import torch
import torchtext

# Test VocabTransform scriptable
vocab = {'here': 1, 'we': 2, 'are':3}
transform1 = torchtext.experimental.transforms.VocabTransform(vocab)
transform1(['here', 'we', 'are'])
jit_method = torch.jit.script(transform1)

# Test TokenizerTransform scriptable
transform1 = torchtext.experimental.transforms.TokenizerTransform('basic_english')
transform1('here we are')
jit_method = torch.jit.script(transform1)
```

torchtext/experimental/datasets/new_text_classification.py
```diff
@@ -0,0 +1,297 @@
...
+def _create_data_from_csv(data_path):
+    data = []
+    with io.open(data_path, encoding="utf8") as f:
+        reader = unicode_csv_reader(f)
+        for row in reader:
+            data.append((row[0], ' '.join(row[1:])))
+    return data
+
+
+def build_vocab(dataset, transform):
```
bentrevett: Should this take a kwargs argument that will pass arguments to the vocab constructor inside `build_vocab_from_iterator`? This will allow us to do something like:

```
train, test = AG_NEWS()
transform1 = TokenizerTransform('basic_english')
train, valid = torch.utils.data.random_split(train, [90_000, 10_000]) #not exact numbers
vocab = build_vocab(train, transform1, max_size = 25_000, min_freq = 2)
```

It will also mean `build_vocab_from_iterator` will also need to be modified to accept kwargs too.

zhangguanheng66: This could be added. However, the wrap-up to build `vocab` is pretty simple and users now have the flexibility to do that themselve.

vincentqb:
```
import torchtext
from torchtext.experimental.datasets import AG_NEWS
from torchtext.experimental.transforms import TokenizerTransform, VocabTransform, Compose
from torchtext.experimental.datasets.new_text_classification import build_vocab

# Import raw text strings
train, test = AG_NEWS()

# Build tokenizer transform
transform1 = TokenizerTransform('basic_english')

transformed_train = ...

# Build vocab transform
vocab = VocabTransform(transformed_train)

# A new dataset with raw text strings + label/string transforms
from torchtext.experimental.datasets.new_text_classification import TextClassificationDataset
new_train = TextClassificationDataset(train.data, [int, transform1, vocab])
```

cpuhrsch:
How about 
```
import torchtext
from torchtext.experimental.datasets import AG_NEWS
from torchtext.experimental.transforms import TokenizerTransform, VocabTransform, Compose
from torchtext.experimental.datasets.new_text_classification import build_vocab

# Import raw text strings
train, _ = AG_NEWS()

def tokenizer(raw_text):
    splits = raw_text.split
    label, row = int(splits[0]), splits[1:]
    return label, TokenizerTransform('basic_english')(row) #Assuming this has 0 init cost

tokenized_train = apply(train, tokenizer) # For lack of a better function name

# Build vocab transform
vocab = VocabTransform(tokenized_train)

# Assuming datasets don't need to be reset after consumed
new_train = apply(tokenized_train, vocab) 
```

vincentqb:
@fmassa points to [tf](https://www.tensorflow.org/api_docs/python/tf/data/Dataset)

zhangguanheng66:
Some offline discussions:

-  Transforms are in general some callable objects and used to map the inputs according to a "contract". For example, a tokenizer transform defines the "contract" to convert a string to a list of tokens.
-  Vocab transform contains a "dictionary" object which maps tokens to ids.
-  A separate folder is created to save the raw text datasets. For the existing datasets in torchtext library, Some standard transforms and the raw text datasets are wrapped together to support "one-command" data loading.

torchtext/experimental/datasets/raw_text_classification.py
```diff
@@ -0,0 +1,250 @@
...
+def _create_data_from_csv(data_path):
+    data = []
+    with io.open(data_path, encoding="utf8") as f:
+        reader = unicode_csv_reader(f)
+        for row in reader:
+            data.append((int(row[0]), ' '.join(row[1:])))
+    return data
+
+
+class RawTextDataset(torch.utils.data.Dataset):
```
cpuhrsch: I think you can do this by putting it into a "raw" folder and drop the Raw prefix

torchtext/experimental/datasets/text_classification.py
```diff
@@ -1,93 +1,455 @@
...
+                                           (label_transform, text_transform))
+                 for item in data_select)
 
-def _setup_datasets(dataset_name, root='.data', ngrams=1, vocab=None,
-                    removed_tokens=[], tokenizer=None,
-                    data_select=('train', 'test')):
 
-    if isinstance(data_select, str):
-        data_select = [data_select]
+def AG_NEWS(*args, **kwargs):
```
cpuhrsch: If someone does "help(AG_NEWS())"  will it render the arguments correctly? Because of the forwarding this might not work.

torchtext/experimental/transforms.py
```diff
@@ -0,0 +1,154 @@
...
+        Examples:
+            >>> tok_transform = torchtext.experimental.transforms.TokenizerTransform()
+            >>> tok_transform('here we are')
+            >>> ['here', 'we', 'are']
+        """
+        # type: (str) -> List[str]
+        return self.tokenizer(str_input)
+
+
+class VocabTransform(torch.nn.Module):
```
cpuhrsch: It seems like the point here is to map the vocab across a given list of input tokens and then return the list. A user could technically do that herself pretty easily (as you can see by the amount of code). Maybe we can add more value to this by also supporting nested streams of input tokens?

torchtext/experimental/datasets/raw_text_classification.py
cpuhrsch: And then also reference this from the other datasets. So you can create an instance within the higher level datasets that consumes this and share some of the logic such as ```_create_data_from_csv```.

.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
-ignore = E402,E722,W503,W504
+ignore = E402,E722,W503,W504, F821
```
cpuhrsch: Nit: extra space

hudeven:
@zhangguanheng66 the new APIs look great! I have a few questions:
1) How to handle dense/categorical feature with the new API?
2) As the data for language model could be too large to fit in memory. We might have to use IterableDataset. Could you also have a demo for IterableDataset as well?
3) How to support custom batching logic and custom sampling logic for IterableDataset?

torchtext/experimental/datasets/raw/text_classification.py
```diff
@@ -0,0 +1,250 @@
+import torch
+import io
+from torchtext.utils import download_from_url, extract_archive, unicode_csv_reader
+
+URLS = {
```
cpuhrsch, line 5: In torchaudio there recently has been issues with corrupted downloads. Either within this PR or at least as a follow-up item we should look into md5 verification of downloads to make sure the user is actually getting the correct data.

torchtext/experimental/datasets/raw/text_classification.py
```diff
@@ -0,0 +1,250 @@
...
+
+    for fname in extracted_files:
+        if fname.endswith('train.csv'):
+            train_csv_path = fname
+        if fname.endswith('test.csv'):
+            test_csv_path = fname
+
+    train_data = _create_data_from_csv(train_csv_path)
+    test_data = _create_data_from_csv(test_csv_path)
+    return (RawTextDataset(train_data),
```
cpuhrsch, line 66: We should check whether the returned objects here actually carry the doc string of the calling class. I'll look into this a bit.

torchtext/experimental/datasets/raw/text_classification.py
```diff
@@ -0,0 +1,250 @@
...
+        if 'urls' in fname:
+            continue
+        elif key in fname and ('pos' in fname or 'neg' in fname):
+            with io.open(fname, encoding="utf8") as f:
+                label = 1 if 'pos' in fname else 0
+                data_set.append((label, f.read()))
+    return data_set
+
+
+def IMDB(root='.data'):
```
cpuhrsch, line 218: This is the only function that accepts an explicit "root" argument.

In general, let's think about how this could be something more generic than a path. In the future users might want to pass other [file like objects](https://docs.python.org/3/glossary.html#term-file-object) as a source for dataset construction.

torchtext/experimental/datasets/raw/text_classification.py
```diff
@@ -0,0 +1,250 @@
...
+def _create_data_from_csv(data_path):
+    data = []
+    with io.open(data_path, encoding="utf8") as f:
+        reader = unicode_csv_reader(f)
+        for row in reader:
+            data.append((int(row[0]), ' '.join(row[1:])))
+    return data
+
+
+class RawTextDataset(torch.utils.data.Dataset):
```
cpuhrsch, line 36: In light of multiprocessing, data parallelism etc. etc. I wonder if it's worthwhile looking into making this an IterableDataset after all and introducing a convenience function that creates a map-style dataset by simply exhausting the iterator and writing out the result. After all this is the raw dataset, so that'd provide an even more general interface. Let's also look into JIT-ability of that.

torchtext/experimental/datasets/raw/text_classification.py
zhangguanheng66: Just check that `object.__doc__` carries the doc string of the calling class (a.k.a. `RawTextDataset` in this case).

torchtext/experimental/datasets/raw/text_classification.py
zhangguanheng66: Switch to `IterableDataset` for raw text dataset.

torchtext/experimental/datasets/raw/text_classification.py
```diff
@@ -0,0 +1,267 @@
...
+    def __init__(self, iterator):
+        """Initiate text-classification dataset.
+        """
+        super(RawTextIterableDataset, self).__init__()
+        self._iterator = iterator
+        self.has_setup = False
+        self.start = 0
+        self.num_lines = None
+
+    def setup_iter(self, start=0, num_lines=None):
```
cpuhrsch, line 48: What's the expected user interface for this? Why not make this part of the constructor?

torchtext/experimental/datasets/text_classification.py
```diff
@@ -1,93 +1,479 @@
...
+    def get_vocab(self):
+        return self.vocab
 
 
 def _setup_datasets(dataset_name, root='.data', ngrams=1, vocab=None,
-                    removed_tokens=[], tokenizer=None,
-                    data_select=('train', 'test')):
+                    tokenizer=None, data_select=('train', 'test')):
+    text_transform = []
+    if not tokenizer:
```
cpuhrsch, line 103: "if tokenizer is None" is more precise. This will also succeed for empty lists and bool values, which isn't the default.

torchtext/experimental/datasets/text_classification.py
```diff
@@ -1,93 +1,479 @@
-import logging
 import torch
-import io
-from torchtext.utils import download_from_url, extract_archive
-from torchtext.data.utils import ngrams_iterator
 from torchtext.data.utils import get_tokenizer
 from torchtext.vocab import build_vocab_from_iterator
-from torchtext.vocab import Vocab
-from torchtext.datasets import TextClassificationDataset
+from torchtext.experimental.datasets.raw import AG_NEWS as RawAG_NEWS
```
cpuhrsch, line 4: Instead of this should be able to just import "raw" and then index into this using the name of the dataset you're calling this from.

torchtext/experimental/datasets/raw/text_classification.py
zhangguanheng66: The default behavior of the raw datasets is to load all the text strings and labels (no complicate APIs). If users like to cache a chunk of the data (for example the workers in `DataLoader`), they have to explicitly call `setup_iter` function and set up the iterator before caching. IMO, this maintains a very simple and clean API at the raw dataset level.

zhangguanheng66:
fix #690

thvasilo:
I'm wondering if I can use the `RawTextIterableDataset` with a csv input.

I see it inherits from IterableDataset, while TabularDataset inherits from torchtext.data.dataset.Dataset->torch.utils.data.Dataset.

So currently since I'm reading my data from a CSV I have to go the old route of creating text and label fields (for a binary classification task).

I'm wondering if I can replace that with these new `RawTextIterableDataset`, and if you have any examples of how to do that.

zhangguanheng66:
`RawTextIterableDataset` actually sets up the iterator to read a CSV file. Then, `TextIterableDataset` caches the iterator for labels and text strings. If you combine those two steps together, it will be very similar to https://github.com/pytorch/text/blob/master/torchtext/datasets/text_classification.py.
In addition to the raw text strings, you have to set up the tokenizer/vocab transforms. With the new dataset abstraction, we hope to get rid of the old utils (like fields).

PR #707, opened 2020-03-13
Bugfix: custom UNK token conversion error

ohke:
Bugfix: https://github.com/pytorch/text/issues/618

Newly, this changes adds `unk_token` argument to build_vocab method for set by Field.
Also, for backward compatibility, this PR leaves `Vocab.UNK` as  default token.

### DIFF
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -34,7 +34,8 @@ class Vocab(object):
     UNK = '<unk>'
 
     def __init__(self, counter, max_size=None, min_freq=1, specials=['<unk>', '<pad>'],
-                 vectors=None, unk_init=None, vectors_cache=None, specials_first=True):
+                 vectors=None, unk_init=None, vectors_cache=None, specials_first=True,
+                 unk_token=None):
         """Create a Vocab object from a collections.Counter.
 
         Arguments:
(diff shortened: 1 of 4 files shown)


torchtext/vocab.py
```diff
@@ -34,7 +34,8 @@ class Vocab(object):
     UNK = '<unk>'
 
     def __init__(self, counter, max_size=None, min_freq=1, specials=['<unk>', '<pad>'],
-                 vectors=None, unk_init=None, vectors_cache=None, specials_first=True):
+                 vectors=None, unk_init=None, vectors_cache=None, specials_first=True,
+                 unk_token=None):
```
zhangguanheng66: Not sure if we want to add extra argument to the API. Ideally we should put all the special tokens in `specials`. `stoi` should return `'<unk>'` id if the token is not found.

bentrevett: How will we know which of the special tokens is the unk token without explicitly passing it as an argument?

zhangguanheng66: I'm not against the hard-code `unk` token and assign it to the vocab class. I just don't like the idea to make the API longer and longer.

ohke: Thanks for your comments.

I don't want to increase the arguments to maintain if possible, but Field class accespts custom `unk_token`. 
Field class adds `unk_token` to `specials` in `build_vocab` method, so there is no problem when accessing via Field object.

We can include arguments check if `unk_token` exists in `specials` , in Vocab initialize method.

ohke: I added one argument validation.

PR #718, opened 2020-04-02
Add a judgement logic when loading dataset

scarydemon2:
When users loading a dataset like AG_NEWS , the "_setup_datasets" function will download directly , But if the user has downloaded the unpacked dataset manually from the given url ，the function won't extract it directly . So I write a judgement logic and related dic UNEXTRACT_NAMES to judge whether a unpacked file is exist.
My implementation is a bit dirty. I hope you can suggest improvements . If you don't need such a logic ,just close this PR

### DIFF
--- a/torchtext/datasets/text_classification.py
+++ b/torchtext/datasets/text_classification.py
@@ -29,7 +29,7 @@ URLS = {
 }
 
 
-UNEXTRACT_NAMES={
+UNEXTRACT_NAMES = {
     'AG_NEWS':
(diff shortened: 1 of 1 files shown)


zhangguanheng66:
The download/extract_archive functions can automatically detect the existing files. If it exists, they will skip.

scarydemon2:
> Since this is not the normal way to utilize the datasets, uses could easily delete the zip files.

Ohhh,you are right,and sounds reasonable .  And this PR is based on a special situation

PR #720, opened 2020-04-02
MultiheadAttention building blocks in torchtext

zhangguanheng66:
We propose to refactor `nn.MultiheadAttention` module as a MHA container:

- InProjContainer
- ScaledDotProduct
- A regular linear layer for out-projection.

The objective is to add more flexibility to try different MHA variants. The new MHA container is capable of
- Drop-in replacement. It's easy to switch from `nn.MultiheadAttention` to MHA container.

To initiate `nn.MultiheadAttention`:
```
mha = nn.MultiheadAttention(embed_dim, nhead)
```
To initiate MHA container:
```
in_proj_container = InProjContainer(Linear(embed_dim, embed_dim), Linear(embed_dim, embed_dim), Linear(embed_dim, embed_dim))
mha = MultiheadAttentionContainer(nhead, in_proj_container, ScaledDotProduct(), Linear(embed_dim, embed_dim))
```
- `attn_output_weights` from MHA container is output without averaging. Therefore, for the drop-in replacement above, users will need to average the attention output weights in order to have the same results as `nn.MultiheadAttention`.
- Compatible with torchscript. Ready to add quantization support.
- Incremental decoding - `bias_k` and `bias_v` will be attached to the sequence dim of `key`/`value`
```
seq_len, bsz = 100, 64
query = key = value = torch.rand((seq_len, bsz, embed_dim))
bias_k = bias_v = torch.rand((1, 1, embed_dim // nhead)).repeat(1, bsz * nhead, 1)
attn_output, attn_weight = MHA(query, key, value, bias_k=bias_k, bias_v=bias_v)
```
- Broadcast and support `query`/`key`/`value` with more than three dimensions. For example, for some CV applications, the input tensors have four dimensions (N, H, W, C) ([link](https://github.com/pytorch/pytorch/issues/34760#issuecomment-601414055))
```
query = torch.rand((seq_len, 1, embed_dim)) # query's batch dim is 1
key = value = torch.rand((3, 3, seq_len, bsz, embed_dim)) # key and value have five dims
attn_output, attn_weight = MHA(query, key, value)
```
- Resolutions for various research requests. For example, an efficient Transformer architecture is proposed in [ref.](https://arxiv.org/pdf/2001.04451.pdf) and https://github.com/pytorch/pytorch/issues/35458. The idea is to share the linear in-projection between query and key with a separate linear layer for value. With the `SharedQK_Proj` class below, we can drop the custom in-projection module in MHA container as
```
class SharedQK_Proj(torch.nn.Module):
	def __init__(self, qk_proj, v_proj):
		super(SharedQK_Proj, self).__init__()
		self.qk_proj = qk_proj
		self.v_proj = qk_proj
		
	def forward(self, q, k, v):
		return self.qk_proj(q), self.qk_proj(k), self.v_proj(v)
		
in_proj_container = SharedQK_Proj(Linear(embed_dim, embed_dim), Linear(embed_dim, embed_dim))
MHA = MultiheadAttentionContainer(nhead, in_proj_container,
                                  ScaledDotProduct(), Linear(embed_dim, embed_dim))
```
Another example is the relative attention implementation introduced in [ref](https://arxiv.org/pdf/1904.09925.pdf). The matrices for relative position distance are added to the  the attention layer (see Equation 4 in the reference). 
```
class RelativeAttention(torch.nn.Module):
	def __init__(self):
		super(Relative2DAttention, self).__init__()
		
	def forward(self, query, key, value, attn_mask=None, bias_k=None, bias_v=None):
		query, key, value = query.transpose(-2, -3), key.transpose(-2, -3), value.transpose(-2, -3)
		attn_output_weights = torch.matmul(query, key.transpose(-2, -1))
		
		# a custom func to calculate relative logits in the sequence dimension
		rel_logits = relative_logits(query)
		attn_output_weights += rel_logits_h
		
		attn_output_weights = torch.nn.functional.softmax(attn_output_weights, dim=-1)
		attn_output = torch.matmul(attn_output_weights, value)
		return attn_output.transpose(-2, -3), attn_output_weights
		
MHA = MultiheadAttentionContainer(nhead, in_proj_container,
                                  RelativeAttention(), Linear(embed_dim, embed_dim))
```

Here is another example to add normalization and dropout in out-projection layer:
```
class CustomOutProj(torch.nn.Module):
	def __init__(self, in_dim, out_dim, dropout=0.1):
		super(CustomOutProj, self).__init__()
		self.out_proj = torch.nn.Linear(in_dim, out_dim)
		self.norm = torch.nn.LayerNorm(out_dim)
		self.dropout = torch.nn.Dropout(dropout)
        
	def forward(self, seq):
		seq = self.out_proj(seq)
		return self.norm(self.dropout(seq))

in_proj_container = InProjContainer(Linear(embed_dim, embed_dim), Linear(embed_dim, embed_dim), Linear(embed_dim, embed_dim))
MHA = MultiheadAttentionContainer(nhead, in_proj_container,
                                  ScaledDotProduct(), CustomOutProj(embed_dim, embed_dim))
```


torchtext/models/multiheadattention.py
```diff
@@ -0,0 +1,236 @@
...
+               (attn_mask.size(-3) != 1 and attn_mask.size(-3) != batch_heads):
+                raise RuntimeError('The size of the attn_mask is not correct.')
+            if attn_mask.dtype != torch.bool:
+                raise RuntimeError('Only bool tensor is supported for attn_mask')
+
+        # Dot product of q, k
+        attn_output_weights = torch.matmul(query, key.transpose(-2, -1))
+        assert attn_output_weights.size(-3) == batch_heads
+        assert attn_output_weights.size(-2) == tgt_len
+        assert attn_output_weights.size(-1) == src_len
```
fmassa, line 228: aren't those asserts redundant, given that we have already asserted the sizes of the other tensors?

torchtext/models/multiheadattention.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        super(MultiheadAttentionContainer, self).__init__()
+        self.query_in_proj = in_proj_tuple[0]
+        self.key_in_proj = in_proj_tuple[1]
+        self.value_in_proj = in_proj_tuple[2]
+        self.attention_layer = attention_layer
+        self.out_proj = out_proj
+
+    def forward(self, query, key, value, attn_mask=None, bias_k=None, bias_v=None):
+        # type: (Tensor, Tensor, Tensor, Optional[Tensor], Optional[Tensor], Optional[Tensor]) -> Tuple[Tensor, Optional[Tensor]]
```
fmassa, line 38: maybe use python3 type annotations?
Like
```python
def forward(self, query: Tensor, key: Tensor, ...):
```

torchtext/models/multiheadattention.py
```diff
@@ -0,0 +1,236 @@
+import torch
+from torch._jit_internal import Tuple, Optional
```
fmassa, line 2: now that python2 is not supported anymore, we can use `from typing import Tuple, Optional` instead, as `typing ` is available in Python3 by default

torchtext/models/multiheadattention.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+Tensor = torch.Tensor
```
fmassa, line 5: nit: I would have done instead
```python
from torch import Tensor
```

but this doesn't make a difference, just stylistic

torchtext/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .multiheadattention import MultiheadInProject, MultiheadOutProject, \
```
fmassa, line 1: In general, I think those abstractions should live in a `ops` or `modules` folder, instead of `models`, as those are not really models but model building blocks

torchtext/models/multiheadattention.py
```diff
@@ -0,0 +1,236 @@
...
+        assert bsz_num_head % self.num_heads == 0, \
+            "Dimension -2 of MultiheadOutProject input must be divisible by num_heads"
+        bsz = bsz_num_head // self.num_heads
+        seq = seq.reshape(seq_len, bsz, self.num_heads * self.head_dim)
+        seq = self.proj_layer(seq)
+        return seq
+
+
+class ScaledDotProduct(torch.nn.Module):
+    __constants__ = ['dropout']
```
fmassa, line 149: I think this is not required by torchscript anymore.

torchtext/models/multiheadattention.py
```diff
@@ -0,0 +1,236 @@
...
+            >>> query = torch.rand((21, bsz, embed_dim))
+            >>> key = value = torch.rand((16, bsz, embed_dim))
+            >>> attn_output, attn_weights = MHA(query, key, value)
+            >>> print(attn_output.shape)
+            >>> torch.Size([21, 64, 10])
+        """
+        super(MultiheadAttentionContainer, self).__init__()
+        self.query_in_proj = in_proj_tuple[0]
+        self.key_in_proj = in_proj_tuple[1]
+        self.value_in_proj = in_proj_tuple[2]
```
fmassa, line 33: Can you explain what's the original idea on passing a tuple of modules here, instead of three modules or a single module?

test/data/test_modules.py
```diff
@@ -0,0 +1,103 @@
...
+from torch.testing import assert_allclose
+from ..common.torchtext_test_case import TorchtextTestCase
+
+
+class TestModels(TorchtextTestCase):
+
+    def test_multiheadattention(self):
+        embed_dim, nhead, tgt_len, src_len, bsz = 10, 5, 6, 10, 64
+        # Build torchtext MultiheadAttention module
+        MHA = MultiheadAttentionContainer((MultiheadInProject(embed_dim, nhead),
```
cpuhrsch, line 14: Is MultiheadInProject really necessary? Why not just torch.nn.functional.linear? or torch.nn.Linear?

test/data/test_modules.py
```diff
@@ -0,0 +1,103 @@
...
+                                                          MHA.out_proj.proj_layer.weight,
+                                                          MHA.out_proj.proj_layer.bias,
+                                                          attn_mask=torch_attn_mask)
+
+        assert_allclose(mha_output, torch_mha_output)
+        # With bias_k and bias_v, src_len needs to plus 1
+        attn_weights = attn_weights.view(bsz, nhead, tgt_len, src_len + 1).sum(dim=1) / nhead
+        assert_allclose(attn_weights, torch_mha_weights)
+
+    def test_broadcast_scaled_dot_product(self):
```
cpuhrsch, line 48: I'd add some "self.assertRaises" to exercises some explicit cases where we expect broadcasting to fail definitely.

torchtext/__init__.py
```diff
@@ -7,6 +8,7 @@
 __version__ = '0.6.0'
 
 __all__ = ['data',
+           'modules',
```
cpuhrsch, line 11: Why modules and not the torch.nn path convention?

torchtext/modules/multiheadattention.py
```diff
@@ -0,0 +1,229 @@
+import torch
+from typing import Tuple, Optional
+
+
+class MultiheadAttentionContainer(torch.nn.Module):
```
cpuhrsch, line 5: If initializes is forward setup such that an instance of this could be a drop-in replacement of an instance of torch.nn.MHA?

torchtext/modules/multiheadattention.py
```diff
@@ -0,0 +1,229 @@
...
+            - seq: :math:`(S, N, E)`
+
+            - Output: :math:`(S, N * H, E / H)`
+
+            where S is the sequence length, H is the number of attention heads, N is the
+            batch size, and E is the embedding dimension.
+        """
+        seq_len, bsz, proj_dim = seq.size()
+        seq = self.proj_layer(seq)
+        seq = seq.reshape(seq_len, bsz * self.num_heads, self.head_dim)
```
cpuhrsch, line 104: I think this is something the container should do

torchtext/modules/multiheadattention.py
```diff
@@ -0,0 +1,229 @@
...
+            - bias_k and bias_v:bias: :math:`(1, N * H, E / H)`
+
+            - Outputs:
+            - attn_output: :math:`(L, N, E)`
+            - attn_output_weights: :math:`(N*num_heads, L, S)`
+
+            where where L is the target length, S is the sequence length, H is the number of attention heads,
+                N is the batch size, and E is the embedding dimension.
+        """
+        q = self.query_in_proj(query)
```
cpuhrsch, line 63: This container doesn't do a lot of work. It requires the users to do all the reshaping and concatenation which was flagged as the hard part

zhangguanheng66:
Some performance results by running `benchmark/mha_block.py` file
```
********************************************************************************
test case GPU with embed_dim, nhead, tgt_len, src_len, bsz: 768 12 128 128 72
starting torchtext.modules.MultiheadAttentionContainer
0.6918396130204201
starting torch.nn.functional.multi_head_attention_forward
0.7520507331937551
********************************************************************************
test case GPU with embed_dim, nhead, tgt_len, src_len, bsz: 64 2 10 10 8
starting torchtext.modules.MultiheadAttentionContainer
0.09328490681946278
starting torch.nn.functional.multi_head_attention_forward
0.07301254756748676
********************************************************************************
test case CPU with embed_dim, nhead, tgt_len, src_len, bsz: 768 12 128 128 72
starting torchtext.modules.MultiheadAttentionContainer
4.999244207516313
starting torch.nn.functional.multi_head_attention_forward
6.2229302898049355
********************************************************************************
test case CPU with embed_dim, nhead, tgt_len, src_len, bsz: 64 2 10 10 8
starting torchtext.modules.MultiheadAttentionContainer
0.09795885905623436
starting torch.nn.functional.multi_head_attention_forward
0.07054279558360577
```

zhangguanheng66:
@Enealor just want to ping you and see if you have any feedback.

torchtext/modules/multiheadattention.py
```diff
@@ -0,0 +1,203 @@
...
+            - attn_mask: :math:`(N * H, L, S)`, positions with ``True`` are not allowed to attend
+                while ``False`` values will be unchanged.
+            - bias_k and bias_v:bias: :math:`(1, N * H, E / H)`
+
+            - Output: :math:`(L, N * H, E / H)`, :math:`(N * H, L, S)`
+
+            where L is the target length, S is the source length, H is the number
+            of attention heads, N is the batch size, and E is the embedding dimension.
+        """
+        if bias_k is not None and bias_v is not None:
```
Enealor, line 135: It looks like `bias_k` and `bias_v` are only used if both are defined. If the user provides only one, it is silently ignored. It would be helpful to explicitly document that either both arguments should be provided or neither argument should be provided. Alternatively, you could add an an exception if just one is provided.

torchtext/modules/multiheadattention.py
```diff
@@ -0,0 +1,203 @@
...
+                attn_mask: Optional[torch.Tensor] = None,
+                bias_k: Optional[torch.Tensor] = None,
+                bias_v: Optional[torch.Tensor] = None) -> Tuple[torch.Tensor, torch.Tensor]:
+        r"""
+
+        Args:
+            query, key, value (Tensor): map a query and a set of key-value pairs to an output.
+                See "Attention Is All You Need" for more details.
+            attn_mask (BoolTensor, optional): 3D mask that prevents attention to certain positions.
+            bias_k and bias_v:bias (Tensor, optional): one more key and value sequence to be added at
```
Enealor, line 47: Is this how you meant to have this documented? It looks more in-line with the others as "bias_k and bias_v (Tensor, optional):"

torchtext/modules/multiheadattention.py
```diff
@@ -0,0 +1,203 @@
...
+                sequence dim (dim=-3). Those are used for incremental decoding.
+
+        Shape:
+            - Inputs:
+            - query: :math:`(L, N, E)`
+            - key: :math:`(S, N, E)`
+            - value: :math:`(S, N, E)`
+            - attn_mask: :math:`(N * H, L, S)`, positions with ``True`` are not allowed to attend
+                while ``False`` values will be unchanged.
+            - bias_k and bias_v:bias: :math:`(1, N * H, E / H)`
```
Enealor, line 57: My comment here is similar to the above comment. Is this meant to be "bias_k and bias_v:"?

Enealor:
It looks good. I noticed a few things but those amount to documentation notes.

zhangguanheng66:
> It looks good. I noticed a few things but those amount to documentation notes.

Thanks for reviewing. I will accommodate those feedbacks and update the doc strings.

benchmark/mha_block.py
```diff
@@ -0,0 +1,75 @@
...
+        attn_mask_2D = torch.randint(0, 2, (tgt_len, src_len)).to(torch.bool).to(device)
+        bias_k = bias_v = torch.rand((1, 1, embed_dim)).to(device)
+        print("starting torchtext.modules.MultiheadAttentionContainer")
+        t0 = time.monotonic()
+        for _ in range(100):
+            mha_output, attn_weights = MHA(query, key, value,
+                                           attn_mask=torch.stack([attn_mask_2D] * (bsz * nhead)),
+                                           bias_k=bias_k.repeat(1, bsz, 1).reshape(1, bsz * nhead, -1),
+                                           bias_v=bias_v.repeat(1, bsz, 1).reshape(1, bsz * nhead, -1))
+        print(time.monotonic() - t0)
```
fmassa, line 33: If you are benchmarking with CUDA, you need to add a `torch.cuda.synchronize()` before and after measuring the time, otherwise the timings won't be correct

benchmark/mha_block.py
```diff
@@ -0,0 +1,75 @@
...
+        for _ in range(100):
+            torch_mha_output, torch_mha_weights = mha_forward(query, key, value,
+                                                              embed_dim, nhead,
+                                                              in_proj_weight, None,
+                                                              bias_k, bias_v,
+                                                              False, 0.0,
+                                                              MHA.out_proj.weight,
+                                                              MHA.out_proj.bias,
+                                                              attn_mask=torch_attn_mask)
+        print(time.monotonic() - t0)
```
fmassa, line 51: Same comment here.

benchmark/mha_block.py
```diff
@@ -0,0 +1,75 @@
...
+
+    print("*" * 80)
+    print("test case GPU with embed_dim, nhead, tgt_len, src_len, bsz:", 768, 12, 128, 128, 72)
```
fmassa, line 54: I believe most of the potential speed benefits from the MHA implemented in PyTorch are only valid when `query = key = value` (because it computes the projections in a single kernel launch for the 3).
Can you add more benchmarks for different sizes in the `query = key = value` case? A for loop would be helpful there, something like
```python
for embed_dim in [256, 768]:
    for ...
        for ...
            print(...)
             _run_benchmark(...)
```

benchmark/mha_block.py
zhangguanheng66: Thanks. Will add them there.

cpuhrsch: The reason for this is that calls into cuda verions of operations are launched asynchronously. Only when you print a Tensor or convert it onto CPU can you be sure all operations have finished. Using synchronize here helps you make sure indeed all the work has finished and you're timing things correctly. Also see [torch.cuda](https://pytorch.org/docs/master/cuda.html).

benchmark/mha_block.py
cpuhrsch: We've run benchmarks on this and it depends on the size of the inputs as well. For large inputs, as you can probably imagine, it shouldn't make much of a difference since the overhead disappears.

torchtext/modules/multiheadattention.py
```diff
@@ -0,0 +1,205 @@
...
+
+        assert k.size(-1) % self.nhead == 0, "key's embed_dim must be divisible by the number of heads"
+        head_dim = k.size(-1) // self.nhead
+        k = k.reshape(src_len, bsz * self.nhead, head_dim)
+
+        assert v.size(-1) % self.nhead == 0, "value's embed_dim must be divisible by the number of heads"
+        head_dim = v.size(-1) // self.nhead
+        v = v.reshape(src_len, bsz * self.nhead, head_dim)
+
+        attn_output, attn_output_weights = self.attention_layer(q, k, v, attn_mask=attn_mask,
```
cpuhrsch, line 81: It seems that for this container in particular there are no assumptions made on the dtype of attn_mask. I think we can relax that constraint. It stems from the fact that ScaledDotProduct needs a BoolTensor as a mask, but not for the container.

torchtext/modules/multiheadattention.py
```diff
@@ -0,0 +1,205 @@
...
+            if (attn_mask.size(-1) != src_len) or (attn_mask.size(-2) != tgt_len) or \
+               (attn_mask.size(-3) != 1 and attn_mask.size(-3) != batch_heads):
+                raise RuntimeError('The size of the attn_mask is not correct.')
+            if attn_mask.dtype != torch.bool:
+                raise RuntimeError('Only bool tensor is supported for attn_mask')
+
+        # Dot product of q, k
+        attn_output_weights = torch.matmul(query, key.transpose(-2, -1))
+        if attn_mask is not None:
+            attn_output_weights.masked_fill_(attn_mask, float('-inf'),)
```
myleott, line 169: I believe for some speech use case they needed to use `-1e8` instead of `-inf` to avoid NaN: https://github.com/pytorch/fairseq/blob/928dc47e7e72f3e6ed96e50942e7fb8892cdcf32/fairseq/modules/transformer_layer.py#L108-L112

Does it make sense to have this be user configurable?

zhangguanheng66: I think I will follow the convention in fairseq. We could add this user configurable later.

cpuhrsch: Also, since this is part of ScaledDotProduct we can create variants of ScaledDotProduct that are more flexible for this kind of stuff. I think we'll end up with a small collection of attention functions and maybe we'll come up with some common building blocks there as well.

netw0rkf10w:
@zhangguanheng66 In the docstrings, it seems `torchtext.models` should be replaced by `torchtext.modules`.

zhangguanheng66:
> @zhangguanheng66 In the docstrings, it seems `torchtext.models` should be replaced by `torchtext.modules`.

I have a PR to update the doc.

benchmark/mha_block.py
netw0rkf10w: @zhangguanheng66 Could you share with us how your implementation performs compared to the PyTorch one after you have fixed the timing? Thanks.

netw0rkf10w:
@zhangguanheng66 There seems to be some discrepancy compared to the PyTorch implementation, which has `bias_k` and `bias_v` as learnable parameters of the MHA (see [here](https://github.com/pytorch/pytorch/blob/75a4862f639de666c66a0db240c993918b80707f/torch/nn/modules/activation.py#L872)). In yours, they are tensors passed as inputs to the MHA's forward function. Is there a good reason for this? Thanks.

zhangguanheng66:
> @zhangguanheng66 There seems to be some discrepancy compared to the PyTorch implementation, which has `bias_k` and `bias_v` as learnable parameters of the MHA (see [here](https://github.com/pytorch/pytorch/blob/75a4862f639de666c66a0db240c993918b80707f/torch/nn/modules/activation.py#L872)). In yours, they are tensors passed as inputs to the MHA's forward function. Is there a good reason for this? Thanks.

In  pytorch MHA, `bias_k` and `bias_v` are learnable variables. For MHA container, those are two tensors attached to key/value for incremental decoding. These are different things.

netw0rkf10w:
@zhangguanheng66 Thanks. I've successfully built a MHA layer using your implementation. Its outputs numerically match the ones of PyTorch implementation (I had to write an auxiliary function to convert the state dict of the latter to the former).

torchtext/modules/multiheadattention.py
```diff
@@ -0,0 +1,213 @@
...
+        if bias_k is not None and bias_v is not None:
+            assert key.size(-1) == bias_k.size(-1) and key.size(-2) == bias_k.size(-2) and bias_k.size(-3) == 1, \
+                "Shape of bias_k is not supported"
+            assert value.size(-1) == bias_v.size(-1) and value.size(-2) == bias_v.size(-2) and bias_v.size(-3) == 1, \
+                "Shape of bias_v is not supported"
+            key = torch.cat([key, bias_k])
+            value = torch.cat([value, bias_v])
+            if attn_mask is not None:
+                _attn_mask = attn_mask
+                attn_mask = torch.nn.functional.pad(_attn_mask, (0, 1))
```
netw0rkf10w, line 142: @zhangguanheng66 Why not simply `attn_mask = torch.nn.functional.pad(attn_mask, (0, 1))`?

zhangguanheng66: Update in the revised_mha PR (link])(https://github.com/pytorch/text/pull/839/commits/10fb9583c0bc46b229323766555f9b089e51441d)

netw0rkf10w:
> > @zhangguanheng66 There seems to be some discrepancy compared to the PyTorch implementation, which has `bias_k` and `bias_v` as learnable parameters of the MHA (see [here](https://github.com/pytorch/pytorch/blob/75a4862f639de666c66a0db240c993918b80707f/torch/nn/modules/activation.py#L872)). In yours, they are tensors passed as inputs to the MHA's forward function. Is there a good reason for this? Thanks.
> 
> In pytorch MHA, `bias_k` and `bias_v` are learnable variables. For MHA container, those are two tensors attached to key/value for incremental decoding. These are different things.

@zhangguanheng66 I am trying to understand `bias_k` and `bias_v`. Could you please point me to some references? Is it the same as [fairseq's incremental decoding](https://fairseq.readthedocs.io/en/latest/_modules/fairseq/models/fairseq_incremental_decoder.html) described below? Thank you very much!

```
Incremental decoding is a special mode at inference time where the Model
    only receives a single timestep of input corresponding to the previous
    output token (for teacher forcing) and must produce the next output
    *incrementally*. Thus the model must cache any long-term state that is
    needed about the sequence, e.g., hidden states, convolutional states, etc.
```

zhangguanheng66:
> > > @zhangguanheng66 There seems to be some discrepancy compared to the PyTorch implementation, which has `bias_k` and `bias_v` as learnable parameters of the MHA (see [here](https://github.com/pytorch/pytorch/blob/75a4862f639de666c66a0db240c993918b80707f/torch/nn/modules/activation.py#L872)). In yours, they are tensors passed as inputs to the MHA's forward function. Is there a good reason for this? Thanks.
> > 
> > 
> > In pytorch MHA, `bias_k` and `bias_v` are learnable variables. For MHA container, those are two tensors attached to key/value for incremental decoding. These are different things.
> 
> @zhangguanheng66 I am trying to understand `bias_k` and `bias_v`. Could you please point me to some references? Is it the same as [fairseq's incremental decoding](https://fairseq.readthedocs.io/en/latest/_modules/fairseq/models/fairseq_incremental_decoder.html) described below? Thank you very much!
> 
> ```
> Incremental decoding is a special mode at inference time where the Model
>     only receives a single timestep of input corresponding to the previous
>     output token (for teacher forcing) and must produce the next output
>     *incrementally*. Thus the model must cache any long-term state that is
>     needed about the sequence, e.g., hidden states, convolutional states, etc.
> ```

Kind of. From from code view, it pads an extra token in the sequence dimension of key/value.

torchtext/modules/multiheadattention.py
```diff
@@ -0,0 +1,213 @@
...
+            if (attn_mask.size(-1) != src_len) or (attn_mask.size(-2) != tgt_len) or \
+               (attn_mask.size(-3) != 1 and attn_mask.size(-3) != batch_heads):
+                raise RuntimeError('The size of the attn_mask is not correct.')
+            if attn_mask.dtype != torch.bool:
+                raise RuntimeError('Only bool tensor is supported for attn_mask')
+
+        # Dot product of q, k
+        attn_output_weights = torch.matmul(query, key.transpose(-2, -1))
+        if attn_mask is not None:
+            attn_output_weights.masked_fill_(attn_mask, -1e8,)
```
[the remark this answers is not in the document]
zhangguanheng66: @netw0rkf10w There are some ongoing discussions about `NaN` output for some special cases. We tried to avoid this when implementing MHA container in torchtext. I believe we will modify this accordingly as https://github.com/pytorch/pytorch/pull/42323 concludes.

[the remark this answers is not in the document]
netw0rkf10w: @zhangguanheng66 Great. Thanks for the information! I'll join that discussion later.

PR #728, opened 2020-04-07
update README file for conda installation

zhangguanheng66:
Since one of the dependency libraries `sentencepiece` is not distributed via conda, we advice users to install `sentencepiece` from `powerai` channel. Fix https://github.com/pytorch/text/pull/726

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -28,6 +28,10 @@ Make sure you have Python 2.7 or 3.5+ and PyTorch 0.4.0 or newer. You can then i
     
 For PyTorch versions before 0.4.0, please use `pip install torchtext==0.2.3`.
 
+Or you can install torchtext using conda::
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -28,6 +28,10 @@ Make sure you have Python 2.7 or 3.5+ and PyTorch 0.4.0 or newer. You can then i
     
 For PyTorch versions before 0.4.0, please use `pip install torchtext==0.2.3`.
 
+Or you can install torchtext using conda::
+    conda install -c pytorch torchtext
+    conda install -c powerai sentencepiece
```
seemethere, line 33: This could also be simplified into one line:

```suggestion
    conda install -c pytorch -c powerai torchtext sentencepiece
```

Issue #717, opened 2020-04-01
[Dependency Bug]`sentencepiece` is not installed as a dependency in conda


z-a-f:
## 🐛 Bug
**Describe the bug**
After installing the `torchtext`, and trying the code below (in the repro section) an error pops up: `ModuleNotFoundError: No module named 'sentencepiece'`. I think the package should be installed automatically as a dependency.

The problem is that the depndency on `sentencepiece` is specified for pip, but not for conda (see attachments)

**To Reproduce**

After installing using

```console
conda install -c pytorch torchtext
```

run the following:

```python
from torchtext import data
from torchtext import datasets

imdb_set = datasets.IMDB(DATA_PATH,
                         text_field=data.Field(),
                         label_field=data.LabelField())
```

```console
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-15-019e470b76d6> in <module>
----> 1 from torchtext import data
      2 from torchtext import datasets
      3 
      4 imdb_set = datasets.IMDB(DATA_PATH,
      5                          text_field=data.Field(),

~/miniconda3/envs/py37-ml/lib/python3.7/site-packages/torchtext/__init__.py in <module>
----> 1 from . import data
      2 from . import datasets
      3 from . import utils
      4 from . import vocab
      5 from . import experimental

~/miniconda3/envs/py37-ml/lib/python3.7/site-packages/torchtext/data/__init__.py in <module>
      8 from .pipeline import Pipeline
      9 from .utils import get_tokenizer, interleave_keys
---> 10 from .functional import generate_sp_model, \
     11     load_sp_model, \
     12     sentencepiece_numericalizer, \

~/miniconda3/envs/py37-ml/lib/python3.7/site-packages/torchtext/data/functional.py in <module>
----> 1 import sentencepiece as spm
      2 import re
      3 
      4 __all__ = [
      5     "generate_sp_model", "load_sp_model",

ModuleNotFoundError: No module named 'sentencepiece'

```

**Expected behavior**

If some core part of the `torchtext` package is using a third-party package, it should be added as a dependency.

**Screenshots**

`conda search --info torchtext` doesn't list `sentencepiece` as a dependency.

![image](https://user-images.githubusercontent.com/4216323/78107971-4c9a3d80-73ab-11ea-9c87-cff345ae03ad.png)

However, `pip show torchtext` does:

[...]

Issue #703, opened 2020-03-02
Windows and Sentence Piece


achapkowski:
## 🐛 Bug
**Describe the bug**

It appears one of the requirements sentencepiece is not built for windows machines.

This causes an import error:

```python
     1 import torch
----> 2 import torchtext
      3 import numpy as np
      4 import pandas as pd

C:\Program Files\ArcGIS\Pro\bin\Python\envs\arcgispro-dev\lib\site-packages\torchtext\__init__.py in <module>
----> 1 from . import data
      2 from . import datasets
      3 from . import utils
      4 from . import vocab
      5 from . import experimental

C:\Program Files\ArcGIS\Pro\bin\Python\envs\arcgispro-dev\lib\site-packages\torchtext\data\__init__.py in <module>
      8 from .pipeline import Pipeline
      9 from .utils import get_tokenizer, interleave_keys
---> 10 from .functional import generate_sp_model, \
     11     load_sp_model, \
     12     sentencepiece_numericalizer, \

C:\Program Files\ArcGIS\Pro\bin\Python\envs\arcgispro-dev\lib\site-packages\torchtext\data\functional.py in <module>
----> 1 import sentencepiece as spm
      2 import re
      3 
      4 __all__ = [
      5     "generate_sp_model", "load_sp_model",

ModuleNotFoundError: No module named 'sentencepiece'
```

**Expected behavior**
module imports without issue.

**Environment**

Please copy and paste the output from our
[environment collection script](https://raw.githubusercontent.com/pytorch/text/master/torchtext/utils/collect_env.py)
(or fill out the checklist below manually).

You can get the script and run it with:
```
wget https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py
# For security purposes, please check the contents of collect_env.py before running it.
python collect_env.py
python -c "import torchtext; print(\"torchtext version is \", torchtext.__version__)"
```

 - PyTorch Version (e.g., 1.0): latest
 - OS (e.g., Linux): Windows
 - How you installed PyTorch (`conda`, `pip`, source): conda
 - Build command you used (if compiling from source):N/A
 - Python version:3.6
 - CUDA/cuDNN version:
 - GPU models and configuration:
 - Any other relevant information: Install torchtext from conda.

zhangguanheng66:
@achapkowski AFAIK, Windows is not officially supported for torchtext. I will reach out to our Windows partner and see if they would like to add it sooner or later.

peterjc123:
[...]

PR #733, opened 2020-04-10
error out if sentencepiece is not installed

zhangguanheng66:
Since sp is not released via conda, if users install torchtext's conda wheels, we don't force them to install sp until they use relevant API.

### DIFF
--- a/torchtext/data/functional.py
+++ b/torchtext/data/functional.py
@@ -35,7 +35,7 @@ def generate_sp_model(filename, vocab_size=20000,
     """
     try:
         import sentencepiece as spm
-    except ImportError:
+    except ModuleNotFoundError:
         print("Please install sentencepiece")
         raise
     spm_training_string = "--input={} \
@@ -64,7 +64,7 @@ def load_sp_model(spm_path):
     """
     try:
         import sentencepiece as spm
-    except ImportError:
+    except ModuleNotFoundError:
         print("Please install sentencepiece")
         raise
     sp_model = spm.SentencePieceProcessor()
--- a/torchtext/data/functional.py
+++ b/torchtext/data/functional.py
@@ -36,8 +36,7 @@ def generate_sp_model(filename, vocab_size=20000,
     try:
         import sentencepiece as spm
     except ModuleNotFoundError:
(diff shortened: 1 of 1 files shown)


torchtext/data/functional.py
```diff
@@ -34,7 +33,11 @@ def generate_sp_model(filename, vocab_size=20000,
         >>> from torchtext.data.functional import generate_sp_model
         >>> generate_sp_model('test.csv', vocab_size=23456, model_prefix='spm_user')
     """
-
+    try:
+        import sentencepiece as spm
+    except ImportError:
+        print("Please install sentencepiece")
```
cpuhrsch, line 39: It's better to use ["ImportWarning"](https://docs.python.org/3.8/library/exceptions.html#ImportWarning) here

torchtext/data/functional.py
```diff
@@ -59,7 +62,11 @@ def load_sp_model(spm_path):
         >>> from torchtext.data.functional import load_sp_model
         >>> sp_model = load_sp_model("m_user.model")
     """
-
+    try:
+        import sentencepiece as spm
+    except ImportError:
+        print("Please install sentencepiece")
```
cpuhrsch, line 68: Same here.

torchtext/data/functional.py
```diff
@@ -34,7 +33,11 @@ def generate_sp_model(filename, vocab_size=20000,
         >>> from torchtext.data.functional import generate_sp_model
         >>> generate_sp_model('test.csv', vocab_size=23456, model_prefix='spm_user')
     """
-
+    try:
+        import sentencepiece as spm
+    except ImportError:
+        print("Please install sentencepiece")
+        raise
```
cpuhrsch, line 40: If you're raising the exception it'll fail after all, right? Did you try this with and without sentencepiece available?

cpuhrsch:
I'd move this logic into the top-level "__init__" so you can be sure it's run immediately into import.

zhangguanheng66:
Fix https://github.com/pytorch/text/issues/717 and partially https://github.com/pytorch/text/issues/703

PR #736, opened 2020-04-22
Simplifications and code formatting for experimental text classification datasets

cpuhrsch:
<work in progress>

### DIFF
--- a/torchtext/experimental/datasets/raw/text_classification.py
+++ b/torchtext/experimental/datasets/raw/text_classification.py
@@ -46,7 +46,7 @@ class RawTextIterableDataset(torch.utils.data.IterableDataset):
 
     def __iter__(self):
         for i, item in enumerate(self._iterator):
-            if i > self.start:
+            if i >= self.start:
                 yield item
             if self.num_lines is not None and i == (self.start + num_lines):
                 break
--- a/torchtext/experimental/datasets/raw/text_classification.py
+++ b/torchtext/experimental/datasets/raw/text_classification.py
@@ -1,7 +1,6 @@
 import torch
 import io
 from torchtext.utils import download_from_url, extract_archive, unicode_csv_reader
-import sys
 
 URLS = {
     'AG_NEWS':
--- a/torchtext/experimental/datasets/raw/text_classification.py
+++ b/torchtext/experimental/datasets/raw/text_classification.py
@@ -36,34 +36,19 @@ class RawTextIterableDataset(torch.utils.data.IterableDataset):
     """Defines an abstraction for raw text iterable datasets.
     """
 
-    def __init__(self, iterator):
+    def __init__(self, iterator, start=0, num_lines=None):
         """Initiate text-classification dataset.
         """
         super(RawTextIterableDataset, self).__init__()
         self._iterator = iterator
-        self.has_setup = False
-        self.start = 0
-        self.num_lines = None
-
-    def setup_iter(self, start=0, num_lines=None):
         self.start = start
         self.num_lines = num_lines
-        self.has_setup = True
 
     def __iter__(self):
-        if not self.has_setup:
-            self.setup_iter()
-
         for i, item in enumerate(self._iterator):
-            if i == self.start:
-                break
-
-        num_lines = self.num_lines if self.num_lines is not None else sys.maxsize
-        for _ in range(num_lines):
-            yield item
-            try:
-                item = next(self._iterator)
-            except StopIteration:
+            if i > self.start:
+                yield item
+            if self.num_lines is not None and i == (self.start + num_lines):
                 break
 
     def get_iterator(self):
--- a/torchtext/experimental/datasets/text_classification.py
+++ b/torchtext/experimental/datasets/text_classification.py
@@ -63,7 +63,9 @@ def _setup_datasets(
     text_transform = []
     if tokenizer is None:
         tokenizer = get_tokenizer("basic_english")
-    text_transform = lambda x: ngrams_iterator(tokenizer(x), ngrams)
+
+    def ngram_tokenizer(x):
+        return ngrams_iterator(tokenizer(x), ngrams)
 
     if isinstance(data_select, str):
         data_select = [data_select]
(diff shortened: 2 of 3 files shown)


torchtext/experimental/datasets/text_classification.py
```diff
@@ -87,32 +52,47 @@ def get_vocab(self):
         return self.vocab
 
 
-def _setup_datasets(dataset_name, root='.data', ngrams=1, vocab=None,
-                    tokenizer=None, data_select=('train', 'test')):
-    text_transform = []
+def _setup_datasets(
```
zhangguanheng66, line 55: Let's consider if we want to keep some functions (or transforms) separated from `_setup_datasets` func. For example, `ngrams_func` defines a "contract" to convert a list of tokens to ngrams. `vocab_func` defines a "contract" from tokens to ids. Although those functions are still quite preliminary, those "contracts" are reusable in the future. When we re-write the datasets (for example, language modeling datasets), we need some similar "contracts" (like tokenizer, vocab). Then, we can unify those "contracts" and add them as `data.transforms`. This is one of the reasons that I want to keep them away from `_setup_datasets`.
Another thing is the layout of `_setup_datasets`. Within the original `_setup_datasets` func, it just combines the "transforms" together as a pipeline. IMO, it's easier for users to follow if they want to create their own custom `_setup_datasets`.
Overall, I think those building blocks separated from `_setup_datasets` func may add some flexibility.

cpuhrsch: Sure, if we want to create a collection of functions that behave according to some contract and have those as user-facing, I definitely think there's value to it. This is similar to get_tokenizer or ngrams_iterator within utils. We just might want to pick a better path. Did you mean that you'd rather have these functions next to the datasets?

zhangguanheng66: For now, since we only experiment those utils here, we could keep those utils next to the datasets. Once we have more utils for language modeling and question-answer datasets, we could start unifying them. Or we could create some experimental utils in the experimental folder, which will evolve over time. Either way works for me, IMO.

cpuhrsch: I'd say we build out the folder structure we want to see at the module level within experimental so it'll be an easier switch / merge later on.

cpuhrsch:
@zhangguanheng66 - let me know what you think about this and if you want to merge it

torchtext/experimental/datasets/text_classification.py
cpuhrsch: There are five generic functions introduced in the current code

vocab_func - returns a function that calls ```__getitem__``` on each entry of a given list using a particular vocab object
totensor - effectively returns torch.tensor. I think this is unnecessary.
ngrams_func - we have ```ngrams_iterator``` already
build_vocab - implements ```build_vocab_from_iterator(transforms(txt) for (_, txt) in data)```
squential_transforms - effectively [functools.reduce](https://docs.python.org/3/library/functools.html#functools.reduce) and similar to nn.Sequential for nn.Modules and torchvision also has a similar combination function.

torchtext/experimental/datasets/raw/text_classification.py
```diff
@@ -35,24 +35,15 @@ class RawTextIterableDataset(torch.utils.data.IterableDataset):
     """Defines an abstraction for raw text iterable datasets.
     """
 
-    def __init__(self, iterator):
+    def __init__(self, iterator, start=0, num_lines=None):
```
cpuhrsch, line 38: I understand the allure of "setup_iter" to be in the multiprocessing context. When passed into DataLoader we will want to use the worker info to forward the underlying iterator and prevent batches with repeated datapoints. However, we don't make use of workerinfo in ```__iter__``` right now. Further, specifying an offset and length could be useful for the constructor in general. This could arguably be skipped and a generic abstraction that returns an iterator at a given offset and length could be introduced (albeit it seems unnecessary given how little code it is).

zhangguanheng66: I think that's fine to remove those argus since most datasets in torchtext fit the memory.

torchtext/experimental/datasets/text_classification.py
zhangguanheng66: I agree that those five functions don't cover a lot of codes and can be replaced. However, I feel like we are telling a consistent story here with some standard components. And we keep this consistent with the other two domain libraries. And we are using the same `language` as pytext since they also use transforms. Plus, I re-use those functionals in SQuAD datasets.
https://github.com/pytorch/text/blob/edb143af512e890f3a00d3e872658f0986f3b23a/torchtext/experimental/datasets/question_answer.py#L5

cpuhrsch: Outside of ```squential_transforms``` the other two are minimal enough that I think that it's ok to repeat them. totensor effectively returns a torch.tensor with a different default dtype. I think we should be careful here, for all these transforms really, until we have a standard and wider plans around transforms and functionals.

cpuhrsch:
Using the small benchmark script I found that this PR gives us a significant speedup. I didn't exactly characterize this (it's running on my laptop), but the magnitude of I think we should not need higher resolution instrumentation.

master
```
AG_NEWS - 6.124168223000001
SogouNews - 199.483479047
DBpedia - 38.38662358899998
YelpReviewPolarity - 70.71134889599998
YelpReviewFull - 86.49273644800002
YahooAnswers - 146.78631064299992
AmazonReviewPolarity - 319.45054842900004
AmazonReviewFull - 293.204584815
IMDB - 24.152421628999946
```

up1
```
AG_NEWS - 6.304364777
SogouNews - 158.734018144
DBpedia - 29.756586211000013
YelpReviewPolarity - 59.01243050400001
YelpReviewFull - 69.66327271199998
YahooAnswers - 114.90884431500001
AmazonReviewPolarity - 237.30875152599998
AmazonReviewFull - 198.39240907499993
IMDB - 19.537461151000002
```

zhangguanheng66:
> Using the small benchmark script I found that this PR gives us a significant speedup. I didn't exactly characterize this (it's running on my laptop), but given the order I think we should not need higher resolution instrumentation. This is for dataset creation only.
> 
> master
> 
> ```
> AG_NEWS - 6.124168223000001
> SogouNews - 199.483479047
> DBpedia - 38.38662358899998
> YelpReviewPolarity - 70.71134889599998
> YelpReviewFull - 86.49273644800002
> YahooAnswers - 146.78631064299992
> AmazonReviewPolarity - 319.45054842900004
> AmazonReviewFull - 293.204584815
> IMDB - 24.152421628999946
> ```
> 
> up1
> 
> ```
> AG_NEWS - 6.304364777
> SogouNews - 158.734018144
> DBpedia - 29.756586211000013
> YelpReviewPolarity - 59.01243050400001
> YelpReviewFull - 69.66327271199998
> YahooAnswers - 114.90884431500001
> AmazonReviewPolarity - 237.30875152599998
> AmazonReviewFull - 198.39240907499993
> IMDB - 19.537461151000002
> ```
> 
> I suspect (but didn't verify) that this is because the transforms during creation are setup to avoid some materializations (e.g. ngrams_iterator).

I assume in both cases you didn't count the time to download/unzip files.

cpuhrsch:
@zhangguanheng66 - no, the underlying archives were already downloaded and also extracted. Creating the raw datasets in this circumstance is very fast. So this extra time is mostly about creating the vocab etc.

PR #737, opened 2020-04-22
[DO NOT LAND] Exercising test infrastructure

cpuhrsch:
I'm trying a few things to see if tests are running as expected.

### DIFF
--- a/torchtext/experimental/datasets/text_classification.py
+++ b/torchtext/experimental/datasets/text_classification.py
@@ -7,12 +7,14 @@ from torchtext.experimental.datasets import raw
 def vocab_func(vocab):
     def _forward(tok_iter):
         return [vocab[tok] for tok in tok_iter]
+
     return _forward
 
 
 def totensor(dtype):
     def _forward(ids_list):
         return torch.tensor(ids_list).to(dtype)
+
     return _forward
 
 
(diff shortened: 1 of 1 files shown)


torchtext/experimental/datasets/text_classification.py
```diff
@@ -87,32 +91,44 @@ def get_vocab(self):
...
+    text_transform = squential_transforms(
+        text_transform, vocab_func(vocab), totensor(dtype=torch.long)
+    )
     label_transform = squential_transforms(totensor(dtype=torch.long))
-    return tuple(TextClassificationDataset(raw_data[item], vocab,
-                                           (label_transform, text_transform))
-                 for item in data_select)
+    return tuple(
+        TextClassificationDataset(
+            raw_data[item], vocab, (label_transform, lambda x: None)
```
cpuhrsch, line 128: 99% of all changes in this file are due to the linter I ran so it's guaranteed to pass the lint tests. But this change here is manual. I'm setting the text_transforms to a function that returns None. The expectation is that this change will cause various tests to fail.

cpuhrsch:
This is still passing even though the tests were updated via https://github.com/pytorch/text/commit/0818a554427b49f7f603e1968093f39afdb8864e by @zhangguanheng66 

I ran the py.test command executed by travis locally and it seems like test_imdb is skipped when decorated by slow.

zhangguanheng66:
> This is still passing even though the tests were updated via [0818a55](https://github.com/pytorch/text/commit/0818a554427b49f7f603e1968093f39afdb8864e) by @zhangguanheng66
> 
> I ran the py.test command executed by travis locally and it seems like test_imdb is skipped when decorated by slow.

With the slow decorator, it will run only under the slow test set and it fails as expected. https://travis-ci.org/github/pytorch/text/jobs/678838711

PR #741, opened 2020-04-24
setup.py CppExtension and sentencepiece bindings

cpuhrsch:


cpuhrsch:
I'll change this PR to include the sentencepiece work and will split out the Vectors optimization into another PR. The Vectors class will need a more detailed rewrite to make sure we don't break someone's code instead of a small extensions that essentially provides a C++ map<std::string, at::Tensor> + lookup

PR #744, opened 2020-04-28
Run Unittest on CircleCI

mthrok:
This PR adds unittest on CircleCI, by mostly copying configuration file from `torachaudio`.

The test configuration is not 100% parity with the one in Travis. `--runslow` flag is not added yet.


mthrok:
@cpuhrsch  (CC @zhangguanheng66 
I added unit test based on your CCI setting, and test is passing. (Yay
It's not completely same as the one running on Travis as `--runslow` flag is not added.
If we would like only one test to run slow tests, (like in Travis) we can do so.

cc @seemethere 
BTW the support script required for binary build is not present in `torchtext`, so `binary_XX` jobs are failing.
What should we do?

mthrok:
Commented out binary build workflow.

cpuhrsch:
We can also turn off travis for tests covered by circleci. Looks like circle is also *a lot* faster. I know that some tests download vector files and such. Does circleci setup a cache for that?

mthrok:
@cpuhrsch 

> We can also turn off travis for tests covered by circleci. Looks like circle is also _a lot_ faster.

Sounds good, but I prefer doing it in a separate PR, unless you have a strong opinion in it.

>  I know that some tests download vector files and such. Does circleci setup a cache for that?

In this configuration CCI caches the following things
 - Miniconda (for `conda` command)
 - Conda environment (of different Python version), which contains dependencies listed in [`environment.yml`](https://github.com/mthrok/text/blob/moto-circleci-unittest/.circleci/unittest/scripts/environment.yml).
    - These dependencies will be updated each time job is ran if a newer version is found.
 -  Spacy English Dictionary.

Things happen after this, that includes PyTorch nightly installation, torch text installation, actual test run (and things downloaded while in test) are **not** cached.

mthrok:
@zhangguanheng66 

> thanks @mthrok. A quick question. Do you add nightly release here?

No, because I do not see existing scripts to be used for nightly release in `torchtext`. If you would like, I can copy and migrate the one from `torchaudio`.

cpuhrsch:
@mthrok - Alright, sounds like we can merge this in its current stage, enable the C++ extension setup to uncomment the binary builds and then follow up with 

a) Disabling travis for tests covered by circleci
b) Caching vectors and other heavy objects specific to torchtext used during testing in circle
c) Setting up nightly builds

What do you think?

mthrok:
> @mthrok - Alright, sounds like we can merge this in its current stage, enable the C++ extension setup to uncomment the binary builds and then follow up with
> 
> a) Disabling travis for tests covered by circleci
> b) Caching vectors and other heavy objects specific to torchtext used during testing in circle
> c) Setting up nightly builds
> 
> What do you think?

Sure, all of them sound reasonable and doable. Let's proceed with it.

PR #745, opened 2020-04-29
use 'cpu' instead of None for Iterator

### LINKED ISSUE #376: Use "cpu" instead of None for representing device
I think using device value as `cpu` instead of `None` will make `Iterator` class compatible with PyTorch device types [1] i.e., "cpu" and "gpu". 

[1] https://pytorch.org/docs/stable/tensor_attributes.html

avinashsai:
Fixes #376 
Points default device to torch cpu instead of None if no device is specified to be compatible with torch device types

### DIFF
--- a/torchtext/data/iterator.py
+++ b/torchtext/data/iterator.py
@@ -3,6 +3,7 @@ import random
 
 import logging
 
+import torch
 from .utils import RandomShuffler
 from .batch import Batch
 from .dataset import Dataset
--- a/torchtext/data/iterator.py
+++ b/torchtext/data/iterator.py
@@ -69,6 +69,8 @@ class Iterator(object):
 
         if device is None:
             device = torch.device('cpu')
+        elif isinstance(device, str):
+            device = torch.device(device)
 
         self.device = device
         self.random_shuffler = RandomShuffler()
--- a/torchtext/data/iterator.py
+++ b/torchtext/data/iterator.py
@@ -3,6 +3,7 @@ import random
 
 import logging
 
+import torch
 from .utils import RandomShuffler
 from .batch import Batch
 from .dataset import Dataset
@@ -64,7 +65,7 @@ class Iterator(object):
             logger.warning("The `device` argument should be set by using `torch.device`"
                            + " or passing a string as an argument. This behavior will be"
                            + " deprecated soon and currently defaults to cpu.")
-            device = None
+            device = torch.device('cpu')
         self.device = device
         self.random_shuffler = RandomShuffler()
 
--- a/torchtext/data/iterator.py
+++ b/torchtext/data/iterator.py
@@ -67,7 +67,7 @@ class Iterator(object):
                            + " deprecated soon and currently defaults to cpu.")
             device = None
 
-        if(device is None):
+        if device is None:
             device = torch.device('cpu')
 
         self.device = device
--- a/torchtext/data/iterator.py
+++ b/torchtext/data/iterator.py
@@ -3,7 +3,6 @@ import random
 
 import logging
 
-import torch
 from .utils import RandomShuffler
 from .batch import Batch
 from .dataset import Dataset
@@ -65,8 +64,11 @@ class Iterator(object):
             logger.warning("The `device` argument should be set by using `torch.device`"
                            + " or passing a string as an argument. This behavior will be"
                            + " deprecated soon and currently defaults to cpu.")
-            if device is None:
-            	device = torch.device('cpu')
+            device = None
+
+        if(device is None):
+        	device = torch.device('cpu')
+
         self.device = device
         self.random_shuffler = RandomShuffler()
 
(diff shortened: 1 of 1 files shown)


torchtext/data/iterator.py
```diff
@@ -64,7 +65,7 @@ def __init__(self, dataset, batch_size, sort_key=None, device=None,
             logger.warning("The `device` argument should be set by using `torch.device`"
                            + " or passing a string as an argument. This behavior will be"
                            + " deprecated soon and currently defaults to cpu.")
-            device = None
+            device = torch.device('cpu')
```
zhangguanheng66, line 68: It makes sense that the default device is cpu if device is None. But if `device` is provided in the init func, you still force to use cpu. Seem not correct. And double check the doc explains this clearly.

avinashsai:
Should I change accordingly or leave it as it is??

zhangguanheng66:
> Should I change accordingly or leave it as it is??

If device is provided in the `__init__` func, it should be assigned to `device`. Otherwise, the `device` is set to the default, which is `cpu`

avinashsai:
I will change accordingly

avinashsai:
Made changes as per your comments

torchtext/data/iterator.py
```diff
@@ -65,7 +65,8 @@ def __init__(self, dataset, batch_size, sort_key=None, device=None,
             logger.warning("The `device` argument should be set by using `torch.device`"
                            + " or passing a string as an argument. This behavior will be"
                            + " deprecated soon and currently defaults to cpu.")
-            device = torch.device('cpu')
```
zhangguanheng66: This is not correct. The `device` is under `if isinstance` statement. For `None` device, it's no way to enter it.

zhangguanheng66:
You have to rebase to the master branch

avinashsai:
changed as per your comments

zhangguanheng66:
The CI tests fail.

avinashsai:
i fixed indentation issue

torchtext/data/iterator.py
zhangguanheng66: nit: could you remove the parentheses in Python conditional?

zhangguanheng66:
The circleci tests need to pass. Please check the log since the errors are relevant.

avinashsai:
Fixed

zhangguanheng66:
nit: could you remove the parentheses in Python conditional?

avinashsai:
Done!!

torchtext/data/iterator.py
```diff
@@ -65,6 +66,10 @@ def __init__(self, dataset, batch_size, sort_key=None, device=None,
                            + " or passing a string as an argument. This behavior will be"
                            + " deprecated soon and currently defaults to cpu.")
             device = None
+
+        if device is None:
+            device = torch.device('cpu')
```
zhangguanheng66, line 71: In the doc, device could also be string. Should we also cover that as:
```
if device is None:
    device = torch.device('cpu')
elif isinstance(device, str):
    device = torch.device(device)
```

avinashsai: I would also like to add one more thing. How about adding a warning if cuda is available and user still preferred 'cpu'
```
if torch.cuda.is_available():
    if device is None or device=='cpu':
        logger.warning("you have gpu available but preferred device as cpu")
if device is None:
    device = torch.device('cpu')
elif isinstance(device, str):
    device = torch.device(device)
```

cpuhrsch: I wouldn't assume that GPU is preferred if it is available. This might be used in some unexpected context. 

Alongside updating the default behavior we should also update the doc string to document this.

avinashsai: Sometimes users forget to pass 'gpu' parameter even if gpu is available. So, if we send a warning, users can rectify it.

cpuhrsch: Yes, but it adds verbosity we commonly don't have. I suggest we do this in a separate PR where we address logging issues in general. If we were to do it for this operator we should also do it for the other ones and that'll required some strategy.

avinashsai: ok then, for now i will ignore and add only required ones.

avinashsai:
@zhangguanheng66 please merge this

cpuhrsch:
@avinashsai - The docstring wasn't updated. Please send a follow-up PR.

cpuhrsch:
```
 device (str or `torch.device`): A string or instance of `torch.device`
            specifying which device the Variables are going to be created on.	            specifying which device the Variables are going to be created on.
            If left as default, the tensors will be created on cpu. Default: None.	            If left as default, the tensors will be created on cpu. Default: None.
```

but it should be ```Default: cpu``` or such.

avinashsai:
@cpuhrsch where should I edit the docs? Iam not able to find the location.

zhangguanheng66:
The default is `None` and `None` will be converted to `cpu` device.

PR #746, opened 2020-04-29
Disable Travis tests of which equivalent run on CCI

mthrok:
@cpuhrsch followup `a)` of #744

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,6 @@ cache:
 matrix:
   fast_finish: true
   include:
-    - env: PYTHON_VERSION="3.5" COVERAGE="true"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -12,9 +12,6 @@ cache:
 matrix:
   fast_finish: true
   include:
-    - env: PYTHON_VERSION="3.5" COVERAGE="true"
-    - env: PYTHON_VERSION="3.6" COVERAGE="true"
-    - env: PYTHON_VERSION="3.6" RUN_FLAKE8="true" SKIP_TESTS="true"
```
cpuhrsch, line 17: This is the test we should have kept around and the next one was to be deleted.

PR #747, opened 2020-04-29
[CCI migration] Disable travis tests except for RUN_FLAKE8

cpuhrsch:


cpuhrsch:
@zhangguanheng66 - This is only temporary. Moto is migrating the testing infrastructure over to circleci (which already runs much faster, albeit without slow tests).

Then once it's running on CCI we can enable their distributed caching for all the ~1GB Vector files and datasets and then hopefully get rid of "slow" tests in general.

For more details see [this comment on the next steps](https://github.com/pytorch/text/pull/744#issuecomment-621468707).

zhangguanheng66:
We should get them covered by CCI ASAP.

cpuhrsch:
A lot of the slow tests already weren't working [as you can see in this log](https://travis-ci.org/github/pytorch/text/jobs/678843090) associated to the PR [that removes all slow tags as an example](https://github.com/pytorch/text/pull/739). The issue is that the slow tests pull in multiple GBs of data from the internet and the tests time out.

zhangguanheng66:
> A lot of the slow tests already weren't working [as you can see in this log](https://travis-ci.org/github/pytorch/text/jobs/678843090) associated to the PR [that removes all slow tags as an example](https://github.com/pytorch/text/pull/739). The issue is that the slow tests pull in multiple GBs of data from the internet and the tests time out.

Typically, I expect those slow tests to finish with warnings because some of them take time to run (see [bottom here](https://travis-ci.org/github/pytorch/text/jobs/681214584)). If the test fails, it means something we need to fix (correct me if I'm wrong).

cpuhrsch:
@zhangguanheng66 - The only issue with these tests when they time out is that they, depending on the test, they seem to never run at all. ```test_vectors_get_vecs``` for example seems to never run because it is taking so long. And even more so, if you remove the slow tags from the tests, it's [the first one to fail](https://travis-ci.org/github/pytorch/text/jobs/678843090) whereas in the [log you reference](https://travis-ci.org/github/pytorch/text/jobs/681214584) it doesn't show up anywhere except at the bottom together with the tests that all seem to have been skipped. Even though I'm a bit surprised it shows the amount of time it took to run it and otherwise doesn't mention the test.

But if this test never runs we should remove it or change it. I think if we cache the vectors it downloads the test might be able to run in a reasonable time and therefore will always run.

PR #750, opened 2020-05-01
20200501 fbsync

cpuhrsch:
This PR pulls in changes from fbsync to get both this repo and the internal repo in sync.


torchtext/vocab.py
```diff
@@ -31,7 +31,7 @@ class Vocab(object):
     # TODO (@mttk): Populate classs with default values of special symbols
     UNK = '<unk>'
 
-    def __init__(self, counter, max_size=None, min_freq=1, specials=['<unk>', '<pad>'],
+    def __init__(self, counter, max_size=None, min_freq=1, specials=('<unk>', '<pad>'),
```
cpuhrsch, line 34: [Mutable default arguments](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments):

"""
A new list is created once when the function is defined, and the same list is used in each successive call.

Python’s default arguments are evaluated once when the function is defined, not each time the function is called (like it is in say, Ruby). This means that if you use a mutable default argument and mutate it, you will and have mutated that object for all future calls to the function as well.
"""

We could use None as a default and construct a list if it is None, but I don't see the need for ```specials``` to be a list based on the rest of the code. [A tuple is not mutable](https://docs.python.org/3.8/library/stdtypes.html#tuples), so we can use it instead.

test/test_vocab.py
```diff
@@ -135,7 +135,7 @@ def test_vocab_download_fasttext_vectors(self):
     def test_vocab_extend(self):
         c = Counter({'hello': 4, 'world': 3, 'ᑌᑎIᑕOᗪᕮ_Tᕮ᙭T': 5, 'freq_too_low': 2})
         # Build a vocab and get vectors twice to test caching.
-        for i in range(2):
+        for _ in range(2):
```
cpuhrsch, line 138: Naming a loop variable and not using it, is a common lint issue. Using ```_``` prevents this.

torchtext/data/utils.py
```diff
@@ -6,7 +6,7 @@
 from functools import partial
 
 
-def _split_tokenizer(x):
+def _split_tokenizer(x): # noqa: F821
     # type: (str) -> List[str]
```
zhangguanheng66, line 10: Since we only support python3 now, we can use python3 type annotation here.

torchtext/experimental/datasets/raw/text_classification.py
```diff
@@ -55,15 +55,9 @@ def __iter__(self):
             self.setup_iter()
 
         for i, item in enumerate(self._iterator):
-            if i == self.start:
```
zhangguanheng66, line 58: Are those changes to resolve the errors for sync? Or I think we should do it in a separate PR.

torchtext/data/utils.py
cpuhrsch: This lint is still being raised internally due to some issues in flake8. This way we suppress the error.

torchtext/experimental/datasets/raw/text_classification.py
cpuhrsch: The lint complained that ```item``` wasn't used even though it is defined. The follow-up code relies on the idea that item will be populated by the loop iteration, which I'm not even sure will happen if self._iterator where to already be depleted, which would then raise an error on line 63.

PR #755, opened 2020-05-06
[BC breaking] Add Sentencepiece torchscript Extension

mthrok:
This PR adds `torchscript` extension `_torchtext.so`, which contains simple interface to `SentencePiece`.

   - SentencePiece `v0.1.86` is used.
   - `libsentencepiece.a` is built right before `_torchtext.so` is compiled.
     The logic for triggering this build from `setuptools` can be found under `build_tools/setup_helpers`.
   - `_torchtext.so` provides interface to train a SentencePiece model and load a model from file.
   - Our custom `SentencePiece` class has the following methods
       - `Encode`
       - `EncodeAsIds`
       - `EncodeAsPieces`
       - ~~is pickle-able **TODO** Add pickle/un-pickle test~~ This is not for simple pickling, but for `torchscrpit` serialization.
  - Our custom `SentencePiece` class replaces the original `SentencePieceProcessor` class, which is the official Python binding using Swig.
      - Due to the limitation of `torchscript`, our custom `SentencePiece` class does not implement special methods `__len__` and `__getitem__`.
      - Other methods provided `SentencePieceProcessor` are not implemented.
  - `generate_sp_model` and `load_sp_model` are replaces with equivalent C++ implementation that handles our custom `SentencePiece` class.
  - Original `load_sp_model` was returning `sentencepiece.SentencePieceProcessor`. (BC-braeaking.)
  - We cannot replace `sentencepiece_tokenizer` and `sentencepiece_numericalizer` with C++ ext yet as C++ ext does not handle generator.

Breaking change:

Previously `torchtext.data.functional.load_sp_model` returned `sentencepiece.SentencePieceProcessor` object, which supported the following methods, in addition to `__len__` and `__getitem__`

```
$ grep '$self->' third_party/sentencepiece/python/sentencepiece.i
    return $self->Load(filename);
    return $self->LoadFromSerializedProto(filename);
    return $self->SetEncodeExtraOptions(extra_option);
    return $self->SetDecodeExtraOptions(extra_option);
    return $self->SetVocabulary(valid_vocab);
    return $self->ResetVocabulary();
    return $self->LoadVocabulary(filename, threshold);
    return $self->EncodeAsPieces(input);
    return $self->EncodeAsIds(input);
    return $self->NBestEncodeAsPieces(input, nbest_size);
    return $self->NBestEncodeAsIds(input, nbest_size);
    return $self->SampleEncodeAsPieces(input, nbest_size, alpha);
    return $self->SampleEncodeAsIds(input, nbest_size, alpha);
    return $self->DecodePieces(input);
    return $self->DecodeIds(input);
    return $self->EncodeAsSerializedProto(input);
    return $self->SampleEncodeAsSerializedProto(input, nbest_size, alpha);
    return $self->NBestEncodeAsSerializedProto(input, nbest_size);
    return $self->DecodePiecesAsSerializedProto(pieces);
    return $self->DecodeIdsAsSerializedProto(ids);
    return $self->GetPieceSize();
    return $self->PieceToId(piece);
    return $self->IdToPiece(id);
    return $self->GetScore(id);
    return $self->IsUnused(id);
    return $self->IsControl(id);
    return $self->IsUnused(id);
    return $self->GetPieceSize();
    return $self->PieceToId(key);
```

The new C++ Extension provides the following methods

```
Encode(input)
EncodeAsIds(input)
EncodeAsPieces(input)
```

### DIFF
--- a/build_tools/setup_helpers/extension.py
+++ b/build_tools/setup_helpers/extension.py
@@ -1,0 +1,117 @@
+import platform
+import subprocess
+from pathlib import Path
+
+from torch.utils.cpp_extension import (
+    CppExtension,
+    BuildExtension as TorchBuildExtension
+)
+
+__all__ = [
+    'get_ext_modules',
+    'BuildExtension',
+]
+
+_ROOT_DIR = Path(__file__).parent.parent.parent.resolve()
+_CSRC_DIR = _ROOT_DIR / 'torchtext' / 'csrc'
+_TP_BASE_DIR = _ROOT_DIR / 'third_party'
+_TP_INSTALL_DIR = _TP_BASE_DIR / 'build'
+
+
+def _get_eca(debug):
+    eca = []
+    if debug:
+        eca += ["-O0", "-g"]
+    else:
+        eca += ["-O3"]
+    return eca
+
+
+def _get_ela(debug):
+    ela = []
+    if debug:
+        if platform.system() == "Windows":
+            ela += ["/DEBUG:FULL"]
+        else:
+            ela += ["-O0", "-g"]
+    else:
+        ela += ["-O3"]
+    return ela
+
+
+def _get_srcs():
+    return [str(p) for p in _CSRC_DIR.glob('**/*.cpp')]
+
+
+def _get_include_dirs():
+    return [
+        str(_CSRC_DIR),
+        str(_TP_INSTALL_DIR / 'include'),
+    ]
+
+
+def _get_library_dirs():
+    return [
+        str(_TP_INSTALL_DIR / 'lib'),
+    ]
+
+
+def _get_libraries():
+    # NOTE: The order of the library listed bellow matters.
+    #
+    # For example, the symbol `sentencepiece::unigram::Model` is
+    # defined in sentencepiece but UNDEFINED in sentencepiece_train.
+    # GCC only remembers the last encountered symbol.
+    # Therefore placing 'sentencepiece_train' after 'sentencepiece' cause runtime error.
+    #
+    # $ nm third_party/build/lib/libsentencepiece_train.a | grep _ZTIN13sentencepiece7unigram5ModelE
+    #                  U _ZTIN13sentencepiece7unigram5ModelE
+    # $ nm third_party/build/lib/libsentencepiece.a       | grep _ZTIN13sentencepiece7unigram5ModelE
+    # 0000000000000000 V _ZTIN13sentencepiece7unigram5ModelE
+    return [
+        'sentencepiece_train',
+        'sentencepiece',
+    ]
+
+
+def _build_sentence_piece():
+    build_dir = _TP_BASE_DIR / 'sentencepiece' / 'build'
+    build_dir.mkdir(exist_ok=True)
+    subprocess.run(
+        args=['cmake', '-DSPM_ENABLE_SHARED=OFF', f'-DCMAKE_INSTALL_PREFIX={_TP_INSTALL_DIR}', '..'],
+        cwd=str(build_dir),
+        check=True,
+    )
+    subprocess.run(
+        args=['make', 'install'],
+        cwd=str(build_dir),
+        check=True,
+    )
+
+
+def _configure_third_party():
+    _build_sentence_piece()
+
+
+_EXT_NAME = 'torchtext._torchtext'
+
+
+def get_ext_modules(debug=False):
+    return [
+        CppExtension(
+            _EXT_NAME,
+            _get_srcs(),
+            libraries=_get_libraries(),
+            include_dirs=_get_include_dirs(),
+            library_dirs=_get_library_dirs(),
+            extra_compile_args=_get_eca(debug),
+            extra_link_args=_get_ela(debug),
+        ),
+    ]
+
+
+class BuildExtension(TorchBuildExtension):
+    def build_extension(self, ext):
+        if ext.name == _EXT_NAME:
+            _configure_third_party()
+        super().build_extension(ext)
--- a/setup.py
+++ b/setup.py
@@ -1,15 +1,18 @@
 #!/usr/bin/env python
-import os
 import io
 import re
+import shutil
+from pathlib import Path
+import distutils.command.clean
 from setuptools import setup, find_packages
+
+from build_tools import setup_helpers
+
+ROOT_DIR = Path(__file__).parent.resolve()
 
 
 def read(*names, **kwargs):
-    with io.open(
-        os.path.join(os.path.dirname(__file__), *names),
-        encoding=kwargs.get("encoding", "utf8")
-    ) as fp:
+    with io.open(ROOT_DIR.joinpath(*names), encoding=kwargs.get("encoding", "utf8")) as fp:
         return fp.read()
 
 
@@ -24,6 +27,27 @@ def find_version(*file_paths):
 
 VERSION = find_version('torchtext', '__init__.py')
 long_description = read('README.rst')
+
+
+class clean(distutils.command.clean.clean):
+    def run(self):
+        # Run default behavior first
+        distutils.command.clean.clean.run(self)
+
+        # Remove torchtext extension
+        for path in (ROOT_DIR / 'torchtext').glob('**/*.so'):
+            print(f'removing \'{path}\'')
+            path.unlink()
+        # Remove build directory
+        build_dirs = [
+            ROOT_DIR / 'build',
+            ROOT_DIR / 'third_party' / 'build',
+        ]
+        for path in build_dirs:
+            if path.exists():
+                print(f'removing \'{path}\' (and everything under it)')
+                shutil.rmtree(str(path), ignore_errors=True)
+
 
 setup_info = dict(
     # Metadata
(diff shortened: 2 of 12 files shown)


benchmark/vectors.py
```diff
@@ -0,0 +1,20 @@
+from torchtext import vocab
+from torchtext import experimental
+from torchtext.data.utils import get_tokenizer
+import time
+
+
+def benchmark_fasttext():
```
zhangguanheng66, line 7: I'd remove the benchmark for vector because it will be a separate PR for the vector work.

torchtext/csrc/sentencepiece.cpp
```diff
@@ -5,51 +5,69 @@
...
+  void Load(const std::string& path) {
+    const auto status = processor_.Load(path);
     if (!status.ok()) {
-        std::stringstream ss;
-        ss << "Failed to load SentencePiece model. Error: ";
-        ss << status.ToString();
-        throw std::runtime_error(ss.str());
+      // TODO: Use proper logging
+      std::cerr << status.ToString() << std::endl;
+      // TODO: Either fail here or propagate the error to Python.
```
cpuhrsch, line 25: pybind catches and translates cpp exceptions: https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html

torchtext/csrc/sentencepiece.cpp
```diff
@@ -5,51 +5,69 @@
...
-    auto status = processor_.LoadFromSerializedProto(content);
+public:
+  std::string model_proto_;
+  SentencePiece() {}
+  /*
+  SentencePiece(const std::string& model) {
+    processor_.LoadFromSerializedProto(model);
+  }
+  */
+  void Load(const std::string& path) {
```
cpuhrsch, line 20: As a design pattern, why not make this a factory function?

torchtext/data/functional.py
```diff
@@ -61,13 +52,7 @@ def load_sp_model(spm_path):
...
-        import sentencepiece as spm
-    except ModuleNotFoundError:
-        raise ImportWarning("Please install sentencepiece")
-    sp_model = spm.SentencePieceProcessor()
-    sp_model.Load(spm_path)
-    return sp_model
+    return torch.ops.torchtext.load_sp_model(spm_path)
 
 
 def sentencepiece_numericalizer(sp_model):
```
zhangguanheng66, line 58: This function is the tokenizer, which tokenizes a list of text strings to a list of index with a pretrained sp model. We should add the cpp extension for this one.

build_tools/setup_helpers/extension.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def _get_eca(debug):
+    eca = []
+    if debug:
+        eca += ["-O0", "-g"]
+    return eca
+
+
+def _get_ela(debug):
```
cpuhrsch, line 28: What about "-O3" etc. for non-debug builds?

setup.py
```diff
@@ -48,11 +72,15 @@ def find_version(*file_paths):
...
     # Package info
-    packages=find_packages(exclude=('test', 'test.*')),
-
-    zip_safe=True,
+    packages=find_packages(exclude=('test*', 'build_tools*')),
+    zip_safe=False,
+    # Extension info
+    ext_modules=setup_helpers.get_ext_modules(),
+    cmdclass={
+        'build_ext': setup_helpers.BuildExtension.with_options(no_python_abi_suffix=True),
```
cpuhrsch, line 81: Why is this abi option necessary?

Also, I'd add a comment that references the issue around dropping symbols we recently referenced (I can dig it up if this doesn't ring a bell).

torchtext/csrc/sentencepiece.cpp
```diff
@@ -0,0 +1,89 @@
...
+namespace torchtext {
+namespace {
+
+struct SentencePiece : torch::CustomClassHolder {
+private:
+  sentencepiece::SentencePieceProcessor processor_;
+
+public:
+  std::string content_;
+  explicit SentencePiece(const std::string &content) : content_(content) {
```
cpuhrsch, line 13: This could be changed to take a ::sentencepiece::<Sentence piece class> via move constructor and then effectively represent a shell that forwards methods to the underlying object.

Do you need this string constructor? Does this match previous behavior? Usually I'd make this a separate factory function and then construct a SentencePiece object that is initialized with whatever ```processor_.LoadFromSerializedProto(content_);``` returns.

torchtext/csrc/sentencepiece.cpp
```diff
@@ -0,0 +1,89 @@
...
+    const auto status = processor_.LoadFromSerializedProto(content_);
+    if (!status.ok()) {
+      throw std::runtime_error("Failed to load SentencePiece model. Error: " +
+                               status.ToString());
+    }
+  }
+
+  std::vector<std::string> Encode(const std::string &input) const {
```
cpuhrsch, line 21: If you want, you could play with iterables here. You pass an object that yields strings and return and object that yields iterators over strings. That could be quite powerful when it comes to avoiding the overhead of writing out a lot of text while processing.

setup.py
mthrok: ah not necessary. It's a reminding custom from what I followed in `torchvision`. (`_C.so` notation.)
I will remove it.

build_tools/setup_helpers/extension.py
mthrok: Sure we can do that. If you give me the list of typical flags that can go here, it would be helpful too.

test/data/test_functional.py
```diff
@@ -66,3 +76,59 @@ def test_simple_space_split(self):
         ref_results = ['test', 'simple', 'space', 'split', 'function']
         self.assertEqual(list(simple_space_split(test_sample))[0],
                          ref_results)
+
+
+class ScriptableSP(torch.jit.ScriptModule):
+    def __init__(self, model_path):
```
cpuhrsch, line 82: I'd make this a factory function.

build_tools/setup_helpers/extension.py
cpuhrsch: Let's start with O3.

torchtext/csrc/sentencepiece.cpp
mthrok: This class has to provide serialization mechanism so that scripted module can be saved.
Ideally, the underlying `SentencePieceProcessor` class should provide such mechanism, but it does not provide interface to serialize the model as protobuf.
Note 1. The lifecycle of the official `SentencePieceProcessor` is 1. train and save to a file then 2. load from the saved file. It does not have a functionality to save the model.
Note 2. If you look at [the source code of `SentencePieceProcessor`](https://github.com/google/sentencepiece/blob/master/src/sentencepiece_processor.h#L462), it almost look like provide serialization interface, but `ModelProto` and all the functionality related to `protobuf` does not exist in the library file.

To workaround this, we need to keep the content of the saved model file (model data serialized protobuf) when loading a model from a file.
So, it is simpler to pass the model content as string to constructor and to instantiate the underlying `SentencePieceProcessor` there. Because, you still need to pass the model content separately for the later serialization.

cpuhrsch: The serialization here is necessary to support pickle? As far as I know that's what we use to save models. We could look into adding custom pickle support [as does pybind11](https://pybind11.readthedocs.io/en/stable/advanced/classes.html#pickling-support) for this. This might also enable us, if they don't already or if there's a need, to apply compression algorithms on the fly and save a smaller model.

torchtext/csrc/sentencepiece.cpp
```diff
@@ -0,0 +1,89 @@
...
+    return processor_.EncodeAsPieces(input);
+  }
+
+  int64_t GetPieceSize() const { return processor_.GetPieceSize(); }
+};
+
+// Registers our custom class with torch.
+static auto sentencepiece =
+    torch::class_<SentencePiece>("torchtext", "SentencePiece")
+        .def(torch::init<std::string>())
```
cpuhrsch, line 42: At a minimum I would remove this unless necessary to prevent this being used in lieu of factory functions.

torchtext/data/functional.py
mthrok: Offline discussion: Handling generator is a more complicated, I will work on it on the next PR.

setup.py
mthrok: Actually, it turned out that we need to provide the full path to `so` file, so removing abi suffix makes it easier to locate the library file.

mthrok:
@zhangguanheng66 @cpuhrsch 

Addressed your feedback.
Regarding replacing `sentencepiece_tokenizer` and `sentencepiece_numericalizer`, we can try in the next step with @cpuhrsch 's idea to pass a generator to C++.

torchtext/csrc/sentencepiece.cpp
zhangguanheng66: Just a reminder. Iterator is not supported by torchscript but we expect those transforms jitable.

mthrok:
@cpuhrsch I squashed the commits and added the list of methods available now and then in the commit message.
Checkout the commit message 👉https://github.com/pytorch/text/pull/755/commits

mthrok:
> Also, maybe it makes sense to wait for **len** and **getitem** to land

I prefer to merge this one then work on separately.

cpuhrsch:
Ok sounds good. Please add a [BC breaking] tag and then this should be good to go. Also, please do an import into fbcode after this.

zhangguanheng66:
We need to update the README file for some instructions to build torchtext from master branch.

PR #759, opened 2020-05-08
[WIP] JIT-able Split,BasicEnglish, SentencePiece and Spacy tokenizer

zhangguanheng66:
Add a generic tokenizer transform with JIT support:

- Basic English
- Simple split
- spaCy
- callable func for third-party tokenizer

### DIFF
--- a/torchtext/experimental/transforms/transforms.py
+++ b/torchtext/experimental/transforms/transforms.py
@@ -1,5 +1,6 @@
 from typing import List
 import torch
+from torchtext.data.functional import load_sp_model
 
 
 class SplitTokenizer(torch.nn.Module):
@@ -40,6 +41,19 @@ class BasicEnglishTokenizer(torch.nn.Module):
                     re_tokens.append(token)
             return re_tokens
         self.tokenizer = basic_english_normalize
+
+    def forward(self, txt: str) -> List[str]:
+        return self.tokenizer(txt)
+
+
+class SentencePieceTokenizer(torch.nn.Module):
+    r""" Initiate a tokenizer transform class
+    """
+
+    def __init__(self, model_path):
+        super(SentencePieceTokenizer, self).__init__()
+        sp_model = load_sp_model(model_path)
+        self.tokenizer = sp_model.EncodeAsPieces
 
     def forward(self, txt: str) -> List[str]:
         return self.tokenizer(txt)
--- a/torchtext/experimental/transforms/transforms.py
+++ b/torchtext/experimental/transforms/transforms.py
@@ -1,0 +1,60 @@
+from typing import List
+import torch
+
+
+class TokenizerTransform(torch.nn.Module):
+    r""" Initiate a tokenizer transform class
+    """
+
+    def __init__(self, tokenizer, language='en'):
+        super(TokenizerTransform, self).__init__()
+        if tokenizer is None:
+            def split_tokenizer(line: str) -> List[str]:  # noqa: F821
+                re_tokens: List[str] = []
+                for token in line.split():
+                    if token != '':
+                        re_tokens.append(token)
+                return re_tokens
+            self.tokenizer = split_tokenizer
+        elif tokenizer == "basic_english":
+            if language != 'en':
+                raise ValueError("Basic normalization is only available for Enlish(en)")
+
+            def basic_english_normalize(line: str) -> List[str]:
+                line = line.lower().replace('\'', ' \' ').replace('\"', '').replace('.', ' . ')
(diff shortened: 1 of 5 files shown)


torchtext/experimental/transforms/transforms.py
```diff
@@ -0,0 +1,60 @@
...
+                for token in line.split():
+                    if token != '':
+                        re_tokens.append(token)
+                return re_tokens
+            self.tokenizer = split_tokenizer
+        elif tokenizer == "basic_english":
+            if language != 'en':
+                raise ValueError("Basic normalization is only available for Enlish(en)")
+
+            def basic_english_normalize(line: str) -> List[str]:
```
cpuhrsch, line 23: We have a generic regex functional already I think right?

cpuhrsch: https://github.com/pytorch/text/blob/master/torchtext/data/functional.py#L125

cpuhrsch: Maybe "custom_regex" isn't the best name

torchtext/experimental/transforms/transforms.py
```diff
@@ -0,0 +1,60 @@
...
+            except ImportError:
+                print("Please install SpaCy. "
+                      "See the docs at https://spacy.io for more information.")
+                raise
+            except AttributeError:
+                print("Please install SpaCy and the SpaCy {} tokenizer. "
+                      "See the docs at https://spacy.io for more "
+                      "information.".format(language))
+                raise
+        elif callable(tokenizer):
```
cpuhrsch, line 54: I don't know why it's necessary to wrap a callable. If the callable isn't JIT-able neither is the transform. If it is, then why use a wrapper? Otherwise I could see a benefit in being able to map over nested structures using a Transform given a simple callable that translates a string to a string. But that could be a separate utility function.

cpuhrsch:
I think if we introduce something like this we should also write up some documentation around standardization similar to [torchaudio's conventions](https://github.com/pytorch/audio#conventions).

torchtext/experimental/transforms/transforms.py
zhangguanheng66: yes that's true. But there is a `try` block in `re`, which is not jitable. That's why I'm re-writing this part.

zhangguanheng66:
> I think if we introduce something like this we should also write up some documentation around standardization similar to [torchaudio's conventions](https://github.com/pytorch/audio#conventions).

Yes. We need a full doc to define the convention.

torchtext/experimental/transforms/transforms.py
```diff
@@ -0,0 +1,86 @@
+from typing import List
+import torch
+from torchtext.data.functional import load_sp_model
+
+
+class SplitTokenizer(torch.nn.Module):
```
cpuhrsch, line 6: I think this should be supported via a RegexTokenizer. It's also not covered by the tests. Even the functionality might be obvious, at the very minimum a test prevents syntax errors.

torchtext/experimental/transforms/transforms.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class BasicEnglishTokenizer(torch.nn.Module):
+    r""" Initiate a tokenizer transform class
+    """
+
+    def __init__(self):
+        super(BasicEnglishTokenizer, self).__init__()
+
+        def basic_english_normalize(line: str) -> List[str]:
```
cpuhrsch, line 32: Effectively this is the functional part of this transform and could be moved into such a file

torchtext/experimental/transforms/transforms.py
zhangguanheng66: Yes. We can re-write the re package with JIT support. It's on the plan.

torchtext/experimental/transforms/transforms.py
zhangguanheng66: I agree.

torchtext/experimental/transforms/transforms.py
cpuhrsch: Let's do that first then before we introduce an abstraction we plan to immediately after remove. Or should we have a SplitTokenizer in the long term as well?

zhangguanheng66: I think we can have different tokenizer transforms around. For example, an sentencepiece transform accept a sp model in the constructor and convert string into tokens. Another one could be convert string into index. With those two, it covers some usage cases in pytext.

PR #761, opened 2020-05-09
Add methods defined on SentencePieceProcessor

mthrok:
This PR adds most of methods define in SentencePieceProcessor Python wrapper.
~~Blocked by https://github.com/pytorch/pytorch/pull/38167~~

- `NBestEncodeAsPieces`
- `NBestEncodeAsIds`
- `SampleEncodeAsPieces`
- `SampleEncodeAsIds`
- `DecodePieces`
- `DecodeIds`
- `GetPieceSize`
- `PieceToId`
- `IdToPiece`
- `GetScore`
- `IsUnknown`
- `IsUnused`
- `IsControl`
- `unk_id`
- `bos_id`
- `eos_id`
- `pad_id`
- `SetEncodeExtraOptions`
- `SetDecodeExtraOptions`
- `SetVocabulary`
- `ResetVocabulary`
- `LoadVocabulary`
- `__len__`
- `__getitem__`

### DIFF
--- a/torchtext/csrc/sentencepiece.cpp
+++ b/torchtext/csrc/sentencepiece.cpp
@@ -49,7 +49,107 @@ public:
     return processor_.EncodeAsPieces(input);
   }
 
+  std::vector<std::vector<std::string>>
+  NBestEncodeAsPieces(const std::string &input,
+                      const int64_t nbest_size) const {
+    return processor_.NBestEncodeAsPieces(input, nbest_size);
+  }
+
+  std::vector<std::vector<int64_t>>
+  NBestEncodeAsIds(const std::string &input, const int64_t nbest_size) const {
+    const auto vals = processor_.NBestEncodeAsIds(input, nbest_size);
+    std::vector<std::vector<std::int64_t>> ret;
+    for (const auto &vec : vals) {
+      ret.emplace_back(std::vector<std::int64_t>(vec.begin(), vec.end()));
+    }
+    return ret;
+  }
+
+  std::vector<std::string> SampleEncodeAsPieces(const std::string &input,
+                                                const int64_t nbest_size,
+                                                const double alpha) const {
+    return processor_.SampleEncodeAsPieces(input, nbest_size, alpha);
+  }
+
+  std::vector<int64_t> SampleEncodeAsIds(const std::string &input,
+                                         const int64_t nbest_size,
+                                         const double alpha) const {
+    const auto val = processor_.SampleEncodeAsIds(input, nbest_size, alpha);
+    return std::vector<int64_t>(val.begin(), val.end());
+  }
+
+  std::string DecodePieces(const std::vector<std::string> &input) const {
+    return processor_.DecodePieces(input);
+  }
+
+  std::string DecodeIds(const std::vector<int64_t> &input) const {
+    const std::vector<int> val(input.begin(), input.end());
+    return processor_.DecodeIds(val);
+  }
+
   int64_t GetPieceSize() const { return processor_.GetPieceSize(); }
+
+  int64_t PieceToId(const std::string &piece) const {
+    return processor_.PieceToId(piece);
+  }
+
+  std::string IdToPiece(const int64_t id) const {
+    return processor_.IdToPiece(id);
+  }
+
+  double GetScore(const int64_t id) const { return processor_.GetScore(id); }
+
+  bool IsUnknown(const int64_t id) const { return processor_.IsUnknown(id); }
+
+  bool IsUnused(const int64_t id) const { return processor_.IsUnused(id); }
+
+  int64_t unk_id() const { return processor_.unk_id(); }
+
+  int64_t bos_id() const { return processor_.bos_id(); }
+
+  int64_t eos_id() const { return processor_.eos_id(); }
+
+  int64_t pad_id() const { return processor_.pad_id(); }
+
+  void SetEncodeExtraOptions(const std::string &extra_option) {
+    const auto status = processor_.SetEncodeExtraOptions(extra_option);
+    if (!status.ok()) {
+      throw std::runtime_error("Failed to set encode extra options. Error: " +
+                               status.ToString());
+    }
+  }
+
+  void SetDecodeExtraOptions(const std::string &extra_option) {
+    const auto status = processor_.SetDecodeExtraOptions(extra_option);
+    if (!status.ok()) {
+      throw std::runtime_error("Failed to set decode extra options. Error: " +
+                               status.ToString());
+    }
+  }
+
+  void SetVocabulary(const std::vector<std::string> &valid_vocab) {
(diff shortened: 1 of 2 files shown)


torchtext/csrc/sentencepiece.cpp
```diff
@@ -39,7 +39,59 @@ struct SentencePiece : torch::CustomClassHolder {
...
+    return processor_.PieceToId(piece);
+  }
+
+  std::string IdToPiece(const int64_t id) const {
+    return processor_.IdToPiece(id);
+  }
+
+  double GetScore(const int64_t id) const { return processor_.GetScore(id); }
+
+  bool IsUnknown(const int64_t id) const { return processor_.IsUnused(id); }
```
cpuhrsch, line 92: Where does this come from?

torchtext/csrc/sentencepiece.cpp
```diff
@@ -39,7 +39,59 @@ struct SentencePiece : torch::CustomClassHolder {
...
+  NBestEncodeAsIds(const std::string &input, const int64_t nbest_size) const {
+    const auto vals = processor_.NBestEncodeAsIds(input, nbest_size);
+    std::vector<std::vector<std::int64_t>> ret;
+    for (const auto &vec : vals) {
+      ret.emplace_back(std::vector<std::int64_t>(vec.begin(), vec.end()));
+    }
+    return ret;
+  }
+
+  std::vector<std::string> SampleEncodeAsPieces(const std::string &input,
```
zhangguanheng66, line 58: With `SampleEncodeAsPieces` and `SampleEncodeAsIds`, we can add them to sentencepiece transforms [here](https://github.com/zhangguanheng66/text/blob/c30188b7b68a50d18bc2908442bd9786b012108a/torchtext/experimental/transforms/transforms.py#L49-L59)

torchtext/csrc/sentencepiece.cpp
mthrok: You can see the list of methods defined on the original SentencePieceProcessor class here. [sentencepiece.py](https://github.com/google/sentencepiece/blob/e8a84a16d13e8bf92892a1cd92e4de3b0d0321fd/python/sentencepiece.py#L101)

zhangguanheng66: Why do we need all the methods here? My understanding is that uses will usually load a sp model and apply it as a tokenizer.

mthrok: This is an attempt to keep BC change minimum. Before we introduce the C++ extension, all the methods were available.
Since we are adding the methods that are not used in library such as `__len__` and `__getitem__`, I thought it's a good idea to add other methods.

cpuhrsch: Ah maybe you wanted to call ```processor_.IsUnknown(id)``` instead of ```processor_.IsUnused(id)``` for the definition of ```bool IsUnknown(const int64_t id)```?

cpuhrsch:
I'm not sure it's necessary to expose these methods for the sake of completion if we don't have a torchtext usecase in mind. This class is not supposed to be used explicitly, but in conjunction with the torchtext transforms.

torchtext/csrc/sentencepiece.cpp
mthrok: Good catch. will fix.

mthrok: fixed.

cpuhrsch:
Let us know when you need a review on this

torchtext/csrc/sentencepiece.cpp
```diff
@@ -48,7 +48,107 @@ struct SentencePiece : torch::CustomClassHolder {
...
   }
 
+  std::vector<std::vector<std::string>>
+  NBestEncodeAsPieces(const std::string &input,
+                      const int64_t nbest_size) const {
+    return processor_.NBestEncodeAsPieces(input, nbest_size);
+  }
+
+  std::vector<std::vector<int64_t>>
+  NBestEncodeAsIds(const std::string &input, const int64_t nbest_size) const {
```
zhangguanheng66, line 58: What's difference between `NBestEncodeAsIds ` and `EncodeAsIds`?

mthrok:
> I was wondering why we need to impose all those methods. Probably our users won't expect those methods from our side and I don't see them in pytext.

@zhangguanheng66  Okay I removed. Once it passes the tests, it's ready to merge.

torchtext/csrc/sentencepiece.cpp
```diff
@@ -58,7 +62,8 @@ static auto sentencepiece =
         .def("Encode", &SentencePiece::Encode)
         .def("EncodeAsIds", &SentencePiece::EncodeAsIds)
         .def("EncodeAsPieces", &SentencePiece::EncodeAsPieces)
-        .def("GetPieceSize", &SentencePiece::GetPieceSize)
+        .def("__len__", &SentencePiece::GetPieceSize)
+        .def("__getitem__", &SentencePiece::PieceToId)
```
zhangguanheng66, line 66: @cpuhrsch Do you think it's make more sense to have `__call__`? This will be callable directly as a transform. Or we could make `EncodeAsIds ` as `__call__`.

cpuhrsch: I think for now this is meant as a simple internal wrapper around this object, which is then wrapped into a functional + transform and as such mirrors the API for the Python sentencepiece model.

cpuhrsch: @mthrok - please correct me if I'm wrong

mthrok: @cpuhrsch Well I just picked up your work, so I do not have a strong preference, but I see `SentencePiece` class as a container/surrogate for the underlying model, and I think natural interpretation of the current implementation is that it is consumed by functionals and transforms.

I do not have strong preference on adding `__call__` but as we see we have `Encode`, `EncodeAsIds` and `EncodeAsPieces` methods for encoding, and users have to pick one appropriate for them, so I do not feel adding `__call__` would provide much value.

hudeven:
LGTM. thanks @mthrok !

zhangguanheng66:
Here are a few python wrappers:

- `encode_as_pieces` tokenizes a sentence into a list of tokens
- `encode_as_ids` tokenizes a sentence into a list of tokens. In general, this wrap could be directly applied to convert a sentence into a tensor that can be sent to the model for training/inference
- `piece_to_id` is equivalent to `stoi` in `torchtext.vocab` class
- `id_to_piece` is equivalent to `itos` in `torchtext.vocab` class
- `decode_pieces` combines a list of tokens into the original sentence
- `decode_ids` combines a list to token ids into the original sentence

Will look other funcs in cpp side.

mthrok:
@zhangguanheng66 I added simple tests for `piece_to_id`, `id_to_piece`, `decode_pieces` and `decode_ids`.

PR #763, opened 2020-05-12
Update installation direction

mthrok:
Preview: https://github.com/mthrok/text/blob/update-readme/README.rst

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,16 +21,24 @@ Note: we are currently re-designing the torchtext library to make it more compat
 Installation
 ============
 
+We recommend Anaconda as Python package management system. Please refer to `pytorch.org <https://pytorch.org/>`_ for the detail of PyTorch installation. The following is the corresponding ``torchtext`` versions and supported Python versions.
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -44,6 +52,21 @@ Alternatively, you might want to use the `Moses <http://www.statmt.org/moses/>`_
...
+
+Building from source
+--------------------
+
+To build torchtext from source, you need ``git``, ``CMake`` and C++11 compiler such as ``g++``.::
+
+    git clone https://github.com/pytorch/text torchtext
+    cd torchtext
+    git submodule update --init --recursive
+    python setup.py clean install
```
zhangguanheng66, line 67: Please double check that the package is installed at global site-packages directory properly. You could try to `import torchtext` outside the `text` folder and see if it works properly.

mthrok: Oops, you are right. The installation does not work.
I think https://github.com/pytorch/text/pull/764 should solve this.

PR #764, opened 2020-05-12
Fix extension import path for buck

mthrok:
When building `torchtext` with buck, the path to `_torchtext.so` has to be normalized.
The fix was taken from [`torchvision`'s equivalent](https://github.com/pytorch/vision/blob/4e138bd5602d7e1b3713998ad213c5e1f16e25ce/torchvision/extension.py#L4-L20)

### DIFF
--- a/torchtext/__init__.py
+++ b/torchtext/__init__.py
@@ -14,9 +14,23 @@ __all__ = ['data',
 
 
 def _init_extension():
+    import os
+    import importlib
     import torch
-    torch.ops.load_library('torchtext/_torchtext.so')
-    torch.classes.load_library('torchtext/_torchtext.so')
+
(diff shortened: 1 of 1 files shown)


zhangguanheng66:
Please double check the global install works properly.

zhangguanheng66:
nit: you might fix the lint errors in the travis test since they are really minor. Thanks.

mthrok:
This does solve library resolution but it leads to another undefined symbol issue. Taking a look at it.

mthrok:
@zhangguanheng66 I verified that the code works fine. The caveat is when building `torchtext` from source, one has to use the same gcc as the one used for building PyTorch. I will update the #763 . We need to add smoketest job, like `torchaudio`.

PR #765, opened 2020-05-12
Temporary fix for flake8

mthrok:
The latest version of flake8 added check for E741 and causing CI to fail.
This PR temporarily fix it by pinning the flake8 version.

```
./torchtext/datasets/translation.py:181:21: E741 ambiguous variable name 'l'
./test/common/torchtext_test_case.py:110:30: E741 ambiguous variable name 'l'
```

### DIFF
--- a/.circleci/unittest/scripts/environment.yml
+++ b/.circleci/unittest/scripts/environment.yml
@@ -1,7 +1,7 @@
 channels:
   - defaults
 dependencies:
-  - flake8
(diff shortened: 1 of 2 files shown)


zhangguanheng66:
Is this something we should report at flake8 repo?

mthrok:
> Is this something we should report at flake8 repo?

I do not think so.

PR #767, opened 2020-05-13
BERT example in torchtext

zhangguanheng66:
Train a BERT model with PyTorch and torchtext, including masked language modeling and next sentence tasks. Fine-tune the BERT model for question-answer task.

There are a few things to do:
- [x] Move some datasets to `torchtex.experimental.datasets`, including SQuAD. Some ongoing PRs https://github.com/pytorch/text/pull/773, https://github.com/pytorch/text/pull/774
- [x] Switch the models in model.py file to the corresponding components in pytorch or torchtext. For example, the MultiheadAttention container in https://github.com/pytorch/text/pull/720
- [x] A more generic metric for F1 and exact score in torchtext
- [x] Remove some duplicate codes across the three tasks.


examples/BERT/data.py
```diff
@@ -0,0 +1,603 @@
+import torch
```
cpuhrsch, line 1: I guess we could start incrementally porting this over into the experimental folder which will also help with cleanup. These datasets seem useful in general.

examples/BERT/metrics.py
```diff
@@ -0,0 +1,72 @@
...
+        pred_str = " ".join(pred_tokens)
+        candidate_score = []
+        for item in ans_tokens:
+            ans_str = " ".join(item)
+            candidate_score.append(int(normalize_txt(ans_str) == normalize_txt(pred_str)))
+        exact_scores.append(max(candidate_score))
+    return 100.0 * sum(exact_scores) / len(exact_scores)
+
+
+def compute_qa_f1(ans_pred_tokens_samples):
```
cpuhrsch, line 43: can this be turned into a generic f1 metric that we can throw into torchtext?

examples/BERT/mlm_task.py
```diff
@@ -0,0 +1,263 @@
...
+        test_data = torch.tensor([vocab.stoi[_id]
+                                 for _id in enwik9[idx2:]]).long()
+        from torchtext.experimental.datasets import LanguageModelingDataset
+        train_dataset = LanguageModelingDataset(train_data, vocab)
+        valid_dataset = LanguageModelingDataset(val_data, vocab)
+        test_dataset = LanguageModelingDataset(test_data, vocab)
+    elif args.dataset == 'BookCorpus':
+        train_dataset, test_dataset, valid_dataset = BookCorpus(vocab)
+
+    train_data = process_raw_data(train_dataset.data, args)
```
cpuhrsch, line 155: I'd move this until the end of the function into a separate ```def train(...)``` function.

examples/BERT/ns_task.py
```diff
@@ -0,0 +1,248 @@
...
+                      'loss {:8.5f} | ppl {:5.2f}'.format(epoch, idx,
+                                                          len(train_dataset) // batch_size,
+                                                          scheduler.get_last_lr()[0],
+                                                          elapsed * 1000 / args.log_interval,
+                                                          cur_loss, math.exp(cur_loss)))
+            total_loss = 0
+            start_time = time.time()
+
+
+def run_main(args, rank=None):
```
cpuhrsch, line 122: It seems like some of this code used here could be shared with the other run_main functions?

cpuhrsch:
Looks pretty good! I think the next steps could include some code deduplication. Looks like you could factor out some more stuff already and also use more stuff from torchtext.

examples/BERT/README.md
```diff
@@ -0,0 +1,56 @@
...
+-------------------------
+
+Train the BERT model with masked language modeling task and next-sentence task. Run the tasks on a local GPU or CPU:
+
+    python mlm_task.py
+    python ns_task.py
+
+or run the tasks on a SLURM powered cluster with Distributed Data Parallel (DDP):
+
+    srun --label --ntasks-per-node=1 --time=400 --mem-per-cpu=5120 --gres=gpu:8 --cpus-per-task 80 --nodes=1 --pty python mlm_task.py --parallel DDP
```
vincentqb, line 16: We should make an explicit decision on whether we should add SLURM support or not. Since SLURM is only one possible technology that users could use, I would lean for removing this from the implementation to keep things leaner. Thoughts?

I only see partial references in torchvision.
```
classification/utils.py:    elif 'SLURM_PROCID' in os.environ:
classification/utils.py:        args.rank = int(os.environ['SLURM_PROCID'])
detection/utils.py:    elif 'SLURM_PROCID' in os.environ:
detection/utils.py:        args.rank = int(os.environ['SLURM_PROCID'])
segmentation/utils.py:    elif 'SLURM_PROCID' in os.environ:
segmentation/utils.py:        args.rank = int(os.environ['SLURM_PROCID'])
video_classification/utils.py:    elif 'SLURM_PROCID' in os.environ:
video_classification/utils.py:        args.rank = int(os.environ['SLURM_PROCID'])
```

examples/BERT/data.py
vincentqb: I second this. I would also help make this pull request shorter.

examples/BERT/README.md
```diff
@@ -0,0 +1,56 @@
...
+This example shows how to train a BERT model with PyTorch and torchtext only. Then, we fine-tune the pre-trained BERT for the question-answer task.
+
+
+Generate pre-trained BERT
+-------------------------
+
+Train the BERT model with masked language modeling task and next-sentence task. Run the tasks on a local GPU or CPU:
+
+    python mlm_task.py
+    python ns_task.py
```
vincentqb, line 12: Should these two tasks be added as separate pull request?

examples/BERT/ns_task.py
```diff
@@ -0,0 +1,248 @@
...
+                tok_type = tok_type.to(device)
+                target_ns_labels = target_ns_labels.to(device)
+            seq_input = seq_input.transpose(0, 1)  # Wrap up by DDP or DataParallel
+            ns_labels = model(seq_input, token_type_input=tok_type)
+            loss = criterion(ns_labels, target_ns_labels)
+            total_loss += loss.item()
+    return total_loss / (len(data_source) // batch_size)
+
+
+def train(train_dataset, model, train_loss_log, device, optimizer, criterion,
```
vincentqb, line 80: nit: `train_one_epoch`, as done [here](https://github.com/pytorch/vision/blob/master/references/classification/train.py#L19)?

examples/BERT/mlm_task.py
```diff
@@ -0,0 +1,263 @@
...
+    if rank is not None:
+        # Chunk training data by rank for different gpus
+        chunk_len = len(train_data) // args.world_size
+        train_data = train_data[(rank * chunk_len):((rank + 1) * chunk_len)]
+    val_data = process_raw_data(valid_dataset.data, args)
+    test_data = process_raw_data(test_dataset.data, args)
+
+    if args.checkpoint != 'None':
+        model.bert_model = torch.load(args.checkpoint)
```
vincentqb, line 164: Given that SLURM was used here, I would have also have expected to capture the sigterm/sigusr1 signal, as done [here](https://github.com/fairinternal/FAIR_Cluster/blob/master/checkpoint/main.py#L146). This allows creating a checkpoint when some signal is received, say when the job is about to timeout.

zhangguanheng66:
> It would be nice if this were divided into at least 3 PRs: one for each of the two tasks, and one for some of the abstractions that could land straight into torchtext (e.g. datasets).

Thanks for the feedback. Yes, there are currently two PRs to land the datasets into torchtext and a separate PR to merge the model. Once done, this PR will be very task-related work.

examples/BERT/README.md
zhangguanheng66: In the specific task, like `ns_task.py` and `mlm_task.py`, we support single machine (either GPU or CPU) and DDP. This command here is to show how to run DDP with SLURM. The implementation of the tasks doesn't assume SLURM as the distributed technology.  @vincentqb

examples/BERT/mlm_task.py
```diff
@@ -0,0 +1,263 @@
...
+                                                          cur_loss, math.exp(cur_loss)))
+            total_loss = 0
+            start_time = time.time()
+
+
+def run_main(args, rank=None):
+    torch.manual_seed(args.seed)
+    if args.parallel == 'DDP':
+        n = torch.cuda.device_count() // args.world_size
+        device = list(range(rank * n, (rank + 1) * n))
```
vincentqb, line 105: nit: this assumes a distribution of GPUs across machines

examples/BERT/utils.py
```diff
@@ -0,0 +1,58 @@
+import torch
+import torch.distributed as dist
+import os
+import torch.multiprocessing as mp
+import math
+
+
+def setup(rank, world_size, seed):
+    os.environ['MASTER_ADDR'] = 'localhost'
+    os.environ['MASTER_PORT'] = '12355'
```
vincentqb, line 10: nit: this is specific to the particular distributed implementation

examples/BERT/ns_task.py
zhangguanheng66: You are right. However, since those three tasks are quite different, I prefer to explicitly set up the `run_main` func

examples/BERT/utils.py
zhangguanheng66: It's set up to SLURM.

examples/BERT/metrics.py
zhangguanheng66: `sample_f1` func can be landed in torchtext.

examples/BERT/data.py
```diff
@@ -0,0 +1,79 @@
+import glob
+import torch
+import logging
+from torchtext.data.utils import get_tokenizer
+import random
+
+
+class LanguageModelingDataset(torch.utils.data.Dataset):
```
cpuhrsch, line 8: Can't you use https://github.com/pytorch/text/blob/master/torchtext/experimental/datasets/language_modeling.py#L15 ?

examples/BERT/data.py
```diff
@@ -0,0 +1,79 @@
...
+            yield x
+
+    def get_vocab(self):
+        return self.vocab
+
+
+###################################################################
+# Set up dataset for book corpus
+###################################################################
+def BookCorpus(vocab, tokenizer=get_tokenizer("basic_english"),
```
cpuhrsch, line 37: Maybe it's worth moving this into experimental as well?

examples/BERT/ns_task.py
```diff
@@ -0,0 +1,251 @@
...
+    shuffle_idx2 = torch.randperm(len(processed_data))
+    num_shuffle = int(len(processed_data) * args.frac_ns)
+    shuffle_zip = list(zip(shuffle_idx1, shuffle_idx2))[:num_shuffle]
+    for (i, j) in shuffle_zip:
+        processed_data[i][1] = processed_data[j][0]
+        processed_data[i][2] = int(0)  # Switch same sentence label to false 0
+    return processed_data
+
+
+def collate_batch(batch, args, cls_id, sep_id, pad_id):
```
cpuhrsch, line 34: You have "collate_batch" earlier as well. Is there some combination possible here with a utils file?

examples/BERT/data.py
zhangguanheng66: Switched to the one in `experimental/datasets`.

examples/BERT/data.py
zhangguanheng66: I was thinking about this. But the original data for BookCorpus comes from FAIR cluster.

PR #772, opened 2020-05-14
Add Windows unit tests

peterjc123:


peterjc123:
cc @mthrok @zhangguanheng66

torchtext/csrc/sentencepiece.cpp
```diff
@@ -1,6 +1,23 @@
 #include <sentencepiece_trainer.h>
 #include <torch/script.h>
 
+#ifdef _WIN32
+#include <Python.h>
+#if PY_MAJOR_VERSION < 3
```
mthrok, line 6: nit: As torchtext dropped the Python 2 support, I feel this redundant. though it should not hurt.

peterjc123: Fixed!

mthrok:
Thank you very much!

PR #773, opened 2020-05-14
Question answer datasets: SQuAD1 and SQuAD2

zhangguanheng66:


torchtext/experimental/datasets/raw/question_answer.py
```diff
@@ -0,0 +1,78 @@
...
+                        processed['answers'] = [""]
+                        processed['answer_start'] = [-1]
+                    yield processed
+
+
+class RawQuestionAnswerDataset(torch.utils.data.IterableDataset):
+    """Defines an abstraction for raw question answer iterable datasets.
+    """
+
+    def __init__(self, iterator, start=0, num_lines=None):
```
cpuhrsch, line 34: The other dataset still has the setup_iter method, so I'd mirror that behavior or change the behavior of the other.

zhangguanheng66: OK, let's keep this consistent for now since there are several ongoing PRs to re-write datasets. Once done, we can send out a separate PR to fix it.

Issue #742, opened 2020-04-24
[HELP WANTED] Re-write datasets in torchtext


zhangguanheng66:
As mentioned in https://github.com/pytorch/text/issues/664, we are working on a new dataset abstraction. The new datasets will be more compatible with pytorch core library and capable of out-of-box libraries (like SentencePiece BPE). 

We have landed several datasets in `torchtext.experimental.datasets` folder to test the new abstraction. Now, we want some help from open-source community. Please sign up here and contribute PRs to re-write those datasets in torchtext. Those datasets in `torchtext/experimental/datasets/text_classification` should be some good examples to follow. Some datasets but not limited include:
- [x] Word language modeling datasets. Already in the experimental folder but need some update to the latest abstraction.
- [x] IMDB
- [x] Text classification datasets (AG_NEWS, SogouNews, DBpedia, YelpReviewPolarity, YelpReviewFull, YahooAnswers, AmazonReviewPolarity, AmazonReviewFull)
- [x] Translation
- [x] Question-answer datasets (e.g. SQuAD [here](https://github.com/zhangguanheng66/torchBERT/blob/e8cd8e5546d46d7e9788969bebb85933e0a91bf0/torchBERT/data.py#L135))
- [x] EnWik9, which's already good
- [x] Sequence tagging dataset in #805

Here is a checklist that you may consider to migrate a dataset:
1. Set up a IterableDataset with an iterator to read raw data [link](https://github.com/pytorch/text/blob/8833b970df08b1b0ec18b774cd0047c4160a9f84/torchtext/experimental/datasets/raw/text_classification.py#L89)
2. Cache out the raw dataset and create a transform pipeline [link](https://github.com/pytorch/text/blob/8833b970df08b1b0ec18b774cd0047c4160a9f84/torchtext/experimental/datasets/text_classification.py#L90)
    - pick up a tokenizer
    - generate the vocab object if not provided by users
    - attach tokenizer + vocab + totensor to the transform pipeline
3. Add unit tests [link](https://github.com/pytorch/text/blob/8833b970df08b1b0ec18b774cd0047c4160a9f84/test/data/test_builtin_datasets.py#L114)
    - check the total length of the dataset
    - check the content of the dataset

Free free to ping me and @cpuhrsch if you want to check some ideas for this. Thanks in advance.

akurniawan:
hi @zhangguanheng66 I'm interested to work on the translation dataset migration

zhangguanheng66:
[...]

PR #774, opened 2020-05-15
experimental.dataset WikiText2, WikiText103, PennTreeBank, WMTNewsCrawl

z-Runmin:
- Update word language modeling datasets #742


torchtext/experimental/datasets/language_modeling.py
```diff
@@ -1,22 +1,38 @@
...
-        'https://s3.amazonaws.com/research.metamind.io/wikitext/wikitext-2-v1.zip',
-    'WikiText103':
-        'https://s3.amazonaws.com/research.metamind.io/wikitext/wikitext-103-v1.zip',
-    'PennTreebank':
-        ['https://raw.githubusercontent.com/wojzaremba/lstm/master/data/ptb.train.txt',
-         'https://raw.githubusercontent.com/wojzaremba/lstm/master/data/ptb.test.txt',
-         'https://raw.githubusercontent.com/wojzaremba/lstm/master/data/ptb.valid.txt']
-}
+
+def vocab_func(vocab):
```
zhangguanheng66, line 8: if `vocab_func`, `totensor`, `build_vocab` are same from those in `torchtext.experimental.datasets.text_classification`, we don't need to repeat here. I pull them to question answer datasets in the same way. https://github.com/pytorch/text/blob/edb143af512e890f3a00d3e872658f0986f3b23a/torchtext/experimental/datasets/question_answer.py#L5

torchtext/experimental/datasets/raw/language_modeling.py
```diff
@@ -0,0 +1,162 @@
...
+    def __init__(self, iterator):
+        """Initiate language modeling dataset.
+        """
+        super(RawTextIterableDataset, self).__init__()
+        self._iterator = iterator
+        self.has_setup = False
+        self.start = 0
+        self.num_lines = None
+
+    def setup_iter(self, start=0, num_lines=None):
```
zhangguanheng66, line 32: we decides to move `start` and `num_lines` to the constructor and remove `def setup_iter`.

torchtext/experimental/datasets/text_classification.py
```diff
@@ -32,7 +32,7 @@ def build_vocab(data, transforms):
     return build_vocab_from_iterator(tok_list)
 
 
-def squential_transforms(*transforms):
+def sequential_transforms(*transforms):
```
zhangguanheng66, line 35: You could send out a separate PR to fix those typos together with `start` and `num_lines` in `TextClassificationDataset`

torchtext/experimental/datasets/language_modeling.py
cpuhrsch: Or let's move them in a functional.py file similar to torchaudio and pytorch

torchtext/experimental/datasets/text_classification.py
z-Runmin: Sure. I'll revert these changes and submit another PR.

torchtext/experimental/datasets/raw/language_modeling.py
z-Runmin: Will submit another PR for this issue.

torchtext/experimental/datasets/language_modeling.py
z-Runmin: Duplicated functions `vocab_func`, `totensor` have been removed.

torchtext/experimental/datasets/language_modeling.py
```diff
@@ -1,22 +1,24 @@
...
+
+
+def build_vocab(data, transforms):
+    tok_list = []
+    for txt in data:
+        tok_list.append(transforms(txt))
+    return build_vocab_from_iterator(tok_list)
+
+
+def sequential_transforms(*transforms):
```
zhangguanheng66, line 15: you can load this one from `experimental.functional`

torchtext/experimental/datasets/language_modeling.py
```diff
@@ -29,27 +31,22 @@ class LanguageModelingDataset(torch.utils.data.Dataset):
...
-            >>> dataset = LanguageModelingDataset(data, vocab)
+            transforms: Text string transforms.
 
         """
 
         super(LanguageModelingDataset, self).__init__()
-        self.data = data
         self.vocab = vocab
+        self.transforms = transforms
+        self.data = torch.cat(tuple(transforms(row) for row in data), axis=0)
```
zhangguanheng66, line 49: To be consistent with text classification datasets, please call the transform func in `__getitem__` func

torchtext/experimental/datasets/language_modeling.py
```diff
@@ -65,63 +62,31 @@ def get_vocab(self):
...
-def _get_datafile_path(key, extracted_files):
-    for fname in extracted_files:
-        if key in fname:
-            return fname
-
-
-def _setup_datasets(dataset_name, tokenizer=get_tokenizer("basic_english"),
-                    root='.data', vocab=None, removed_tokens=[],
-                    data_select=('train', 'test', 'valid')):
+def _setup_datasets(dataset_name, root='.data', vocab=None,
```
zhangguanheng66, line 65: Why do we want to change the order of `tokenizer` and `vocab`? This will be a BC breaking.

torchtext/experimental/datasets/language_modeling.py
```diff
@@ -238,3 +195,10 @@ def PennTreebank(*args, **kwargs):
     """
 
     return _setup_datasets(*(("PennTreebank",) + args), **kwargs)
+
+
+DATASETS = {
+    'WikiText2': raw.WikiText2,
```
zhangguanheng66, line 201: Please follow the convention in https://github.com/pytorch/text/pull/775

torchtext/experimental/datasets/raw/language_modeling.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+
+    def __init__(self, iterator, start=0, num_lines=None):
+        """Initiate language modeling dataset.
+        """
+        super(RawTextIterableDataset, self).__init__()
+        self._iterator = iterator
+        self.start = start
+        self.num_lines = num_lines
+
```
zhangguanheng66, line 29: You can keep `setup_iter` there as text classification datasets as [here](https://github.com/pytorch/text/blob/9bdfb046c37b473accb7749ecc7c8e0e4ed7860f/torchtext/experimental/datasets/raw/text_classification.py#L47) and we will remove them all together once we agree with the abstraction.

torchtext/experimental/datasets/raw/language_modeling.py
```diff
@@ -0,0 +1,146 @@
...
+            if i >= self.start:
+                yield item
+            if (self.num_lines is not None) and (i == (self.start + self.num_lines)):
+                break
+
+    def get_iterator(self):
+        return self._iterator
+
+
+def _get_datafile_path(key, extracted_files):
```
zhangguanheng66, line 41: Since `_get_datafile_path` is called only once, it may not be necessary to separate it from `_setup_datasets` func.

torchtext/experimental/datasets/language_modeling.py
```diff
@@ -210,14 +159,13 @@ def PennTreebank(*args, **kwargs):
     Separately returns the train/test/valid set
 
     Arguments:
+        root: Directory where the datasets are saved. Default: ".data"
+        vocab: Vocabulary used for dataset. If None, it will generate a new
```
zhangguanheng66, line 163: Since you change the order of `vocab` and `tokenizer`, will this be a BC breaking?

z-Runmin: Yes, it's a possible BC breaking if people use `args` instead of `kwargs`.  I restored the previous order of `vocab` and `tokenizer` in commit f433b40 above.

zhangguanheng66: Great. Let me know when you are done with the revision. I will have another look.

torchtext/experimental/datasets/language_modeling.py
z-Runmin: It might be hard to call the transform func in `__getitem__`. For dataset like WikiText2, the raw data `wiki.train.tokens` is stored as a multi-line txt file, and `self.__getitem__[i]` is expeced to output i-th token. Therefore, pre-processing is required to concat and tokenize the multi-line file. Can you advise if we need an additional `preprocess` func?

torchtext/experimental/datasets/raw/language_modeling.py
z-Runmin: `setup_iter` restored in 321871d

torchtext/experimental/datasets/language_modeling.py
z-Runmin: I cant' move `transforms` into `__getitem__`, because `tokenizer` must be applied before `__getitem__`. Otherwise, the unit test https://github.com/pytorch/text/blob/master/test/data/test_builtin_datasets.py#L54-L57 will fail here.

Possible solution
- Move `tokenizer` out of `transforms` 
- OR: split `transforms` into two parts - global transforms / token-level transforms. 

Please advise how I can move forward. Thanks.

zhangguanheng66:
@z-Runmin Please kindly let us know if you need a review.

z-Runmin:
Not sure why unittest_windows failed to install `torchtext`

zhangguanheng66:
> Not sure why unittest_windows failed to install `torchtext`

@peterjc123 Could you take a look at here?

torchtext/experimental/datasets/language_modeling.py
```diff
@@ -26,33 +19,36 @@ class LanguageModelingDataset(torch.utils.data.Dataset):
...
                 torch.tensor([token_id_1, token_id_2, token_id_3, token_id1]).long()
             vocab: Vocabulary object used for dataset.
-
-        Examples:
-            >>> from torchtext.vocab import build_vocab_from_iterator
-            >>> data = torch.tensor([token_id_1, token_id_2,
-                                     token_id_3, token_id_1]).long()
-            >>> vocab = build_vocab_from_iterator([['language', 'modeling']])
-            >>> dataset = LanguageModelingDataset(data, vocab)
+            transforms: Text string transforms.
```
zhangguanheng66, line 35: and docs for `single_line`

torchtext/experimental/datasets/language_modeling.py
zhangguanheng66: I thought we agreed to go by `tokenizer`, `root`, `vocab` in order.

peterjc123:
> > Not sure why unittest_windows failed to install `torchtext`
> 
> @peterjc123 Could you take a look at here? The master branch is green.

@z-Runmin @zhangguanheng66 You might need to rebase your branch on master.

peterjc123:
Looks like the wmt dataset is so large that it couldn't be downloaded within the time limit (30 min).

zhangguanheng66:
> Looks like the wmt dataset is so large that it couldn't be downloaded within the time limit (30 min).

OK, then, I think we can just remove it.

PR #776, opened 2020-05-15
Move experimental text_classification functional into new functional.py

cpuhrsch:


torchtext/experimental/datasets/text_classification.py
```diff
@@ -1,43 +1,13 @@
...
-def ngrams_func(ngrams):
-    def _forward(token_list):
-        _token_list = []
-        for _i in range(ngrams + 1):
-            _token_list += zip(*[token_list[i:] for i in range(_i)])
-        return [' '.join(x) for x in _token_list]
-    return _forward
-
-
-def build_vocab(data, transforms):
```
zhangguanheng66, line 28: I'd leave `build_vocab` in the text classification file because the data is supposed to yield `label` and `txt`, which is not sure for other datasets.

PR #777, opened 2020-05-15
Add utility func to get path to assets

mthrok:
@cpuhrsch @zhangguanheng66 

It turned out that buck runs test in a directory of which full path contains a comma.
When Sentence Piece Trainer is given such a path, it considers there are multiple of paths, and fails.

To workaround this, I added a helper function that facilitates path generation and optionally copy the file to temporary directory and return the path there.
Then I adopted the function to other places where it uses test assets.

### DIFF
--- a/test/common/assets.py
+++ b/test/common/assets.py
@@ -1,0 +1,32 @@
+import os
+import shutil
+import atexit
+import tempfile
+from pathlib import Path
+
+_ASSET_DIR = (Path(__file__).parent.parent / "asset").resolve()
+
+_TEMP_DIR = None
+
+
+def _init_temp_dir():
+    """Initialize temporary directory and register clean up at the end of test."""
+    global _TEMP_DIR
+    _TEMP_DIR = tempfile.TemporaryDirectory()  # noqa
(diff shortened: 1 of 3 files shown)


test/common/assets.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def _init_temp_dir():
+    """Initialize temporary directory and register clean up at the end of test."""
+    global _TEMP_DIR
+    _TEMP_DIR = tempfile.TemporaryDirectory()  # noqa
+    atexit.register(_TEMP_DIR.cleanup)
+
+
+def get_asset_path(*path_components, use_temp_dir=False):
```
cpuhrsch, line 19: You *could* make this a context ```with get_assest``` that'll then also cleans up the temp file if it goes out of scope (if that is intended).

mthrok: Yes, that's the typical approach, but that way user of this function has to be aware of temporary files and directories. This way, they are abstracted out and users need not care.

mthrok:
thanks!

PR #778, opened 2020-05-15
Add __call__ to Vocab class

cpuhrsch:


cpuhrsch:
@zhangguanheng66 - I was relying on the dataset tests for the experimental datasets, but you're right. I'll add an explicit test.

erip:
This seems safe to close since Vocabs are now `nn.Module`s with `__call__` and `forward`, cc @parmeet

PR #779, opened 2020-05-15
Separate test that require network access in test/data/test_field.py

mthrok:
Separating `TestNestedField.test_build_vocab` which requires network access, from the rest so that we can exclude it when running test with `buck`.


mthrok:
Going to merge with #780

zhangguanheng66:
Can we combine those tests in a separate file and exclude them from `buck`?

PR #780, opened 2020-05-15
Separate test that require network in test/data/test_builtin_datasets

mthrok:
Separating `TestDataset.test_imdb` which requires network access, from the rest so that we can exclude it when running test with buck.

### DIFF
--- a/test/data/test_builtin_datasets_imdb.py
+++ b/test/data/test_builtin_datasets_imdb.py
@@ -1,0 +1,28 @@
+import torch
+from torch.testing import assert_allclose
+
+from ..common.torchtext_test_case import TorchtextTestCase
+
+
+class TestDataset(TorchtextTestCase):
+    def test_imdb(self):
+        # This test requires network access if no cache is found in `.data` dir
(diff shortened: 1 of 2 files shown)


test/data/test_builtin_datasets_imdb.py
```diff
@@ -0,0 +1,28 @@
+import torch
+from torch.testing import assert_allclose
+
+from ..common.torchtext_test_case import TorchtextTestCase
+
+
+class TestDataset(TorchtextTestCase):
```
zhangguanheng66, line 7: You may want to change the name of the class

mthrok:
> Are we going to end up many individual files for each builtin datasets that requires network?

Good point, let me merge them.

cpuhrsch:
Is this still in flight or can it be merged?

mthrok:
> Is this still in flight or can it be merged?

@cpuhrsch 

Let me rework on this one. There are 21 tests (out of 98) that has to be isolated, so I will put them all in the same file.

PR #782, opened 2020-05-18
Do not skip sacremoses tokenizer test

mthrok:
Seems like the issue had been resolved. https://github.com/alvations/sacremoses/issues/61

Test run time: ~= 0.4 second.


mthrok:
thanks!

PR #783, opened 2020-05-19
Separate tests which require external resource

mthrok:
Tests that require dataset (either via network access or cache) as following are moved to `test/test_build.py`.

```
test_vocab_vectors_custom_cache (test.test_vocab.TestVocab)
test_vocab_extend (test.test_vocab.TestVocab)
test_vocab_download_glove_vectors (test.test_vocab.TestVocab)
test_vocab_download_fasttext_vectors (test.test_vocab.TestVocab)
test_vocab_download_custom_vectors (test.test_vocab.TestVocab)
test_vocab_download_charngram_vectors (test.test_vocab.TestVocab)
test_vectors_get_vecs (test.test_vocab.TestVocab)
test_get_tokenizer_spacy (test.data.test_utils.TestUtils)
test_json_dataset_one_key_multiple_fields (test.data.test_dataset.TestDataset)
test_csv_file_no_header_one_col_multiple_fields (test.data.test_dataset.TestDataset)
test_build_vocab (test.data.test_field.TestNestedField)
```

In addition to that, the following test modules contain only such tests. so they can simply exclude in `fbcode`.
- `test/test_utils.py` (unchanged)
- `test/data/test_subword.py` (unchanged)

PR #784, opened 2020-05-19
Move `test_get_tokenizer_moses` to test_build

mthrok:
Follow up of #783 . `sacremoses` is not available in fbcode. This is the last piece.


mthrok:
Thanks!

PR #786, opened 2020-05-20
Fix csrc header for autodeps

mthrok:
Fix for `fbcode autodeps`.
Although the header file `sentencepiece_processor.h` is included by `sentencepiece_trainer.h`, without explicitly including `sentencepiece_processor.h`, `autodeps` cannot find the dependencies properly.


mthrok:
thanks!

PR #788, opened 2020-05-25
experimental vocab with JIT support

zhangguanheng66:


zhangguanheng66:
@peterjc123 Could you take a look at the Windows failure? Appreciate if you could give me some hints.

peterjc123:
There is some problem with the new Windows image on CircleCI. Just change the tag from stable to temporary and it will be fine. Related PR in pytorch https://github.com/pytorch/pytorch/pull/38909.

torchtext/csrc/vocab.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+  std::string itos(const int64_t &input) const {
+    return _vocab_index_vec[input];
+  }
+};
+
+static auto textvocab =
+    torch::class_<TextVocab>("torchtext", "TextVocab")
+        .def(torch::init<std::vector<std::string>, std::string>())
+        .def("stoi", &TextVocab::stoi)
```
cpuhrsch, line 44: I wonder if we want to get rid of these method names and instead rely on the __call__ operator. Maybe it's possible to split this into two objects. One translates tokens to id, the other represents the corresponding inverse function. For the Vocab class as it currently stands I think we also allocate memory for both even if only one is needed. Here you're using the search.

zhangguanheng66: This is just a prototype to show @nayef211 how the torch binding works. Let's discuss some API design for the milestone 1.

PR #794, opened 2020-05-26
Update smoke test to use image from Docker hub

mthrok:


.circleci/config.yml
```diff
@@ -259,13 +256,10 @@ jobs:
...
           no_output_timeout: "1h"
           command: |
-            set +x
-            export AWS_ACCESS_KEY_ID=${ECR_AWS_ACCESS_KEY}
-            export AWS_SECRET_ACCESS_KEY=${ECR_AWS_SECRET_ACCESS_KEY}
-            eval $(aws ecr get-login --region us-east-1 --no-include-email)
             set -x
-            docker tag ${image_name}:${CIRCLE_WORKFLOW_ID} 308535385114.dkr.ecr.us-east-1.amazonaws.com/${image_name}:${CIRCLE_WORKFLOW_ID}
-            docker push 308535385114.dkr.ecr.us-east-1.amazonaws.com/${image_name}:${CIRCLE_WORKFLOW_ID}
+            docker tag ${image_name}:${CIRCLE_WORKFLOW_ID} ${image_name}:latest
```
seemethere, line 260: Oh we really shouldn't be using docker images here as intermediary storage.

I'd prefer it if we just removed this and used circleci workspaces.

Here's some documentation on how to do that: https://circleci.com/blog/deep-diving-into-circleci-workspaces/

mthrok: @seemethere or can we just do build and push in one step?

mthrok: @seemethere I merged the steps so that there is no need to pass data between steps.

seemethere: Oh I see, this build would still fail though since you don't actually have credentials in order to push these images, let me see if I can resolve that for you.

.circleci/config.yml
```diff
@@ -251,21 +248,14 @@ jobs:
...
-          no_output_timeout: "1h"
-          command: |
-            set +x
-            export AWS_ACCESS_KEY_ID=${ECR_AWS_ACCESS_KEY}
-            export AWS_SECRET_ACCESS_KEY=${ECR_AWS_SECRET_ACCESS_KEY}
-            eval $(aws ecr get-login --region us-east-1 --no-include-email)
             set -x
-            docker tag ${image_name}:${CIRCLE_WORKFLOW_ID} 308535385114.dkr.ecr.us-east-1.amazonaws.com/${image_name}:${CIRCLE_WORKFLOW_ID}
-            docker push 308535385114.dkr.ecr.us-east-1.amazonaws.com/${image_name}:${CIRCLE_WORKFLOW_ID}
+            cd .circleci/smoke_test/docker && docker build . -t ${image_name}:${CIRCLE_WORKFLOW_ID}
```
seemethere, line 255:
```suggestion
            echo "${DOCKER_HUB_TOKEN}" | docker login --username "${DOCKER_HUB_USERNAME}" --password-stdin
            cd .circleci/smoke_test/docker && docker build . -t ${image_name}:${CIRCLE_WORKFLOW_ID}
```

seemethere: To actually use these variables you'll need to also add the `org-member` security context to the build job for these.

mthrok: Applied the suggestions.

mthrok:
@seemethere Merging this to test the smoke test. Let me know if you have further follow ups.

PR #795, opened 2020-05-27
Improve error message when extension is not found

mthrok:


torchtext/__init__.py
```diff
@@ -32,7 +32,7 @@ def _init_extension():
     extfinder = importlib.machinery.FileFinder(lib_dir, loader_details)
     ext_specs = extfinder.find_spec("_torchtext")
     if ext_specs is None:
-        raise ImportError
+        raise ImportError("Extension _torchtext is not found.")
```
zhangguanheng66, line 35: Should we be more explicit?

mthrok: I am open to your suggestion, but if the extension is not found then it most likely means that there is something wrong with installation. Not sure what is the better message to give out.

While I was working on building `torchtext` on fbcode, I encountered this and raising bare `ImportError` without a message was more confusing because I could not tell which module is causing `ImportError`. so as long as it is clear that it's coming from `torchtext`, I am okay with any message.

zhangguanheng66: How about "cpp extension is not found"? something similar.

mthrok:
Rebased to see the effect of #796

mthrok:
Thanks!

PR #796, opened 2020-05-27
Use new circleci image for Windows jobs

peterjc123:
The old image is not available anymore.


peterjc123:
cc @zhangguanheng66 @mthrok

mthrok:
Thanks!

Issue #760, opened 2020-05-08
Is the `.data` directory supposed to be auto-created when using the experimental datasets?


thvasilo:
I just tried running the following on a fresh instance:

```python
import torch
import torchtext

# import datasets
from torchtext.experimental.datasets import IMDB

SEED = 11

torch.manual_seed(SEED)
torch.backends.cudnn.deterministic = True

# use the default tokenizer
train_dataset, test_dataset = IMDB()
vocab = train_dataset.get_vocab()
```

and got the error `RuntimeError: Download directory .data does not exist. Did you create it?
`

I expected that the library would create the directory on its own, like it does for the `.vector_cache` directory. Is that not the case?

zhangguanheng66:
Fixed by #797

PR #797, opened 2020-05-28
Create root directory automatically for download_from_url if not exists

z-Runmin:
Address https://github.com/pytorch/text/issues/760

### DIFF
--- a/torchtext/utils.py
+++ b/torchtext/utils.py
@@ -78,12 +78,11 @@ def download_from_url(url, path=None, root='.data', overwrite=False):
         root, filename = os.path.split(path)
 
     if not os.path.exists(root):
-        if root == ".data":
-            os.makedirs(".data")
-        else:
+        try:
+            os.makedirs(root)
+        except:
             raise RuntimeError(
-                "Download directory {} does not exist. "
-                "Did you create it?".format(root))
+                "Can't create the download directory {}.".format(root))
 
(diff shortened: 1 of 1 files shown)


torchtext/utils.py
```diff
@@ -78,9 +78,12 @@ def _process_response(r, root, filename):
         root, filename = os.path.split(path)
 
     if not os.path.exists(root):
-        raise RuntimeError(
-            "Download directory {} does not exist. "
-            "Did you create it?".format(root))
+        if root == ".data":
```
zhangguanheng66, line 81: should we use `try`/`except`? And `root` could be a user-defined directory.

z-Runmin: Sure. Update in 091e263

torchtext/utils.py
```diff
@@ -78,12 +78,11 @@ def _process_response(r, root, filename):
         root, filename = os.path.split(path)
 
     if not os.path.exists(root):
-        if root == ".data":
-            os.makedirs(".data")
-        else:
+        try:
+            os.makedirs(root)
+        except:
```
zhangguanheng66, line 83: To be save, `except` comes with some error message. I guess you can try `except OSError:` but please double check.

z-Runmin: Yes, I switched to the default `OSError`

PR #798, opened 2020-05-28
Fix test_generate_sp_model for stress test

mthrok:
When multiple instances of test_generate_sp_model run,
intermediate model data have to be unique and properly cleaned up.

See also T67599333

### DIFF
--- a/test/common/assets.py
+++ b/test/common/assets.py
@@ -1,32 +1,8 @@
-import os
-import shutil
-import atexit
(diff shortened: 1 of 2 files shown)


mthrok:
> LGTM. Just make sure the templfile is cleaned up properly for both internal and external tests, since we don't `os.remove` here anymore.

Yup, the context manager does it for you.

PR #801, opened 2020-05-28
Fix pip install in smoke test

mthrok:


mthrok:
merging this as it is required for testing nightly

Issue #791, opened 2020-05-26
Regular Expression not JIT-able


zhangguanheng66:
## 🚀 Feature
The `re` library in python is not JIT-able, which is used by the tokenizers in torchtext. 

**Motivation**

With a JIT-able `re` library, the `basic_english` tokenizer will have JIT support.

```
import torch
import torchtext
from typing import List
import re

def regex_tokenizer(line: str) -> List[str]:
	line = line.lower()	
	line = re.compile(r'\"').sub('', line)
	line = re.compile(r'\(').sub(' ( ', line)
	return line.split()
	
jit_tokenizer = torch.jit.script(regex_tokenizer)
```

Error:
```
UnsupportedNodeError: try blocks aren't supported:
  File "/private/home/zhangguanheng/anaconda3/lib/python3.7/re.py", line 275
    if isinstance(flags, RegexFlag):
        flags = flags.value
    try:
    ~~~ <--- HERE
        return _cache[type(pattern), pattern, flags]
    except KeyError:
'compile' is being compiled since it was called from 'regex_tokenizer'
  File "<ipython-input-1-e544d1c55cf5>", line 8
def regex_tokenizer(line: str) -> List[str]:
        line = line.lower()
        line = re.compile(r'\"').sub('', line)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
        line = re.compile(r'\(').sub(' ( ', line)
        return line.split()
```

PR #802, opened 2020-05-28
Added custom regex class to wrap C++11 regex library

Nayef211:
### Description
- Created a new c++ regex class to wrap the c++11 `std::regex` library
- Having this will allow for tokenizers, such as `basic_english` tokenizer,  which currently relies on the `re` library to also be JIT-able when using this new class 
- Created new `BasicEnglishNormalize` and `RegexTokenizer` transforms in the experimental folder using the c++ regex implementation
- Refer to #791 for more info

### Follow Ups
- Uncomment more complex test cases once [#38207](https://github.com/pytorch/pytorch/issues/38207) gets resolved


.gitignore
```diff
@@ -121,4 +121,7 @@ venv.bak/
 *.swp
 *.swo
 
-torchtext/version.py
\ No newline at end of file
+torchtext/version.py
+
+# deleted files
+torchtext/.nfs*
```
cpuhrsch, line 127: Is this from an editor?

torchtext/experimental/functional.py
```diff
@@ -1,4 +1,29 @@
...
+    Args:
+        line (str): a line of text to tokenize.
+        patterns_list (List[Tuple[str, str]]): a list of tuples (ordered pairs) which contain the regex pattern string as the first element and the replacement string as the second element.
+
+    Returns:
+        tokens (List[str]): a list of tokens after splitting on whitespace.
+    """
+
+    line = line.lower()
+    regex = torch.classes.torchtext.Regex('')
```
cpuhrsch, line 21: I'd guess that creating this object is example, because the regex needs to be compiled.

torchtext/experimental/functional.py
```diff
@@ -1,4 +1,29 @@
...
+    for pattern_re, replaced_str in patterns_list:
+        regex.UpdateRe(pattern_re)
```
cpuhrsch, line 23: We could have a factory function that takes a list of regex and returns a list of Regex objects instead of this. Or maybe there's a way of writing a single regex from this list?

torchtext/experimental/transforms.py
```diff
@@ -0,0 +1,81 @@
...
+        self.patterns_list = patterns_list
+
+    def forward(self, line: str) -> List[str]:
+        r"""
+        Args:
+            line (str): a line of text to tokenize.
+        Returns:
+            List[str]: a list of tokens after normalizing and splitting on whitespace.
+        """
+        return F.regex_tokenizer(line, self.patterns_list)
```
cpuhrsch, line 81: This functional creates state on every invocation (the Regex object) - why could create that Regex object during init here and then feed into the functional at each time.

test/data/test_functional.py
```diff
@@ -74,6 +83,43 @@ def test_sentencepiece_tokenizer(self):
         self.assertEqual(list(spm_generator([test_sample]))[0],
                          ref_results)
 
+    def test_BasicEnglishNormalize(self):
```
cpuhrsch, line 83: Do you have a test that tries to jit this?

torchtext/experimental/transforms.py
```diff
@@ -0,0 +1,81 @@
...
+
+class RegexTokenizer(nn.Module):
+    r"""Regex tokenizer for a string sentence.
+
+    Steps include
+    - lowercasing
+    - applying all regex replacements defines in patterns_list
+
+    Args:
+        patterns_list (List[Tuple[str, str]]): a list of tuples (ordered pairs) which contain the regex pattern string as the first element and the replacement string as the second element.
```
zhangguanheng66, line 67: You can include an example to use the func/class, like this
https://github.com/pytorch/text/blob/e70955309ead681f924fecd36d759c37e3fdb1ee/torchtext/experimental/datasets/text_classification.py#L137-L143

test/data/test_functional.py
Nayef211: I actually simplified my test case and added the jit test which is fully working. Once the PR fixing the string split lands, I will make my test case more complicated.

.gitignore
Nayef211: I removed this. I was using it to ignore some .nfs files that were polluting the directory

torchtext/csrc/regex.cpp
```diff
@@ -0,0 +1,48 @@
...
+public:
+  // re_str_ holds the serialized regex string passed at the initialization.
+  // We need this because we need to be able to serialize the model so that we
+  // can save the scripted object. pickle will get the
+  // serialized model from this re_str_ member, thus it needs to be public.
+  std::string re_str_;
+
+  Regex(const std::string &re_str) { UpdateRe(re_str_); }
+
+  void UpdateRe(const std::string &re_str) {
```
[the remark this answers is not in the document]
Nayef211: Discussed about this offline. Seems simpler to create one object and update the regex pattern instead of creating an arbitrary number of objects using a factory method.

torchtext/experimental/functional.py
Nayef211: Moved this to a class attribute instead that's being passed in

torchtext/experimental/functional.py
Nayef211: Refer to above comment

torchtext/experimental/transforms.py
```diff
@@ -0,0 +1,105 @@
...
+        >>> test_sample = 'Basic Regex Tokenization for a Line of Text'
+        >>> patterns_list = [
+            (r'\'', ' \'  '),
+            (r'\"', '')]
+        >>> regex_tokenizer = RegexTokenizer(patterns_list)
+        >>> jit_regex_tokenizer = torch.jit.script(regex_tokenizer)
+        >>> tokens = jit_regex_tokenizer.forward(test_sample)
+    """
+
+    patterns_list: List[Tuple[str, str]]
```
zhangguanheng66, line 90: why do we need this?

Nayef211: I believe this is the new API for typing attributes in Pytorch1.2. Refer to https://pytorch.org/docs/stable/jit.html#attributes

torchtext/experimental/transforms.py
```diff
@@ -0,0 +1,105 @@
...
+    Examples:
+        >>> import torch
+        >>> from torchtext.experimental.transforms import RegexTokenizer
+        >>> test_sample = 'Basic Regex Tokenization for a Line of Text'
+        >>> patterns_list = [
+            (r'\'', ' \'  '),
+            (r'\"', '')]
+        >>> regex_tokenizer = RegexTokenizer(patterns_list)
+        >>> jit_regex_tokenizer = torch.jit.script(regex_tokenizer)
+        >>> tokens = jit_regex_tokenizer.forward(test_sample)
```
zhangguanheng66, line 87: Do you need to call `.forward` function? You should be able to call `jit_regex_tokenizer(test_sample)` directly. Please check it as it works for most nn.Module class.

torchtext/experimental/transforms.py
```diff
@@ -0,0 +1,105 @@
...
+            - replace ':' with single space
+            - replace multiple spaces with single space
+
+    Examples:
+        >>> import torch
+        >>> from torchtext.experimental.transforms import BasicEnglishNormalize
+        >>> test_sample = 'Basic English Normalization for a Line of Text'
+        >>> basic_english_normalize = BasicEnglishNormalize()
+        >>> jit_basic_english_normalize = torch.jit.script(basic_english_normalize)
+        >>> tokens = jit_basic_english_normalize.forward(test_sample)
```
zhangguanheng66, line 38: Do you need to call `.forward` function? You should be able to call `jit_regex_tokenizer(test_sample)` directly. Please check it as it works for most `nn.Module` class.

torchtext/experimental/transforms.py
Nayef211: Done

torchtext/experimental/transforms.py
Nayef211: Done

docs/source/experimental_datasets.rst
```diff
@@ -1,5 +1,5 @@
 torchtext.experimental.datasets
-================================
+===============================
```
cpuhrsch, line 2: Why is this changing?

test/data/test_functional.py
```diff
@@ -74,6 +80,96 @@ def test_sentencepiece_tokenizer(self):
         self.assertEqual(list(spm_generator([test_sample]))[0],
                          ref_results)
 
+    # TODO(Nayef211): uncomment and replace the test below with this once once
```
cpuhrsch, line 83: nit: "once once"

test/data/test_functional.py
```diff
@@ -74,6 +80,96 @@ def test_sentencepiece_tokenizer(self):
...
+
+    #     jit_regex_tokenizer = torch.jit.script(regex_tokenizer)
+    #     jit_tokens = jit_regex_tokenizer(test_sample)
+
+    #     self.assertEqual(jit_tokens, ref_results)
+    #     self.assertEqual(jit_tokens, eager_tokens)
+
+    def test_RegexTokenizer(self):
+        test_sample = 'Basic Regex Tokenization for a Line of Text'
+        ref_results = ['Basic', 'Regex', 'Tokenization',
```
cpuhrsch, line 148: This isn't really engaging any of the patterns in the list

torchtext/experimental/transforms.py
cpuhrsch: You'll still need to update the comment

torchtext/experimental/functional.py
```diff
@@ -1,4 +1,33 @@
...
+        line (str): a line of text to tokenize.
+        patterns_list (List[Tuple[str, str]]): a list of tuples (ordered pairs) which contain the regex pattern string
+            as the first element and the replacement string as the second element.
+        regex (torch.classes.torchtext.Regex): a custom c++ Regex object.
+        lowercase (bool, optional): a flag to determine whether to lowercase the input text.
+    Returns:
+        tokens (List[str]): a list of tokens after splitting on whitespace.
+    """
+
+    if lowercase:
```
cpuhrsch, line 23: Why do we need this as part of this?

torchtext/experimental/transforms.py
Nayef211: Should be fixed now

docs/source/experimental_datasets.rst
Nayef211: basically sphinx was complaining during `make html`. in `rst`, the title (in this case accomplished with `=` sign) needs to be the same length or greater than the text above it

test/data/test_functional.py
Nayef211: Fixed

torchtext/experimental/functional.py
Nayef211: removed this function as part of our discussion

test/data/test_functional.py
Nayef211: you're right. let me update this test

torchtext/experimental/transforms.py
```diff
@@ -0,0 +1,114 @@
...
+            (r'\)', ' ) '),
+            (r'\!', ' ! '),
+            (r'\?', ' ? '),
+            (r'\;', ' '),
+            (r'\:', ' '),
+            (r'\s+', ' ')]
+
+        regex_objects = map(lambda pattern_tuple: torch.classes.torchtext.Regex(pattern_tuple[0]), patterns_list)
+        replacement_strings = map(lambda pattern_tuple: pattern_tuple[1], patterns_list)
+        self.regex_and_replacement_string_pairs = list(zip(regex_objects, replacement_strings))
```
cpuhrsch, line 60: nit: You don't need to store this imo, you could just do

```
for regex, replacement_string in zip(regex_objects, replacement_strings):
```
below

torchtext/experimental/transforms.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+
+    regex_and_replacement_string_pairs: List[Tuple[torch.classes.torchtext.Regex, str]]
+
+    def __init__(self, patterns_list: List[Tuple[str, str]]):
+        super(RegexTokenizer, self).__init__()
+
+        regex_objects = map(lambda pattern_tuple: torch.classes.torchtext.Regex(pattern_tuple[0]), patterns_list)
+        replacement_strings = map(lambda pattern_tuple: pattern_tuple[1], patterns_list)
+        self.regex_and_replacement_string_pairs = list(zip(regex_objects, replacement_strings))
```
cpuhrsch, line 102: Same here

cpuhrsch:
cc @mthrok for reference around discussion for C++ extensions in general

torchtext/experimental/transforms.py
Nayef211: so I wasn't actually storing `regex_objects` or `replacement_strings ` as attributes. since this tokenizer could possibly get called many times for different lines of text, wouldn't it make sense to store `regex_and_replacement_string_pairs` as an attribute so that we don't have to call `zip`? Also avoids the issue of storing 2 different lists when logically the lists are related and should be stored as a list of ordered pairs. LMK what you think

torchtext/experimental/transforms.py
Nayef211: look at comment above ^

torchtext/experimental/transforms.py
cpuhrsch: There really shouldn't be any performance penalties, but it's a minor thing and indeed a thing you can avoid if you set it up like this.

PR #803, opened 2020-05-28
Install conda package from workspace

mthrok:
When running smoke test, proper way to install conda package is to install from workspace, instead of installing from archive. https://github.com/conda/conda/issues/1884 Otherwise the dependencies are not properly installed https://app.circleci.com/pipelines/github/pytorch/text/340/workflows/1a453eb2-4480-4f49-babf-34e5230697f2/jobs/4401/steps .

For this, we pass the whole `conda-bld` directory from conda package build job to upload job then smoke test job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,9 +80,9 @@ jobs:
       - store_artifacts:
           path: /opt/conda/conda-bld/linux-64
       - persist_to_workspace:
-          root: /opt/conda/conda-bld/linux-64
+          root: /opt/conda
           paths:
-            - "*"
+            - "conda-bld/*"
 
   binary_windows_wheel:
     <<: *binary_common
@@ -120,9 +120,9 @@ jobs:
       - store_artifacts:
           path: C:/tools/miniconda3/conda-bld/win-64
       - persist_to_workspace:
-          root: C:/tools/miniconda3/conda-bld/win-64
+          root: C:/tools/miniconda3
           paths:
-            - "*"
+            - "conda-bld/*"
 
   binary_macos_wheel:
     <<: *binary_common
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -214,8 +214,7 @@ jobs:
           command: |
             set -x
             source /usr/local/etc/profile.d/conda.sh && conda activate python${PYTHON_VERSION}
-            conda install -v -y -c pytorch-nightly pytorch
-            conda install -v -y $(ls ~/workspace/torchtext*.tar.bz2)
+            conda install -v -y -c pytorch-nightly pytorch file://$HOME/workspace/conda-bld
```
seemethere, line 217:
```suggestion
            conda install -v -y -c pytorch-nightly -c file://$HOME/workspace/conda-bld torchtext
```

With the `file://` local repository it should automatically download and install `pytorch` since it's listed as a dependency.

Note this will actually only install the local package if the version we are installing locally is greater than the latest version found in the `pytorch-nightly` channel

mthrok: > Note this will actually only install the local package if the version we are installing locally is greater than the latest version found in the pytorch-nightly channel

Can `torchtext` in `pytorch-nightly` channel have the same version as `local build`?

Instead of installing `pytorch` and `torchtext` at the same time, will installing them separately, enabling each channel separately make sure that local `torchtext` is installed? 

i.e.

```
conda install -v -y -c pytorch-nightly pytorch
conda install -v -y -c file://$HOME/workspace/conda-bld torchtext
```

seemethere: Yes that could work, I only suggested doing it all in one command because of convenience.

> Can torchtext in pytorch-nightly channel have the same version as local build?

That's a good question, I'm not entirely sure how conda would prioritize certain channels over another. Doing the two step process is probably preferable if you're expecting them to potentially have the same versions.

mthrok: Sounds good. Updated.

mthrok:
Thanks!

mthrok:
Nightly build & smoke test worked ! https://app.circleci.com/pipelines/github/pytorch/text/345/workflows/a12396a9-9800-460c-96bf-738a6259fb66

PR #805, opened 2020-06-04
Experimental sequence tagging datasets

akurniawan:
Adding new abstraction for sequence tagging datasets with new API


torchtext/experimental/datasets/raw/sequence_tagging.py
```diff
@@ -0,0 +1,165 @@
...
+                yield item
+            if (self.num_lines is not None) and (i == (self.start +
+                                                       self.num_lines)):
+                break
+
+    def get_iterator(self):
+        return self._iterator
+
+
+def UDPOS(train_filename="en-ud-tag.v2.train.txt",
```
zhangguanheng66, line 109: I'm not familiar with the tagging datasets. But in default, we generate all the train/test/valid sets with the raw datasets.

torchtext/experimental/datasets/sequence_tagging.py
```diff
@@ -0,0 +1,220 @@
...
+                data_list[idx].append(col)
+
+    for it in data_list:
+        vocabs.append(build_vocab_from_iterator(it))
+
+    return vocabs
+
+
+def _setup_datasets(dataset_name,
+                    train_filename,
```
zhangguanheng66, line 32: Do we need to specify those filenames?

torchtext/experimental/datasets/sequence_tagging.py
```diff
@@ -0,0 +1,220 @@
...
+                                              valid_filename=valid_filename,
+                                              test_filename=test_filename,
+                                              root=root)
+    raw_data = {
+        "train": [line for line in train] if train else None,
+        "valid": [line for line in val] if val else None,
+        "test": [line for line in test] if test else None
+    }
+
+    word_transform = None
```
zhangguanheng66, line 50: nit: to keep consistent, I would call `word_transform` as `text_transform`

torchtext/experimental/datasets/sequence_tagging.py
```diff
@@ -0,0 +1,220 @@
...
+    if vocabs is None:
+        if "train" not in data_select:
+            raise TypeError("Must pass a vocab if train is not selected.")
+        vocabs = _build_vocab(raw_data["train"], word_transform)
+    else:
+        if not isinstance(vocabs, list):
+            raise TypeError("vocabs must be an instance of list")
+
+        # Find data that's not None
+        notnone_data = None
```
zhangguanheng66, line 63: what's this part?

torchtext/experimental/datasets/sequence_tagging.py
```diff
@@ -0,0 +1,220 @@
...
+            By default, all the three datasets (train, test, valid) are generated. Users
+            could also choose any one or two of them, for example ('train', 'test') or
+            just a string 'train'. If 'train' is not in the tuple or string, a vocab
+            object should be provided which will be used to process valid and/or test
+            data.
+        root: Directory where the datasets are saved. Default: ".data"
+        vocabs: A list of voabularies for each columns in the dataset. Must be in an
+            instance of List
+            Default: None
+        word_tokenizer: The tokenizer used to preprocess word column in raw text data
```
zhangguanheng66, line 158: `tokenizer`?

torchtext/experimental/datasets/sequence_tagging.py
```diff
@@ -0,0 +1,220 @@
...
+            Default: None
```
zhangguanheng66, line 159: To keep consistent, we have `root`, `vocab`, `tokenizer`, `data_select` in order.

torchtext/experimental/datasets/sequence_tagging.py
```diff
@@ -0,0 +1,220 @@
...
+                                              test_filename=test_filename,
+                                              root=root)
+    raw_data = {
+        "train": [line for line in train] if train else None,
+        "valid": [line for line in val] if val else None,
+        "test": [line for line in test] if test else None
+    }
+
+    word_transform = None
+    if word_tokenizer:
```
zhangguanheng66, line 51: What happen if `tokenizer` is None?

torchtext/experimental/datasets/sequence_tagging.py
```diff
@@ -0,0 +1,220 @@
...
+          data_select=("train", "valid", "test"),
+          root=".data",
+          vocabs=None,
+          word_tokenizer=None):
+    """ Universal Dependencies English Web Treebank
+
+    Separately returns the training and test dataset
+
+    Arguments:
+        train_filename: Filename for training dataset.
```
zhangguanheng66, line 141: Same like the above, do we need those filenames?

test/data/test_builtin_datasets.py
```diff
@@ -122,3 +122,61 @@ def test_imdb(self):
         old_vocab = train_dataset.get_vocab()
         new_vocab = Vocab(counter=old_vocab.freqs, max_size=2500)
         new_train_data, new_test_data = IMDB(vocab=new_vocab)
+
+    def test_sequence_tagging(self):
```
zhangguanheng66, line 126: If the CI tests for datasets don't take a lot of time to run, we should include tests to cover both datasets: `UDPOS` and `CoNLL2000Chunking`.

torchtext/experimental/datasets/sequence_tagging.py
akurniawan: this is the sample of the data
```
Al	PROPN	NNP
-	PUNCT	HYPH
Zaman	PROPN	NNP
:	PUNCT	:
American	ADJ	JJ
forces	NOUN	NNS
killed	VERB	VBD
Shaikh	PROPN	NNP
Abdullah	PROPN	NNP
al	PROPN	NNP
-	PUNCT	HYPH
```

we don't need a tokenizer as per usual as the words themselves are already tokenized since the beginning, so I believe setting up a default tokenizer wouldn't make sense in this case

torchtext/experimental/datasets/sequence_tagging.py
akurniawan: sure, please see the reasoning below on why the tokenizer is set as `None`

torchtext/experimental/datasets/sequence_tagging.py
akurniawan: removed

torchtext/experimental/datasets/sequence_tagging.py
akurniawan: removed

torchtext/experimental/datasets/sequence_tagging.py
akurniawan: changed

test/data/test_builtin_datasets.py
akurniawan: done

torchtext/experimental/datasets/sequence_tagging.py
akurniawan: I'm adding this part to validate whether the vocabularies passed through the argument has similar length with the number of colums from the dataset.
If we look from the example as follows
```
Al	PROPN	NNP
-	PUNCT	HYPH
Zaman	PROPN	NNP
:	PUNCT	:
American	ADJ	JJ
forces	NOUN	NNS
killed	VERB	VBD
Shaikh	PROPN	NNP
Abdullah	PROPN	NNP
al	PROPN	NNP
-	PUNCT	HYPH
```
it would be problematic if the given vocabularies (from argument) do not cover every columns

torchtext/experimental/datasets/sequence_tagging.py
zhangguanheng66: If we don't need a tokenizer for this dataset, should we just delete them?

akurniawan: I was thinking to keep them so that it allows user to try different tokenizer for the word, such as adding capability to use char level embedding instead of word. Or do you have other way to achieve similar results?

akurniawan: After some testing it seems to use char level embedding is a bit hard, might be better to use transform function instead. Still not pretty sure so I just remove the tokenizer for now

test/data/test_builtin_datasets.py
```diff
@@ -157,6 +157,108 @@ def test_multi30k(self):
...
 
+    def test_udpos_sequence_tagging(self):
+        from torchtext.experimental.datasets import UDPOS
+
+        # smoke test to ensure imdb works properly
+        train_dataset, valid_dataset, test_dataset = UDPOS()
+        self.assertEqual(len(train_dataset), 12543)
+        self.assertEqual(len(valid_dataset), 2002)
+        self.assertEqual(len(test_dataset), 2077)
+        assert_allclose(train_dataset[0][0][:10],
```
zhangguanheng66, line 168: Just to bring your attention. We have a PR to support `self.assertEqual` for torch.tensor. https://github.com/pytorch/text/pull/822

torchtext/experimental/datasets/raw/sequence_tagging.py
```diff
@@ -0,0 +1,134 @@
...
+    "UDPOS":
+    'https://bitbucket.org/sivareddyg/public/downloads/en-ud-v2.zip',
+    "CoNLL2000Chunking": [
+        'https://www.clips.uantwerpen.be/conll2000/chunking/train.txt.gz',
+        'https://www.clips.uantwerpen.be/conll2000/chunking/test.txt.gz'
+    ]
+}
+
+
+def _create_data_from_iob(data_path, separator="\t"):
```
zhangguanheng66, line 15: Do we return or yield something from this func?

torchtext/experimental/datasets/sequence_tagging.py
```diff
@@ -0,0 +1,155 @@
...
+    Currently, we only support the following datasets:
+        - UDPOS
+        - CoNLL2000Chunking
+    """
+    def __init__(self, data, vocab, transforms):
+        """Initiate sequence tagging dataset.
+        Arguments:
+            data: a list of word and its respective tags. Example:
+                [[word, POS, dep_parsing label, ...]]
+            vocab: Vocabulary object used for dataset.
```
zhangguanheng66, line 83: Should we use `vocabs` and explicitly require a same number of vocabs and columns?

torchtext/experimental/datasets/sequence_tagging.py
```diff
@@ -0,0 +1,155 @@
...
+            transforms: a list of string transforms for words and tags.
+        """
+
+        super(SequenceTaggingDataset, self).__init__()
+        self.data = data
+        self.vocab = vocab
+        self.transforms = transforms
+
+    def __getitem__(self, i):
+        line = []
```
zhangguanheng66, line 93: nit: can this be simplified as `return [self.transforms[idx](self.data[i][idx]) for idx in range(self.transforms)]?

torchtext/experimental/datasets/sequence_tagging.py
```diff
@@ -0,0 +1,155 @@
...
+            vocab: Vocabulary object used for dataset.
+            transforms: a list of string transforms for words and tags.
+        """
+
+        super(SequenceTaggingDataset, self).__init__()
+        self.data = data
+        self.vocab = vocab
+        self.transforms = transforms
+
+    def __getitem__(self, i):
```
zhangguanheng66, line 92: Should we check that the data[i] and transforms have same number of columns here?

torchtext/experimental/datasets/sequence_tagging.py
```diff
@@ -0,0 +1,155 @@
...
+
+    return vocabs
+
+
+def _setup_datasets(dataset_name,
+                    root=".data",
+                    vocabs=None,
+                    data_select=("train", "valid", "test")):
+    train, val, test = DATASETS[dataset_name](root=root)
+    raw_data = {
```
zhangguanheng66, line 32: you might add some checks here
https://github.com/pytorch/text/blob/2aa8858272d9b46b71194616ab948a087081207f/torchtext/experimental/datasets/text_classification.py#L80-L83

akurniawan: done

torchtext/experimental/datasets/sequence_tagging.py
akurniawan: done

torchtext/experimental/datasets/sequence_tagging.py
akurniawan: done

torchtext/experimental/datasets/sequence_tagging.py
akurniawan: done

torchtext/experimental/datasets/raw/sequence_tagging.py
akurniawan: yes, please take a look at line 22. due to the nature of the data, each sentences is separated with an empty line, therefore we will return one sentence if we found one. and I just add new commit to return leftovers

test/data/test_builtin_datasets.py
akurniawan: noted, will wait that PR merged firs then

zhangguanheng66: You could add [TODO] tag in the comment line and move forward with this PR.

torchtext/experimental/datasets/sequence_tagging.py
```diff
@@ -75,31 +80,44 @@ class SequenceTaggingDataset(torch.utils.data.Dataset):
...
-        self.vocab = vocab
+        self.vocabs = vocabs
         self.transforms = transforms
 
+        if len(self.data[0]) != len(self.vocabs):
+            raise ValueError("vocabs must hahve the same number of columns "
+                             "as the data")
+
+        if len(self.data[0]) != len(self.transforms):
+            raise ValueError("vocabs must hahve the same number of columns "
```
zhangguanheng66, line 106: Is this a typo? transforms?

akurniawan:
@zhangguanheng66 have already fixed all the comments. though for the codecoverage not really sure what can be added to improve the coverage

zhangguanheng66:
I just merged the testcase PR https://github.com/pytorch/text/pull/822

zhangguanheng66:
@seemethere Any ideas about the codecov failure?

seemethere:
No idea, I can't view the codecov either

zhangguanheng66:
> No idea, I can't view the codecov either

@seemethere I add a PR to remove codecov and this is consistent with the other two domain libraries. https://github.com/pytorch/text/pull/832

akurniawan:
@zhangguanheng66 thanks, the unittest have been updated with the latest patch

akurniawan:
@zhangguanheng66 done merging #832, thanks for the fix!

zhangguanheng66:
@akurniawan Thanks for the contribution.

PR #813, opened 2020-06-05
switch experimental ngrams_func to data.utils.ngrams_iterator

zhangguanheng66:


test/data/test_utils.py
```diff
@@ -50,3 +50,7 @@ def test_text_nomalize_function(self):
                 ref_lines.append(line)
 
         self.assertEqual(ref_lines, test_lines)
+
+    def test_ngrams_func(self):
+        func = ngrams_func(2)
```
cpuhrsch, line 55: We could add more tests here

PR #820, opened 2020-06-10
Creating new JITable Vectors class

Nayef211:
## Description
- Creating a new Vectors class in experimental that is JITable and more perfomant
- Using a custom c++ class for the underlying vector implementation
- Refactoring original cache function out of the Vectors class
- The vectors class defines an `unk_token` as a `torch.zeros` tensor if one isn't provided. If empty `tokens` and `vectors` are passed in without passing in a `unk_token`, a `ValueError` is raised.


torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,23 @@
+import torch
```
zhangguanheng66, line 1: You can find `download_from_url` and `extract_archive` func [Here](https://github.com/pytorch/text/blob/master/torchtext/utils.py).

torchtext/csrc/vectors.cpp
```diff
@@ -0,0 +1,59 @@
...
+  // model so that we can save the scripted object. Pickle will get the
+  // serialized model from these members, thus they needs to be public.
+  std::vector<std::string> tokens_;
+  std::vector<torch::Tensor> vectors_;
+
+  explicit Vectors(std::vector<std::string> tokens,
+                   std::vector<torch::Tensor> vectors)
+      : tokens_(tokens), vectors_(vectors) {
+    for (std::size_t i = 0; i < tokens.size(); i++) {
+      stoi_[tokens[i]] = static_cast<int>(i);
```
zhangguanheng66, line 23: Why don't we save the `vectors` directly in `stoi_`, like `std::unordered_map<std::string, std::vector<torch::Tensor>>`

Nayef211: Good call. I'll go ahead and do that. I think we still need to store the vectors in `vectors_` for the `__getstate__` and `__setstate__` methods.

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def __init__(self, tokens, vectors):
+        self.unk_tensor = torch.zeros(vectors[0].size())
+        self.vectors = torch.classes.torchtext.Vectors(tokens, vectors)
+
+    def __getitem__(self, token):
+        if self.vectors.TokenExists(token):
+            return self.vectors.GetItem(token)
+        else:
+            return self.unk_tensor
```
cpuhrsch, line 21: I'd fold this into CppVectors

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,21 @@
+import torch
+
+
+class Vectors(object):
```
cpuhrsch, line 4: file factory function (write this in Python for now) - check for support file formats.
Example uses white space separated entries
```
def vectors_from_tsv_file(path):
    tuples = map(lambda x: (x.split()[0], torch.tensor(x.split()[1:], dtype=torch.float)) open(path).read())
    words = map(lambda x: x[0], tuples)
    vectors = map(lambda x: [1], tuples)
    return Vectors(words, vectors)
```

Potential future factory functions
vectors_from_url (folds download + extract_archive if necessary)

Later on we could add a dictionary of pretrained vectors with associated URL, which is currently [folded into vocab.py](https://github.com/pytorch/text/blob/master/torchtext/vocab.py#L469).

To get vectors from the internet people could then do

vectors = vectors_from_url(torchtext.experimental.vectors.URLs['Glove'])

torchtext/experimental/vectors.py
Nayef211: Done

torchtext/experimental/vectors.py
Nayef211: Created the factory method for creating vectors to files. Lmk what you think

test/experimental/test_vectors.py
```diff
@@ -0,0 +1,62 @@
...
+)
+
+
+class TestVectors(TorchtextTestCase):
+
+    def test_has_unk(self):
+        tensorA = torch.Tensor([1, 0])
+        expected_unk_tensor = torch.Tensor([0, 0])
+
+        tokens = ['a']
```
cpuhrsch, line 18: Try initializing without any tokens at all (i.e. an empty list)

torchtext/csrc/vectors.cpp
```diff
@@ -0,0 +1,68 @@
...
+  std::vector<std::string> tokens_;
+  std::vector<torch::Tensor> vectors_;
+  torch::Tensor unk_tensor_;
+
+  explicit Vectors(std::vector<std::string> tokens,
+                   std::vector<torch::Tensor> vectors, torch::Tensor unk_tensor)
+      : tokens_(tokens), vectors_(vectors), unk_tensor_(unk_tensor) {
+    for (std::size_t i = 0; i < tokens.size(); i++) {
+      // guarding against size mismatch of vectors and tokens
+      if (i >= vectors.size()) {
```
cpuhrsch, line 25: Why do this on every loop instead of looking at tokens.size() vs. vectors.size() at the beginning? Is this to detect duplicates? If so, why not at the end?

torchtext/csrc/vectors.cpp
```diff
@@ -0,0 +1,68 @@
...
+  // serialized model from these members, thus they needs to be public.
+  std::vector<std::string> tokens_;
+  std::vector<torch::Tensor> vectors_;
+  torch::Tensor unk_tensor_;
+
+  explicit Vectors(std::vector<std::string> tokens,
+                   std::vector<torch::Tensor> vectors, torch::Tensor unk_tensor)
+      : tokens_(tokens), vectors_(vectors), unk_tensor_(unk_tensor) {
+    for (std::size_t i = 0; i < tokens.size(); i++) {
+      // guarding against size mismatch of vectors and tokens
```
Nayef211, line 24: @zhangguanheng66 @cpuhrsch lmk what you think of this part. Without adding this check, I'd get a segfault during the instantiation of the cpp `Vectors` class.

torchtext/csrc/vectors.cpp
```diff
@@ -0,0 +1,68 @@
...
+      if (i >= vectors.size()) {
+        break;
+      }
+
+      stovectors_[tokens[i]] = vectors_[i];
```
cpuhrsch, line 29: Preallocate memory for the expected amount of keys. See https://en.cppreference.com/w/cpp/container/unordered_map/reserve

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Args:
+        file_like_object (FileObject): a file like object to read data from.
+        unk_tensor (int): a 1d tensors representing the vector associated with an unknown token
+    Returns:
+        Vectors: a Vectors object.
+    """
+    readCSV = csv.reader(file_like_object, delimiter=',')
+
+    tuples = list(map(lambda x: (x[0], torch.tensor([float(c) for c in x[1].split()], dtype=torch.float)), readCSV))
```
cpuhrsch, line 16: It should be better to just do a for-loop and append. The code I wrote below was meant of brevity (where map probably actually didn't help).

The reason for that is, that within the current code you're creating two lists of length equal to the file length instead of just one.

torchtext/csrc/vectors.cpp
Nayef211: Done

torchtext/csrc/vectors.cpp
Nayef211: Moved this check to the top and now throwing a `runtime_error`

torchtext/csrc/vectors.cpp
```diff
@@ -0,0 +1,68 @@
...
+  }
+};
+
+// Registers our custom class with torch.
+static auto vectors =
+    torch::class_<Vectors>("torchtext", "Vectors")
+        .def(torch::init<std::vector<std::string>, std::vector<torch::Tensor>,
+                         torch::Tensor>())
+        .def("GetItem", &Vectors::GetItem)
+        .def_pickle(
```
zhangguanheng66, line 47: `AddItem` func to attach custom symbols here?

test/experimental/test_vectors.py
```diff
@@ -0,0 +1,62 @@
...
+        self.assertEqual(vectorsObj['b'].tolist(), tensorB.tolist())
+        self.assertEqual(vectorsObj['c'].tolist(), expected_unk_tensor.tolist())
+
+    def test_vectors_from_file(self):
+        asset_name = 'vectors_test.csv'
+        asset_path = get_asset_path(asset_name)
+        f = open(asset_path, 'r')
+        vectorsObj = vectors_from_csv_file(f)
+
+        expected_tensorA = torch.Tensor([1, 0, 0])
```
zhangguanheng66, line 56: Do we support both long() and float() tensor for the vector class?

test/experimental/test_vectors.py
Nayef211: Done

torchtext/csrc/vectors.cpp
Nayef211: @zhangguanheng66  what should this do if we try to add an item for which the token already exists in our `vector`?

test/experimental/test_vectors.py
Nayef211: hmm let me convert everything to float for now

torchtext/csrc/vectors.cpp
zhangguanheng66: That's a good question. One possibility is that users want to overwrite an existing token with a new vector. Since we make this func public, users should know what they are doing here. @cpuhrsch Do you think if we need a warning message here?

test/experimental/test_vectors.py
zhangguanheng66: I think it's fine to cover both (actually we should not put any assumptions here). Just make sure you have covered one or two cases for each category.

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,43 @@
...
+
+    return Vectors(tokens, vectors, unk_tensor=unk_tensor)
+
+
+class Vectors(object):
+    r"""Creates a vectors object which maps tokens to vectors.
+
+    Arguments:
+        tokens (List[str]): a list of tokens.
+        vectors (List[torch.Tensor]): a list of 1d tensors representing the vector associated with each token.
```
zhangguanheng66, line 30: Should we also explain our warnings in the doc here (e.g. tokens and vectors are required to have same length; or empty list will happen ...)?

torchtext/csrc/vectors.cpp
Nayef211: okay I think thats fine then. the default behavior will be to overwrite the existing vector

torchtext/experimental/vectors.py
Nayef211: I will add a section for `Raises:`

test/experimental/test_vectors.py
Nayef211: Isn't it important to stick to one type for now since the factory functions will create a `float` type tensor (that's what I am doing now at least). Or do we want to pass in an optional argument to the factory function that specifies the type?

test/experimental/test_vectors.py
```diff
@@ -0,0 +1,100 @@
...
+
+from test.common.assets import get_asset_path
+from test.common.torchtext_test_case import TorchtextTestCase
+from torchtext.experimental.vectors import (
+    Vectors,
+    vectors_from_csv_file
+)
+
+
+class TestVectors(TorchtextTestCase):
```
zhangguanheng66, line 12: Should we test the jitability here? A JIT-able vector object should convert a token into vector on jit mode.

test/experimental/test_vectors.py
Nayef211: @zhangguanheng66 lmk what you think?

test/experimental/test_vectors.py
Nayef211: Added a new test for this

torchtext/experimental/vectors.py
cpuhrsch: Change ```class Vectors(object)``` to inherit from ```torch.nn.Module```.

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,60 @@
+import csv
+import torch
+from torch import Tensor
+import torch.nn as nn
+
+
+def vectors_from_csv_file(file_like_object, unk_tensor=None):
```
cpuhrsch, line 7: nit: you probably want "from_file_object" instead of "from_csv_file" or such

torchtext/csrc/vectors.cpp
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  torch::Tensor GetItem(const std::string &token) const {
+    if (stovectors_.find(token) != stovectors_.end()) {
+      return stovectors_.at(token);
+    }
+    return unk_tensor_;
+  }
+
+  void AddItem(const std::string &token, const torch::Tensor &vector) {
```
cpuhrsch, line 51: If there's a dtype constraint I'd enforce it here

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,60 @@
...
+        unk_tensor (torch.Tensor): a 1d tensors representing the vector associated with an unknown token.
+
+    Raises:
+
+        ValueError: if `vectors` is empty and a default `unk_tensor` isn't provided.
+        RuntimeError: if `tokens` and `vectors` have different sizes or `tokens` has duplicates.
+
+    """
+
+    def __init__(self, tokens, vectors, unk_tensor=None):
```
cpuhrsch, line 44: Document and check that the dtype is float

torchtext/experimental/vectors.py
Nayef211: Done

torchtext/csrc/vectors.cpp
Nayef211: Added this

torchtext/experimental/vectors.py
Nayef211: Done

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,83 @@
...
+        r"""
+        Args:
+            token (str): the token used to lookup the corresponding vector.
+        Returns:
+            vector (Tensor): a tensor (the vector) corresponding to the associated token.
+        """
+        return self.vectors.GetItem(token)
+
+    @torch.jit.export
+    def __setitem__(self, token: str, vector: Tensor):
```
zhangguanheng66, line 71: Just wondering if we should add a test to `torch.save()` and `torch.load()`.

Nayef211: I added a test for this. Can you take another look?

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,83 @@
+import csv
+import torch
+from torch import Tensor
+import torch.nn as nn
+
+
+def vectors_from_file_object(file_like_object, unk_tensor=None):
+    r"""Create a Vectors object from a csv file like object.
```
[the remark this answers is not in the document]
Nayef211: Done

PR #821, opened 2020-06-11
add translation datasets to __init__ file

zhangguanheng66:


zhangguanheng66:
@ akurniawan Could you take a good at this patch?

PR #831, opened 2020-06-18
Added new factory functions for Vectors class

Nayef211:
## Description
- Added 2 new factory functions for loadings FastText and GloVe Vectors
- Added functions for loading and saving pickled Vectors to disc for performance
- Added unit tests to test factory function


torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,211 @@
...
+    tokens = []
+    vectors = []
+    for row in readCSV:
+        tokens.append(row[0])
+        vectors.append(torch.tensor([float(c) for c in row[1].split()], dtype=torch.float))
+
+    return Vectors(tokens, vectors, unk_tensor=unk_tensor)
+
+
+def save_vectors(vector_obj, path_pt):
```
zhangguanheng66, line 135: What's the difference with `torch.save`? Why not use it here?

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,211 @@
...
+from torch import Tensor
+import torch.nn as nn
+
+from torchtext.utils import (
+    download_from_url,
+    extract_archive
+)
+
+
+def fast_text(language="en", unk_tensor=None):
```
zhangguanheng66, line 15: Can we consolidate `fast_text` and `glo_ve ` func? Seems like several parts are similar.

Nayef211: Yup I was thinking about this. Do you think I should create more utility functions? And if so I was also wondering about the best place to put them. I.e. should I keep them in the `vectors.py` file or create a new file under `experimental/utils.py`?

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,211 @@
...
+    downloaded_file_path = download_from_url(url)
+    extracted_file_path = extract_archive(downloaded_file_path)[0]
+
+    csv_file_name, _ext = os.path.splitext(extracted_file_path)
+    csv_file_path = '{}.csv'.format(csv_file_name)
+
+    # skip if csv file already exists
+    if not os.path.exists(csv_file_path):
+        # format downloaded file into csv format expected by `vectors_from_file_object()`
+        with io.open(extracted_file_path, encoding="utf8") as f1:
```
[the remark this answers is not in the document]
Nayef211: This is because we need to modify the downloaded file to be consistent with the file_like format that `vectors_from_file_object()` method is expecting. Do you have suggestions as to how else I could do this?

torchtext/experimental/vectors.py
Nayef211: `torch.save()` basically saves a pickled model to disc. are you suggesting to use the `torch.save()` function directly in my method?

torchtext/experimental/vectors.py
zhangguanheng66: for now, you can create some private funcs in vectors file here. No need to push them to utils unless they are very generic.

torchtext/experimental/vectors.py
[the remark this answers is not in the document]
zhangguanheng66: I was thinking about downing and read the file to load tokens and vectors directly without `vectors_from_file_object` func.

torchtext/experimental/vectors.py
zhangguanheng66: I mean `torch.save(vector_obj, "vector.pt")` should already work here, right? So we don't expect users to use `save_vectors` func.

Nayef211: Oh I meant to make this a private function. I was thinking I might be doing a few other checks when loading files to make sure the file exists. I can just call `torch.save` and `torch.load` directly though because it doesn't seem like it warrants its own private function.

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,188 @@
...
+    vectors_obj = Vectors(tokens, vectors, unk_tensor=unk_tensor)
+    torch.save(vectors_obj, vectors_file_path)
+    return vectors_obj
+
+
+def glo_ve(name="840B", dim=300, unk_tensor=None, root='.data'):
+    r"""Create a GloVe Vectors object.
+
+    Args:
+        name (str): the language to use for FastText.
```
zhangguanheng66, line 79: nit: FastText?

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,188 @@
...
+            vectors.append(vector)
+
+    return tokens, vectors, dup_tokens
+
+
+def fast_text(language="en", unk_tensor=None, root='.data'):
+    r"""Create a fast text Vectors object.
+
+    Args:
+        language (str): the language to use for FastText.
```
zhangguanheng66, line 48: We should be explicit for the language options supported.

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,188 @@
...
+
+    if dup_tokens:
+        raise ValueError("Found duplicate tokens in file: {}".format(str(dup_tokens)))
+
+    vectors_obj = Vectors(tokens, vectors, unk_tensor=unk_tensor)
+    torch.save(vectors_obj, vectors_file_path)
+    return vectors_obj
+
+
+def glo_ve(name="840B", dim=300, unk_tensor=None, root='.data'):
```
zhangguanheng66, line 75: We should be explicit for the dim options supported. And please include it in the doc.

torchtext/experimental/vectors.py
```diff
@@ -0,0 +1,188 @@
...
+                continue
+
+            stoi[token] = len(vectors)
+            tokens.append(token)
+            vectors.append(vector)
+
+    return tokens, vectors, dup_tokens
+
+
+def fast_text(language="en", unk_tensor=None, root='.data'):
```
zhangguanheng66, line 44: Should we be consistent with the legacy code for the pre-trained vector names? like FastText, GloVe.

torchtext/experimental/vectors.py
Nayef211: I don't think the `dim` param value is being used for anything actually. In the past it was being used for the cached `.pt` file_name. I've gone ahead and removed it. Lmk if you think its necessary.

torchtext/experimental/vectors.py
Nayef211: I think I was adhering to the Python snake_case formatting for function names. In the legacy code we named it using PascalCase because thats how classes are usually formatted. Do you think it still makes sense to change the naming even though this has become a function rather than a class?

torchtext/experimental/vectors.py
Nayef211: Thanks for catching this! Fixed

torchtext/experimental/vectors.py
Nayef211: I believe there is actually a list of 157 languages supported. I will add the URL where they can find out what the options are (i.e. https://fasttext.cc/docs/en/language-identification.html)

.gitignore
```diff
@@ -81,6 +81,9 @@ docs/_build/
 # PyBuilder
 target/
 
+# thirdparty dirs
+third_party/*/
```
cpuhrsch, line 85: Where is this coming from?

.gitignore
```diff
@@ -109,6 +112,9 @@ venv.bak/
 # Rope project settings
 .ropeproject
 
+# VSCode project settings
+.vscode
```
cpuhrsch, line 116: What about this/

test/experimental/test_vectors.py
```diff
@@ -135,3 +140,41 @@ def test_vectors_from_file(self):
...
+                'world': [-0.32423, -0.098845, -0.0073467],
+            }
+
+            for word in expected_fasttext_simple_en.keys():
+                self.assertEqual(vectors_obj[word][:3], expected_fasttext_simple_en[word])
+
+    def test_glo_ve(self):
+        # copy the asset file into the expected download location
+        asset_name = 'glove.840B.300d.zip'
+        asset_path = get_asset_path(asset_name)
```
cpuhrsch, line 166: Is this cached?

test/experimental/test_vectors.py
```diff
@@ -135,3 +140,41 @@ def test_vectors_from_file(self):
...
+            # The first 3 entries in each vector.
+            expected_fasttext_simple_en = {
+                'the': [-0.065334, -0.093031, -0.017571],
+                'world': [-0.32423, -0.098845, -0.0073467],
+            }
+
+            for word in expected_fasttext_simple_en.keys():
+                self.assertEqual(vectors_obj[word][:3], expected_fasttext_simple_en[word])
+
+    def test_glo_ve(self):
```
cpuhrsch, line 163: should this be glove or glo_ve?

torchtext/experimental/vectors.py
```diff
@@ -1,7 +1,113 @@
...
+    with open(file_path, 'rb') as f:
+        for line in tqdm(f):
+            token, entries = line.rstrip().split(b" ", 1)
+            vector = torch.tensor([float(c) for c in entries.split(b" ")], dtype=torch.float)
+
+            try:
+                if isinstance(token, bytes):
+                    token = token.decode('utf-8')
+            except UnicodeDecodeError:
+                logger.info("Skipping non-UTF8 token {}".format(repr(token)))
```
cpuhrsch, line 29: This should be an option. Also, don't we already have a unicode csv iterator?

torchtext/experimental/vectors.py
```diff
@@ -1,7 +1,113 @@
...
+    """
+    url = 'https://dl.fbaipublicfiles.com/fasttext/vectors-wiki/wiki.{}.vec'.format(language)
+    vectors_file_path = os.path.join(root, os.path.splitext(os.path.basename(url))[0] + '.pt')
+    if os.path.isfile(vectors_file_path):
+        return(torch.load(vectors_file_path))
+
+    downloaded_file_path = download_from_url(url, root=root)
+    tokens, vectors, dup_tokens = _load_token_and_vectors_from_file(downloaded_file_path)
+
+    if dup_tokens:
```
cpuhrsch, line 68: Isn't this printing it twice?

torchtext/experimental/vectors.py
```diff
@@ -1,7 +1,113 @@
...
+        raise ValueError("Found duplicate tokens in file: {}".format(str(dup_tokens)))
+
+    vectors_obj = Vectors(tokens, vectors, unk_tensor=unk_tensor)
+    torch.save(vectors_obj, vectors_file_path)
+    return vectors_obj
+
+
+def glo_ve(name="840B", unk_tensor=None, root='.data'):
```
cpuhrsch, line 76: I guess it's consistent to use glo_ve when the camelcase is GloVe, but I'd still say unexpected for most users

torchtext/experimental/vectors.py
```diff
@@ -1,7 +1,113 @@
...
+    urls = {
+        '42B': 'http://nlp.stanford.edu/data/glove.42B.300d.zip',
+        '840B': 'http://nlp.stanford.edu/data/glove.840B.300d.zip',
+        'twitter.27B': 'http://nlp.stanford.edu/data/glove.twitter.27B.zip',
+        '6B': 'http://nlp.stanford.edu/data/glove.6B.zip',
+    }
+
+    url = urls[name]
+    vectors_file_path = os.path.join(root, os.path.splitext(os.path.basename(url))[0] + '.pt')
+    if os.path.isfile(vectors_file_path):
```
cpuhrsch, line 101: I wish we could do something stronger here such as using md5 to make sure the data isn't corrupted.l

torchtext/experimental/vectors.py
```diff
@@ -1,7 +1,113 @@
...
+            vector = torch.tensor([float(c) for c in entries.split(b" ")], dtype=torch.float)
+
+            try:
+                if isinstance(token, bytes):
+                    token = token.decode('utf-8')
+            except UnicodeDecodeError:
+                logger.info("Skipping non-UTF8 token {}".format(repr(token)))
+                continue
+
+            if token in stoi:
```
cpuhrsch, line 32: This seems like something that should be done outside of this function as a separate operation. Also, doesn't the Vectors class already do this? If we mean to remove duplicate elements because we expect it to be for a particular set of Vectors, we should only remove duplicates in that case, because in other cases it might be a bug.

Even more so, ideally we know what element we expect to be duplicated and explicitly check plus make a choice about this.

.gitignore
Nayef211: This is also in the PR for the regex library where we add `re2` submodule. We also have a build_dir that gets created in the third_party folder. I was thinking git should be ignoring these changes

Nayef211:
> Do we support CharNGram?
> https://github.com/pytorch/text/blob/50576f32f081f718faa1ec454888ff11b9893227/torchtext/vocab.py#L493

I believe we discussed coming back to CharNGram later as this would require more work. Do you want me to create a factory method for this now as well? @cpuhrsch

torchtext/experimental/vectors.py
Nayef211: I think so but I wasn't sure if we could use this if we are loading in the file as a binary?

test/experimental/test_vectors.py
Nayef211: I thought GloVe stood for global vectors which are 2 seperate words. Do you think `glove` makes more sense

torchtext/experimental/vectors.py
Nayef211: yup you're right. I removed the other print within `_load_token_and_vectors_from_file`

torchtext/experimental/vectors.py
Nayef211: do you think `glove` would make more sense then?

.gitignore
Nayef211: We should be ignoring `.vscode` project settings. Is this bad?

test/experimental/test_vectors.py
Nayef211: I am storing this in the assets folder I believe

torchtext/experimental/checksums_fast_text.json
```diff
@@ -0,0 +1,178 @@
+{
```
zhangguanheng66, line 1: Do fasttext and glove checksums need to be in two separate files? Should we have an asset folder to save it?

torchtext/experimental/vectors.py
```diff
@@ -58,22 +60,29 @@ def fast_text(language="en", unk_tensor=None, root='.data'):
...
-
-    downloaded_file_path = download_from_url(url, root=root)
+    cached_vectors_file_path = os.path.join(root, os.path.splitext(os.path.basename(url))[0] + '.pt')
+    if os.path.isfile(cached_vectors_file_path):
+        return(torch.load(cached_vectors_file_path))
+
+    # TODO(Nayef211): determine how to specify a relative file path
+    checksum = None
+    if validate_file:
+        with open("/private/home/nayef211/torchtext/torchtext/experimental/checksums_fast_text.json", 'r') as f:
```
zhangguanheng66, line 68: the folder directory?

torchtext/experimental/vectors.py
```diff
@@ -83,30 +92,38 @@ def glo_ve(name="840B", unk_tensor=None, root='.data'):
...
-    vectors_file_path = os.path.join(root, os.path.splitext(os.path.basename(url))[0] + '.pt')
-    if os.path.isfile(vectors_file_path):
-        return(torch.load(vectors_file_path))
-
-    downloaded_file_path = download_from_url(url, root=root)
+    cached_vectors_file_path = os.path.join(root, os.path.splitext(os.path.basename(url))[0] + '.pt')
+    if os.path.isfile(cached_vectors_file_path):
+        return(torch.load(cached_vectors_file_path))
+
+    # TODO(Nayef211): determine how to specify a relative file path
```
zhangguanheng66, line 120: One of the solutions is to hard-code the filename and hashing as a dictionary and save them in `vectors.py` file.

torchtext/experimental/checksums_fast_text.json
Nayef211: Added the scripts to compute checksums to `experimental/asset` folder

torchtext/experimental/vectors.py
Nayef211: removed json files so this is not needed anymore

torchtext/experimental/vectors.py
Nayef211: Done!

docs/source/experimental_vectors.rst
```diff
@@ -18,3 +18,13 @@ torchtext.experimental.vectors
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. autofunction:: vectors_from_file_object
+
+:hidden:`fast_text`
```
zhangguanheng66, line 22: updated? `FastText`?

docs/source/experimental_vectors.rst
```diff
@@ -18,3 +18,13 @@ torchtext.experimental.vectors
...
+~~~~~~~~~~~~~~~~~~~
+
+.. autofunction:: fast_text
+
+:hidden:`glo_ve`
```
zhangguanheng66, line 27: See above.

torchtext/experimental/vectors.py
```diff
@@ -1,7 +1,519 @@
...
+from tqdm import tqdm
+
+from torchtext.utils import (
+    download_from_url,
+    extract_archive
+)
+
+logger = logging.getLogger(__name__)
+
+CHECKSUMS_GLOVE = {
```
zhangguanheng66, line 17: I would move those two checksums to the end of the file.

.gitignore
```diff
@@ -81,6 +81,12 @@ docs/_build/
 # PyBuilder
 target/
 
+# thirdparty dirs
+third_party/*/
+
+# computed checksum files
+torchtext/experimental/asset/.checksums/
```
zhangguanheng66, line 88: Do we still have those files?

test/experimental/test_vectors.py
```diff
@@ -135,3 +139,41 @@ def test_vectors_from_file(self):
...
+            shutil.copy(asset_path, data_path)
+            vectors_obj = FastText(root=dir_name, validate_file=False)
+
+            # The first 3 entries in each vector.
+            expected_fasttext_simple_en = {
+                'the': [-0.065334, -0.093031, -0.017571],
+                'world': [-0.32423, -0.098845, -0.0073467],
+            }
+
+            for word in expected_fasttext_simple_en.keys():
```
zhangguanheng66, line 159: Should we cover a jit case in the test?

test/experimental/test_vectors.py
```diff
@@ -135,3 +139,41 @@ def test_vectors_from_file(self):
...
+            vectors_obj = GloVe(root=dir_name, validate_file=False)
+
+            # The first 3 entries in each vector.
+            expected_glove = {
+                'the': [0.27204, -0.06203, -0.1884],
+                'people': [-0.19686, 0.11579, -0.41091],
+            }
+
+            for word in expected_glove.keys():
+                self.assertEqual(vectors_obj[word][:3], expected_glove[word])
```
zhangguanheng66, line 179: same as above.

.gitignore
Nayef211: Yup so these get generated by the script I checked in to compute checksums. The checksum dictionary needs to be manually copied from the generated JSON file and pasted into the `vectors.py` file after running the script

torchtext/experimental/vectors.py
Nayef211: Sure

docs/source/experimental_vectors.rst
Nayef211: Done

docs/source/experimental_vectors.rst
Nayef211: Done

test/experimental/test_vectors.py
Nayef211: Done

test/experimental/test_vectors.py
Nayef211: Done

PR #839, opened 2020-06-23
Add batch_first support in MHA and update docs

zhangguanheng66:


torchtext/modules/multiheadattention.py
```diff
@@ -3,14 +3,18 @@
...
 class MultiheadAttentionContainer(torch.nn.Module):
-    def __init__(self, nhead, in_proj_container, attention_layer, out_proj):
+    def __init__(self, nhead, in_proj_container, attention_layer, out_proj, batch_first=False):
         r""" A multi-head attention container
 
         Args:
             nhead: the number of heads in the multiheadattention model
             in_proj_container: A container of multi-head in-projection linear layers (a.k.a nn.Linear).
-            attention_layer: The attention layer.
+            attention_layer: The custom attention layer. The input sent from MHA container to the attention layer
```
cpuhrsch, line 12: Does this also take care of broadcasting?

torchtext/modules/multiheadattention.py
```diff
@@ -78,20 +86,26 @@ def forward(self, query: torch.Tensor, key: torch.Tensor, value: torch.Tensor,
...
         scaled dot product attention.
 
         Args:
             dropout (float): probability of dropping an attention weight.
+            batch_first: If ``True``, then the input and output tensors are provided
+                as `(batch, seq, feature)`. Default: ``False``
 
         Examples::
-            >>> SDP = torchtext.models.ScaledDotProduct(0.1)
+            >>> SDP = torchtext.modules.ScaledDotProduct(dropout=0.1)
```
cpuhrsch, line 108: Should we mirror the pytorch path conventions here?

torchtext.nn and torchtext.nn.functional?

zhangguanheng66: The other two domains use `torchvision/audio.models`.

torchtext/modules/multiheadattention.py
zhangguanheng66: The custom attention layer needs to take care of broadcasting. Updated the doc to reflect this.

cpuhrsch: I'd then augment the shape to ```(..., seq, batch, feature)``` and explain what that means and also that it's optional, i.e. enough to only handle 3-dim.

torchtext/modules/multiheadattention.py
cpuhrsch: Yes, but this isn't a model, right?

zhangguanheng66: OK, Will fix it.

PR #841, opened 2020-06-23
Merge latest fbsync with master 2020/06/23

Nayef211:
Resolved all merged conflicts locally


Nayef211:
> So we don't have any updates from internal branch, right?

You are right. Will go ahead and close this

Issue #843, opened 2020-06-23
Cannot select embedding size for experimental 6B and twitter.27B GloVe vectors


bentrevett:
GloVe embeddings for 6B and twitter.27B are available in multiple dimension sizes, however due to [this](https://github.com/pytorch/text/blob/master/torchtext/experimental/vectors.py#L139) line only the first file (vectors) in each of the extracted archive is loaded.

This is fine for the 420B and 840B GloVe vectors as they only contain one size of embeddings. However, for the 6B and twitter.27B  vectors there are four sizes of embeddings available, but only the first in each archive (the smallest size) is obtained from the new experimental vector class. 

It would be nice to be able to get the other 6B and twitter.27B vectors. Potentially a `dim` argument can be added to the `GloVe` function?

zhangguanheng66:
@Nayef211 I remember we had some discussions about the dim argument when merging the glove vector PR.

Nayef211:
> @Nayef211 I remember we had some discussions about the dim argument when merging the glove vector PR.

Yes I think I should have looked at this more carefully before. I'll create a PR to fix this!

PR #846, opened 2020-06-24
Fixing embedding selection issue in experimental glove

Nayef211:
## Description
- Fixing issue where the first file path being extracted from downloaded GloVe `zip` is loaded by default
- Now the correct file will be loaded based on the specified `dim` parameter 
- Refer to #843


torchtext/experimental/vectors.py
```diff
@@ -103,6 +105,7 @@ def GloVe(name="840B", unk_tensor=None, root='.data', validate_file=True):
             - 840B
             - twitter.27B
             - 6B
+        dim (int): the dimension for the GloVe dataset to load.
```
zhangguanheng66, line 108: Let's be more explicit about the dim options for each category.

torchtext/experimental/vectors.py
```diff
@@ -120,14 +123,16 @@ def GloVe(name="840B", unk_tensor=None, root='.data', validate_file=True):
...
-        '6B': 'https://nlp.stanford.edu/data/glove.6B.zip',
+        "42B": "https://nlp.stanford.edu/data/glove.42B.300d.zip",
+        "840B": "https://nlp.stanford.edu/data/glove.840B.300d.zip",
+        "twitter.27B": "https://nlp.stanford.edu/data/glove.twitter.27B.zip",
+        "6B": "https://nlp.stanford.edu/data/glove.6B.zip",
     }
 
     url = urls[name]
-    cached_vectors_file_path = os.path.join(root, os.path.splitext(os.path.basename(url))[0] + '.pt')
+    file_name = "glove.{}.{}d.txt".format(name, str(dim))
```
zhangguanheng66, line 133: Should we check if dim is support for the specific name?

zhangguanheng66:
Fix #843

torchtext/experimental/vectors.py
Nayef211: Sure. I'll list out what the possible options are.

torchtext/experimental/vectors.py
Nayef211: Do you think I should have a dict like this? And then check that the file name exists within the dict? https://github.com/pytorch/text/blob/fd764ad02968e6310d28ee3c496e1d1d1157aa3e/torchtext/vocab.py#L523

zhangguanheng66:
The `binary_macos_conda_py3.8` test fails again?

Nayef211:
> The `binary_macos_conda_py3.8` test fails again?

Yup this seems to be happening for all my PRs. Could it be another issue on Pytorch's side?

zhangguanheng66:
> > The `binary_macos_conda_py3.8` test fails again?
> 
> Yup this seems to be happening for all my PRs. Could it be another issue on Pytorch's side?
Very likely.

PR #847, opened 2020-06-24
bump nightlies to 0.8.0

seemethere:


zhangguanheng66:
The circleci fails. Is this because of the pytorch binary release?

seemethere:
Yes I think it should be resolved whenever the pytorch nightly version gets bumped, but since it didn't fail for other versions on macOPS we can probably ignore.

PR #848, opened 2020-06-24
Cherry picking release pipeline improvements

seemethere:
These were commits that were made during the `v0.7.0-rc1` release and should be put onto the master branch as well.

These are all already committed to `release/0.7`


.circleci/config.yml
```diff
@@ -230,7 +230,7 @@ jobs:
           command: |
             set -x
             source /usr/local/etc/profile.d/conda.sh && conda activate python${PYTHON_VERSION}
-            conda install -v -y -c pytorch-nightly pytorch
+            conda install -v -y -c pytorch-test -c pytorch-nightly pytorch
```
zhangguanheng66, line 233: Do the future circleci tests depend on the RC package?

PR #851, opened 2020-06-25
Fix experimental translation dataset

akurniawan:
Fix missing import in the dataset


akurniawan:
@zhangguanheng66 this is a fix for translation dataset. There is a missing Vocab import (required for validation in case users provide their own vocabulary objects)

PR #853, opened 2020-06-25
Nayef211/update vectors

Nayef211:
## Description
- Explicitly check that all dimensions are of the same size when loading tokens from a file
- Use `_load_token_and_vectors_from_file` method for the underlying implementation of  `vectors_from_file_object ` for consistency. Raise error when there is a duplicate token present in the csv file.
- Fixed bug where a tuple wasn't being appended to `dup_tokens`
- Fixed logic for dupe check for `GloVe`
- Updated test csv asset to reflect new format


test/asset/vectors_test.csv
```diff
@@ -1,2 +1,2 @@
-a,1 0 0
-b,0 1 0
\ No newline at end of file
+a 1 0 0
+b 0 1 0
```
zhangguanheng66, line 2: why do we remove comma from csv file?

Nayef211: Readded this

PR #854, opened 2020-06-26
Creating new JITable Vocab class

Nayef211:
## Description
- Creating a new Vocab class in experimental that is JITable and more performant
- Using a custom c++ class for the underlying vocab implementation

## Future Scope
- Making certain functions like `stoi()` and `itos()` private
- Adding factory methods for reading from CSV file. Also include example of how to use this in the docstring
- Creating benchmarking code to compare new implementation with exsiting Vocab implementation
- Refactoring cpp Vocab class with optimized implementation (may involve using [`FastText` dictionary implementation](https://github.com/facebookresearch/fastText/blob/master/src/dictionary.cc#L163))


test/experimental/test_vocab.py
```diff
@@ -0,0 +1,164 @@
...
+        c = OrderedDict(sorted_by_freq_tuples)
+
+        v = Vocab(c, specials=['<pad>', '<eos>'])
+        self.assertEqual(v['<unk>'], 0)
+        self.assertEqual(v['<pad>'], 1)
+        self.assertEqual(v['<eos>'], 2)
+        self.assertEqual(v['a'], 3)
+        self.assertEqual(v['b'], 4)
+
+    def test_vocab_set_item(self):
```
zhangguanheng66, line 46: I took a look at torchtext.Vocab and there are no similar func. There is also no such func in [pytext/ScriptVocabulary](https://github.com/facebookresearch/pytext/blob/master/pytext/torchscript/vocab.py#L11).

I think a reasonable scenario is that users can always build a new vocab with the custom order or new tokens. It might be too complicate for us to maintain such capability.

test/experimental/test_vocab.py
```diff
@@ -0,0 +1,164 @@
...
+        token_to_freq = {'a': 2, 'b': 2, 'c': 2}
+        sorted_by_freq_tuples = sorted(token_to_freq.items(), key=lambda x: x[1], reverse=True)
+        c = OrderedDict(sorted_by_freq_tuples)
+
+        v = Vocab(c, specials=['<pad>', '<eos>'])
+        self.assertEqual(len(v), 6)
+
+    def test_vocab_basic(self):
+        token_to_freq = {'hello': 4, 'world': 3, 'ᑌᑎIᑕOᗪᕮ_Tᕮ᙭T': 5, 'freq_too_low': 2}
+        sorted_by_freq_tuples = sorted(token_to_freq.items(), key=lambda x: x[1], reverse=True)
```
zhangguanheng66, line 85: what happen if there are tokens with same frequency?

torchtext/csrc/vocab.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  int64_t __getitem__(const std::string &token) const {
+    const auto &item = stoi_.find(token);
+    if (item != stoi_.end()) {
+      return item->second;
+    }
+    return unk_index_;
+  }
+
+  void addToken(const std::string &token) {
```
zhangguanheng66, line 46: As my comments above, we need more discussions for this func.

torchtext/experimental/vocab.py
```diff
@@ -0,0 +1,97 @@
...
+import torch.nn as nn
+
+logger = logging.getLogger(__name__)
+
+
+class Vocab(nn.Module):
+    r"""Creates a vocab object which maps tokens to indices.
+
+    Arguments:
+        ordered_dict (collections.OrderedDict): object holding the frequencies of each token found in the data.
```
zhangguanheng66, line 13: I think we can be more explicit here. For example, what happen if there are tokens with same frequency.

torchtext/experimental/vocab.py
```diff
@@ -0,0 +1,97 @@
...
+        min_freq: The minimum frequency needed to include a token in the vocabulary.
+            Values less than 1 will be set to 1. Default: 1.
+        specials: The tuple of special tokens (e.g., padding or eos) that will be prepended to the vocabulary.
+            The first value should always be the unknown token Default: ['<unk'>, '<pad>']
```
zhangguanheng66, line 17: This line is not clear. The specials will be prepended to the vocabulary list so the order of the special tokens in `specials` matters, right? And what happen if `<unk>' is not in `specials`? What happen if users assign a custom unknown token?

test/experimental/test_vocab.py
Nayef211: I see so you're suggesting to not keep the functionality to allow users to update an index once the vocab has been created?

test/experimental/test_vocab.py
Nayef211: I believe the sorted function keeps it in the same order that it appears in. I can do some more testing on this to verify. But the whole point is that it is up to the user to provide the ordering of tokens when building the `OrderedDict`. Once the dict has been passed into the vocab class, we respect the ordering of this dict.

torchtext/experimental/vocab.py
Nayef211: Yes I can specify that the ordering of the specials token matter. The comment about the unk token is incorrect. I have a seperate `unk_token` parameter that the user needs to pass in. I will update the comment to reflect this!

torchtext/experimental/vocab.py
Nayef211: Hmm so refer to my comments above for respecting the ordering of the dictionary that is passed in.

torchtext/csrc/vocab.cpp
Nayef211: > Are we going to support a similar func, like [lookup_indices_1d](https://github.com/facebookresearch/pytext/blob/2c6a2b602cbb9e784c5cb4abdb21d26983f12d83/pytext/torchscript/vocab.py#L35)?
> 
> Do we want to have a func for [lookup_word](https://github.com/facebookresearch/pytext/blob/2c6a2b602cbb9e784c5cb4abdb21d26983f12d83/pytext/torchscript/vocab.py#L102)?

Let's discuss this further. We can probably add this in follow up PRs.

test/experimental/test_vocab.py
zhangguanheng66: I think it's not necessary. Do you see this kind of func in other vocab?

torchtext/experimental/vocab.py
cpuhrsch: Does this particular class documentation make the explicit to the user?

Nayef211: Sure I think that's a good idea. I just added 2 lines explaining this

torchtext/experimental/vocab.py
```diff
@@ -0,0 +1,145 @@
...
+        Args:
+            token (str): the token used to lookup the corresponding index.
+
+        Returns:
+            index (int): the index corresponding to the associated token.
+        """
+        return self.vocab[token]
+
+    @torch.jit.export
+    def insert_token(self, token: str, index: int) -> None:
```
cpuhrsch, line 73: I'd remove the "_token" part unless you have an explicit reason. It's more in line with python's List.

torchtext/experimental/vocab.py
```diff
@@ -0,0 +1,145 @@
...
+        Args:
+            tokens (List[str]): the tokens used to lookup their corresponding `indices`.
+
+        Returns:
+            indices (List[int]): the 'indices` associated with `tokens`.
+        """
+        return self.vocab.lookup_indices(tokens)
+
+    @torch.jit.export
+    def get_stoi(self) -> Dict[str, int]:
```
cpuhrsch, line 132: Do we really still want these?

Nayef211: I think it made testing the functions a lot easier. Now I'm not sure how often the list of tokens inside the vocab will be needed by the end user. Do you see any negatives in keeping this?

torchtext/experimental/vocab.py
Nayef211: Okay I was trying to be more consistent with the other function names which also have `_token` at the end. But I can just remove `_token` across the board

torchtext/experimental/vocab.py
```diff
@@ -0,0 +1,145 @@
...
+            they are added into the vocabulary at last. Default: True.
+
+    Raises:
+        ValueError: if a default `unk_token` isn't provided.
+    """
+
+    def __init__(self, ordered_dict, min_freq=1, unk_token='<unk>', specials=('<unk>', '<pad>'), specials_first=True):
+        super(Vocab, self).__init__()
+
+        if not unk_token:
```
zhangguanheng66, line 32: Are we expect users to pass `None` to the keyword argument `unk_token` here?

test/experimental/test_vocab.py
```diff
@@ -0,0 +1,209 @@
...
+    def test_vocab_len(self):
+        token_to_freq = {'a': 2, 'b': 2, 'c': 2}
+        sorted_by_freq_tuples = sorted(token_to_freq.items(), key=lambda x: x[1], reverse=True)
+        c = OrderedDict(sorted_by_freq_tuples)
+        v = Vocab(c)
+
+        self.assertEqual(len(v), 5)
+
+    def test_vocab_basic(self):
+        token_to_freq = {'hello': 4, 'world': 3, 'ᑌᑎIᑕOᗪᕮ_Tᕮ᙭T': 5, 'freq_too_low': 2}
```
zhangguanheng66, line 83: I might miss it. Do we have a test case starting from a list of tokens?

torchtext/experimental/vocab.py
zhangguanheng66, line 73: Do we want to maintain `insert_token` func in vocab class? Any usage case to support it?

torchtext/experimental/vocab.py
```diff
@@ -0,0 +1,145 @@
...
+        ordered_dict (collections.OrderedDict): object holding the frequencies of each token found in the data.
+        min_freq: The minimum frequency needed to include a token in the vocabulary.
+            Values less than 1 will be set to 1. Default: 1.
+        specials: The tuple of special tokens (e.g., padding or eos) that will be prepended/postpended to the vocabulary.
+            based on the `specials_first` flag. The ordering of the tuple will be preserved. Default: ('<unk>', '<pad>')
+        specials_first: Whether to add special tokens into the vocabulary at first. If it is False,
+            they are added into the vocabulary at last. Default: True.
+
+    Raises:
+        ValueError: if a default `unk_token` isn't provided.
```
zhangguanheng66, line 26: Suggest to add two examples in the doc to build vocab from a list of ordered unique tokens and a list of raw text tokens (with repeat in this case).

torchtext/experimental/vocab.py
Nayef211: We generally don't expect this. I was just being explicit here in case the user did pass in a `None` we would fail elegantly.

test/experimental/test_vocab.py
Nayef211: What do you mean by starting off with a list of tokens? Because the `Vocab` class expects an `OrderedDict` as the input. So even if we did start off with a list of tokens, we would have to probably get a dictionary of tokens to frequency and possibly sort this dictionary in a similar manner that I am doing now to get an ordered list of tuples which are finally fed into an OrderedDict class.

Does this make sense?

torchtext/experimental/vocab.py
Nayef211: Yes I think Steven specifically requested this during our design meeting. I.e. they wanted to be able to insert special tokens into specific indices after constructing a `Vocab`.

torchtext/experimental/vocab.py
Nayef211: I will go ahead and add that example for a list of ordered unique tokens. In terms of reading tokens from raw text file, I will add that to the followup PR with the factory function `vocab_from_file_object `

zhangguanheng66: yeap. We can add that func in the followup PR.

PR #859, opened 2020-06-29
Experimental bucket by sequence length sampler

akurniawan:
@zhangguanheng66 I'm proposing a sampler class with similar functionality as the [BucketIterator](https://github.com/pytorch/text/blob/bcb9104680eb9dc978a6bbcc2b9ca46cf2bdbed9/torchtext/data/iterator.py#L241). Let me know what you think of this. Thanks!


zhangguanheng66:
> @zhangguanheng66 I'm proposing a sampler class with similar functionality as the [BucketIterator](https://github.com/pytorch/text/blob/bcb9104680eb9dc978a6bbcc2b9ca46cf2bdbed9/torchtext/data/iterator.py#L241). Let me know what you think of this. Thanks!

Thanks @akurniawan . Since there have been several issues related to sampler, I agree it's worthy some investigation. And Sampling by lengths is an important tool for NLP problem.

akurniawan:
@zhangguanheng66 I initially put the `seqlen_sampler` under the dataset folder as I was following torchvision. But it has now moved to `utils.samplers`

test/data/test_utils_samplers.py
```diff
@@ -0,0 +1,57 @@
...
+
+        def collate_fn(batch):
+            return pad_sequence(batch, batch_first=True)
+
+        dataset = MyDataset()
+        batch_sampler = BucketByLengthBatchSampler(dataset, [3, 5, 10], 5)
+        iterator = DataLoader(dataset, batch_sampler=batch_sampler, collate_fn=collate_fn)
+
+        for x in iterator:
+            print(x)
```
zhangguanheng66, line 57: I think no one check the printout in CI test. It might be better to use assert check.

torchtext/experimental/utils/samplers/seqlen_sampler.py
```diff
@@ -0,0 +1,73 @@
+from typing import List
```
zhangguanheng66, line 1: I think we can have a `simpler.py` file in `experimental/utils` folder and have all the samplers there.

torchtext/experimental/utils/samplers/seqlen_sampler.py
```diff
@@ -0,0 +1,73 @@
...
+
+import numpy as np
+from torch.utils.data import Dataset, IterableDataset
+from torch.utils.data.sampler import RandomSampler, Sampler, SequentialSampler
+
+
+class BucketByLengthBatchSampler(Sampler):
```
zhangguanheng66, line 8: A shorter name?

torchtext/experimental/utils/samplers/seqlen_sampler.py
```diff
@@ -0,0 +1,73 @@
...
+    """Defines a batch sampler that batches examples of similar lengths together and
+    minimizes amount of padding needed.
+    Arguments:
+        data_source: data source to sample from.
+        bucket_boundaries: upper length boundaries to merge sentences with length
```
zhangguanheng66, line 13: How does this parameter work for batching examples according to the length?

torchtext/experimental/utils/samplers/seqlen_sampler.py
```diff
@@ -0,0 +1,73 @@
...
+        batch_size: int = 32,
+        seq_dim: int = 0,
+        shuffle: bool = True,
+    ):
+        if isinstance(data_source, IterableDataset):
+            raise TypeError("Currently does not support IterableDataset!")
+
+        self.data_source = data_source
+        self.seq_dim = seq_dim
+        self.bucket_boundaries = bucket_boundaries + [np.inf]
```
zhangguanheng66, line 44: In general, we don't want to add dependence on `np`.

torchtext/experimental/utils/samplers/seqlen_sampler.py
akurniawan: I have added more documentation on the docstring. Basically I'm borrowing the idea from tensorflow's `bucket_by_sequence_length`. The way it works is by categorizing raw data by putting them in a bucket whose lengths are in the upperbound range ``bucket_boundaries``. For ``bucket_boundaries`` = [5, 10], there will be three different buckets that will consist of sentences whose lengths are less than 5, between 5 and 10, and more than 10.

torchtext/experimental/utils/samplers/seqlen_sampler.py
akurniawan: change it to `BucketBatchSampler`, let me know if you have other preferences

torchtext/experimental/utils/samplers/seqlen_sampler.py
zhangguanheng66: we use `float('-inf')` in modules/multiheadattention.py.

akurniawan:
By that you mean with a real dataset? Because I have already put a dummy dataset with DataLoader on the test file

zhangguanheng66:
> By that you mean with a real dataset? Because I have already put a dummy dataset with DataLoader on the test file

Right. I missed that. Then, we are fine with that.

akurniawan:
thanks! let me know if you some other thoughts, otherwise this is ready for review.

torchtext/experimental/utils/samplers.py
```diff
@@ -0,0 +1,76 @@
...
+
+        self.buckets = []
+        for _ in range(len(bucket_boundaries) + 1):
+            self.buckets.append([])
+
+    def __iter__(self):
+        for idx in self.sampler:
+            row = self.data_source[idx]
+            for bidx, boundary in enumerate(self.bucket_boundaries):
+                if row.size(self.seq_dim) <= boundary:
```
nth-attempt, line 62: by using `row.size` we are assuming that data_source is a tensor. Can we change this to `len(row[self.seq_dim])`

akurniawan: thanks for the feedback. I have replaced the way to get the sequence length by allowing users to pass a function. The problem using `len` is that it might not work the way you suggest. For list we need to use `len(row[0][self.seq_dim])` if the sequence length is in the second dimension. I think by passing user's defined function will be more elegant to solve this problem. What do you think?

nth-attempt: Yes, that is more elegant :)

akurniawan:
@zhangguanheng66 are we still interested in continuing this PR?

zhangguanheng66:
> @zhangguanheng66 are we still interested in continuing this PR?

Since we are going to remove `Field` class and associated legacy code in v0.8.0 release, I'm wondering if we can make this Sampler more generic. What I mean here is a sampler which groups the examples with similar lengths together. This is useful for some NLP problems by padding similar length sentence together. How do you feel about the idea?

akurniawan:
@zhangguanheng66 Thank you for the idea. If you don't mind, could you help to elaborate more on the `groups the examples with similar lengths together`? I would like to understand the difference between your idea and the current implementation on this PR

zhangguanheng66:
> @zhangguanheng66 Thank you for the idea. If you don't mind, could you help to elaborate more on the `groups the examples with similar lengths together`? I would like to understand the difference between your idea and the current implementation on this PR

They are very similar. One difference is the current implementation in this PR requires a boundary or "bucket". What happen if users don't provide the buckets?

akurniawan:
> They are very similar. One difference is the current implementation in this PR requires a boundary or "bucket". What happen if users don't provide the buckets?

Got it. So basically you're thinking to instead of users providing the "bucket" we provide it to them like the one that has already implemented on the legacy code?

zhangguanheng66:
> > They are very similar. One difference is the current implementation in this PR requires a boundary or "bucket". What happen if users don't provide the buckets?
> 
> Got it. So basically you're thinking to instead of users providing the "bucket" we provide it to them like the one that has already implemented on the legacy code?

Kind of. I think this is more useful for the text problems. How do you think about it?

akurniawan:
> > > They are very similar. One difference is the current implementation in this PR requires a boundary or "bucket". What happen if users don't provide the buckets?
> > 
> > 
> > Got it. So basically you're thinking to instead of users providing the "bucket" we provide it to them like the one that has already implemented on the legacy code?
> 
> Kind of. I think this is more useful for the text problems. How do you think about it?

Makes sense. I have updated the implementation to use the sorting approach and ready for another review. Please let me know if you have some other concerns or ideas

nth-attempt:
If I understand this correct, this approach will return the smaller sequence lengths first even with `shuffle=True`. That may not the the ideal behavior?

akurniawan:
@nth-attempt depends on the number of the data. If the dataset is small (<= batch_size * 100) then yes, using `shuffle=True` will not changing anything. However, with dataset in bigger size (> batchsize * 100) shuffling will work since currently we are sorting within batch and not the whole dataset

PR #862, opened 2020-06-30
Optimizing cpp Vectors class to use 2d tensor

Nayef211:
## Description
- Updating Vectors class to use 2d tensor along with a `unordered map<string, Tensor.view()>`
- After conducting performance benchmarks, it was found that this implementation gives the best mix of performance for lookup and construction
- Added functionality for looking up vectors for several tokens using `get_vectors_by_tokens`
- Added tests for the new function

### Existing FastText - Not Jit Mode
**Construction time from file:** 778.5076372753829
**Construction time from cache:** 6.0479023195803165
**Lookup time:** 132.09116122685373

### FastText Experimental
**Construction time:** 670.8132877256721
**Construction time from cache:** 39.3408519141376
#### FastText Experimental - Not Jit Mode
**Lookup time:** 113.68280134536326
#### FastText Experimental - Jit Mode
**Lookup time:** 224.18116508424282

### DIFF
--- a/test/experimental/test_vectors.py
+++ b/test/experimental/test_vectors.py
@@ -177,49 +177,49 @@ class TestVectors(TorchtextTestCase):
         self.assertEqual(vectors_obj['b'], expected_tensorB)
         self.assertEqual(vectors_obj['not_in_it'], expected_unk_tensor)
 
-    # def test_fast_text(self):
-    #     # copy the asset file into the expected download location
-    #     # note that this is just a file with the first 100 entries of the FastText english dataset
-    #     asset_name = 'wiki.en.vec'
-    #     asset_path = get_asset_path(asset_name)
+    def test_fast_text(self):
+        # copy the asset file into the expected download location
+        # note that this is just a file with the first 100 entries of the FastText english dataset
+        asset_name = 'wiki.en.vec'
+        asset_path = get_asset_path(asset_name)
 
-    #     with tempfile.TemporaryDirectory() as dir_name:
-    #         data_path = os.path.join(dir_name, asset_name)
-    #         shutil.copy(asset_path, data_path)
-    #         vectors_obj = FastText(root=dir_name, validate_file=False)
-    #         jit_vectors_obj = torch.jit.script(vectors_obj)
+        with tempfile.TemporaryDirectory() as dir_name:
+            data_path = os.path.join(dir_name, asset_name)
+            shutil.copy(asset_path, data_path)
+            vectors_obj = FastText(root=dir_name, validate_file=False)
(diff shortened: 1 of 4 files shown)


torchtext/csrc/vectors.cpp
```diff
@@ -2,79 +2,83 @@
...
+            [](const c10::intrusive_ptr<Vectors> &self) -> std::tuple<
+                std::vector<std::string>, torch::Tensor, torch::Tensor> {
+              std::tuple<std::vector<std::string>, torch::Tensor, torch::Tensor>
                   states(self->tokens_, self->vectors_, self->unk_tensor_);
               return states;
             },
-            // __setstate__
-            [](std::tuple<std::vector<std::string>, std::vector<torch::Tensor>,
+            // __getstate__
+            [](std::tuple<std::vector<std::string>, torch::Tensor,
```
zhangguanheng66, line 81: Do we want to support a lookup func for a list of tokens?

Nayef211:
> I would be explicit about the speedup of the performance in the summary of PR.

Adding this to the PR description

torchtext/csrc/vectors.cpp
Nayef211: Do you think this would be helpful? I can add this functionality in a followup PR?

Nayef211: Done!

test/experimental/test_vectors.py
```diff
@@ -162,49 +177,49 @@ def test_vectors_from_file(self):
...
-            # The first 3 entries in each vector.
-            expected_glove = {
-                'the': [0.27204, -0.06203, -0.1884],
-                'people': [-0.19686, 0.11579, -0.41091],
-            }
-
-            for word in expected_glove.keys():
-                self.assertEqual(vectors_obj[word][:3], expected_glove[word])
-                self.assertEqual(jit_vectors_obj[word][:3], expected_glove[word])
+    # def test_fast_text(self):
```
zhangguanheng66: Why do we disable those tests?

Nayef211: Sorry this was just an intermediate commit. This should have been fixed.

torchtext/csrc/vectors.cpp
cpuhrsch: Didn't we want to do this in a separate PR? Can you split it out please? The title doesn't indicate it's part of it and there's no design discussion around the name or potential overload mechanisms for existing getitem or maybe even a generic map function.

Nayef211:
> See inline comment

I removed the `get_vectors_by_tokens` function and am adding a new PR for that. Could I get a stamp?

Issue #664, opened 2019-12-06
Experimental datasets in torchtext 0.5.0 release


zhangguanheng66:
## Motivation and summary of the current issues in torchtext

Based on the feedback from users, there are several issues existing in torchtext, including
- Several components and functionals are unclear and difficult to adopt. For example, `Field` class couples tokenizer, vocabulary, split, batching and sampling, padding, and numericalization together. The current `Field` class works as a "black box", and users are confused about what's going on within the class. Instead, those components should be divided into several basic building blocks. This is more consistent with PyTorch core library, which grants users the freedom to build the models and pipelines with orthogonal components.
- Incompatible with [DataLoader](https://pytorch.org/docs/master/data.html#torch.utils.data.DataLoader) and [Sampler](https://pytorch.org/docs/master/data.html#torch.utils.data.Sampler) in [torch.utils.data](https://pytorch.org/docs/master/data.html#module-torch.utils.data). The current datasets in torchtext are not compatible with PyTorch core library. Some custom modules/functions in torchtext (e.g. `Iterator`, `Batch`, `splits`) should be replaced by the corresponding modules in `torch.utils.data`.

## New datasets in `torchtext.experimental.datasets`

We have re-written several datasets in `torchtext.experimental.datasets` which were using the new abstractions. The old version of the datasets are still available in `torchtext.datasets` and the new datasets are opt-in.
- Sentiment analysis dataset (https://github.com/pytorch/text/pull/651)
  * IMDB
- Language modeling datasets (https://github.com/pytorch/text/pull/624), including
  * WikiText2
  * WikiText103
  * PennTreebank

### API for new datasets
To load the new datasets, simply call the dataset API, as follow:
```
from torchtext.experimental.datasets import IMDB
train_dataset, test_dataset = IMDB()
```
To specify a tokenizer:
```
from torchtext.data.utils import get_tokenizer
tokenizer = get_tokenizer("spacy")
train_dataset, test_dataset = IMDB(tokenizer=tokenizer)
```
If you just need the test set (must pass a `Vocab` object!):
```
vocab = train_dataset.get_vocab()
test_dataset, = IMDB(tokenizer=tokenizer, vocab=vocab, data_select='test')
```

### Legacy code
The old IMDB dataset is still available in the folder `torchtext.datasets`. You can use the legacy datasets, as follow:
```
import torchtext.data as data
TEXT = torchtext.data.Field(lower=True, include_lengths=True, batch_first=True)
[...]

PR #863, opened 2020-07-01
Add retire warning to legacy code

zhangguanheng66:


torchtext/data/batch.py
```diff
@@ -19,6 +20,7 @@ class Batch(object):
 
     def __init__(self, data=None, dataset=None, device=None):
         """Create a Batch from a list of examples."""
+        warnings.warn('Batch class will retire in 0.8.0 release', RuntimeWarning)
```
mthrok, line 23: According to [this comment](https://github.com/pytorch/audio/pull/523#discussion_r405673957), PyTorch's convention for deprecation is `UserWarning`.

torchtext/data/example.py
```diff
@@ -9,6 +10,7 @@ class Example(object):
     """
     @classmethod
     def fromJSON(cls, data, fields):
+        warnings.warn('Example class will retire in 0.8.0 release', RuntimeWarning)
```
vincentqb, line 13: For all of these warnings, the user could ask themselves many questions. Is there any context that should be given? What is the user to do instead? Should the user file an issue if they can't do what they want?

torchtext/data/batch.py
zhangguanheng66: Thanks

torchtext/data/example.py
zhangguanheng66: We have an issue to explain this change. I might add this to the warning message. And in 0.7.0 release note, I will introduce this again.

vincentqb: Would adding the link in the message be a good idea? btw can you share the link here in the comment? :)

zhangguanheng66: https://github.com/pytorch/text/issues/664

vincentqb: Based on offline discussion: Please use a standard message for all. The message needs to say: (1) users are recommended to move to the new interface by doing abc, but (2) they can keep using the same functionality without the warning by doing xyz.

torchtext/data/batch.py
```diff
@@ -19,6 +20,7 @@ class Batch(object):
 
     def __init__(self, data=None, dataset=None, device=None):
         """Create a Batch from a list of examples."""
+        warnings.warn('Batch class will retire in 0.8.0 release and stay in torchtext.legacy. See 0.7.0 release note for the replacement.', UserWarning)
```
vincentqb, line 23: Can you add the link to the release notes?

zhangguanheng66: For now, we don't have a link. And I'm thinking if it's proper to add a link in warning message.

vincentqb: That's a good question, but I'm also not a fan of mentioning the release notes alone :)

torchtext/data/example.py
```diff
@@ -9,6 +10,7 @@ class Example(object):
     """
     @classmethod
     def fromJSON(cls, data, fields):
+        warnings.warn('Example class will retire in 0.8.0 release and stay in torchtext.legacy. See 0.7.0 release note for the replacement.', UserWarning)
```
cpuhrsch, line 13: I'd change the phrasing of this a bit.

"Example class will retire in 0.8.0 release and stay in torchtext.legacy. See 0.7.0 release note for the replacement"

to 

"Example class will be retired in the 0.8.0 release and moved to torchtext.legacy. Please see 0.7.0 release notes for further information."

cpuhrsch: Did we settle on UserWarning? cc @gchanan

zhangguanheng66: With some offline discussions, `UserWarning` is proper, though pytorch started to use `FutureWarning`.

torchtext/data/example.py
```diff
@@ -9,6 +10,7 @@ class Example(object):
     """
     @classmethod
     def fromJSON(cls, data, fields):
+        warnings.warn('Example class will be retired in the 0.8.0 release and moved to torchtext.legacy. Please see 0.7.0 release note for further information.', UserWarning)
```
cpuhrsch, line 13: nit: It's "notes" not "note".

PR #864, opened 2020-07-02
Experimental machine translation example

akurniawan:
This is a PR for new torchtext API in machine translation use case. This includes:

- Sample on how to build character and word representation
- Embedding model for character representation
- Seq2seq model from https://pytorch.org/tutorials/beginner/torchtext_translation_tutorial.html

@zhangguanheng66 let me know what you think of this. Thanks!


examples/machine_translation/model.py
```diff
@@ -0,0 +1,151 @@
...
+        super().__init__()
+
+        self.input_dim = input_dim
+        self.emb_dim = emb_dim
+        self.enc_hid_dim = enc_hid_dim
+        self.dec_hid_dim = dec_hid_dim
+        self.dropout = dropout
+
+        self.embedding = embedding
+        self.rnn = nn.GRU(emb_dim, enc_hid_dim, bidirectional=True)
```
zhangguanheng66, line 22: Just for the record. I will add or switch transformer model in 2020 H2 as many OSS users require an example to show how to use transformer decoder. 
https://discuss.pytorch.org/t/nn-transformer-explaination/53175/14
https://discuss.pytorch.org/t/how-to-use-nn-transformerdecoder-at-inference-time/49484

examples/machine_translation/train.py
```diff
@@ -0,0 +1,119 @@
...
+        start_time = time.time()
+
+        train_loss = train(model, train_iterator, optimizer, criterion, clip)
+        valid_loss = evaluate(model, valid_iterator, criterion)
+
+        end_time = time.time()
+
+        epoch_mins, epoch_secs = epoch_time(start_time, end_time)
+
+        print(f"Epoch: {epoch+1:02} | Time: {epoch_mins}m {epoch_secs}s")
```
zhangguanheng66, line 113: Should we use `bleu_score` for a metric?
https://github.com/pytorch/text/blob/bcb9104680eb9dc978a6bbcc2b9ca46cf2bdbed9/torchtext/data/metrics.py#L35

examples/machine_translation/README.md
```diff
@@ -0,0 +1,9 @@
+# This is an example to create a machine translation dataset and train a translation model.
+
```
zhangguanheng66, line 2: Should we include a metric for the test/valid datasets with the trained model? See my comments about `bleu_score` below.

akurniawan: That would be better. However, I may not be able to run a full blown training as my resource is quite limited. Do you have any suggestions?

zhangguanheng66: Never mind. I will find time and work on it this half. Then, I can update this. Just to make sure that you set up the model/training correctly by checking the learning curve.

akurniawan: Got it. Sorry for the trouble 🙏

examples/machine_translation/train.py
akurniawan: done

examples/machine_translation/README.md
akurniawan: borrowed a resource to run 10 epochs, already putting the result on README. wdyt?

examples/machine_translation/train.py
```diff
@@ -0,0 +1,156 @@
...
+
+from dataset import get_dataset
+from embedding import WordCharCNNEmbedding
+from model import Attention, Decoder, Encoder, Seq2Seq
+from torchtext.data.metrics import bleu_score
+from torchtext.vocab import Vocab
+from utils import collate_fn, count_parameters, epoch_time, seed_everything
+
+# Ensure reproducibility
+seed_everything(42)
```
zhangguanheng66, line 18: We should allow users to set up the see, like [this](https://github.com/pytorch/text/blob/441edff8743cc03d2d8676029f5bea3c0ed76486/examples/BERT/mlm_task.py#L240)

examples/machine_translation/train.py
```diff
@@ -0,0 +1,156 @@
...
+
+    # enc_emb_dim = 256
+    # dec_emb_dim = 256
+    # enc_hid_dim = 512
+    # dec_hid_dim = 512
+    # attn_dim = 64
+    # enc_dropout = 0.5
+    # dec_dropout = 0.5
+
+    enc_emb_dim = 300
```
zhangguanheng66, line 110: We use `parser` to set up those hyperparameters, like [this](https://github.com/pytorch/text/blob/441edff8743cc03d2d8676029f5bea3c0ed76486/examples/BERT/mlm_task.py#L219)

examples/machine_translation/utils.py
```diff
@@ -0,0 +1,62 @@
+import itertools
```
zhangguanheng66, line 1: Are there any generic building blocks that we could add to the library to support other pipeline?

examples/machine_translation/utils.py
```diff
@@ -0,0 +1,62 @@
...
+
+    return final_padding
+
+
+def pad_words(input, pad_idx=1):
+    txt = list(map(torch.tensor, input))
+    return pad_sequence(txt, True, pad_idx)
+
+
+def collate_fn(batch):
```
zhangguanheng66, line 42: In general, we put `collate_fn` together with DataLoader in a same file.

examples/machine_translation/utils.py
akurniawan: `pad_chars` and `pad_words` would essentially be useful for other pipeline for sure, but for `pad_chars` not really sure how often people are actually using it compares to subword/word level. the rest (`epoch_time`, `count_parameters`, and `seed_everything`) are typically available on training framework such as pytorch-lightning, ignite, etc, I don't think we need to re-implement them in torchtext, what do you think?

examples/machine_translation/utils.py
akurniawan: moved to train.py

examples/machine_translation/train.py
akurniawan: added argparse

examples/machine_translation/train.py
akurniawan: merged with other hyperparams with argparse

akurniawan:
@zhangguanheng66 just to let you know, this is ready for review. Thanks!

examples/machine_translation/README.md
zhangguanheng66: It might be a little bit too long. Should we just include the final test result?

examples/machine_translation/README.md
```diff
@@ -0,0 +1,141 @@
...
+Users will use the training data in the raw file from Multi30k dataset to train a machine translation model with the character composition method.
+
+To try the example, simply run the following commands:
+
+```bash
+python train_char.py
+```
+
+For character level training, and
```
zhangguanheng66, line 11: What's the difference between "character" level vs "word" level training?

examples/machine_translation/char_dataset.py
```diff
@@ -0,0 +1,121 @@
...
+        [init_sent_token],
+        [eos_sent_token],
+    ]
+    for line in data:
+        tokens = list(itertools.chain.from_iterable(transforms(line[index])))
+        tok_list.append(tokens)
+    return build_vocab_from_iterator(tok_list)
+
+
+def char_vocab_func(vocab):
```
zhangguanheng66, line 36: Is this something like subword method? Do you think if sentencepiece method is easier and more helpful?

examples/machine_translation/char_dataset.py
```diff
@@ -0,0 +1,121 @@
+import itertools
+import re
+
+import torch
+from torch.utils.data import DataLoader
+
+from torchtext.data.utils import get_tokenizer
+from torchtext.experimental.datasets.translation import (DATASETS,
```
zhangguanheng66, line 8: You should be able to load the raw datasets from `torchtext.experimental.datasets.raw` directly.

examples/machine_translation/char_dataset.py
akurniawan: By subword method you mean tokenizing sentence into a list of subwords? Then no. This is a function with similar functionality with `vocab_func`, converting a list of strings to their index, except instead of converting words to string we are converting a list of chars to their index. To give you an example


```
sentence: I love torchtext
character representation: [['I'], ['l', 'o', 'v', 'e'], ['t', 'o', 'r', 'c', 'h', 't', 'e', 'x', 't']]
char_vocab_func(character representation): [[1], [2, 3, 4], [5, 6, 7, 8, 9, 10, 11, 12, 13]]
```

examples/machine_translation/char_dataset.py
akurniawan: So just to be clear, it's not recommended/intended to use the raw datasets from `torchtext.experimental.datasets`?

akurniawan:
@zhangguanheng66 I got `AttributeError: 'Vocab' object has no attribute 'insert_tokens'` in following the script. I have tried to pull the latest changes but seems still no `insert_tokens`, on vocab object

examples/machine_translation/README.md
akurniawan: added more explanation

examples/machine_translation/README.md
akurniawan: Instead of removing the whole training metrics from the docs I trim it so that I only include the first and the last training output to give users some idea on the loss value while running the example

zhangguanheng66:
@akurniawan Just to check in and see if you have problems.

akurniawan:
@zhangguanheng66 yes, I have been trying to follow the code that you have given above, the one created by @cpuhrsch. It throws an error `AttributeError: 'Vocab' object has no attribute 'insert_tokens'`. I take a look at both vocab objects (one in the legacy and one in the experimental) and both don't have `insert_tokens` and `insert` method as well. I'm not sure if that was only a pseudocode to insert tokens after building vocabulary or it's a real methods that haven't been implemented yet. So as of now I'm still keeping the old way to build the vocab

zhangguanheng66:
> @zhangguanheng66 yes, I have been trying to follow the code that you have given above, the one created by @cpuhrsch. It throws an error `AttributeError: 'Vocab' object has no attribute 'insert_tokens'`. I take a look at both vocab objects (one in the legacy and one in the experimental) and both don't have `insert_tokens` and `insert` method as well. I'm not sure if that was only a pseudocode to insert tokens after building vocabulary or it's a real methods that haven't been implemented yet. So as of now I'm still keeping the old way to build the vocab

Could you point out where `insert_tokens` is used in your code? And you are using the vocab class in the main folder, not experimental folder, right?

akurniawan:
> Could you point out where `insert_tokens` is used in your code?

It's not being used right now on the code as it throws the error I mentioned earlier. But I was following [this](https://gist.github.com/cpuhrsch/2ce6358ce8c44caacb47d0ad180d3923) implementation on line 80 and 84.

> And you are using the vocab class in the main folder, not experimental folder, right?

yes correct

zhangguanheng66:
> > Could you point out where `insert_tokens` is used in your code?
> 
> It's not being used right now on the code as it throws the error I mentioned earlier. But I was following [this](https://gist.github.com/cpuhrsch/2ce6358ce8c44caacb47d0ad180d3923) implementation on line 80 and 84.
> 
> > And you are using the vocab class in the main folder, not experimental folder, right?
> 
> yes correct

The Vocab in the main folder doesn't have `insert_token` method, however, the one in experimental folder has. Overall, if you don't need to insert tokens after building the vocab instance, you don't need to use that part of code.

akurniawan:
> The Vocab in the main folder doesn't have `insert_token` method, however, the one in experimental folder has. Overall, no sure why you got that error because you never call `insert_token` method. Could you show me the full error message chain?

Sorry for not being clear. So basically this is what I did. After you made the comment for using `map` and `partial` to tokenize the character-level representation, I tried to make changes in my local by following exactly the example code (the one that contains `insert_tokens` and `insert` method for inserting vocabulary). It throws error that both `insert_tokens` and `insert` were not found. Then because of that error, I removed the implementation back. The difference between the example you gave and the current implementation is on the way we insert the special tokens.

On the example code, we do it this way
```py
    train, _, _ = DATASETS[dataset_name]()
    src_char_vocab = build_char_vocab(train, src_char_transform, index=0)
    src_char_vocab.insert_tokens([init_word_token, eos_word_token, init_sent_token, eos_sent_token], 0)

    train, _, _ = DATASETS[dataset_name]()
    tgt_char_vocab = build_char_vocab(train, tgt_char_transform, index=1)
    tgt_char_vocab.insert_tokens([init_word_token, eos_word_token, init_sent_token, eos_sent_token], 0)

    train, _, _ = DATASETS[dataset_name]()
    tgt_word_vocab = build_vocab_from_iterator(iter(map(lambda tgt_word_transform(x[0]), train)))
    tgt_word_vocab.insert(eos_word_token, 0)
    tgt_word_vocab.insert(init_word_token, 0)
```
We have `insert_tokens` and `insert` to insert special tokens both for char and word level. When I run this, it throws errors as both methods are not available. Therefore, I revert back and still use the following way

```py
def build_word_vocab(data, transforms, index, init_token="<w>", eos_token="</w>"):
    tok_list = [[init_token], [eos_token]]
    return build_vocab_from_iterator(tok_list + list(map(lambda x: transforms(x[index]), data)))


def build_char_vocab(
    data, transforms, index, init_word_token="<w>", eos_word_token="</w>", init_sent_token="<s>", eos_sent_token="</s>",
):
    tok_list = [
        [init_word_token],
        [eos_word_token],
        [init_sent_token],
        [eos_sent_token],
    ]
    for line in data:
        tokens = list(itertools.chain.from_iterable(transforms(line[index])))
        tok_list.append(tokens)
    return build_vocab_from_iterator(tok_list)
```
Where I don't use `insert_tokens` and `insert` method to put the special tokens

zhangguanheng66:
> > The Vocab in the main folder doesn't have `insert_token` method, however, the one in experimental folder has. Overall, no sure why you got that error because you never call `insert_token` method. Could you show me the full error message chain?
> 
> Sorry for not being clear. So basically this is what I did. After you made the comment for using `map` and `partial` to tokenize the character-level representation, I tried to make changes in my local by following exactly the example code (the one that contains `insert_tokens` and `insert` method for inserting vocabulary). It throws error that both `insert_tokens` and `insert` were not found. Then because of that error, I removed the implementation back. The difference between the example you gave and the current implementation is on the way we insert the special tokens.
> 
> On the example code, we do it this way
> 
> ```python
>     train, _, _ = DATASETS[dataset_name]()
>     src_char_vocab = build_char_vocab(train, src_char_transform, index=0)
>     src_char_vocab.insert_tokens([init_word_token, eos_word_token, init_sent_token, eos_sent_token], 0)
> 
>     train, _, _ = DATASETS[dataset_name]()
>     tgt_char_vocab = build_char_vocab(train, tgt_char_transform, index=1)
>     tgt_char_vocab.insert_tokens([init_word_token, eos_word_token, init_sent_token, eos_sent_token], 0)
> 
>     train, _, _ = DATASETS[dataset_name]()
>     tgt_word_vocab = build_vocab_from_iterator(iter(map(lambda tgt_word_transform(x[0]), train)))
>     tgt_word_vocab.insert(eos_word_token, 0)
>     tgt_word_vocab.insert(init_word_token, 0)
> ```
> 
> We have `insert_tokens` and `insert` to insert special tokens both for char and word level. When I run this, it throws errors as both methods are not available. Therefore, I revert back and still use the following way
> 
> ```python
> def build_word_vocab(data, transforms, index, init_token="<w>", eos_token="</w>"):
>     tok_list = [[init_token], [eos_token]]
>     return build_vocab_from_iterator(tok_list + list(map(lambda x: transforms(x[index]), data)))
> 
> 
> def build_char_vocab(
>     data, transforms, index, init_word_token="<w>", eos_word_token="</w>", init_sent_token="<s>", eos_sent_token="</s>",
> ):
>     tok_list = [
>         [init_word_token],
>         [eos_word_token],
>         [init_sent_token],
>         [eos_sent_token],
>     ]
>     for line in data:
>         tokens = list(itertools.chain.from_iterable(transforms(line[index])))
>         tok_list.append(tokens)
>     return build_vocab_from_iterator(tok_list)
> ```
> 
> Where I don't use `insert_tokens` and `insert` method to put the special tokens

Feel free to use whatever you see it works.

akurniawan:
> Feel free to use whatever you see it works.

Cool, it's ready for review then

PR #866, opened 2020-07-02
Added new factory functions for Vocab class

Nayef211:
## Description
- Added new factory function for creating a Vocab object from a csv file
- Added unit tests to test factory function
- Disabling GloVe tests due to URL breakage


torchtext/csrc/vocab.cpp
```diff
@@ -0,0 +1,142 @@
...
+      throw std::runtime_error(
+          "Specified index " + std::to_string(index) +
+          " is out of bounds of the size of itos dictionary: " +
+          std::to_string(itos_.size()) + ".");
+    }
+
+    return itos_[index];
+  }
+
+  std::vector<std::string> lookupTokens(const std::vector<int64_t> &indices) {
```
zhangguanheng66, line 94: Not sure if `itos` is really necessary, especially a list of index to a list of tokens.

torchtext/experimental/vocab.py
```diff
@@ -0,0 +1,177 @@
...
+            raise ValueError("A default unk token wasn't provided.")
+
+        tokens = []
+        for token, freq in ordered_dict.items():
+            if freq >= min_freq:
+                tokens.append(token)
+
+        # assume unk_token and special tokens dont appear in ordered_dict
+        if specials_first:
+            tokens = [unk_token] + list(specials) + tokens
```
zhangguanheng66, line 79: `specials` tokens don't include `unk_token`, which is in counterintuitive. Here we are assuming that '<unk>' is always indexed by "0". What's the solution for the following case:
a "pre-trained" vocab from a list of tokens - ['<pad>', '<sep>', '<unk>', 'this', 'is', 'pytorch']

torchtext/csrc/vocab.cpp
Nayef211: So the `itos` is necessary because when we are serializing a Vocab we need to somehow maintain the ordering for all the tokens (this is currently done by the `itos` vector). Without the `itos` we wouldn't be able to deserialize the saved model in the correct order.

torchtext/experimental/vocab.py
Nayef211: This is a good point. After our offline discussion, it seems like the best approach is to ensure that the `unk_token` is in the `specials` list and that the name of the `unk_token` used is provided.

torchtext/experimental/vocab.py
```diff
@@ -1,12 +1,57 @@
...
+def _infer_shape(f):
+    num_lines = 0
+    for line in f:
+        num_lines += 1
+    f.seek(0)
+    return num_lines
+
+
+def vocab_from_file_object(file_like_object, **kwargs):
+    r"""Create a `Vocab` object from a csv file like object.
```
cpuhrsch, line 21: I'd separate the "csv" and "file like object" part. I'd say this accepts a file like object and then describe the data protocol which is separate lines in csv format.

Nayef211: Done!

torchtext/experimental/vocab.py
```diff
@@ -1,12 +1,58 @@
...
+    for line in f:
+        num_lines += 1
+    f.seek(0)
+    return num_lines
+
+
+def vocab_from_file_object(file_like_object, **kwargs):
+    r"""Create a `Vocab` object from a file like object.
+
+    Note that the tensor corresponding to each vector is of type `torch.float`.
```
zhangguanheng66, line 23: Can this be a txt file, which is a column of tokens? Or it must be a csv file?

Nayef211: This could be a `txt` file. Let me update the docs to reflect that!

Nayef211:
> Let disable the glove vector tests in a separate PR and link to the glove repo issue. It's easier for us to track.

Done!

PR #867, opened 2020-07-03
Experimental pytext vocab examples

Nayef211:
## Description
- Added example for pytext `ScriptVocab`
- Added example for pytext `build_fairseq_vocab`
- Added code to benchmark pytext's `Vocab` and `ScriptVocab` against torchtext's `ScriptVocab`

## Benchmark Results
For each benchmark we have 3 different types of looksups:
- single token lookups
- list lookup for single tokens list
- list lookup for line by line tokens list

```
Tokens size: 5193609
Tokens list size: 120000

Pytext Vocabulary
Construction time: 0.02325898502022028

Pytext Script Vocabulary
Construction time: 0.26593813486397266

Experimental Script Vocabulary
Construction time: 0.1811453679110855

Pytext Vocabulary - Eager Mode
Lookup time: 6.751540813129395
Lookup time: 4.871603666804731
Lookup time: 4.9949317870195955

Pytext ScriptVocab - Eager Mode
Lookup time: 369.67893458600156
Lookup time: 31.241876950021833
Lookup time: 40.4979264291469

Experimental ScriptVocab - Eager Mode
Lookup time: 81.39808336622082
Lookup time: 7.71178909111768
Lookup time: 14.842586216982454

Pytext ScriptVocab - Jit Mode
Lookup time: 362.0857468259055
Lookup time: 31.17033457197249
Lookup time: 41.13285271497443

Experimental ScriptVocab - Jit Mode
Lookup time: 216.8087242541369
Lookup time: 7.972830525133759
Lookup time: 15.07782193692401

Pytext ScriptVocab - Jit For Loop
Lookup time: 4.288848091848195
Lookup time: 4.175843314966187
Lookup time: 5.102244307054207

Experimental ScriptVocab - Jit For Loop
Lookup time: 33.920223012100905
Lookup time: 7.243474407820031
Lookup time: 8.581724852789193
```


examples/vocab/script_vocab.py
```diff
@@ -0,0 +1,113 @@
+from collections import OrderedDict
+import torch
+from torchtext.experimental.vocab import Vocab
+from typing import List, Optional
+
+
+class ScriptVocab(Vocab):
```
zhangguanheng66, line 7: How about the special symbols?
https://github.com/facebookresearch/pytext/blob/75acf0d0508e1b75346ba0950b2432b8bb53af40/pytext/torchscript/vocab.py#L24-L28

Nayef211: I added these now!

examples/data_pipeline/benchmark_pytext_vocab.py
```diff
@@ -0,0 +1,185 @@
...
+
+import torch
+from torchtext.experimental.datasets import AG_NEWS
+
+from pytext_vocab import ScriptVocab as ExperimentalScriptVocabulary
+from pytext.torchscript.vocab import ScriptVocabulary as PytextScriptVocabulary
+from pytext.data.utils import Vocabulary as PytextVocabulary
+
+
+def _run_benchmark_lookup(tokens, vocab, num_iters=1):
```
zhangguanheng66, line 13: Can we consolidate those `_run_benchmark_lookup*` funcs?

Nayef211: I did my best to consolidate some of these functions. I still need to keep separate functions for the JITd for loop. Lmk what you think

benchmark/benchmark_pytext_vocab.py
```diff
@@ -0,0 +1,199 @@
...
+from collections import (Counter, OrderedDict)
+import time
+from typing import List, Union
+
+# this is needed because we want to add 'torchtext/examples/data_pipeline' directory to the
+# `sys.path` variable in order to import the pytext_vocab (since its not a module)
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "examples", "data_pipeline"))
+
+from pytext_vocab import ScriptVocab as ExperimentalScriptVocabulary
+from pytext.torchscript.vocab import ScriptVocabulary as PytextScriptVocabulary
```
zhangguanheng66, line 12: Do we add the dependency of the pytext library.

examples/data_pipeline/pytext_vocab.py
```diff
@@ -0,0 +1,183 @@
+# from collections import OrderedDict
+
+# from fairseq.data.dictionary import Dictionary
+import torch
+from torchtext.experimental.vocab import Vocab
+from typing import List, Optional
+
+
+# TODO: This function is commented out since we don't have fairseq as a dependency in torchtext.
```
cpuhrsch, line 9: The dependencies only really apply at a package level, it's ok to add it for the examples.

benchmark/benchmark_pytext_vocab.py
Nayef211: I didn't do this because I saw that your pipeline benchmarking PR with the `ScriptVocabulary` from pytext didn't add the dependency either. Do you think I should? Especially because this benchmark would just be for an example class that we aren't maintaining anyways?

zhangguanheng66: We probably won't add pytext dependency because we don't use any work from pytext in our main library.

examples/data_pipeline/pytext_vocab.py
zhangguanheng66: And should we uncomment `build_fairseq_vocab` func below as an example for pytext to load fairseq vocab?

Nayef211: Done

Nayef211:
> Should we move the ScriptVocab class to the benchmark folder? Previously, I thought I will use it in my pipeline work. But now I just use the one from pytext library so we can safely move it back to benchmark folder.

I moved the file to `experimental/vocab`.

PR #885, opened 2020-07-15
Adding function for reading vectors file in cpp

Nayef211:
## Description
- Moving `_load_token_and_vectors_from_file()` function from reading vectors file from python to C++
- Using `at::launch` multithreading for loading into tokens and vectors
- Creating a new factory method for constructing Vectors
- Creating a new C++ constructor to directly create a Vectors object from output of `_load_token_and_vectors_from_file()`
- Updating tests to use new Vectors factory method

### Construction from file benchmark
```
Existing FastText - Not Jit Mode 
Construction time: 875.7055841560941
FastText Experimental
Construction time: 29.879692571004853
```

### Construction from cache benchmark
```
Existing FastText - Not Jit Mode
Construction time: 7.549301242921501
FastText Experimental
Construction time: 8.950354253873229
```

### Lookup Benchmark
```
FastText Existing - Eager Mode
Lookup time: 173.31978496885858

FastText Experimental - Eager Mode
Lookup time: 88.13051658077165

FastText Experimental - Jit Mode
Lookup time: 199.40745506016538
```

## All Experiments
image


torchtext/experimental/vectors.py
```diff
@@ -111,7 +112,7 @@ def FastText(language="en", unk_tensor=None, root=".data", validate_file=True):
         checksum = CHECKSUMS_FAST_TEXT.get(url, None)
 
     downloaded_file_path = download_from_url(url, root=root, hash_value=checksum)
-    tokens, vectors, dup_tokens = _load_token_and_vectors_from_file(downloaded_file_path)
+    tokens, vectors, dup_tokens = torch.ops.torchtext._load_token_and_vectors_from_file(downloaded_file_path, ord(' '), 5)
```
cpuhrsch, line 115: Why is using "ord" necessary?

torchtext/csrc/vectors.cpp
```diff
@@ -70,6 +78,213 @@ struct Vectors : torch::CustomClassHolder {
...
+  }
+
+  LoadedVectorsTuple out_tuple;
+  concat_loaded_vectors_tuples(tuples, num_lines, vector_dim, &out_tuple);
+  return out_tuple;
+}
+
+// Registers our custom op with torch.
+TORCH_LIBRARY(torchtext, m) {
+  m.def("_load_token_and_vectors_from_file",
```
zhangguanheng66, line 284: why do we have underline here?

cpuhrsch: This means it's private, i.e. not supposed to be used directly as a public API.

torchtext/experimental/vectors.py
Nayef211: Spoke to the torchscript team and `char` isn't a supported type. Also thought using a string would be too hacky as there is the potential for the user to pass in multiple characters

torchtext/csrc/vectors.cpp
```diff
@@ -70,6 +78,213 @@ struct Vectors : torch::CustomClassHolder {
...
+    std::getline(s, token, static_cast<char>(delimiter_ascii));
+
+    // read every value of the vector and
+    // store it in a string variable, 'vec_val'
+    while (std::getline(s, vec_val, ' ')) {
+      vec_float.push_back(std::stof(vec_val));
+    }
+
+    if (vec_float.size() == 1) {
+      std::cout << "Skipping token " << token
```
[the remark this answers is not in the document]
Nayef211: I tried to use glog (i.e. LOG(INFO)), but it doesn't seem like it was setup. But sure I can log to cerr.

torchtext/experimental/vectors.py
```diff
@@ -111,13 +111,13 @@ def FastText(language="en", unk_tensor=None, root=".data", validate_file=True):
         checksum = CHECKSUMS_FAST_TEXT.get(url, None)
 
     downloaded_file_path = download_from_url(url, root=root, hash_value=checksum)
-    tokens, vectors, dup_tokens = _load_token_and_vectors_from_file(downloaded_file_path)
+    tokens, vectors, dup_tokens = torch.ops.torchtext._load_token_and_vectors_from_file(downloaded_file_path, ord(' '), 10)
```
cpuhrsch, line 114: Why do you have to use "ord" here instead of passing the object as a string?

torchtext/csrc/vectors.cpp
```diff
@@ -70,6 +87,231 @@ struct Vectors : torch::CustomClassHolder {
...
+  std::unordered_set<std::string> tokens_set;
+  std::string line, token, vec_val;
+  int64_t num_vecs_loaded = 0;
+
+  tokens.reserve(num_lines);
+  vectors.reserve(num_lines);
+
+  // get to line we care about
+  for (int64_t i = 0; i < start_line; i++) {
+    std::getline(fin, line);
```
cpuhrsch, line 160: You can probably just use http://www.cplusplus.com/reference/istream/istream/ignore/ instead of actually writing something into the line. Ignore characters one newline at a time.

torchtext/experimental/vectors.py
cpuhrsch: In theory a delimiters of multiple characters isn't necessarily bad, right? I think it's more general to pass a string and then error out if it has more than length 1, so that in the future we could extend support to more complex delimiters instead of inherently limiting ourselves to one character.

torchtext/experimental/vectors.py
Nayef211: Read comment above

torchtext/experimental/vectors.py
```diff
@@ -149,18 +83,19 @@ def GloVe(name="840B", dim=300, unk_tensor=None, root=".data", validate_file=Tru
...
 
     Raises:
         ValueError: if unexpected duplicate tokens are found in GloVe file.
 
     """
-    dup_token_glove_840b = [("����������������������������������������������������������������������"
-                             "����������������������������������������������������������������������"
-                             "����������������������������������������������������������������������"
-                             "����������������������������������������������������������������������"
-                             "������������������������������������������������������", 140649)]
```
cpuhrsch, line 163: What happened here?

torchtext/experimental/vectors.py
```diff
@@ -233,37 +168,38 @@ def vectors_from_file_object(file_like_object, delimiter=",", unk_tensor=None):
         ValueError: if duplicate tokens are found in FastText file.
 
     """
-    tokens, vectors, dup_tokens = _load_token_and_vectors_from_file(file_like_object.name, delimiter=delimiter)
```
cpuhrsch, line 236: Still not a big fan of this "ord(delimiter)" thing. Why not just a string?

torchtext/csrc/vectors.cpp
```diff
@@ -43,6 +58,12 @@ struct Vectors : torch::CustomClassHolder {
     if (item != stovec_.end()) {
       return item->value();
     }
+    const auto &item_index = stoindex_.find(token);
```
cpuhrsch, line 61: This change deserves verifying that there's no regression in performance.

torchtext/csrc/vectors.cpp
```diff
@@ -60,16 +81,195 @@ struct Vectors : torch::CustomClassHolder {
...
+    } else {
+      num_lines++;
+    }
+  }
+  return std::make_tuple(num_lines, num_header_lines, vector_dim);
+}
+
+void _infer_offsets(const std::string &file_path, int64_t num_lines,
+                    int64_t chunk_size, int64_t num_header_lines,
+                    std::vector<size_t> *offsets) {
```
cpuhrsch, line 137: Why not pass by reference?

torchtext/csrc/vectors.cpp
```diff
@@ -60,16 +81,195 @@ struct Vectors : torch::CustomClassHolder {
...
+    // read the token
+    std::getline(fin, token, static_cast<char>(delimiter_ascii));
+    tokens->push_back(token);
+
+    std::string vec_val;
+    // read the vector
+    for (int64_t j = 0; j < vector_dim; j++) {
+      fin >> vec_val;
+      const char *tmp_str = vec_val.c_str();
+      int processed_characters_count;
```
cpuhrsch, line 179: We could TORCH_CHECK that the number of processed characters matches the length of the string.

torchtext/experimental/vectors.py
Nayef211: Reposting my answer from your past comments

> Spoke to the torchscript team and char isn't a supported type. Also thought using a string would be too hacky as there is the potential for the user to pass in multiple characters

Basically, the `std::getline(s, word, static_cast<char>(delimiter_ascii));` expects a char for the delimiter. Do you still think it would be better to use a string in this case?

torchtext/csrc/vectors.cpp
```diff
@@ -60,16 +81,195 @@ struct Vectors : torch::CustomClassHolder {
...
+  IndexDict dict;
+  StringList dup_tokens;
+  std::tie(dict, dup_tokens) =
+      _concat_vectors(chunk_tokens, data_tensor, num_header_lines, num_lines);
+
+  torch::Tensor unk_tensor;
+  if (opt_unk_tensor) {
+    unk_tensor = *opt_unk_tensor;
+  } else {
+    unk_tensor = torch::zeros({vector_dim});
```
cpuhrsch, line 265: I'd also explicitly specify that you want float32 as a dtype.

torchtext/csrc/vectors.cpp
```diff
@@ -81,15 +281,13 @@ static auto vectors =
...
         .def_pickle(
             // __setstate__
-            [](const c10::intrusive_ptr<Vectors> &self)
-                -> std::tuple<std::vector<std::string>, torch::Tensor,
-                              torch::Tensor> {
-              std::tuple<std::vector<std::string>, torch::Tensor, torch::Tensor>
-                  states(self->tokens_, self->vectors_, self->unk_tensor_);
-              return states;
+            [](const c10::intrusive_ptr<Vectors> &self) -> std::tuple<
+                Dict<std::string, int64_t>, torch::Tensor, torch::Tensor> {
```
cpuhrsch, line 285: I'm not sure it's a good idea to serialize the dict vs. just the items, in terms of performance but also BC compatibility, since Dict is a JIT type.

torchtext/csrc/vectors.cpp
Nayef211: I was just trying to adhere to Google's C++ standard,

> In fact it is a very strong convention in Google code that input arguments are values or const references while output arguments are pointers.

https://drake.mit.edu/styleguide/cppguide.html#Reference_Arguments

torchtext/experimental/vectors.py
Nayef211: I believe we discussed the fact that torchbind doesn't allow inner tuple return types which is why I had to remove the tuple of (dup_tokens, line_number). If you still think this is very important, I could return 2 separate `std::vectors`, one containing dup_tokens and the other containing dup line numbers?

torchtext/csrc/vectors.cpp
Nayef211: Sure, I will rerun our benchmark

torchtext/csrc/vectors.cpp
Nayef211: Basically forgot to merge in the latest changes from master. It should be fixed now

torchtext/csrc/vectors.cpp
mthrok: > I was just trying to adhere to Google's C++ standard,
> 
> > In fact it is a very strong convention in Google code that input arguments are values or const references while output arguments are pointers.
> 
> https://drake.mit.edu/styleguide/cppguide.html#Reference_Arguments

I do not see the corresponding one in the official style guide.

https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

Nayef211:
> Should we use this func as an opportunity to benchmark the func to read strings backed by pybind vs torchbind?

I don't want to keep increasing the scope of this PR. Let's do this in a follow up PR?

torchtext/experimental/vectors.py
cpuhrsch: Yes, I think a string would be better and more future proof.

torchtext/csrc/vectors.cpp
Nayef211: Will do this in the follow-up PR that adds the double conversion library!

.gitmodules
```diff
@@ -5,5 +5,4 @@
 [submodule "third_party/re2"]
 	path = third_party/re2
 	url = https://github.com/google/re2
-	ignore = dirty
-
+	ignore = dirty
```
cpuhrsch, line 8: This seems like an unrelated change

third_party/CMakeLists.txt
```diff
@@ -8,4 +8,4 @@ endif()
 project(thirdparty CXX)
 set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 
-add_subdirectory(re2)
```
cpuhrsch, line 11: Same here

torchtext/experimental/vectors.py
cpuhrsch: This change seems to remove the ability to also check the line number the duplicate was found on. That's an important check, so whatever need to do to maintain that seems important.

.gitmodules
Nayef211: I think this is just a missing newline issue. Should be fixed

torchtext/csrc/vectors.cpp
Nayef211: You're right maybe they changed the official guide since the last time I read over it! Updated to use pass by reference.

PR #889, opened 2020-07-16
Updating serialization for experimental Vocab and Vectors

Nayef211:
## Description
- Updated serialization format for experimental vectors and vocab
- We now use a tuple of 
  - `version_string` (to determine the number of parameters in the vocab constructor)
  - `std::vector<int64_t>`
  - `std::vector<std::string>`
  - `std::vector<torch::Tensor>`


torchtext/csrc/vectors.cpp
```diff
@@ -81,21 +112,21 @@ static auto vectors =
         .def("__len__", &Vectors::__len__)
         .def_pickle(
             // __setstate__
-            [](const c10::intrusive_ptr<Vectors> &self)
-                -> std::tuple<std::vector<std::string>, torch::Tensor,
-                              torch::Tensor> {
-              std::tuple<std::vector<std::string>, torch::Tensor, torch::Tensor>
-                  states(self->tokens_, self->vectors_, self->unk_tensor_);
+            [](const c10::intrusive_ptr<Vectors> &self) -> VectorsStates {
+              std::string version_str = "0.0.1";
```
cpuhrsch, line 116: I'd make the version a constexpr at the top of the file

cpuhrsch: for more visibility

torchtext/csrc/vectors.cpp
```diff
@@ -70,6 +75,32 @@ struct Vectors : torch::CustomClassHolder {
   int64_t __len__() { return stovec_.size(); }
 };
 
+c10::intrusive_ptr<Vectors> _get_vectors_from_states(VectorsStates states) {
+  auto state_size = std::tuple_size<decltype(states)>::value;
+  if (state_size == 0) {
+    throw std::runtime_error(
+        "Expected deserialized Vectors to have a version string.");
```
cpuhrsch, line 82: Is this still the right error message?

torchtext/csrc/vocab.cpp
```diff
@@ -111,6 +115,34 @@ struct Vocab : torch::CustomClassHolder {
...
+    throw std::runtime_error(
+        "Expected deserialized Vocab to have a version string.");
+  } else if (state_size != 4) {
+    throw std::runtime_error(
+        "Expected deserialized Vocab to have 4 states but found only " +
+        std::to_string(state_size) + " states.");
+  }
+
+  auto &version_str = std::get<0>(states);
+  // auto &integers = std::get<1>(states);
```
cpuhrsch, line 130: I'd still load these and double check that they are empty

torchtext/csrc/vocab.cpp
```diff
@@ -125,18 +157,24 @@ static auto vocab =
...
         .def("get_itos", &Vocab::get_itos)
         .def_pickle(
-            // __getstate__
-            [](const c10::intrusive_ptr<Vocab> &self)
-                -> std::tuple<std::vector<std::string>, std::string> {
-              std::tuple<std::vector<std::string>, std::string> states(
-                  self->itos_, self->unk_token_);
+            // __setstate__
+            [](const c10::intrusive_ptr<Vocab> &self) -> VocabStates {
+              std::string version_str = "0.0.1";
```
cpuhrsch, line 162: You could turn this into a function "_get_states_from_vocab" or such. Then we have the save and load both abstracted into separate functions.

torchtext/csrc/vectors.cpp
Nayef211: Removed this altogether

torchtext/csrc/vectors.cpp
```diff
@@ -3,12 +3,19 @@
 #include <torch/script.h>
 
 using c10::Dict;
+using c10::IValue;
```
cpuhrsch, line 6: nit: Are you still using IValues? If not, you can remove this.

torchtext/csrc/vectors.cpp
```diff
@@ -3,12 +3,19 @@
...
 namespace torchtext {
 namespace {
 
+typedef std::tuple<std::string, std::vector<int64_t>, std::vector<std::string>,
+                   std::vector<torch::Tensor>>
+    VectorsStates;
+
 struct Vectors : torch::CustomClassHolder {
 public:
+  const std::string version_str_ = "0.0.1";
```
cpuhrsch, line 17: I think you can make this a constexpr, if you want:  https://stackoverflow.com/questions/46100310/how-to-declare-constexpr-c-string

torchtext/csrc/vectors.cpp
```diff
@@ -3,12 +3,19 @@
 #include <torch/script.h>
 
 using c10::Dict;
+using c10::IValue;
 
 namespace torchtext {
 namespace {
 
+typedef std::tuple<std::string, std::vector<int64_t>, std::vector<std::string>,
```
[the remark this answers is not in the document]
Nayef211: This is true and I guess this is a tradeoff that we have to accept if we want to take this approach. Do you have other suggestions?

torchtext/csrc/vectors.cpp
Nayef211: Got it. I wasn't fully aware of what the difference btw `const` and `consexpr` was but from what I understand, a constexpr variable must be initialized at compile time rather than run time which should help the program run faster and use less memory.

Nayef211: Not going to pursue this as per our offline conversation. Using `constexpr` leads to needing to use literals (i.e. `char[]`). This means we need to do conversions to std::string anyways which isn't worth the potential perf gains.

torchtext/csrc/vectors.cpp
```diff
@@ -70,6 +76,45 @@ struct Vectors : torch::CustomClassHolder {
...
+  return states;
+}
+
+c10::intrusive_ptr<Vectors> _get_vectors_from_states(VectorsStates states) {
+  auto state_size = std::tuple_size<decltype(states)>::value;
+  if (state_size != 4) {
+    throw std::runtime_error(
+        "Expected deserialized Vectors to have 4 states but found only " +
+        std::to_string(state_size) + " states.");
+  }
```
mthrok, line 97: Hi @Nayef211 Could you help me understand this? As far as I understand, `std::tuple_size<decltype(states)>` is resolved at compile time so `state_size` is alway 4.

mthrok: or is there a case that Python can pass tuple with different size?

Nayef211: Hi @mthrok. I don't think there's any Python involved in this part of the code. The `_set_vectors_states ` states function always creates a tuple of size 4 when a `vector` is serialized, and the `_get_vectors_from_states ` always expects 4 states during deserialization. `std::tuple_size<decltype(states)>` actually allows us to detect the number of items inside of states before deserializing. The reason this is done is for backward compatibility since in the future we may change the number of items in the tuple that we serialize in the `states` tuple. If the number of items has indeed changed, then we can return an error message to the user so they understand why they can't deserialize their serialized vector.

Lmk if that helps!

mthrok: @Nayef211 Grad to hear back from you. Hope you are doing well.

I understand the purpose of the check, and what I wonder is does the check really prevent the expected undesirable situation? The function does not seem to implement runtime polymorphism, so the questions I am asking myself are;
1. When passing a different type of object (say, tuple with different number) from Python side, does the program execution reach this function? Does PyBind11 find a way to pass a different type to this function?
2. If PyBind11 manages to pass the wrong object to this function, (say it uses some sort of `reinterpret_cast`-ing), isn't the object still represented as the correct `VectorsStates` ?

So the question to you is "Did you test such undeniable case and verified that the check here detect the situation"? (By asking this, I am not accusing you of anything, I would like to check it with you before I do the test by myself.)

PR #901, opened 2020-07-28
Update data_pipeline example

zhangguanheng66:


examples/data_pipeline/pipelines.py
```diff
@@ -51,37 +51,40 @@ def build_batch_torchtext_vocab(vocab_file):
...
     # Insert token in vocab to match a pretrained vocab
-    pipeline = TextDataPipeline(tokenizer, VocabTransform(vocab))
+    pipeline = TextSequentialTransforms(tokenizer, VocabTransform(vocab), ToLongTensor())
     jit_pipeline = torch.jit.script(pipeline)
     print('jit Hugging Face pipeline success!')
     return pipeline, jit_pipeline
 
 
 def build_pytext_vocab_pipeline(vocab_file):
+    from pytext.torchscript.vocab import ScriptVocabulary
```
Nayef211, line 79: Is there a reason you do the import in the functions instead of at the top of the file?

zhangguanheng66: If we don't run `build_pytext_vocab_pipeline` func, we don't need to add pytext dependency.

PR #907, opened 2020-08-05
.circleci: Only use the channel we want for test

seemethere:
Issues were arising when testing release candidates where we would
segfault based on binaries installed from an incorrect channel. This
makes it so that we install from the pytorch-test channel on tags

Similar to https://github.com/pytorch/audio/pull/859


zhangguanheng66:
Will merge after we fix CI on the master branch.

PR #910, opened 2020-08-06
Added regex tokenizer implementation in cpp

Nayef211:
## Description
- Moved implementation of `RegexTokenizer` into cpp for performance gains
- Using the cpp `RegexTokenizer` for `BasicEnglishNormalize`
- Using compile RE2 patterns for `Tokenizer`
- Added benchmarking code for `BasicEnglishNormalize`

### Legacy
```
Vocab - Eager Mode
Lookup time: 3.0771866608411074
```

### `torchbind`
```
Lookup time: 15.13926866510883

Vocab Experimental - Jit Mode
Lookup time: 14.73105155909434
```

## All Experiments
image


benchmark/benchmark_basic_english_normalize.py
```diff
@@ -0,0 +1,36 @@
+import time
+
+import torch
+from torchtext.experimental.datasets import AG_NEWS
+from torchtext.experimental.transforms import BasicEnglishNormalize
+from torchtext.data.utils import get_tokenizer
+
+
+def benchmark_experimental_vocab():
```
zhangguanheng66, line 9: nit: vocab?

benchmark/benchmark_basic_english_normalize.py
```diff
@@ -0,0 +1,36 @@
...
+        print("Lookup time:", time.monotonic() - t0)
+
+    train, = AG_NEWS(data_select='train')
+    
+    existing_basic_english_tokenizer = get_tokenizer("basic_english")
+    experimental_basic_english_normalize = BasicEnglishNormalize()
+    experimental_jit_basic_english_normalize = torch.jit.script(experimental_basic_english_normalize)
+
+    # existing eager lookup
+    print("Vocab - Eager Mode")
```
zhangguanheng66, line 23: nit: vocab?

benchmark/benchmark_basic_english_normalize.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def benchmark_experimental_vocab():
+    def _run_benchmark_lookup(train, tokenizer):
+        t0 = time.monotonic()
+        for (label, text) in train.data:
+            tokenizer(text)
+        print("Lookup time:", time.monotonic() - t0)
+
+    train, = AG_NEWS(data_select='train')
```
zhangguanheng66, line 16: Since you are calling `torchtext.experimental.datasets.AG_NEWS`, do you receive the processed data (a.k.a. tensor)? Should we use `torchtext.experimental.datasets.raw.AG_NEWS`?

examples/data_pipeline/pipelines.py
```diff
@@ -63,13 +63,17 @@ def token_iterator(vocab_file):
 
 
 def build_huggingface_vocab_pipeline(hf_vocab_file):
-    tokenizer = BasicEnglishNormalize()
+    # tokenizer = BasicEnglishNormalize()
```
zhangguanheng66, line 66: Why do we switch the tokenizer in this case?

torchtext/csrc/regex.h
```diff
@@ -0,0 +1,12 @@
+#include <string>
+#include <torch/script.h>
```
zhangguanheng66, line 2: Are you using this one for pybind2 benchmark?

torchtext/csrc/basic_english_normalize.cpp
```diff
@@ -0,0 +1,49 @@
+#include "basic_english_normalize.h"
+#include <sstream>
+
+namespace torchtext {
+
+BasicEnglishNormalize::BasicEnglishNormalize() {
+  for (const auto &pattern : patterns_) {
+    regex_objects_.push_back(Regex(pattern));
```
cpuhrsch, line 8: Since we have static patterns, I wonder if there are static initializers for RE2 regex objects that pre-compile the regex at compile time.

cpuhrsch: For example, see here: https://chromium.googlesource.com/external/github.com/google/re2/+/refs/heads/master/re2/re2.h#132

```

// Example: precompile pattern for faster matching:
--
  | //    RE2 pattern("h.*o");
  | //    while (ReadLine(&str)) {
  | //      if (RE2::FullMatch(str, pattern)) ...;
  | //    }
  | //


```

Nayef211: Updated this to construct precompiled patterns during construction. Lmk what you think!

torchtext/csrc/regex.h
Nayef211: I have a separate branch for pybind11 that I will create a PR with once this is merged in. Thats the branch I used to get the pybind11 numbers

benchmark/benchmark_basic_english_normalize.py
Nayef211: Fixed

benchmark/benchmark_basic_english_normalize.py
Nayef211: I was doing `train.data` before to get the `(label, text)` tuple but I guess using an iterator for the raw dataset may be the better way to go about it.

Nayef211:
> LGTM. One question for the basic english normalization. Why do we need the cpp extension for `basic_english_normalize` in addition to `Regex` cpp extension? Besides moving those replacement patterns to cpp in `basic_english_normliaze`, other functionals are very similar, right? For example, `RE2::GlobalReplace`. In this case, do we save time for the construction of `basic_english_normalize`? Does this affect the performance of the `forward` function?

The main reason for doing this is that we can have the for loop in the forward function occur on the cpp side instead. It also allows us to use a list precompiled patterns directly rather than create a number of cpp Regex objects. From the experiment results in the description you can see that just by moving the for-loop into cpp we get a speedup from 98 seconds to 25 seconds (which is 4x speedup)

torchtext/csrc/regex.cpp
```diff
@@ -1,26 +1,16 @@
+#include "regex.h"
```
cpuhrsch, line 1: Don't use relative includes

torchtext/csrc/regex.cpp
```diff
@@ -1,26 +1,15 @@
-#include <re2/re2.h>
-#include <string>
-#include <torch/script.h>
+#include <regex.h>
```
zhangguanheng66, line 1: What are those two files `regex.cpp` and `regex.h` for? Where are we use them?

Nayef211: Basically I was thinking of removing the `Regex` class  now since we aren't using it anymore. But @cpuhrsch suggested doing this in a followup PR once we've had more time to evaluate whether we want a light wrapper around the `RE2` library for future use cases. Does that sound reasonable?

zhangguanheng66: Sounds good. Let's keep this as the plan.

PR #914, opened 2020-08-09
Moving function for reading Vocab file to cpp

Nayef211:
## Description
- Moving `_load_vocab_from_file()` function for reading vocab file from python to C++
- Using `at::launch` multithreading for reading tokens
- Creating a new factory method for constructing Vocab
- Creating a new C++ constructor to directly create a Vocab object from output of `_load_vocab_from_file()`
- Updating tests to use new Vocab factory method

## Benchmark Results
image


torchtext/csrc/register_bindings.cpp
```diff
@@ -0,0 +1,80 @@
+#include <regex_tokenizer.h>
+#include <torch/script.h>
+#include <vectors.h>
+#include <vocab.h>
+
+namespace torchtext {
```
zhangguanheng66, line 6: Are we moving sentencepiece here?

torchtext/experimental/vocab.py
```diff
@@ -33,30 +22,25 @@ def vocab_from_file_object(file_like_object, **kwargs):
...
-            ordered_dict[token] = 1
-
-    return Vocab(ordered_dict, **kwargs)
+    vocab_obj = torch.ops.torchtext._load_vocab_from_file(file_like_object.name, unk_token, min_freq, num_cpus)
+    return Vocab(vocab_obj)
 
 
-class Vocab(nn.Module):
-    r"""Creates a vocab object which maps tokens to indices.
+def vocab(ordered_dict, min_freq=1, unk_token='<unk>'):
```
zhangguanheng66, line 42: Why do we move this out of Vocab constructor?

zhangguanheng66, line 42: It is counterintuitive to me. We have Vocab class but we don't prefer users to call it directly. In `torchtext.vocab.py`, we have `build_vocab_from_iterator` func. Should we do the similar thing here like `build_vocab`? @cpuhrsch

torchtext/csrc/register_bindings.cpp
Nayef211: Let me do that in the followup PR since I need to create a header file for sentencepiece

torchtext/experimental/vocab.py
Nayef211: I think this is because we want to follow a similar convention to pytorch `Tensors`. Users are not meant to call the `torch.Tensors` class directly but instead create a tensor object using `torch.tensor()` factory method. This also allows us to directly create the underlying cpp `Vocab` object and pass it into the python `Vocab` class instead of having to pass in params to the `Vocab` class after loading tokens from a file (reduces unnecessary FFI btw cpp->python->cpp->python).

Maybe @cpuhrsch can fill in anything I missed

zhangguanheng66: This pattern "cpp->python->cpp->python" only add minor overhead to the performance of the construction time, right? IMO, `vocab` func does exactly same with what `Vocab` constructor did in previous commit.

torchtext/csrc/vocab.cpp
```diff
@@ -98,6 +98,160 @@ std::vector<int64_t> Vocab::lookup_indices(const StringList &tokens) {
 c10::Dict<std::string, int64_t> Vocab::get_stoi() const { return stoi_; }
 StringList Vocab::get_itos() const { return itos_; }
 
+inline int64_t divup(int64_t x, int64_t y) { return (x + y - 1) / y; }
```
cpuhrsch, line 101: There seems to be significant codeshare between this and vocab. Can we just share these functions with the other file?

torchtext/experimental/vocab.py
Nayef211: I think the FFI's can actually add significant performance overheads due to the copy constructor being called multiple times. The overhead also increases linearly as the size of the Vocab increases. 

Using this factory pattern would also allows us to decouple the cpp Vocab class from the python Vocab class. In the future if we wanted to choose between 2 different underlying cpp classes during construction, we can do so more easily without having to rewrite our `Vocab `constructor.

torchtext/csrc/vocab.cpp
Nayef211: Added some of the duplicated code into `torchtext::impl` namespace. Can you take one more look and lmk what you think?

PR #915, opened 2020-08-11
Create header files for all C++ classes

Nayef211:
## Description
- Creating header files for all C++ classes
- Moving all torchtext binding registrations to a separate file

### DIFF
--- a/torchtext/csrc/register_bindings.cpp
+++ b/torchtext/csrc/register_bindings.cpp
@@ -95,7 +95,8 @@ static auto vectors =
 static auto registry =
     torch::RegisterOperators()
         .op("torchtext::generate_sp_model", &generate_sp_model)
-        .op("torchtext::_load_token_and_vectors_from_file", &_load_token_and_vectors_from_file)
+        .op("torchtext::_load_token_and_vectors_from_file",
+            &_load_token_and_vectors_from_file)
         .op(torch::RegisterOperators::options()
                 .schema("torchtext::load_sp_model(str path) -> "
                         "__torch__.torch.classes.torchtext.SentencePiece model")
--- a/torchtext/csrc/register_bindings.cpp
+++ b/torchtext/csrc/register_bindings.cpp
@@ -1,4 +1,5 @@
 #include <regex_tokenizer.h>
+#include <sentencepiece.h>
 #include <torch/script.h>
 #include <vectors.h>
 #include <vocab.h>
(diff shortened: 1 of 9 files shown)


torchtext/csrc/vectors.cpp
```diff
@@ -203,7 +181,7 @@ void parse_chunk(const std::string &file_path, size_t offset,
 
 std::tuple<IndexMap, StringList>
 _concat_vectors(std::vector<std::shared_ptr<StringList>> chunk_tokens,
-                int64_t num_header_lines, int64_t num_lines) {
+                const int64_t num_header_lines, const int64_t num_lines) {
```
zhangguanheng66, line 184: we are not passing by reference. Do we still need `const` here?

torchtext/csrc/vectors.cpp
```diff
@@ -11,101 +11,79 @@
 #include <stdexcept>
```
zhangguanheng66, line 11: Why not move those header files to `vectors.h` together

torchtext/csrc/vectors.cpp
Nayef211: Good call, I can remove this

torchtext/csrc/vocab.cpp
```diff
@@ -1,119 +1,106 @@
+#include <ATen/Parallel.h>
 #include <stdexcept>
```
zhangguanheng66, line 2: Same question about the header file.

torchtext/csrc/vectors.cpp
Nayef211: I think in general, it is good practice to not include gratuitious `#include` directives. A header file should not be included in a `.h` file if the `.h` file itself doesn't need those other header files.

@cpuhrsch can correct me if I am wrong

torchtext/csrc/vocab.cpp
Nayef211: Same as above ^

torchtext/csrc/vectors.cpp
zhangguanheng66: OK, thanks for the explanation.

torchtext/csrc/vectors.cpp
mthrok: If it's `const`, it's better to be annotated as `const`. Compilers might find an opportunity to optimize something. Also that will prevent accidental overwrite.

torchtext/csrc/register_bindings.cpp
```diff
@@ -0,0 +1,105 @@
+#include <regex_tokenizer.h>
+#include <sentencepiece.h>
+#include <torch/script.h>
+#include <vectors.h>
+#include <vocab.h>
+
+namespace torchtext {
+
```
mthrok, line 8: I think now that the official documentation is updated to use `TORCH_LIBRARY` macro, which makes the registration cleaner and simpler, we can also migrate to this new approach, unless there is some issue about it. Did you try it?

torchtext/csrc/vectors.cpp
Nayef211: Readded

torchtext/csrc/vectors.cpp
mthrok: > I think in general, it is good practice to not include gratuitious `#include` directives. A header file should not be included in a `.h` file if the `.h` file itself doesn't need those other header files.
> 
> @cpuhrsch can correct me if I am wrong

I second that.

torchtext/csrc/register_bindings.cpp
Nayef211: Will do this in a follow up PR!

PR #916, opened 2020-08-11
SentencePieceTokenizer and SentencePieceProcessor

zhangguanheng66:
Pretrained SPM transforms: Generate a pretrained sentencepiece model. The sp model was trained with `torchtext.datasets.WikiText103`, `torchtext.datasets.EnWik9` and BookCorpus. Both BPE and unigram methods were used to train the model (for more details please refer to SentencePiece GitHub [link](https://github.com/google/sentencepiece)). We provide the pretrained model with a different size of the vocabulary (i.e. 15000, 25000, 50000).

SentencePieceTokenizer: a tokenizer transform that splits a sentence string into tokens. A pretrained sp model is passed to the constructor of the transform. The transform is JITable.

SentencePieceProcessor: a tokenizer + vocab transform that converts a sentence string into a list of token ids. A pretrained sp model is passed to the constructor of the transform. The transform is JITable.

Add dual registration for sentencepiece (a.k.a. pybind and torchbind).


torchtext/experimental/transforms.py
```diff
@@ -144,22 +147,169 @@ class TextSequentialTransforms(nn.Sequential):
...
                 _module = _module.to_ivalue()
             module_list.append((str(_idx), _module))
         return TextSequentialTransforms(OrderedDict(module_list))
+
+
+_pretrained_spm = ['text_unigram_15000', 'text_unigram_25000', 'text_unigram_50000',
+                   'text_bpe_15000', 'text_bpe_25000', 'text_bpe_50000']
+
+
+class PretrainedSPTokenizer(nn.Module):
```
cpuhrsch, line 183: Why is this necessary instead of using torch.load on a JIT'd instance of our sentencepiece class?

cpuhrsch: Or I guess, why have a wrapper class like this at all instead of just a list of URLs?

torchtext/experimental/transforms.py
```diff
@@ -144,22 +147,169 @@ class TextSequentialTransforms(nn.Sequential):
...
+
+        Examples:
+            >>> spm_transform.decoder(['▁the', '▁pre', 'trained', '▁sp', '▁model', '▁names'])
+            >>> 'the pretrained sp model names'
+        """
+
+        return self.sp_model.DecodePieces(tokens)
+
+
+class PretrainedSPTransform(nn.Module):
```
cpuhrsch, line 239: Same thing here. Also what are these model pretrained on? what's the dataset, what do the numbers mean, are there going to be more? what are the settings used and can the user get them somewhere?

torchtext/experimental/transforms.py
```diff
@@ -144,22 +147,169 @@ class TextSequentialTransforms(nn.Sequential):
...
+    def forward(self, ids: List[int]) -> Tensor:
+        r"""
+        Args:
+            ids: the input list of ids
+
+        Examples:
+            >>> to_tensor([9, 1546, 18811, 2849, 61, 2759, 2202])
+            >>> tensor([    9,  1546, 18811,  2849,    61,  2759,  2202])
+        """
+        return torch.tensor(ids).to(torch.long)
```
cpuhrsch, line 315: The torch tensor constructor has a dtype argument if you want to avoid this additional copy. You can do

torch.tensor(ids, dtype=torch.long).

But you're probably better served to use torch.as_tensor, which can deal with all kinds of data like numpy, list, tuple, scalar, ndarray etc.: https://pytorch.org/docs/master/generated/torch.as_tensor.html#torch.as_tensor

I understand that this wrapper is really just needed for JIT support?

torchtext/experimental/transforms.py
zhangguanheng66: This is a standard transform that can be used together with other transform. Users can download and load a spm but still need a wrapper (cannot jit a method of JITable class).  User can also use this transform for any local pretrained spm. For example, pytext might have a pretrained spm so they can use this one as a standard interface.

torchtext/experimental/transforms.py
zhangguanheng66: The spm provided here was trained by `bookcorpus`, `enwik9`, `wiki103` datasets. I will update the doc strings to reflect this.

torchtext/experimental/transforms.py
zhangguanheng66: Yes, it's for the JIT support. And also people have a one-stop transform to do tokenization + vocab.

torchtext/experimental/transforms.py
cpuhrsch: Sure, but you can just write a factory function that returns an instance of https://github.com/pytorch/text/blob/084b0298e1c9886f5cbf40298bbda35a6841becc/torchtext/csrc/sentencepiece.cpp instead of this wrapper.

In fact, it already exists ```load_sp_model```. You can do ```load_sp_model(download(URL))``` and just have a list of URLs. The result of this will still be a JIT-able instance.

torchtext/experimental/transforms.py
```diff
@@ -174,6 +175,162 @@ def to_ivalue(self):
...
+            >>> spm_transform.decoder([['▁the', '▁pre', 'trained', '▁sp', '▁model', '▁names']])
+            >>> ['the pretrained sp model names']
+        """
+        string_list: List[str] = []
+        for tokens in tokens_list:
+            string_list.append(self.sp_model.DecodePieces(tokens))
+        return string_list
+
+
+class SentencePieceTransform(nn.Module):
```
cpuhrsch, line 261: Should the name imply that this is a tokenizer?

torchtext/experimental/transforms.py
```diff
@@ -174,6 +175,162 @@ def to_ivalue(self):
...
+
+    Args:
+       spm_model: the sentencepiece model instance
+
+    Examples:
+        >>> import torch
+        >>> from torchtext.experimental.transforms import SentencePieceTransform
+        >>> from torchtext.experimental.transforms import pretrained_spm
+        >>> sp_model = pretrained_spm('text_unigram_25000')
+        >>> spm_transform = SentencePieceTransform(sp_model)
```
cpuhrsch, line 272: I wouldn't encourage users to call into the initialization method directly, instead we should continue to use the factory function pattern, so that we can change the init signature and also add new features without introducing bc breaking changes.

torchtext/experimental/transforms.py
```diff
@@ -174,6 +175,162 @@ def to_ivalue(self):
...
+
+    """
+    if spm_model in _pretrained_spm:
+        spm_model = download_from_url('https://pytorch.s3.amazonaws.com/models/text/pretrained_spm/{}.model'.format(spm_model))
+        return load_sp_model(spm_model)
+    else:
+        raise RuntimeError('The pretrained sentencepiece model is not valid')
+
+
+class SentencePieceTokenizer(nn.Module):
```
cpuhrsch, line 211: We have this pattern going on where there is a class wrapper for each tokenizer and vocab, mostly just to support the JIT.

```
spm = sentencepiece_tokenizer(<file-like object)
def transforms(line):
    x = spm(line)
```

Cant' do this because transforms won't be Jit-able.

So instead we've been writing these wrappers, maybe we can think of a more generic solution?

torchtext/experimental/transforms.py
```diff
@@ -174,6 +175,148 @@ def to_ivalue(self):
...
+            ids: the integers list for decoder
+
+        Examples:
+            >>> spm_transform.decoder([9, 1546, 18811, 2849, 2759, 2202])
+            >>> 'the pretrained sp model names'
+        """
+        return self.sp_model.DecodeIds(ids)
+
+
+class ToLongTensor(nn.Module):
```
cpuhrsch, line 297: Separate this out

torchtext/experimental/transforms.py
cpuhrsch: Instead of this, you could add a method to SentencePiece like
```
def to_tokenizer(self):
    return SentencePieceTransform(self)
```

Then you can have 
```
sp_model = pretrained_spm('text_unigram_25000')
spm_transform = sp_model.to_tokenizer()
```

cpuhrsch: Or you can have a factory function like

```
def sentence_piece_transform(spm_model: SentencePiece):
    return SentencePieceTransform(spm_model)
```

torchtext/experimental/transforms.py
zhangguanheng66: This is the reason to have the wrapper.

torchtext/experimental/transforms.py
```diff
@@ -174,6 +175,143 @@ def to_ivalue(self):
...
+
+    Examples:
+        >>> import torch
+        >>> from torchtext.experimental.transforms import SentencePieceTokenizer
+        >>> from torchtext.experimental.transforms import pretrained_spm
+        >>> sp_model = pretrained_spm('text_unigram_25000')
+        >>> spm_tokenizer = sentencepiece_tokenizer(sp_model)
+        >>> jit_spm_tokenizer = torch.jit.script(spm_tokenizer)
+    """
+    return SentencePieceTokenizer(spm_model)
```
fmassa, line 225: can you explain the reasoning behind this function? why not just have the class?

torchtext/experimental/transforms.py
```diff
@@ -174,6 +175,143 @@ def to_ivalue(self):
...
+    def forward(self, line: str) -> List[str]:
+        r"""
+        Args:
+            line: the input sentence string
+
+        Examples:
+            >>> spm_tokenizer('the pretrained sp model names')
+            >>> ['▁the', '▁pre', 'trained', '▁sp', '▁model', '▁names']
+        """
+        return self.sp_model.EncodeAsPieces(line)
```
fmassa, line 248: Is there a particular reason why the `forward` performs the encoding? Why not just have something which is not a `nn.Module`, and have a `encode` / `decode` methods?

torchtext/experimental/transforms.py
```diff
@@ -174,6 +175,143 @@ def to_ivalue(self):
...
+
+    Examples:
+        >>> import torch
+        >>> from torchtext.experimental.transforms import SentencePieceTransform
+        >>> from torchtext.experimental.transforms import pretrained_spm
+        >>> sp_model = pretrained_spm('text_unigram_25000')
+        >>> spm_transform = sentencepiece_transform(sp_model)
+        >>> jit_spm_tokenizer = torch.jit.script(spm_transform)
+    """
+    return SentencePieceTransform(spm_model)
```
fmassa, line 277: Same question here, do we need this function?

torchtext/experimental/transforms.py
```diff
@@ -174,6 +175,143 @@ def to_ivalue(self):
...
+    def decode(self, ids: List[int]) -> str:
+        r"""
+        Args:
+            ids: the integers list for decoder
+
+        Examples:
+            >>> spm_transform.decoder([9, 1546, 18811, 2849, 2759, 2202])
+            >>> 'the pretrained sp model names'
+        """
+        return self.sp_model.DecodeIds(ids)
```
fmassa, line 312: I wonder if there is something special about the way SentencePiece applies the tokenization / transform that would make it be part of the name of the class?

torchtext/experimental/transforms.py
```diff
@@ -174,6 +175,143 @@ def to_ivalue(self):
...
+
+
 class VocabTransform(nn.Module):
```
fmassa, line 315: Unrelated: do we want / need the `Transform` name here? Isn't everything in this file a transform anyway?

fmassa (changes requested):
I have a few questions regarding the API, let me know what you think

torchtext/experimental/transforms.py
zhangguanheng66: As we discussed above, users will call this factory func, instead of the original class (a.k.a. `SentencePieceTokenizer`). We will have more flexibility to add new arguments without worry about BC breaking.

torchtext/experimental/transforms.py
zhangguanheng66: This is basically a wrapper to have `sp_model.EncodeAsPieces` func JITable. And we have `nn.Module` class for the experimental transforms in torchtext.

torchtext/experimental/transforms.py
zhangguanheng66: This one combines the two steps together: tokenizer + vocab.

torchtext/experimental/transforms.py
```diff
@@ -176,6 +177,143 @@ def to_ivalue(self):
...
+    """
+    if spm_model in _pretrained_spm:
+        spm_model = download_from_url('https://pytorch.s3.amazonaws.com/models/text/pretrained_spm/{}.model'.format(spm_model))
+        return load_sp_model(spm_model)
+    else:
+        raise RuntimeError('The pretrained sentencepiece model is not valid')
+
+
+def sentencepiece_tokenizer(spm_model):
+    r"""factory function to generate SentencePieceTokenizer with pretrained sp model
```
cpuhrsch, line 214: nit: capitalization of "factory"
I'd say this is generating a SentencePieceTokenizer "from a pretrained sp model". Since this is documentation we can also write out SentencePiece instead of sp.

torchtext/experimental/transforms.py
cpuhrsch: Factory functions allow us to extend the SentencePieceTransform constructor to something more sophisticated. Right now it holds a reference to the full spm_model, but maybe in the future we can split out something smaller from that model or maybe we have some extra internal settings we need to include.

One discussion we might want to have is whether this should actually be a method of SentencePieceModel

torchtext/experimental/transforms.py
```diff
@@ -176,6 +177,143 @@ def to_ivalue(self):
...
+        >>> from torchtext.experimental.transforms import SentencePieceTransform
+        >>> from torchtext.experimental.transforms import pretrained_spm
+        >>> sp_model = pretrained_spm('text_unigram_25000')
+        >>> spm_transform = sentencepiece_transform(sp_model)
+        >>> jit_spm_tokenizer = torch.jit.script(spm_transform)
+    """
+    return SentencePieceTransform(spm_model)
+
+
+class SentencePieceTransform(nn.Module):
```
cpuhrsch, line 282: Should we call this a Vocab?

zhangguanheng66: Not really. It goes beyond Vocab as this transform combine Vocab/Tokenizer together.

cpuhrsch: We could also add an nn.Module wrapper + forward method for the C++ SentencePiece class instead of introducing another class, that I think does what a user would expect SentencePiece to do by default.

cpuhrsch:
We also already have https://github.com/pytorch/text/blob/53e3ae293359bc2a629d27e195e99d1e7f7b85c3/torchtext/data/functional.py#L64 - maybe we can do some consolidation?

benchmark/benchmark_sentencepiece.py
```diff
@@ -0,0 +1,37 @@
...
+    # experimental sentencepiece model with pybind
+    train, _ = raw.DATASETS[args.dataset]()
+    sp_model = load_pybind_sp_model(sp_model_path)
+    print("SentencePiece EncodeAsIds - pybind")
+    _run_benchmark(train, sp_model.EncodeAsIds)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='SentencePiece benchmark')
+    parser.add_argument('--dataset', type=str, default='AG_NEWS',
```
cpuhrsch, line 34: Can you post some results? Just for reference.

torchtext/csrc/register_bindings.cpp
```diff
@@ -26,9 +26,14 @@ PYBIND11_MODULE(_torchtext, m) {
       .def("forward", &RegexTokenizer::forward);
 
   py::class_<SentencePiece>(m, "SentencePiece")
+      .def(py::init<std::string>())
+      .def("return_content",
```
cpuhrsch, line 30: I'd just make this private, since it's specific to to_ivalue only.

cpuhrsch (approved):

test/experimental/test_transforms.py
```diff
@@ -13,6 +18,50 @@
...
+        self.assertEqual(spm_transform.decode(ref_results), test_sample)
+        self.assertEqual(jit_spm_transform.decode(ref_results), test_sample)
+
+    def test_sentencepiece_tokenizer(self):
+        model_path = get_asset_path('spm_example.model')
+        spm_tokenizer = sentencepiece_tokenizer(model_path)
+        jit_spm_tokenizer = torch.jit.script(spm_tokenizer.to_ivalue())
+        test_sample = 'SentencePiece is an unsupervised text tokenizer and detokenizer'
+        ref_results = ['\u2581Sent', 'ence', 'P', 'ie', 'ce', '\u2581is',
+                       '\u2581an', '\u2581un', 'super', 'vis', 'ed', '\u2581text',
```
cpuhrsch, line 39: Why do we use "\u2581an" when in other tests we use the actual UTF-8 character?

cpuhrsch: Offline discussion: This is the whitespace symbol. How does the user know that this is the case?

zhangguanheng66: Updated the doc of SentencepieceTokenizer transform.

torchtext/csrc/register_bindings.cpp
zhangguanheng66: Done

benchmark/benchmark_sentencepiece.py
zhangguanheng66: Added.

PR #917, opened 2020-08-12
Adding pybind registration for all C++ classes

Nayef211:
## Description
- Adding pybind11 registration for all C++ classes
- Updating Vocab, Vectors, BasicEnglishNormalize, RegexTokenizer to use pybind11 registered cpp classes by default in eager mode
- Added `to_ivalue` methods in python to return a copy of the class registered using torchbind
- Added `is_jitable` property to classes
- Added factory methods for creating RegexTokenizer and BasicEnglishNormalize
- Updated tests to reflect changes

## Known Bugs
- Refer to https://github.com/pytorch/text/issues/933 for bug with propagating C++ errors to Python when using pybind11 in an Windows environment

### Vocab Benchmark
```
Vocab - Eager Mode
Lookup time: 2.397123485803604

Lookup time: 8.786815668921918

Vocab Experimental - Jit Mode
Lookup time: 232.09749766811728
```

### Vectors Benchmark
```
FastText Existing - Eager Mode
Lookup time: 198.23392064962536

FastText Experimental - Eager Mode
Lookup time: 9.97914463095367

FastText Experimental - Jit Mode
Lookup time: 263.4094475242309
```

### Basic English Normalize Benchmark
```
BasicEnglishNormalize - Eager Mode
Tokenization time: 3.9534205556847155

BasicEnglishNormalize Experimental - Eager Mode
Tokenization time: 5.013230613898486

BasicEnglishNormalize Experimental - Jit Mode
Tokenization time: 17.305227379780263
```


test/experimental/test_vectors.py
```diff
@@ -130,10 +133,6 @@ def test_errors(self):
         tokens = ['a', 'b', 'c']
         vecs = torch.stack((tensorA, tensorB,), 0)
 
-        with self.assertRaises(RuntimeError):
```
zhangguanheng66, line 133: So we are not covering this `assertRaises` anymore?

torchtext/csrc/vectors.cpp
```diff
@@ -11,36 +11,36 @@
 #include <sstream>
 #include <stdexcept>
 #include <string>
-#include <torch/script.h>
 #include <vectors.h>
 
 namespace torchtext {
 
-Vectors::Vectors(const IndexMap &stoindex, const torch::Tensor vectors,
+Vectors::Vectors(const IndexMap &stoi, const torch::Tensor vectors,
```
zhangguanheng66, line 18: So those changes are out of the pybind scope for this PR, right?

Nayef211: You are right. I just created another PR with these changes https://github.com/pytorch/text/pull/921

test/experimental/test_vectors.py
Nayef211: No and this is because technically the `vectors` Tensor and the `tokens` can have different sizes since the Tensor can be sparse. This could occur when reading from a file because we initialize the tensor to be of the number of lines in the file. However, if there are duplicate tokens found in the file, we just skip the row corresponding to that line within the tensor.

torchtext/csrc/register_bindings.cpp
```diff
@@ -8,17 +8,6 @@
 #include <vectors.h>
 #include <vocab.h>
 
-// If we are in a Windows environment, we need to define
```
zhangguanheng66: Why do we remove this?

Nayef211: I was just testing to see if the registration had anything to do with the error. Readded this

zhangguanheng66:
@peterjc123 Just wondering if you have any suggestions for the failed CI tests. Thx a lot.

torchtext/csrc/register_bindings.cpp
peterjc123: It is okay to remove this registration now because it is not a Python extension anymore. But I'm curious about the crash.

torchtext/csrc/vectors.cpp
peterjc123: I think the crash is very likely coming from the modifications of this file.

zhangguanheng66: @Nayef211 did we merge those changes in a separate PR and the PR passes the CI test?

Nayef211: Yes we did. I just merged in the latest master now

zhangguanheng66: > I think the crash is very likely coming from the modifications of this file.

Thanks @peterjc123, could you be more specific?

torchtext/csrc/register_bindings.cpp
Nayef211: Hey @peterjc123 I removed the python Windows init and now we are getting different test failures. Could you take another look?

Nayef211:
> Please do a quick search and add print out before all the `std:: runtime_error`.

I don't think we need to add this for functions returning an `intrusive_pointer` since that can only be registered with torchbind. But I will go ahead and add this just in case.

torchtext/csrc/vocab.cpp
```diff
@@ -77,6 +91,11 @@ void Vocab::insert_token(const std::string &token, const int64_t &index) {
 
 std::string Vocab::lookup_token(const int64_t &index) {
   if (index < 0 || index > static_cast<int64_t>(itos_.size())) {
+#ifdef _MSC_VER
```
zhangguanheng66, line 94: Just for the record, we need further investigation (probably some help from MS) on the propagation of RuntimeError through pybind.

Nayef211: Yup added a issue to the PR description to keep track of this https://github.com/pytorch/text/issues/933

torchtext/csrc/sentencepiece.cpp
```diff
@@ -1,14 +1,5 @@
 #include <sentencepiece.h>
 
-#ifdef _WIN32
```
cpuhrsch, line 3: Why are you deleting this?

Nayef211: According to @peterjc123 comments above
> It is okay to remove this registration now because it is not a Python extension anymore. But I'm curious about the crash.

Do you think we still need this?

PR #918, opened 2020-08-12
Updated vocab to use `order_preserving_flat_hash_map` implementation

Nayef211:
## Description
- Updating Vocab class to use `order_preserving_flat_hash_map`
- Added benchmark for single tokens list and line-by-line tokens list lookups
- Added `lookup_indices()` function to legacy Vocab for benchmarking

## Bug
We ran into a bug when using `order_preserving_flat_hashmap` where doing something like 
```
ska_ordered::order_preserving_flat_hash_map<std::string, int64_t> stoi;
stoi["example"] = stoi.size();
```
results in `map["example"]` having a value of `1` on Windows and a value of `0` on Linux/Mac. As a workaround we are setting `stoi["example"] = itos.size();` where `itos` is a list of unique tokens within `stoi`.

### Existing Vocab - Eager Mode
```
Lookup time: 2.363746049813926
Lookup time: 2.438106880057603
Lookup time: 2.609188372734934
```

### Experimental Vocab - `c10::dict`
```
Lookup time: 113.60653045214713
Lookup time: 10.016905480064452
Lookup time: 18.412074860651046

Vocab Experimental - Jit Mode
Lookup time: 233.7020576535724
Lookup time: 9.94018982956186
Lookup time: 17.678338350262493
```

### Experimental Vocab - `order_preserving_flat_hash_map`
```
Lookup time: 100.76871072594076
Lookup time: 7.151833652984351
Lookup time: 14.777397811412811

Vocab Experimental - Jit Mode
Lookup time: 216.60021552396938
Lookup time: 7.096281274687499
Lookup time: 13.794224750250578
```

### DIFF
--- a/torchtext/csrc/vocab.cpp
+++ b/torchtext/csrc/vocab.cpp
@@ -163,8 +163,19 @@ _concat_tokens(std::vector<std::shared_ptr<StringList>> chunk_tokens,
     }
   }
 
-  // create tokens list and stoindex map
   int64_t index = 0;
+  // insert unk_token if not present
+  if (tokens_freq.find(unk_token) == tokens_freq.end()) {
+    std::cerr << "The `unk_token` " << unk_token
+              << " wasn't found in the `ordered_dict`. Adding the `unk_token` "
+                 "to the beginning of the Vocab."
+              << std::endl;
+
+    tokens.emplace_back(unk_token);
+    stoindex[unk_token] = index;
+  }
+
+  // create tokens list and stoindex map
   for (size_t i = 0; i < chunk_tokens.size(); i++) {
     auto &subset_tokens = *chunk_tokens[i];
     for (size_t j = 0; j < subset_tokens.size(); j++) {
@@ -176,18 +187,6 @@ _concat_tokens(std::vector<std::shared_ptr<StringList>> chunk_tokens,
       }
     }
   }
-
-  // insert unk_token if not present
-  if (tokens_freq.find(unk_token) == tokens_freq.end()) {
-    std::cerr << "The `unk_token` " << unk_token
-              << " wasn't found in the `ordered_dict`. Adding the `unk_token` "
-                 "to the end of the Vocab."
-              << std::endl;
-
-    tokens.emplace_back(unk_token);
-    stoindex[unk_token] = index;
-  }
-
   return std::make_tuple(std::move(stoindex), std::move(tokens));
 }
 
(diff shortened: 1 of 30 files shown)


zhangguanheng66:
@peterjc123 The windows tests fail but the linux ones pass. We are wondering if there is something special for `AssertEqual` in Windows.

Nayef211:
> @peterjc123 The windows tests fail but the linux ones pass.

@peterjc123 any idea what could be going on?

torchtext/csrc/vocab.h
```diff
@@ -3,7 +3,9 @@
 namespace torchtext {
 
 typedef std::vector<std::string> StringList;
-typedef c10::Dict<std::string, int64_t> IndexDict;
+typedef ska_ordered::order_preserving_flat_hash_map<std::string, int64_t>
```
zhangguanheng66, line 6: I'm wondering if we should limit the scope of the PR and only change `IndexDict` here to order flat hash map. Not touching those things related to `get_stoi` func.

torchtext/csrc/vocab.cpp
```diff
@@ -20,24 +20,24 @@ Vocab::Vocab(const StringList &tokens, const std::string &unk_token)
...
+    return item->second;
   }
   return unk_index_;
 }
 
 void Vocab::append_token(const std::string &token) {
   if (stoi_.find(token) == stoi_.end()) {
-    stoi_.insert(std::move(token), stoi_.size());
+    stoi_[std::move(token)] = stoi_.size();
   }
```
zhangguanheng66, line 41: I guess we do need to update `itos` func here.

torchtext/csrc/vocab.cpp
```diff
@@ -20,24 +20,24 @@ Vocab::Vocab(const StringList &tokens, const std::string &unk_token)
...
-    return item->value();
+    return item->second;
   }
   return unk_index_;
 }
 
 void Vocab::append_token(const std::string &token) {
   if (stoi_.find(token) == stoi_.end()) {
-    stoi_.insert(std::move(token), stoi_.size());
+    stoi_[std::move(token)] = stoi_.size();
```
zhangguanheng66, line 40: And I think we need to check this step under the windows environment. We need to make sure `stoi_.size()` is correct when setting `stoi_['b']`.

torchtext/csrc/vocab.cpp
Nayef211: Good catch!

zhangguanheng66:
Just to notify our MSFT partner team about the bug mentioned in the context of PR. Here are what we changed to pass the CI test ([link](https://github.com/pytorch/text/pull/918/commits/0bc831b4478e2d8a9d15a9e7e18286cdd298f0aa)).
@peterjc123 @maxluk @nbcsm @guyang3532 @gunandrose4u @smartcat2010 @mszhanyi

PR #921, opened 2020-08-12
Updating variable names for C++ Vocab and Vectors

Nayef211:
## Description
- Updated variable name from `stoindex` to `stoi`

### DIFF
--- a/torchtext/csrc/vocab.cpp
+++ b/torchtext/csrc/vocab.cpp
@@ -6,9 +6,9 @@
 
 namespace torchtext {
 
-Vocab::Vocab(const StringList &tokens, const IndexDict &stoindex,
+Vocab::Vocab(const StringList &tokens, const IndexDict &stoi,
              const std::string &unk_token, const int64_t unk_index)
-    : itos_(std::move(tokens)), stoi_(std::move(stoindex)),
+    : itos_(std::move(tokens)), stoi_(std::move(stoi)),
       unk_index_(std::move(unk_index)), unk_token_(std::move(unk_token)) {}
 
 Vocab::Vocab(const StringList &tokens, const std::string &unk_token)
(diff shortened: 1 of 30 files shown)


torchtext/csrc/vocab.cpp
```diff
@@ -20,24 +20,24 @@ Vocab::Vocab(const StringList &tokens, const std::string &unk_token)
       throw std::runtime_error("Duplicate token found in tokens list: " +
                                tokens[i]);
     }
-    stoi_.insert(std::move(tokens[i]), i);
+    stoi_[std::move(tokens[i])] = i;
```
zhangguanheng66, line 23: I'm wondering if we should limit the scope of the PR and just change the variable name. To debug the windows failure, we might move `insert`, `->value()` etc back. We just need to limit the scope of the PR if we want to debug the wire discrepancy across different platforms.

torchtext/csrc/vocab.cpp
```diff
@@ -96,7 +96,18 @@ std::vector<int64_t> Vocab::lookup_indices(const StringList &tokens) {
   return indices;
 }
 
-c10::Dict<std::string, int64_t> Vocab::get_stoi() const { return stoi_; }
+std::unordered_map<std::string, int64_t> Vocab::get_stoi() const {
```
zhangguanheng66, line 99: Same here. We can limit the scope of the PR and add this change later for debug purpose.

torchtext/csrc/vocab.cpp
```diff
@@ -152,15 +163,15 @@ _concat_tokens(std::vector<std::shared_ptr<StringList>> chunk_tokens,
...
 
-  // create tokens list and stoindex map
+  // create tokens list and stoi map
   int64_t index = 0;
   for (size_t i = 0; i < chunk_tokens.size(); i++) {
     auto &subset_tokens = *chunk_tokens[i];
     for (size_t j = 0; j < subset_tokens.size(); j++) {
       if (tokens_freq[subset_tokens[j]] >= min_freq &&
-          !stoindex.contains(subset_tokens[j])) {
+          stoi.find(subset_tokens[j]) == stoi.end()) {
```
zhangguanheng66, line 172: Same for the changes here.

PR #926, opened 2020-08-13
Moved unk_token to beginning of Vocab by default

Nayef211:
## Description
- Moved `unk_token` to beginning of Vocab class by default
- Updated unit tests

### DIFF
--- a/torchtext/csrc/vocab.cpp
+++ b/torchtext/csrc/vocab.cpp
@@ -152,8 +152,19 @@ _concat_tokens(std::vector<std::shared_ptr<StringList>> chunk_tokens,
     }
   }
 
-  // create tokens list and stoindex map
   int64_t index = 0;
+  // insert unk_token if not present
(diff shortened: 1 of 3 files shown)


torchtext/csrc/vocab.cpp
```diff
@@ -152,8 +152,19 @@ _concat_tokens(std::vector<std::shared_ptr<StringList>> chunk_tokens,
     }
   }
 
-  // create tokens list and stoindex map
   int64_t index = 0;
+  // insert unk_token if not present
```
zhangguanheng66, line 156: Are those lines duplicate? I thought we add the `unk_token` in L76 of `vocab.py`.

Nayef211: These are not duplicate. The C++ one is for loading a vocab from a file. The python one is creating a Vocab with a OrderedDict as the input.

zhangguanheng66: Right. I forgot another path to load vocab.

PR #928, opened 2020-08-13
Create Vocab from raw text file

Nayef211:
## Description
- Created function to create and load a Vocab from any raw text file
- Using a JITed tokenizer to tokenize each line within the raw text file
- Fixed bug with incrementing index when inserting `unk_token` to front
- Updated tests and documentation to reflect changes

## Benchmark
Loading `train.csv` file into Vocab

```
ag_news
Construction time: 3.4189829449169338

db_pedia 
Construction time: 20.468725885264575

yahoo_answers train.csv
Construction time: 131.20680678170174
```


torchtext/experimental/vocab.py
```diff
@@ -6,12 +6,43 @@
...
+    Examples:
+        >>> from torchtext.experimental.vocab import vocab_from_raw_text_file_object
+        >>> from torchtext.experimental.transforms import basic_english_normalize
+        >>> f = open('vocab.txt', 'r')
+        >>>     tokenizer = basic_english_normalize()
+        >>> tokenizer = basic_english_normalize()
+        >>> jit_tokenizer = torch.jit.script(tokenizer.to_ivalue())
+        >>> v = vocab_from_raw_text_file_object(f, jit_tokenizer)
+    """
+    vocab_obj = _load_vocab_from_raw_text_file(file_like_object.name, unk_token, min_freq, num_cpus, jited_tokenizer)
```
cpuhrsch, line 42: ".name" assumes it's backed by a physical file. If that's the assumption I'd changed the name of function and the description.

torchtext/experimental/vocab.py
```diff
@@ -6,12 +6,43 @@
...
-    _load_vocab_from_file
+    _load_vocab_from_file,
+    _load_vocab_from_raw_text_file
 )
 
 logger = logging.getLogger(__name__)
 
 
+def vocab_from_raw_text_file_object(file_like_object, jited_tokenizer, min_freq=1, unk_token='<unk>', num_cpus=32):
+    r"""Create a `Vocab` object from a raw text file like object.
```
cpuhrsch, line 17: I don't think this gives the functionality (such as applying a generic, JIT'd tokenizer in parallel), any justice. Also num_cpus=32 seems quite aggressive as a default, I'd set it to 4 to reflect a rough guess of what most PC can support.

cpuhrsch:
Any data on speedup over the current method?

cpuhrsch:
Also, this can be immediately used to back the implementation of [text_classification datasets](https://github.com/pytorch/text/blob/master/torchtext/datasets/text_classification.py). Whether we want to do this in this or another PR is up to you.

torchtext/csrc/vocab.cpp
```diff
@@ -244,8 +268,61 @@ Vocab _load_vocab_from_file(const std::string &file_path,
...
+  return Vocab(std::move(tokens), std::move(stoi), unk_token, unk_index);
+}
+
+Vocab _load_vocab_from_raw_text_file(const std::string &file_path,
+                                     const std::string &unk_token,
+                                     const int64_t min_freq,
+                                     const int64_t num_cpus, py::object fn) {
+  std::cerr << "[INFO] Reading file " << file_path << std::endl;
+
+  torch::jit::script::Module module = *torch::jit::as_module(fn);
```
cpuhrsch, line 301: I think it's better to use the copy constructor for this or at a minimum verify that you can do concurrent access for the forward method and not cause any data races.

torchtext/experimental/vocab.py
Nayef211: Sure I will change this to `vocab_from_raw_text_file` instead

torchtext/experimental/vocab.py
Nayef211: Updated the description to reflect this. Also changed num cpus to 4

torchtext/csrc/vocab.cpp
Nayef211: Done

Nayef211:
> Any data on speedup over the current method?

Seems like a factor of 2 improvement over the past implementation going off of moving `vocab_from_file` to C++
image

Nayef211:
> Also, this can be immediately used to back the implementation of [the experimental text classification datasets](https://github.com/pytorch/text/blob/master/torchtext/experimental/datasets/text_classification.py). Whether we want to do this in this or another PR is up to you.

Let me do this in a follow-up PR!

PR #934, opened 2020-08-14
Fix wreorder warnings with cpp files

Nayef211:
## Description
- Fixing wreorder warnings with cpp files


Nayef211:
> Could you sync with the master branch?

Done

PR #942, opened 2020-08-24
Modify the default parameters of csv reader owing the unreasonable case.

aseaday:
In the default settings of csv reader, the double quote is the default quote symbol which may cause some problems when dealing with sentences, because the corpus usually has a double quote as an inner symbol in the sentence such as conversations.


zhangguanheng66:
FYI, we will retire the legacy code in torchtext by October, which includes those datasets in `torchtext/data.dataset.py`

aseaday:
> FYI, we will retire the legacy code in torchtext by October, which includes those datasets in `torchtext/data.dataset.py`

Where could find the new API doc? We are implementing something based on the torchtext.

PR #952, opened 2020-08-31
JITable padding function

zhangguanheng66:
For text data processing, a batch of the text sequences end up with different length after tokenization/vocab steps. To convert them into a tensor, those integer lists need to be padded properly. This PR introduces a JITable padding func which pads a list of integer list to a 2d tensor. The function returns a pair of tensors: the padded tensor and the key padding mask tensor (optional). The mask tensor has the same shape as the padded tensor, with a value of False in the position of non-pad values and a value of True in the position of pads (consistent with `torch.nn.MultiheadAttention` and `torch.nn.Transformer`).

Here is a quick review of the similar func:
- torchtext.data.field.pad ([link](https://github.com/pytorch/text/blob/fd764ad02968e6310d28ee3c496e1d1d1157aa3e/torchtext/data/field.py#L235))
- pytext has pad_2d and pad_2d_mask ([link1](https://www.internalfb.com/intern/diff/D21367323/), [link2](https://github.com/facebookresearch/pytext/blob/master/pytext/torchscript/utils.py)) and non-jitable pad ([link](https://github.com/facebookresearch/pytext/blob/49a45b7f53312e01d4f828cbad13eda4846ff117/pytext/data/utils.py#L52))

Integrate the PadTransform with the sentencepiece pipeline in `examples/data_pipeline`. A follow-up task will integrate the padding function with the BERT example.


torchtext/experimental/transforms.py
```diff
@@ -251,3 +251,56 @@ def to_ivalue(self):
...
             return VectorTransform(vector)
         return self
+
+
+class PadTransform(nn.Module):
+    r"""Pad sequences
+
+    Args:
+        pad_id: the id for padding token
+        bos_token_id: the id for bos token (Default: None)
```
cpuhrsch, line 261: Maybe spell out bos

torchtext/experimental/transforms.py
```diff
@@ -251,3 +251,56 @@ def to_ivalue(self):
...
+        eos_token_id: the id for eos token (Default: None)
+        return_key_padding_mask: flag to output key_padding_mask as output (Defualt: True)
+
+    Example:
+        >>> pad_transform = PadTransform(2)
```
cpuhrsch, line 266: I'd set a flag to a non-default value just for illustration

torchtext/experimental/transforms.py
```diff
@@ -251,3 +251,56 @@ def to_ivalue(self):
...
+
+    def __init__(self, pad_id, bos_token_id=None, eos_token_id=None, return_key_padding_mask=True):
+        super(PadTransform, self).__init__()
+        self.pad_id = pad_id
+        self.bos_token_id = bos_token_id
+        self.eos_token_id = eos_token_id
+        self.return_key_padding_mask = return_key_padding_mask
+
+    @torch.jit.export
+    def forward(self, seq_batch: List[List[int]]) -> Tuple[torch.Tensor, Optional[Tensor]]:
```
cpuhrsch, line 280: We have efficient padding layers including types of padding https://pytorch.org/docs/master/nn.html#padding-layers

cpuhrsch:
What about supporting code like https://github.com/pytorch/text/blob/master/examples/BERT/mlm_task.py#L13

cpuhrsch:
Or this: https://github.com/pytorch/text/blob/master/examples/BERT/qa_task.py#L37

torchtext/experimental/transforms.py
```diff
@@ -251,3 +251,56 @@ def to_ivalue(self):
...
+        >>> pad_transform = PadTransform(2)
+        >>> seq_batch = [[5, 4, 5, 6, 7], [1, 3], [7, 5, 8]]
+        >>> pad_seq, key_padding_mask = pad_transform(seq_batch)
+        >>> jit_pad_transform = torch.jit.script(pad_transform)
+    """
+
+    def __init__(self, pad_id, bos_token_id=None, eos_token_id=None, return_key_padding_mask=True):
+        super(PadTransform, self).__init__()
+        self.pad_id = pad_id
+        self.bos_token_id = bos_token_id
```
cpuhrsch, line 275: This bos and eos tokens aren't necessarily and unrelated to padding.

cpuhrsch (changes requested):
Please see comments

torchtext/experimental/transforms.py
```diff
@@ -251,3 +251,57 @@ def to_ivalue(self):
...
+    @torch.jit.export
+    def forward(self, seq_batch: List[List[int]]) -> Tuple[torch.Tensor, Optional[Tensor]]:
+        r"""Pad a list of integer lists. The individual integer list might have different length
+        such that the padding function will add padding id to the end of list for the same length.
+
+        Args:
+            seq_batch: a list of integer lists. Type: List[List[int]]
+
+        Outputs:
+            padded_sequence, key_padding_mask Type: Tuple[torch.Tensor, Optional[Tensor]]
```
cpuhrsch, line 288: key_padding_mask is MHA specific. Is that necessary?

cpuhrsch (changes requested):
1. Integrate with BERT
2. Remove need for eos_token_id 
3. Look into torch.nn.utils.rnn.pad_sequence
4. Remove explicit reference to MHA by using key_padding_mask

torchtext/experimental/transforms.py
zhangguanheng66: fixed.

torchtext/experimental/transforms.py
```diff
@@ -251,3 +251,55 @@ def to_ivalue(self):
...
+        pad_id: the id for padding token
+        return_key_padding_mask: flag if to output padding_mask as output (Default: True)
+
+    Example:
+        >>> pad_id = 2
+        >>> pad_transform = PadTransform(pad_id)
+        >>> seq_batch = [[5, 4, 5, 6, 7], [1, 3], [7, 5, 8]]
+        >>> pad_seq, padding_mask = pad_transform(seq_batch)
+        >>> jit_pad_transform = torch.jit.script(pad_transform)
+    """
```
fmassa, line 270: What is the reasoning about adding this new transform, instead of using `torch.nn.utils.rnn.pack_padded_sequence`, which returns a `PackedSequence` object that contains exactly the same information (maybe mashed in a different way)?

fmassa (changes requested):
I have one high-level question that I would like answered before we move forward with this PR. There might be redundancy between what we already provide in PyTorch and what we are providing here.
If that's the case, we should use `torch.nn.utils.rnn.pack_padded_sequence` and massage the `batch_sizes` to return a boolean mask.

torchtext/experimental/transforms.py
cpuhrsch: This is supposed to be easier to use and it's a Module that can be JIT'd

torchtext/experimental/transforms.py
```diff
@@ -251,3 +251,55 @@ def to_ivalue(self):
...
+
+        Note:
+            The padding_mask tensor has the same shape [len(seq_batch), max_seq_len], with a value of False in
+            the position of non-pad values and a value of True in the position of pads. len(seq_batch) is the number
+            of input sequences and max_seq_len is the maximum length of the input sequences.
+        """
+        max_seq_len = max([seq.size(0) for seq in seq_batch])
+        trailing_dims = seq_batch[0].size()[1:]
+        padding_mask = torch.zeros(len(seq_batch), max_seq_len)
+        output_batch = seq_batch[0].new_full((len(seq_batch), max_seq_len) + trailing_dims, self.pad_id)
```
cpuhrsch, line 298: Use the rnn pad function

torchtext/experimental/transforms.py
```diff
@@ -248,3 +248,53 @@ def to_ivalue(self):
             vector = self.vector.to_ivalue()
             return VectorTransform(vector)
         return self
+
+
+class PadTransform(nn.Module):
```
cpuhrsch, line 253: Maybe drop "Transform" in the end since it's already in the transform file.

torchtext/experimental/transforms.py
```diff
@@ -422,3 +422,53 @@ def to_ivalue(self):
...
+
+    @torch.jit.export
+    def forward(self, seq_batch: List[Tensor]) -> Tuple[torch.Tensor, Optional[Tensor]]:
+        r"""Pad a list of tensors in the dim of (seq_dim, ...). The individual tensor has different length
+        (i.e. seq_dim) such that the padding function will add padding id to the end of list for the same length. It
+        assumes that the dimensions after seq_dim of the tensors are same. And the tensors have same dtype, which is
+        the dtype of output padded tensor.
+
+        Args:
+            seq_batch: a list of torch.tensor. Type: List[Tensor]
```
cpuhrsch, line 456: Why the name seq_batch? Can we expand support to n-dim Tensors? [DE;TR has generic padding+masking code that can be used for Tensors of any dimension.](https://github.com/facebookresearch/detr/blob/master/util/misc.py#L315) Maybe that's more broadly applicable and could be used in more circumstances, maybe it's overengineering.

cpuhrsch: If this is meant to work on only 1-dim Tensors, check the entires and throw an error. We can add support for n-dim Tensors later on.

zhangguanheng66: Here are a review of the padding functions with the JIT support in pytext library: [pad_2d](https://github.com/facebookresearch/pytext/blob/master/pytext/torchscript/utils.py#L181-L188), [pad_2d_float](https://github.com/facebookresearch/pytext/blob/8472355cb10293c833f603a461915a702311084d/pytext/torchscript/utils.py#L192-L199), [pad_2d_mask](https://github.com/facebookresearch/pytext/blob/8472355cb10293c833f603a461915a702311084d/pytext/torchscript/utils.py#L76-L107). For now, I think we can limit this PR to 1-dim tensors.

torchtext/experimental/transforms.py
```diff
@@ -422,3 +422,59 @@ def to_ivalue(self):
...
+            The padding_mask tensor has the same shape [len(tensor_list), max_seq_len], with a value of False in
+            the position of non-pad values and a value of True in the position of pads. len(tensor_list) is the number
+            of input sequences and max_seq_len is the maximum length of the input sequences.
+        """
+        seq_len: List[int] = []
+        for seq in tensor_list:
+            if seq.dim() == 1:
+                seq_len.append(seq.size(0))
+            else:
+                raise RuntimeError("Pad transform supports 1-D torch.tensor only")
```
cpuhrsch, line 471: nit: torch.Tensor is the type, torch.tensor is the factory function.

torchtext/experimental/transforms.py
```diff
@@ -422,3 +422,59 @@ def to_ivalue(self):
...
+        max_seq_len = max(seq_len)
+        padding_mask = torch.zeros(len(tensor_list), max_seq_len)
```
cpuhrsch, line 473: You could allocate as a BoolTensor right away by using ```torch.zeros([...], dtype=torch.byte)```.

cpuhrsch:
We [often](https://github.com/pytorch/fairseq/blob/54b934417d95baa1b0076089c61bde32728e34cf/fairseq/data/audio/raw_audio_dataset.py#L92) [see](https://github.com/facebookresearch/ParlAI/blob/8200396cdd08cfd26b01fe52b4a3bd0654081182/parlai/agents/drqa/utils.py#L143) [users](https://github.com/facebookresearch/detr/blob/master/util/misc.py#L306) [write](https://github.com/pytorch/vision/blob/24f16a338391d6f45aa6291c48eb6d5513771631/references/detection/utils.py#L102) [one-off](https://github.com/pytorch/audio/blob/3250d3df168c956389bd16956aa458ce111570d0/examples/pipeline_wav2letter/datasets.py#L90) implementations to represent a list of Tensors as a single Tensor plus a Tensor mask. Maybe we can turn this into something very generic within the core.

torchtext/experimental/transforms.py
cpuhrsch: Also, you don't need to do this if ```self.return_key_padding_mask``` is None.

cpuhrsch (changes requested):
Please try using this in the BERT example first. This can also land in the BERT example first.

torchtext/experimental/transforms.py
```diff
@@ -422,3 +422,59 @@ def to_ivalue(self):
...
+
+    Example:
+        >>> pad_id = 2
+        >>> pad_transform = Pad(pad_id)
+        >>> tensor_list = [torch.tensor([5, 4, 5, 6, 7]), torch.tensor([1, 3]), torch.tensor([7, 5, 8])]
+        >>> pad_seq, padding_mask = pad_transform(tensor_list)
+        >>> jit_pad_transform = torch.jit.script(pad_transform)
+    """
+
+    def __init__(self, pad_id, return_key_padding_mask=True):
```
cpuhrsch, line 443: I'd make a choice here. Either return a data Tensor with padding values or return a data Tensor plus a mask Tensor.

The latter choice allows you to write code such

```
data, mask = Pad(0, True)
data.masked_fill(~mask, pad_id)
```

which means in essence specifying a padding is an optimization, but not a necessity.

erip:
This seems safe to close since we have an equivalent JIT'd transform, cc @parmeet

PR #960, opened 2020-09-02
Update data pipelines with the latest APIs of building blocks

zhangguanheng66:


examples/data_pipeline/pipelines.py
```diff
@@ -8,10 +8,8 @@
 from torchtext.experimental.transforms import (
     basic_english_normalize,
     TextSequentialTransforms,
-    VocabTransform,
-    VectorTransform,
```
Nayef211, line 12: Why are we getting rid of the transforms? Is this because these building blocks have their own `forward` method defined?

examples/data_pipeline/transforms.py
```diff
@@ -6,19 +6,6 @@
 from torch import Tensor
 
 
-class TextClassificationPipeline(nn.Module):
```
Nayef211, line 9: Is this removed because it wasn't being used?

examples/data_pipeline/pipelines.py
zhangguanheng66: Switch to the `forward` func. It works for a list of tokens on JIT mode.

examples/data_pipeline/transforms.py
zhangguanheng66: We moved them to the experimental.transforms file.

PR #961, opened 2020-09-02
Transform for a batch of data

zhangguanheng66:
This PR is to extend a few transforms to work with a data batch. For example, `Vocab` forward func is now working with a list of token list, instead of a list of tokens. This is more consistent with pytext. It also makes the transforms connected with padding func.


examples/data_pipeline/transforms.py
```diff
@@ -73,5 +81,24 @@ class ToLongTensor(nn.Module):
...
         return torch.tensor(tokens).to(torch.long)
+
+
+def iterate_batch(pipeline):
+    def func(data_batch):
+        return [pipeline(data) for data in data_batch]
+    return func
+
+
+class TextClassificationPipeline(nn.Module):
```
Nayef211, line 94: Is this class being used anywhere?

examples/data_pipeline/pipelines.py
```diff
@@ -24,7 +24,7 @@ def build_sp_pipeline(spm_file):
 
     # Insert token in vocab to match a pretrained vocab
     vocab.insert_token('<pad>', 1)
-    pipeline = TextSequentialTransforms(tokenizer, vocab, ToLongTensor())
+    pipeline = TextSequentialTransforms(tokenizer, vocab)
```
Nayef211, line 27: Why are we removing the call to `ToLongTensor `?

examples/data_pipeline/pipelines.py
```diff
@@ -68,7 +68,7 @@ def build_text_vocab_pipeline(hf_vocab_file):
     vocab = vocab_from_file(f)
 
     # Insert token in vocab to match a pretrained vocab
-    pipeline = TextSequentialTransforms(tokenizer, vocab, ToLongTensor())
+    pipeline = TextSequentialTransforms(tokenizer, vocab)
```
Nayef211, line 71: Same as above

examples/data_pipeline/transforms.py
zhangguanheng66: I deleted it by accident in previous PR. It's used in the `batch_torchtext` pipeline.

examples/data_pipeline/pipelines.py
zhangguanheng66: Because right now, `vocab` object returns a list of token list with different length. So when it calls `ToLongTensor`, it will fail. After I land padding func https://github.com/pytorch/text/pull/952, I will add padding func together with `ToLongTensor`.

examples/data_pipeline/transforms.py
Nayef211: I see this makes sense

examples/data_pipeline/pipelines.py
Nayef211: Okay that sounds good. 

NIT: might be worth adding a TODO comment just to keep track of this

Nayef211 (approved):

PR #962, opened 2020-09-02
Updated all pipelines to include legacy and experimental implementations

Nayef211:
## Description
- Updated pipelines to include legacy and experimental implementations for benchmarking
- Added `script_vocab` factory method
- Added transforms for Pytext `Vocab` and Pytext `ScriptVocab`


examples/data_pipeline/transforms.py
```diff
@@ -87,16 +87,27 @@ def __init__(self, vocab):
         super(PyTextVocabTransform, self).__init__()
         self.vocab = vocab
 
+    def forward(self, tokens):
+        return [self.vocab.idx[token] if token in self.vocab.idx.keys() else 0 for token in tokens]
+
+
+class PyTextScriptVocabTransform(nn.Module):
```
zhangguanheng66, line 77: What's the difference for `PyTextScriptVocabTransform` and `VocabTransform`?

examples/data_pipeline/transforms.py
```diff
@@ -87,16 +87,27 @@ def __init__(self, vocab):
...
+    r"""Vocab transform
+    """
+
+    def __init__(self, vocab):
+        super(PyTextScriptVocabTransform, self).__init__()
+        self.vocab = vocab
+
     def forward(self, tokens: List[str]) -> List[int]:
```
zhangguanheng66, line 85: Why not `List[List[str]]`?

torchtext/experimental/functional.py
```diff
@@ -9,6 +9,13 @@ def func(tok_iter):
     return func
 
 
+def vector_func(vector):
```
zhangguanheng66, line 12: Is this for the experimental vector? Why not to write this as `nn.Module`?

zhangguanheng66 (commented):
Please double check that all the transforms and pipeline work with a batch of data.

torchtext/experimental/functional.py
Nayef211: This is because this will only be used with the legacy pipeline. I believe you did something similar with `vocab_func`

examples/data_pipeline/transforms.py
Nayef211: The `VocabTransform ` calls  `self.vocab.lookup_indices(tokens)` but the `PyTextScriptVocabTransform` calls  `self.vocab.lookup_indices_1d(tokens)`. The function names are different between the two

examples/data_pipeline/transforms.py
Nayef211: I will update this to use the batch format

torchtext/experimental/functional.py
zhangguanheng66: let's not add this for now because it's related to the legacy code. If you need a function like this, just add it to `examples/data_pipelines/transform.py`.

zhangguanheng66:
sync with master branch to fix the windows tests.

zhangguanheng66 (approved):

PR #963, opened 2020-09-03
Optimize vocab from file construction 

Nayef211:
## Description
- Updated vocab from file construction to use counters instead of lists to reduce memory footprint
- Updated final tokens list to be sorted primarily by frequency and secondarily by lexicographic order

## Benchmarks
image


zhangguanheng66 (commented):
LGTM. Just make sure that it works with the latest `basic_english_normalize`, which now accepts a list of strings, not a single string.

Nayef211:
> LGTM. Just make sure that it works with the latest `basic_english_normalize`, which now accepts a list of strings, not a single string.

Will do that in a follow up PR. I'm keeping this PR to the optimizations when using counters.

zhangguanheng66 (approved):
LGTM. Let's wait for the CI tests.

zhangguanheng66:
Eli just update the master branch to fix the windows unittest.

PR #965, opened 2020-09-03
Update load_sp_model func to accept a file object

zhangguanheng66:
Test against `pytext.utils.file_io.PathManager` and it works with `spm = load_sp_model(PathManager.open('m_user.model', 'rb'))`

### DIFF
--- a/test/data/test_functional.py
+++ b/test/data/test_functional.py
@@ -68,7 +68,7 @@ class TestFunctional(TorchtextTestCase):
     def test_sentencepiece_tokenizer(self):
         test_sample = 'SentencePiece is an unsupervised text tokenizer and detokenizer'
         model_path = get_asset_path('spm_example.model')
-        sp_model = load_sp_model(model_path)
+        sp_model = load_sp_model(open(model_path, 'rb'))
         self.assertEqual(sp_model.GetPieceSize(), 20000)
         spm_generator = sentencepiece_tokenizer(sp_model)
 
--- a/torchtext/data/functional.py
+++ b/torchtext/data/functional.py
@@ -49,14 +49,14 @@ def load_sp_model(spm):
         output: a SentencePiece model.
 
     Examples:
-        >>> from torchtext.data.functional import load_spmodel
-        >>> spmodel = load_spmodel("m_user.model")
-        >>> spmodel = load_spmodel(open("m_user.model", 'rb'))
+        >>> from torchtext.data.functional import load_sp_model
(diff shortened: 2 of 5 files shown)


zhangguanheng66:
@hudeven Let me know if this API works for the internal cases.

torchtext/data/functional.py
```diff
@@ -39,20 +39,26 @@ def generate_sp_model(filename, vocab_size=20000,
...
         >>> sp_model = load_sp_model("m_user.model")
+        >>> sp_model = load_sp_model(open("m_user.model", 'rb'))
     """
-    return torch.ops.torchtext.load_sp_model(spm_path)
+    if isinstance(spm, str):
+        return torch.ops.torchtext.load_sp_model(spm)
+    elif isinstance(spm, io.BufferedReader):
+        return torch.ops.torchtext.load_sp_model_string(spm.read())
+    else:
+        raise RuntimeError('the input of the load_sp_model func is not supported.')
```
hudeven, line 61: @zhangguanheng66  PathManager.open() returns "_io.TextIOWrapper" type and will fail here. How about remove the check?
```suggestion
    return torch.ops.torchtext.load_sp_model_string(spm.read())
```

zhangguanheng66: I notice that and was thinking about adding that check. 
However, with `PathManager.open() `, it will return f as `_io.TextIOWrapper`. It seems not working with sentencepiece constructor. It needs to be binary reading where  `PathManager.open(, 'rb')` returns `io.BufferedReader`.

hudeven: @zhangguanheng66 you are right! It LGTM. Note: there is a test failure

hudeven (approved):

torchtext/data/functional.py
cpuhrsch, line 61: If the type is not detect you can specify the type found in the error message to make it easier to debug.

test/data/test_functional.py
```diff
@@ -68,7 +68,7 @@ def test_sentencepiece_numericalizer(self):
     def test_sentencepiece_tokenizer(self):
         test_sample = 'SentencePiece is an unsupervised text tokenizer and detokenizer'
         model_path = get_asset_path('spm_example.model')
-        sp_model = load_sp_model(model_path)
+        sp_model = load_sp_model(open(model_path, 'rb'))
```
cpuhrsch, line 71: assertRaises an unsupported type

PR #973, opened 2020-09-10
Remove data batch

zhangguanheng66:
This PR changes the input format of the building blocks in `experimental.transforms`, `experimental.vocab`, and `experimental.vectors` from a batch of data to a single data. For example, for `BasicEnglishNormalize` transform, the original format of the forward function is ["input is a list of string", "data batch"]. The new format of the forward function is "input is a single string". 

The data batch is more consistent with the transforms in pytext where the data from Hive are in the form of batch. In this way, the data processing pipeline will be placed in `collate_fn` which processes a batch of data from DataLoader.

However, the single data or line-by-line are more compatible with `torch.utils.data.Dataset`, which is used by the other two domain libraries. It is also more compatible with the existing experimental datasets. In this way, the data processing pipeline will be placed before DataLoader and in the `__getitem__` func within Dataset.


cpuhrsch:
We can introduce batch support later on in various ways

examples/data_pipeline/pipelines.py
```diff
@@ -5,9 +5,7 @@
     PretrainedSPVocab,
```
cpuhrsch, line 5: We should check if there is a need to update the performance numbers in the README file.

On that note, it should be easier to read if we aggregate those benchmark numbers into a table and also add the disclaimer that these numbers are for our development reference only.

zhangguanheng66: Done.

cpuhrsch (approved):

PR #975, opened 2020-09-12
Add build_vocab_from_iterator factory

zhangguanheng66:
Update `torchtext.experimental.vocab` with:
- add `build_vocab_from_iterator` func to generate experimental Vocab from an iterator.


torchtext/experimental/vocab.py
```diff
@@ -71,6 +71,23 @@ def vocab_from_file(file_object, min_freq=1, unk_token='<unk>', num_cpus=4):
     return Vocab(vocab_obj)
 
 
+def build_vocab_from_iterator(iterator):
```
fmassa, line 74: Any particular reason for not exposing the same arguments as `vocab` for `min_freq` and `unk_token`?

torchtext/experimental/vocab.py
```diff
@@ -71,6 +71,23 @@ def vocab_from_file(file_object, min_freq=1, unk_token='<unk>', num_cpus=4):
...
+    Build a Vocab from an iterator.
+
+    Arguments:
+        iterator: Iterator used to build Vocab. Must yield list or iterator of tokens.
+    """
+
+    counter = Counter()
+    for tokens in iterator:
+        counter.update(tokens)
+    sorted_by_freq_tuples = sorted(counter.items(), key=lambda x: x[1], reverse=True)
```
fmassa, line 85: Is sorting of the sequences a requirement / necessity? If yes, why not move it to `vocab` itself?

torchtext/experimental/vocab.py
```diff
@@ -71,6 +71,23 @@ def vocab_from_file(file_object, min_freq=1, unk_token='<unk>', num_cpus=4):
...
 
+def build_vocab_from_iterator(iterator):
+    """
+    Build a Vocab from an iterator.
+
+    Arguments:
+        iterator: Iterator used to build Vocab. Must yield list or iterator of tokens.
+    """
+
+    counter = Counter()
```
fmassa, line 82: If we don't expose `min_freq` to `vocab`, then this function effectively becomes the same thing as
```python
ordered_dict = OrderedDict([(v, k) for k, v in enumerate(sorted(set(iterator)))])
```

fmassa (commented):
Added a few comments, let me know what you think

torchtext/experimental/vocab.py
zhangguanheng66: The `iterator` will yield repeated tokens so we need a counter first and rank the high frequency tokens first. `set` will remove the frequency of the tokens.

torchtext/experimental/vocab.py
zhangguanheng66: We could. It's just the design of the vocab API.

torchtext/experimental/vocab.py
zhangguanheng66: There is an equivalent func in `torchtext.vocab` and we are maintaining the same behavior here. We could add `min_freq` and `unk_token` here.

bentrevett: `min_freq` and `unk_token` are exposed in `load_vocab_from_file` and `load_vocab_from_raw_text_file` so does it not make sense to expose them (as well as `pad_token`) in `build_vocab_from_iterator`?

zhangguanheng66: I can add those two for sure.

cpuhrsch (changes requested):
We've had discussions and a lot of back and forth about adding this by default in the past. Let's please have a high level decisions instead of updating our Vocabs to include or exclude this token by default. Really it shouldn't exist at all unless the user provides it. How is this necessary?

zhangguanheng66:
> We've had discussions and a lot of back and forth about adding this by default in the past. Let's please have a high level decisions instead of updating our Vocabs to include or exclude this token by default. Really it shouldn't exist at all unless the user provides it. How is this necessary?

I took a quick search among the vocabulary classes in other libraries:
- `pytext.torchscript.vocab.ScriptVocabulary` has attributes - `unk`, `pad`, `eos`, `bos`, `mask` ([link](https://github.com/facebookresearch/pytext/blob/3a55d553ddbbd76ba72467790f6ed634be306bc7/pytext/torchscript/vocab.py#L25-L29))
- `fairseq.data.dictionary.Dictonary` has attributes - `unk`, `pad`, `bos`, `eos` ([link](https://github.com/pytorch/fairseq/blob/23d8502bdde88a3e58e0910e2ee49834f8478b39/fairseq/data/dictionary.py#L24-L27))
- `allennlp.data.vocabulary.Vocabulary` has attributes - `oov`, `pad` ([link](https://github.com/allenai/allennlp/blob/2d5f24bd94ba229741a1a17e09b540a374447c77/allennlp/data/vocabulary.py#L238-L239))
- `torchtext.vocab` has attributes - `pad`, `unk` ([link](https://github.com/pytorch/text/blob/0302ea933b2141eb3cd495d8e60fa57d7f008811/torchtext/vocab.py#L45-L46))

cpuhrsch:
I think we have an opportunity to be more generic by not making these assumptions on default tokens. Even the unk token is something we might want to remove and instead add a method to specify a fallback token when something isn't found. That allows the user to get an error, if they want, and construct a perfect Vocab that fails on a miss.

zhangguanheng66:
> I think we have an opportunity to be more generic by not making these assumptions on default tokens. Even the unk token is something we might want to remove and instead add a method to specify a fallback token when something isn't found. That allows the user to get an error, if they want, and construct a perfect Vocab that fails on a miss.

OK, I will remove the `pad` token from the PR.

torchtext/experimental/vocab.py
cpuhrsch: We should always sort the tokens by frequency so that tokens that are accessed frequently are close to each other in the hash table, which can improve caching.

zhangguanheng66: We do sorting the tokens by frequency here when constructs a vocab instance. Francisco's suggestion is to move the sorting step to the vocab itself, not staying outside.

torchtext/experimental/vocab.py
```diff
@@ -77,6 +77,25 @@ def vocab_from_file(file_object, min_freq=1, unk_token='<unk>', num_cpus=4):
     return Vocab(vocab_obj)
 
 
+def build_vocab_from_iterator(iterator, min_freq=1, unk_token='<unk>'):
```
cpuhrsch, line 80: I'd want us to remove "unk_token" with a more generic solution in a follow-up PR so that we can also support Vocabs that fail if a key isn't present.

cpuhrsch (approved):

torchtext/experimental/vocab.py
zhangguanheng66: Will do it in the follow-up PR for the unknown token.

PR #980, opened 2020-09-14
[DO NOT REVIEW] Tolongtensor

zhangguanheng66:


erip:
This seems safe to close since we have an equivalent JIT'd transform, cc @parmeet

PR #981, opened 2020-09-14
packaging: Use cpu directory by default

seemethere:
Default to using the `cpu` directory since the base one is not reliable


zhangguanheng66 (commented):
LGTM

PR #984, opened 2020-09-16
add test and update error message for functional.py::load_sp_model

astaff:
This PR:
- adds test for unsupported input type `test_sentencepiece_unsupported_input_type` ([review comment](https://github.com/pytorch/text/pull/965#discussion_r485065199))
- updates error message listing valid input types and type of input given. ([review comment](https://github.com/pytorch/text/pull/965#discussion_r485064957))

Notes:
- Should we raise [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) instead of [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError)?
- Type annotation and static type checking is another option to do that.

### DIFF
--- a/test/data/test_functional.py
+++ b/test/data/test_functional.py
@@ -83,9 +83,10 @@ class TestFunctional(TorchtextTestCase):
     def test_sentencepiece_unsupported_input_type(self):
         with self.assertRaisesRegex(
             TypeError,
-            'Unsupported type for spm argument: dict. ' + \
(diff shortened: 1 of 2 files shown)


zhangguanheng66 (changes requested):
LGTM. Please sync with the master branch so the failed CI tests will go away.

zhangguanheng66:
I like `TypeError`.

cpuhrsch (approved):

PR #988, opened 2020-09-18
Dataset construction benchmark and various fixes.

cpuhrsch:
This PR introduces 

a) benchmarks for dataset construction to track performance improvements for that.
b) a dictionary of datasets (called DATASETS) to torchtext.experimental.datasets to make it easier for users to iterate over and access datasets by their name without having to cycle through the subfolders such as torchtext.experimental.datasets.text_classification. 
c) the use of iterators for building Vocabs within the experimental datasets instead of materializing the entire list again for the purpose of building a Vocab only.
d) passing language modeling datasets line by line to the build_vocab_from_iterator factory
e) return numericalized language models (already happens for single line versions)
f) only processes parts of dataset actually returned for language model datasets

We could split this up further, but it's a very small PR.

Benchmark results WIP. Will edit once they finish.

Follow up items:
Change our parallelized, C++ vocab constructor to accept a list of sentences (which are already materialized for our default datasets) and build Counter for torchtext.vocab.Vocab in parallel.


torchtext/experimental/datasets/__init__.py
```diff
@@ -6,23 +6,25 @@
 from .translation import Multi30k, IWSLT, WMT14
 from .question_answer import SQuAD1, SQuAD2
 
-__all__ = ['LanguageModelingDataset',
```
cpuhrsch, line 9: I think including "LanguageModelingDataset" might have been a bug? cc @zhangguanheng66

cpuhrsch: If not we should include the others (e.g. TextClassification) datasets as well

torchtext/experimental/datasets/language_modeling.py
```diff
@@ -74,31 +75,19 @@ def _setup_datasets(dataset_name, tokenizer=None, root='.data', vocab=None,
...
-        train, = raw.DATASETS[dataset_name](root=root, data_select=('train',))
-        if single_line:
-            raw_data = {'train': [" ".join([txt for txt in train]), ]}
-        else:
-            raw_data = {'train': [txt for txt in train]}
-    else:
-        train, test, valid = raw.DATASETS[dataset_name](root=root, data_select=('train', 'test', 'valid'))
-        # Cache raw text iterable dataset
-        if single_line:
-            raw_data = {'train': [" ".join([txt for txt in train]), ],
```
cpuhrsch, line 87: All of this was being processed, even if the given data ends up being discarded due to data_select

cpuhrsch:
This PR is getting too large. I'm going to split it up.

PR #989, opened 2020-09-19
Dataset construction benchmarks

cpuhrsch:
This PR introduces basic benchmarks for dataset construction to track performance improvements.

Benchmark results WIP. Will edit once they finish.

EDIT: Benchmarks can currently not finish due to bug in translation dataset (data_select keyword isn't being forwarded)


torchtext/experimental/datasets/__init__.py
```diff
@@ -6,23 +6,25 @@
...
+            'AmazonReviewFull': AmazonReviewFull,
+            'UDPOS': UDPOS,
+            'CoNLL2000Chunking': CoNLL2000Chunking,
+            'Multi30k': Multi30k,
+            'IWSLT': IWSLT,
+            'WMT14': WMT14,
+            'SQuAD1': SQuAD1,
+            'SQuAD2': SQuAD2}
+
+__all__ = ['LanguageModelingDataset'] + list(map(str, DATASETS.keys()))
```
zhangguanheng66, line 30: Not sure if we still want to expose `LanguageModelingDataset` to users.

zhangguanheng66 (approved):
Add a comment. Other than that, LGTM

torchtext/experimental/datasets/__init__.py
cpuhrsch: I'll make that a separate PR.

PR #991, opened 2020-09-19
Fix documentation and naming of torchtext.experimental.vocab factory functions.

cpuhrsch:
a) The documentation doesn't clearly state that one factory function is meant to be used to construct a Vocabulary from a dataset (e.g. AG_NEWS) and another is meant to be used to load a vocabulary from a text file that contains a list of tokens (as you might construct when filtering and sorting text files on the command line).
b) factory function vocab_from_file_object doesn't exist.

### DIFF
--- a/torchtext/csrc/register_bindings.cpp
+++ b/torchtext/csrc/register_bindings.cpp
@@ -63,7 +63,7 @@ PYBIND11_MODULE(_torchtext, m) {
   m.def("_load_token_and_vectors_from_file",
         &_load_token_and_vectors_from_file);
   m.def("_load_vocab_from_file", &_load_vocab_from_file);
-  m.def("_load_vocab_from_raw_text_file", _load_vocab_from_raw_text_file);
+  m.def("_build_vocab_from_text_file", _build_vocab_from_text_file);
 }
 
 // Registers our custom classes with torch.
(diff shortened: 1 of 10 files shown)


zhangguanheng66 (approved):
LGTM.

zhangguanheng66:
there was a back and forth that the names of the functions were changed from `load_from_file_object` to `load_from_file`

torchtext/csrc/register_bindings.cpp
```diff
@@ -63,7 +63,7 @@ PYBIND11_MODULE(_torchtext, m) {
   m.def("_load_token_and_vectors_from_file",
         &_load_token_and_vectors_from_file);
   m.def("_load_vocab_from_file", &_load_vocab_from_file);
```
cpuhrsch, line 65: This should be "_load_vocab_from_text_file"

torchtext/csrc/vocab.h
```diff
@@ -41,10 +41,9 @@ VocabStates _set_vocab_states(const c10::intrusive_ptr<Vocab> &self);
 Vocab _load_vocab_from_file(const std::string &file_path,
```
zhangguanheng66, line 41: `_load_vocab_from_file` and `_load_vocab_from_raw_text_file` are different. `_load_vocab_from_raw_text_file` requires a tokenizer, which works on the raw text file. `_load_vocab_from_file` will load the vocab from a column of vocabulary tokens saved in the text file.

cpuhrsch: Yes and this naming is meant to disambiguate that by calling "_load_vocab_from_raw_text_file" "build_vocab_from_text_file" (using build just as we do for build_from_iterator) and by calling "_load_vocab_from_file" "load_vocab_from_text_file" similar to how we'd call torch.load to load a model (except here it's a specific plain-text format).

rshraga:
closing stale PR

PR #992, opened 2020-09-19
Avoid unnecessary memory allocations in language modeling datasets.

cpuhrsch:
Language modeling datasets construct *all* datasets even if only a subset is constructed. It also stores the fully numericalized version of the dataset if it's stored as "a single line" (word by word), but not otherwise


torchtext/experimental/datasets/language_modeling.py
```diff
@@ -65,40 +58,52 @@ def _setup_datasets(dataset_name, tokenizer=None, root='.data', vocab=None,
...
+            for token in tokenizer(line):
+                ids.append(vocab[token])
+            if filter_empty and len(ids) == 0:
+                continue
+            yield torch.tensor(ids, dtype=torch.long)
+
+    raw_data = {}
+    for name in raw_iter:
+        # Materialize datasets
+        raw_data[name] = [torch.tensor(text_transform(txt), dtype=torch.long) for txt in raw_iter[name]]
```
zhangguanheng66, line 99: Do we decide to call `transform` during `__getitem__` func or within `_setup_datasets` func? For now, the `_setup_datasets` func in the experimental datasets just generate the transforms and we pass transforms and data to the dataset abstraction (for example, `LanguageModelingDataset`).

cpuhrsch: It's inconsistent. For the single_line version the transform is applied during construction (init function of LanguageModelingDataset), else it's applied at runtime during get_item.

zhangguanheng66: I think we need a decision for that. With the transformation during the construction, we spend more time on dataset construction, but less time for `__getitem__` (a.k.a. training time). With the opposite, we have a faster construction.

cpuhrsch: This also depends on whether we defined "get_vocab" as a read-only method.

rshraga:
closing stale PR

PR #993, opened 2020-09-19
Avoid unnecessary memory allocations in translation datasets and fix critical bugs.

cpuhrsch:
By default the constructor allocates data for all datasets, even if only a subset was constructed. 

The error handling could be written a bit more concisely and give a user feedback to fix the issue more easily.

The data_select keyword was also not forwarded and the raw version doesn't accept data_select.

The DATASETS dictionary also referenced the raw version instead of the processed.

The removed_tokens flag has no effect and was removed.

The current implementation sets "train", "test", "valid" as the default data_select, even though the test assumes train, valid, test.

This PR also creates a per-dataset subfolder for the long list of Multi30k, etc. files to make it easier for the user to keep structure in the .data folder. Maybe there are some downsides to this?

### DIFF
--- a/torchtext/experimental/datasets/raw/translation.py
+++ b/torchtext/experimental/datasets/raw/translation.py
@@ -131,7 +131,8 @@ def _setup_datasets(dataset_name,
     # We stand to download a lot of files for some of
     # these datasets. Creating a subfolder makes that more
     # manageable.
-    download_root = os.path.join(root, dataset_name)
+    download_root = root
+    # download_root = os.path.join(root, dataset_name)
     os.makedirs(download_root, exist_ok=True)
     extracted_files = []
     if isinstance(URLS[dataset_name], list):
--- a/torchtext/experimental/datasets/raw/translation.py
+++ b/torchtext/experimental/datasets/raw/translation.py
@@ -7,6 +7,7 @@ from collections import defaultdict
 
 from torchtext.utils import (download_from_url, extract_archive,
                              unicode_csv_reader)
+from torchtext.experimental.datasets.raw.utils import process_data_select
 
 URLS = {
     'Multi30k': [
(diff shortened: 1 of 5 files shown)


torchtext/experimental/datasets/translation.py
```diff
@@ -21,72 +21,61 @@ def _setup_datasets(dataset_name,
...
+        raw_iter = build_raw_iter()
+    vocab = tuple(vocab_)
+    logging.info('src Vocab has {} entries'.format(len(vocab[0])))
+    logging.info('tgt Vocab has {} entries'.format(len(vocab[1])))
+
+    raw_data = {}
+    for name in raw_iter:
+        raw_data[name] = list(raw_iter[name])
+
+    def build_transform(vocab, tokenizer):
```
zhangguanheng66, line 67: Is this same as `sequential_transforms`?

cpuhrsch: No, because it allows tokenizer to be a generator.
```
def sequential_transforms(*transforms):
    def func(txt_input):
        for transform in transforms:
            txt_input = transform(txt_input)
        return txt_input
```

requires tokenizer to materialize the entire list of tokens, whereas here we're iterating over the token the tokenizer returns and are then looking it up immediately within the vocab without ever storing it in a list.

torchtext/experimental/datasets/raw/translation.py
```diff
@@ -120,27 +120,33 @@ def _setup_datasets(dataset_name,
...
 
-    src_train, tgt_train = train_filenames
-    src_eval, tgt_eval = valid_filenames
-    src_test, tgt_test = test_filenames
+    filenames = {'train': train_filenames, 'valid': valid_filenames, 'test': test_filenames}
 
+    # We stand to download a lot of files for some of
+    # these datasets. Creating a subfolder makes that more
+    # manageable.
+    download_root = root
```
cpuhrsch, line 134: For some reason the test fails saying that 1000 lines were expected for "val", but 1014 were produced instead. The underlying raw dataset (val.de/val.en) has 1014 lines. I suspect that maybe the files we're downloading changed and the CI is using an older, cached version of them?

cpuhrsch:
This PR is getting too big, I'll split it up further.

PR #994, opened 2020-09-20
Use DATASETS consistently across experimental datasets

cpuhrsch:
The DATASETS dictionaries sometimes points to the raw versions in torchtext.experimental.datasets and sometimes to the local one, this PR 

This PR
1) opts for having DATASETS point to the datasets within the local file, instead of somethings raw.
2) removes 'LanguageModelingDataset' from experimental.datasets since the other types of datasets don't expose their factory class.
3) introduces torchtext.experimental.datasets.raw.DATASETS


zhangguanheng66 (approved):

torchtext/experimental/datasets/raw/__init__.py
```diff
@@ -6,22 +6,25 @@
...
+            'IWSLT': IWSLT,
+            'WMT14': WMT14,
+            'WikiText2': WikiText2,
+            'WikiText103': WikiText103,
+            'PennTreebank': PennTreebank,
+            'WMTNewsCrawl': WMTNewsCrawl,
+            'SQuAD1': SQuAD1,
+            'SQuAD2': SQuAD2}
+
+__all__ = list(map(str, DATASETS.keys()))
```
cpuhrsch, line 30: This should be sorted to result in the same __all__ each time since the keys of a dictionary aren't guaranteed to be ordered.

PR #995, opened 2020-09-20
Use consistent data_select default

cpuhrsch:
Right now the default sometime is "train", "test", "valid" and sometimes (but more commonly) "train", "valid", "test". We should pick a single convention (this PR opts for the latter) to avoid confusion. This also affects one of the tests.


zhangguanheng66 (approved):

test/data/test_builtin_datasets.py
```diff
@@ -52,7 +52,7 @@ def test_wikitext2(self):
         cachefile = os.path.join(self.project_root, ".data", "wikitext-2-v1.zip")
         conditional_remove(cachefile)
 
-        train_dataset, test_dataset, valid_dataset = WikiText2()
+        train_dataset, valid_dataset, test_dataset = WikiText2()
```
cpuhrsch, line 55: Why isn't this necessary for Wikitext103?

PR #1000, opened 2020-09-22
Define RawTextIterableDataset as common dataset for all raw iterators and store NUM_LINES.

cpuhrsch:
All datasets used within the experimental raw datasets are in fact very similar.

### DIFF
--- a/torchtext/experimental/datasets/raw/common.py
+++ b/torchtext/experimental/datasets/raw/common.py
@@ -19,7 +19,7 @@ class RawTextIterableDataset(torch.utils.data.IterableDataset):
     def setup_iter(self, start=0, num_lines=None):
         self.start = start
         self.num_lines = num_lines
-        if self.start + self.num_lines < self.full_num_lines:
+        if num_lines and self.start + self.num_lines > self.full_num_lines:
             raise ValueError("Requested start {} and num_lines {} exceeds available number of lines {}".format(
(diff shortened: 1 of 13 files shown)


torchtext/experimental/datasets/raw/language_modeling.py
```diff
@@ -16,37 +16,6 @@
 }
 
 
-class RawTextIterableDataset(torch.utils.data.IterableDataset):
-    """Defines an abstraction for raw text iterable datasets.
-    """
-
-    def __init__(self, iterator, start=0, num_lines=None):
```
cpuhrsch, line 23: This is an API breaking change for the constructor, but then we're not exposing this class to users.

torchtext/experimental/datasets/raw/common.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        super(RawTextIterableDataset, self).__init__()
+        self.name = name
+        self.full_num_lines = full_num_lines
+        self._iterator = iterator
+        self.has_setup = False
+        self.start = 0
+        self.num_lines = None
+
+    def setup_iter(self, start=0, num_lines=None):
```
cpuhrsch, line 19: We could also add [DataLoader support](https://github.com/pytorch/text/blob/master/examples/text_classification/iterable_train.py#L184).

zhangguanheng66 (approved):
LGTM

PR #1001, opened 2020-09-22
const string for c10::string_view

cpuhrsch:
This PR typedefs strings that are meant to be constant, i.e. read-only. They can then be optionally replaced by std::string_view.

Also adds the ever so important "#pragma once" to the header files.

### DIFF
--- a/torchtext/csrc/regex_tokenizer.cpp
+++ b/torchtext/csrc/regex_tokenizer.cpp
@@ -3,8 +3,8 @@
 
 namespace torchtext {
 
-RegexTokenizer::RegexTokenizer(const std::vector<std::string> &patterns,
-                               const std::vector<std::string> &replacements,
+RegexTokenizer::RegexTokenizer(ConstStringList patterns,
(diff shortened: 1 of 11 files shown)


torchtext/csrc/regex_tokenizer.cpp
```diff
@@ -3,8 +3,8 @@
 
 namespace torchtext {
 
-RegexTokenizer::RegexTokenizer(const std::vector<std::string> &patterns,
```
zhangguanheng66, line 6: nit: why not use const reference for functions?

cpuhrsch: Which functions?

zhangguanheng66: For example, the arguments of the Regex::Sub func are not passed by reference ([link](https://github.com/pytorch/text/pull/1001/files#diff-bbb231ee24be9ba47a93c8b6332cca7cR9)).

PR #1007, opened 2020-09-24
Allow AG_NEWS test to pass if the dataset link is down

zhangguanheng66:
Since we don't host the datasets, the link used for AG_NEWS occasionally fails. With the PR, the test is allowed to pass if the link is down.


cpuhrsch:
Try using
https://raw.githubusercontent.com/mhjabreel/CharCnn_Keras/master/data/ag_news_csv/train.csv
https://raw.githubusercontent.com/mhjabreel/CharCnn_Keras/master/data/ag_news_csv/test.csv

PR #1011, opened 2020-09-26
Update AG_NEWS url

zhangguanheng66:
Update AG_NEWS with a more stable URL.

### DIFF
--- a/torchtext/experimental/datasets/raw/text_classification.py
+++ b/torchtext/experimental/datasets/raw/text_classification.py
@@ -4,8 +4,8 @@ from torchtext.experimental.datasets.raw.common import RawTextIterableDataset
 
 URLS = {
     'AG_NEWS':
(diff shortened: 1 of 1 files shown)


torchtext/experimental/datasets/raw/text_classification.py
```diff
@@ -64,9 +65,11 @@ def get_iterator(self):
 
 
 def _setup_datasets(dataset_name, root='.data'):
-    dataset_tar = download_from_url(URLS[dataset_name], root=root)
-    extracted_files = extract_archive(dataset_tar)
-
+    if dataset_name == 'AG_NEWS':
+        extracted_files = [download_from_url(url, root=root) for url in URLS[dataset_name]]
```
cpuhrsch, line 69: In a separate PR where we're adding data_select support to this function we need to make sure that we're getting the right link. One option here is to replace the list for the links with a dictionary that is keyed on the data subset (e.g. 'train', 'test').

cpuhrsch (approved):

PR #1012, opened 2020-09-28
Uppercase pretrained_sp_model name

zhangguanheng66:


cpuhrsch (approved):
[This](https://www.python.org/dev/peps/pep-0008/#constants) is the reference to the Python styleguide entry that dictates this change.

cpuhrsch (approved):

PR #1021, opened 2020-10-06
Add explicit argus to the dataset constructors in the experimental folder

zhangguanheng66:
Add explicit arguments to the dataset constructors for better docs.


torchtext/experimental/datasets/language_modeling.py
```diff
@@ -61,7 +61,7 @@ def get_vocab(self):
 
 
 def _setup_datasets(dataset_name, tokenizer=None, root='.data', vocab=None,
```
cpuhrsch, line 63: I'd remove the default arguments and make everything positional. All callsites must provide values for all keywords, so removing the default arguments here should also improve the certainty that indeed all arguments were forwarded.

cpuhrsch: Same applies to all other _setup_datasets functions

cpuhrsch (approved):

PR #1022, opened 2020-10-06
Add more tests to cover the subset of datasets

zhangguanheng66:
This PR includes:
- Add more tests to cover the scenario for the subset of datasets

### DIFF
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -5,6 +5,7 @@ import glob
 import shutil
 import torchtext.data as data
 import torch
+import torchtext
 from ..common.torchtext_test_case import TorchtextTestCase
 
 
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -188,9 +188,9 @@ class TestDataset(TorchtextTestCase):
 
         # Add test for the subset of the standard datasets
         train_dataset, = torchtext.experimental.datasets.raw.Multi30k(data_select=('train'))
-        self._helper_test_func(len(train_dataset), 29000, next(iter(train_dataset)),
-                               ('Zwei junge weiße Männer sind im Freien in der Nähe vieler Büsche.',
-                                'Two young  White males are outside near many bushes.'))
+        self._helper_test_func(len(train_dataset), 29000, ' '.join(next(iter(train_dataset))),
+                               ' '.join(['Zwei junge weiße Männer sind im Freien in der Nähe vieler Büsche.',
+                               'Two young  White males are outside near many bushes.']))
         del train_dataset, test_dataset
         train_dataset, = Multi30k(data_select=('train'))
         self._helper_test_func(len(train_dataset), 29000, train_dataset[20],
@@ -257,7 +257,8 @@ class TestDataset(TorchtextTestCase):
                                ([262, 16, 5728, 45, 289, 701, 1160, 4436, 10660, 585],
                                 [6, 20, 8, 10, 8, 8, 24, 13, 8, 15]))
         train_iter, = torchtext.experimental.datasets.raw.UDPOS(data_select=('train'))
-        self._helper_test_func(len(train_iter), 12543, next(iter(train_iter))[0][:5], ['Al', '-', 'Zaman', ':', 'American'])
+        self._helper_test_func(len(train_iter), 12543, ' '.join(next(iter(train_iter))[0][:5]),
+                               ' '.join(['Al', '-', 'Zaman', ':', 'American']))
         del train_iter
 
     def test_conll_sequence_tagging(self):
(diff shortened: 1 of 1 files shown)


test/data/test_builtin_datasets.py
```diff
@@ -89,9 +83,13 @@ def test_penntreebank(self):
         tokens_ids = [vocab[token] for token in 'the player characters rest'.split()]
         self.assertEqual(tokens_ids, [2, 2550, 3344, 1125])
 
+        # Add test for the subset of the standard datasets
+        train_dataset, test_dataset = PennTreebank(data_select=('train', 'test'))
+        train_iter, test_iter = torchtext.experimental.datasets.raw.PennTreebank(data_select=('train', 'test'))
```
cpuhrsch, line 88: I'd check whether the returned iterator then is also indeed for training or testing as you'd expect. That way we make sure this subset selection is working.

zhangguanheng66: Fixed

test/data/test_builtin_datasets.py
```diff
@@ -84,14 +89,29 @@ def test_penntreebank(self):
...
 
         vocab = train_dataset.get_vocab()
         tokens_ids = [vocab[token] for token in 'the player characters rest'.split()]
         self.assertEqual(tokens_ids, [2, 2550, 3344, 1125])
 
+        # Add test for the subset of the standard datasets
+        train_dataset, test_dataset = PennTreebank(data_select=('train', 'test'))
+        self.assertEqual(len(train_dataset), 924412)
+        self.assertEqual(len(test_dataset), 82114)
+        self.assertEqual(train_dataset[20:25], torch.tensor([9919, 9920, 9921, 9922, 9188]).long())
```
cpuhrsch, line 104: nit: [torch.tensor](https://pytorch.org/docs/master/generated/torch.tensor.html#torch.tensor) has a dtype flag, so you can write ```torch.tensor([...], dtype=torch.int64)```.

test/data/test_builtin_datasets.py
```diff
@@ -56,11 +46,26 @@ def test_wikitext2(self):
...
 
         vocab = train_dataset.get_vocab()
         tokens_ids = [vocab[token] for token in 'the player characters rest'.split()]
         self.assertEqual(tokens_ids, [2, 286, 503, 700])
 
+        # Add test for the subset of the standard datasets
+        train_dataset, test_dataset = torchtext.experimental.datasets.raw.WikiText2(data_select=('train', 'test'))
+        next(iter(train_dataset))
+        self.assertEqual(next(iter(train_dataset)), ' = Valkyria Chronicles III = \n')
+        next(iter(test_dataset))
```
cpuhrsch, line 61: based on offline discussion: Another way to get rid of the iterator that's being purposefully overwritten by this naming is to use ```del train_iter```

test/data/test_builtin_datasets.py
```diff
@@ -56,11 +46,26 @@ def test_wikitext2(self):
...
+        self.assertEqual(train_dataset[20:25], torch.tensor([5024, 89, 21, 3, 1838]).long())
+        self.assertEqual(test_dataset[30:35], torch.tensor([914, 4, 36, 11, 569]).long())
+        self.assertEqual(valid_dataset[40:45], torch.tensor([925, 8, 2, 150, 8575]).long())
 
         vocab = train_dataset.get_vocab()
         tokens_ids = [vocab[token] for token in 'the player characters rest'.split()]
         self.assertEqual(tokens_ids, [2, 286, 503, 700])
 
+        # Add test for the subset of the standard datasets
+        train_dataset, test_dataset = torchtext.experimental.datasets.raw.WikiText2(data_select=('train', 'test'))
```
cpuhrsch, line 58: You can also add a length test to this.

test/data/test_builtin_datasets.py
```diff
@@ -124,13 +154,31 @@ def test_imdb(self):
...
+        self.assertEqual(text[:25], 'I rented I AM CURIOUS-YEL')
+
     def test_multi30k(self):
         from torchtext.experimental.datasets import Multi30k
         # smoke test to ensure multi30k works properly
         train_dataset, valid_dataset, test_dataset = Multi30k()
         self.assertEqual(len(train_dataset), 29000)
         self.assertEqual(len(valid_dataset), 1014)
         self.assertEqual(len(test_dataset), 1000)
+        self.assertEqual(train_dataset[20], (torch.tensor([3, 443, 2530, 46, 17478, 7422, 7, 157, 9, 11, 5848, 2]).long(),
```
cpuhrsch, line 176: ```:%s/torch.tensor([\(.*\)]).long()/torch.tensor(\[\1\], dtype=torch.int64/g```
vim regex search replace pattern to save you some time.

test/data/test_builtin_datasets.py
```diff
@@ -18,13 +18,12 @@ def conditional_remove(f):
 
 
 class TestDataset(TorchtextTestCase):
+    def helper_test_func(self, length, target_length, results, target_results):
```
cpuhrsch, line 21: Using the [style guide](https://www.python.org/dev/peps/pep-0008/#designing-for-inheritance) we should probably prefix this with a leading underscore.

cpuhrsch (approved):

test/data/test_builtin_datasets.py
cpuhrsch: As a suggestion, you could have one helper for datasets and one for iterators if you want to distinguish between these two cases to make writing code easier. You could also move the tensor construction ```torch.tensor([...], dtype=torch.int64)``` into these helpers.

PR #1027, opened 2020-10-09
Remove <unk> token and index from experimental Vocab

zhangguanheng66:


zhangguanheng66:
cc @bentrevett Please review the PR and let us know if you have any other suggestions.

bentrevett:
@zhangguanheng66 All looks good to me.

cpuhrsch:
Maybe "default" is a better name than "fallback" since it's akin to the default kwarg passed to [dict.get](https://docs.python.org/3/library/stdtypes.html#dict.get).

torchtext/csrc/register_bindings.cpp
```diff
@@ -51,12 +51,13 @@ PYBIND11_MODULE(_torchtext, m) {
...
   py::class_<Vocab>(m, "Vocab")
-      .def(py::init<std::vector<std::string>, std::string>())
+      .def(py::init<std::vector<std::string>>())
       .def_readonly("itos_", &Vocab::itos_)
-      .def_readonly("unk_token_", &Vocab::unk_token_)
       .def("__getitem__", &Vocab::__getitem__)
       .def("__len__", &Vocab::__len__)
       .def("insert_token", &Vocab::insert_token)
+      .def("set_default_index", &Vocab::set_default_index)
+      .def("return_default_index", &Vocab::return_default_index)
```
cpuhrsch, line 60: nit: I'd expect "get" instead of "return" as a counterpart for a function that uses "set", which is more in line of the concepts of "getters and setters".

test/experimental/test_vocab.py
```diff
@@ -17,21 +17,23 @@ def tearDown(self):
...
+            v['not_in_it']
+        with self.assertRaises(RuntimeError):
+            v['<unk>']
 
-        # check if new_unk is mapped to the first index
-        self.assertEqual(v['<new_unk>'], 0)
-        self.assertEqual(v['not_in_it'], 0)
+        v.insert_token('not_in_it', 0)
+        v.set_default_index(0)
+        self.assertEqual(v.get_default_index(), 0)
```
cpuhrsch, line 36: You probably also want to check what numbers these tokens correspond to

```
        self.assertEqual(v['not_in_it'], 0)
        self.assertEqual(v['<unk>'], 0)
```

cpuhrsch:
Related to this would be a test that verifies the behavior of insert_token(```<unk>```) if ```<unk>``` is already part of the vocabulary.

test/experimental/test_vocab.py
```diff
@@ -17,21 +17,25 @@ def tearDown(self):
...
         torch.jit._recursive.concrete_type_store = torch.jit._recursive.ConcreteTypeStore()
 
-    def test_has_unk(self):
+    # we separate out these errors because Windows runs into seg faults when propagating
+    # exceptions from C++ using pybind11
+    @unittest.skipIf(platform.system() == "Windows", "Test is known to fail on Windows.")
+    def test_has_no_unk(self):
         c = OrderedDict()
         v = vocab(c)
+        self.assertEqual(v.get_default_index(), -1)
```
cpuhrsch, line 26: I think it's better if this were to return "None". You should be able to do this easily by using c10::optional<int64_t> instead of int64_t in the C++ code.

cpuhrsch:
As an aside, while we're introducing this for the Vocab we should probably as a follow-up also introduce the same concepts to the Vectors class

cpuhrsch (changes requested):
Before we merge this we also need to support resassignment. A special token might show up in the dataset and ends up inadvertently being mapped to the wrong index. For example, <unk> might show up in the dataset used to build this Vocab, but really the user wants it to be mapped to index "0" (which is what Vocab currrently does).

PR #1029, opened 2020-10-11
Update the names of the experimental factory functions for vocab and vectors

zhangguanheng66:
This PR is to standardize a few names of the factory functions. We add a verb to the factory function to indicate the purpose. The word "build" means creating a block, while "load" mean loading a pre-generated block from an object (like a file). 

experimental.vectors:
- `vectors` -> `build_vectors`
- `vectors_from_file_object` -> `load_vectors_from_file_object`

experimental.vocab
- `vocab_from_file` -> `load_vocab_from_file`
- `vocab_from_raw_text_file` -> `build_vocab_from_raw_text_file`


cpuhrsch (changes requested):
I'd constraint this to vocab and vectors only and solely to distinguish between loading a prebuilt vocab / vector from a file and build a vocab / vector based on a text file.

torchtext/experimental/vectors.py
```diff
@@ -144,7 +144,7 @@ def GloVe(name="840B", dim=300, unk_tensor=None, root=".data", validate_file=Tru
     return vectors_obj
 
 
-def vectors_from_file_object(file_like_object, delimiter=",", unk_tensor=None, num_cpus=10):
+def load_vectors_from_file_object(file_like_object, delimiter=",", unk_tensor=None, num_cpus=10):
```
cpuhrsch, line 147: This actually assumes the file_object comes from a path via open. We should just restrict this to accept paths only. We should then also add support for [pathlib](https://docs.python.org/3/library/pathlib.html) while we're at it.

benchmark/benchmark_experimental_vocab.py
```diff
@@ -6,8 +6,8 @@
 from torchtext.experimental.datasets import AG_NEWS
 from torchtext.experimental.vocab import (
     vocab as VocabExperimental,
-    vocab_from_file,
-    vocab_from_raw_text_file
+    load_vocab_from_file,
+    build_vocab_from_raw_text_file
```
cpuhrsch, line 10: Let's change "build_vocab_from_raw_text_file" to "build_vocab_from_text_file" until we have a better definition of raw

cpuhrsch (approved):

cpuhrsch (approved):

PR #1049, opened 2020-10-20
Update doc strings

zhangguanheng66:


cpuhrsch (approved):
Please add a short description if there was any commands involved

PR #1055, opened 2020-10-26
Remove sentencepiece from setup.py file

zhangguanheng66:
Since we added the sentencepiece dependency from the source code, we don't need it in the setup.py file anymore.


cpuhrsch (approved):
It's ok to merge this right now, since sentencepiece is indeed not required for the library as an explicit dependency, but please in a follow-up PR bring it back for testing purposes.

PR #1060, opened 2020-10-28
Add checksum to the raw datasets

zhangguanheng66:
Update raw datasets with:

- checksum md5
- NUM_LINES

### DIFF
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -63,10 +63,11 @@ class TestDataset(TorchtextTestCase):
         self.assertEqual(tokens_ids, [2, 286, 503, 700])
 
         # Add test for the subset of the standard datasets
-        train_dataset, test_dataset = torchtext.experimental.datasets.raw.WikiText2(data_select=('train', 'test'))
-        self._helper_test_func(len(train_dataset), 36718, next(iter(train_dataset)), ' \n')
-        self._helper_test_func(len(test_dataset), 4358, next(iter(test_dataset)), ' \n')
-        del train_dataset, test_dataset
+        train_iter, valid_iter, test_iter = torchtext.experimental.datasets.raw.WikiText2(data_select=('train', 'valid', 'test'))
+        self._helper_test_func(len(train_iter), 36718, next(iter(train_iter)), ' \n')
+        self._helper_test_func(len(valid_iter), 3760, next(iter(valid_iter)), ' \n')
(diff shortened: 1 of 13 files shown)


test/data/test_builtin_datasets.py
```diff
@@ -65,7 +65,7 @@ def test_wikitext2(self):
         # Add test for the subset of the standard datasets
         train_dataset, test_dataset = torchtext.experimental.datasets.raw.WikiText2(data_select=('train', 'test'))
         self._helper_test_func(len(train_dataset), 36718, next(iter(train_dataset)), ' \n')
-        self._helper_test_func(len(test_dataset), 36718, next(iter(test_dataset)), ' \n')
+        self._helper_test_func(len(test_dataset), 4358, next(iter(test_dataset)), ' \n')
```
cpuhrsch, line 68: We just need to spread this to the other datasets as well and then this is good to go

cpuhrsch (approved):

cpuhrsch (approved):

mthrok:
Question: Was there a discussion on the justification of the use of MD5? MD5 is fast but known to easily collide especially when two strings are similar. I know this might not be an immediate threat but wonder what you think.

https://www.mscs.dal.ca/~selinger/md5collision/

mthrok:
Also, I am wondering when checksum is computed. (not immediately clear from the CC)
Does it happen at download time? at the time dataset object is instantiated? or every time files are loaded?

zhangguanheng66:
> Also, I am wondering when checksum is computed. (not immediately clear from the CC)
> Does it happen at download time? at the time dataset object is instantiated? or every time files are loaded?

It's calculated after downloading.

zhangguanheng66:
> Also, I am wondering when checksum is computed. (not immediately clear from the CC)
> Does it happen at download time? at the time dataset object is instantiated? or every time files are loaded?

We have a follow-up PR to switch to SHA2.

zhangguanheng66:
> Also, I am wondering when checksum is computed. (not immediately clear from the CC)
> Does it happen at download time? at the time dataset object is instantiated? or every time files are loaded?

We have a follow-up PR to switch to SHA2 after benchmarking the performance of those two checksum.

mthrok:
> > Also, I am wondering when checksum is computed. (not immediately clear from the CC)
> > Does it happen at download time? at the time dataset object is instantiated? or every time files are loaded?
> 
> We have a follow-up PR to switch to SHA2 after benchmarking the performance of those two checksum.

sounds good. thanks for the response.

PR #1061, opened 2020-10-28
[BC BREAKING]Remove single_line support from all the language modeling datasets

zhangguanheng66:
This is a BC breaking PR. Remove the single_line support to all the language modeling datasets. To obtain all the tokens in a single line
```
train_data = torch.cat(tuple(filter(lambda t: t.numel() > 0, train_dataset)))
```

### DIFF
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -51,6 +51,9 @@ class TestDataset(TorchtextTestCase):
         conditional_remove(cachefile)
 
         train_dataset, valid_dataset, test_dataset = WikiText2()
+        train_dataset.data = torch.cat(tuple(filter(lambda t: t.numel() > 0, train_dataset.data)))
+        valid_dataset.data = torch.cat(tuple(filter(lambda t: t.numel() > 0, valid_dataset.data)))
+        test_dataset.data = torch.cat(tuple(filter(lambda t: t.numel() > 0, test_dataset.data)))
         self._helper_test_func(len(train_dataset), 2049990, train_dataset[20:25],
                                [5024, 89, 21, 3, 1838])
         self._helper_test_func(len(test_dataset), 241859, test_dataset[30:35],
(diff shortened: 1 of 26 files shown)


torchtext/experimental/datasets/language_modeling.py
```diff
@@ -215,6 +217,10 @@ def WMTNewsCrawl(tokenizer=None, root='.data', vocab=None, data_select=('train')
             vocabulary based on the train data set.
         data_select: a string or tuple for the returned datasets
             (Default: ('train',))
+        single_line: whether to return all tokens in a single line.
+            (Default: True)
+            By default, all lines in raw text file are concatenated into a single line.
```
cpuhrsch, line 222: nit: files

cpuhrsch (changes requested):
I don't think the single_line argument is necessary at all. It turns the dataset from a line-by-line fashion into a token-by-token fashion. We should double-check that it's very easy to convert a line-by-line dataset into a token-by-token dataset and then just add that code as an example in the BERT example or possibly turn it into a utility.

test/data/test_builtin_datasets.py
```diff
@@ -51,6 +51,9 @@ def test_wikitext2(self):
         conditional_remove(cachefile)
 
         train_dataset, valid_dataset, test_dataset = WikiText2()
+        train_dataset.data = torch.cat(tuple(filter(lambda t: t.numel() > 0, train_dataset.data)))
```
cpuhrsch, line 54: I don't think the .data field is necessary for this and we should probably even consider making it private.

torchtext/experimental/datasets/language_modeling.py
```diff
@@ -145,23 +141,23 @@ def WikiText103(tokenizer=None, root='.data', vocab=None, data_select=('train',
...
-            (Default: True)
-            By default, all lines in raw text file are concatenated into a single line.
-            Use `single_line = False` if one wants to get data line by line.
 
     Examples:
         >>> from torchtext.experimental.datasets import WikiText103
         >>> from torchtext.data.utils import get_tokenizer
         >>> tokenizer = get_tokenizer("spacy")
         >>> train_dataset, valid_dataset, test_dataset = WikiText103(tokenizer=tokenizer)
+        >>> # convert dataset into a single line
```
cpuhrsch, line 150: I don't think this is necessary right here. We could however include this in one of our tutorials to show how easy it can be.

cpuhrsch:
Please update description and title according to content

cpuhrsch (approved):
Looks good, please update the title to reflect this change more accurately before merging.

PR #1062, opened 2020-10-29
Add logging for experimental datasets

zhangguanheng66:
Add logging for the experimental datasets:

- text classification
- language modeling
- question answer
- sequence tagging
- translation

The logging information include:

- building vocabulary
- vocabulary length
- building datasets

### DIFF
--- a/torchtext/experimental/datasets/language_modeling.py
+++ b/torchtext/experimental/datasets/language_modeling.py
@@ -5,7 +5,7 @@ from torchtext.vocab import build_vocab_from_iterator
 from torchtext.experimental.datasets.raw import language_modeling as raw
 from torchtext.experimental.datasets.raw.common import check_default_set
 
-_logger = logging.getLogger(__name__)
+logger_ = logging.getLogger(__name__)
 
 
 def build_vocab(data, transforms):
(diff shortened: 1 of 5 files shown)


mthrok (commented):
Nice! 

In the library codebase, instead of calling `logging.info`, defining a logger dedicated for the module is considered to be a good practice.

For example,

```python
# On top level
_LG = logging.getLogger(__name__)

# anywhere in the code
_LG.info('foobar')
```

torchtext/experimental/datasets/language_modeling.py
```diff
@@ -76,7 +77,9 @@ def _setup_datasets(dataset_name, tokenizer, root, vocab, data_select, single_li
             raw_train, = raw.DATASETS[dataset_name](root=root, data_select=('train',), year=year, language=language)
         else:
             raw_train, = raw.DATASETS[dataset_name](root=root, data_select=('train',))
+        logging.info('Building Vocab based on train data')
         vocab = build_vocab(raw_train, tokenizer)
+    logging.info('Vocab has {} entries'.format(len(vocab)))
```
mthrok, line 82: Instead of building the string before passing it to the logger object, a common way is to let the logger object construct the string.

```python
info('Vocab has %s entries', len(vocab))
```

torchtext/experimental/datasets/language_modeling.py
```diff
@@ -1,4 +1,5 @@
 import torch
+import logging
```
mthrok, line 2: nit: PEP-8 recommends to group imports into three groups (standard modules, third party modules, the module this code is part of), separated by one new line.

```python
import logging

import torch

import torchtext...
```

torchtext/experimental/datasets/language_modeling.py
zhangguanheng66: added

cpuhrsch:
Please add a description to the PR

cpuhrsch (approved):
Since this is just spreading a pattern across torchtext to be consistent, I think it's fine. But we'll revisit logging in a cross-domain context.

PR #1078, opened 2020-11-11
[BC Breaking] Remove unicode_csv_reader from raw translation datasets

zhangguanheng66:
For the raw translation datasets, we mistakenly used `unicode_csv_reader` to preprocess the string. Just read the file and yield the raw strings. See a relevant issue https://github.com/pytorch/text/issues/1081


cpuhrsch (changes requested):
This is a big change, because it causes the entire dataset to be different. Before landing this, let's figure out what exactly the change in output is (e.g. all these new lines) and how doing this is "more correct" than what we're currently doing. Also please flag this as BC breaking.

zhangguanheng66:
Link to a relevant issue https://github.com/pytorch/text/issues/1081

cpuhrsch (approved):
Update the description to indicate that the csv reader was used erroneously.

PR #1080, opened 2020-11-11
switch to_ivalue to __prepare_scriptable__

zhangguanheng66:
After merging https://github.com/pytorch/pytorch/pull/49242, the relevant building blocks and pipelines in torchtext are switching to `__prepare_scriptable__`. Users are no longer required to call to_ivalue when `torch.jit.script`.


examples/data_pipeline/transforms.py
```diff
@@ -24,11 +24,11 @@ def forward(self, tokens: List[str]) -> List[int]:
...
         self.vocab.insert_token(token, index)
 
-    def to_ivalue(self):
-        if hasattr(self.vocab, 'to_ivalue'):
+    def __prepare_scriptable__(self):
+        if hasattr(self.vocab, '__prepare_scriptable__'):
             sp_model = self.sp_model
             new_module = PretrainedSPVocab(sp_model)
-            new_module.vocab = self.vocab.to_ivalue()
+            new_module.vocab = self.vocab.__prepare_scriptable__()
```
cpuhrsch, line 31: Isn't prepare_scriptable recursive to members of modules? Will we even need to call this manually? Do we even need this method at all?

examples/data_pipeline/transforms.py
```diff
@@ -57,9 +57,9 @@ def __init__(self, vocab):
     def forward(self, tokens: List[str]) -> List[int]:
         return self.vocab.lookup_indices_1d(tokens)
 
-    def to_ivalue(self):
-        if hasattr(self.vocab, 'to_ivalue'):
-            vocab = self.vocab.to_ivalue()
+    def __prepare_scriptable__(self):
+        if hasattr(self.vocab, '__prepare_scriptable__'):
```
cpuhrsch, line 61: Same here

test/data/test_functional.py
```diff
@@ -94,14 +94,14 @@ def test_BasicEnglishNormalize(self):
...
-        jit_basic_eng_norm = torch.jit.script(basic_eng_norm.to_ivalue())
+        jit_basic_eng_norm = torch.jit.script(basic_eng_norm)
         experimental_jit_tokens = jit_basic_eng_norm(test_sample)
 
         basic_english_tokenizer = data.get_tokenizer("basic_english")
         eager_tokens = basic_english_tokenizer(test_sample)
 
         assert not basic_eng_norm.is_jitable
-        assert basic_eng_norm.to_ivalue().is_jitable
+        assert basic_eng_norm.__prepare_scriptable__().is_jitable
```
cpuhrsch, line 104: I'd interact with this via torch.jit.script. These methods are not meant to be called directly.

test/data/test_functional.py
```diff
@@ -121,7 +121,7 @@ def test_basicEnglishNormalize_load_and_save(self):
 
         with self.subTest('torchscript'):
             save_path = os.path.join(self.test_dir, 'ben_torchscrip.pt')
-            ben = basic_english_normalize().to_ivalue()
+            ben = basic_english_normalize().__prepare_scriptable__()
```
cpuhrsch, line 124: I'd interact with this via torch.jit.script. These methods are not meant to be called directly.

test/data/test_functional.py
```diff
@@ -149,11 +149,11 @@ def test_RegexTokenizer(self):
         r_tokenizer = regex_tokenizer(patterns_list)
         eager_tokens = r_tokenizer(test_sample)
 
-        jit_r_tokenizer = torch.jit.script(r_tokenizer.to_ivalue())
+        jit_r_tokenizer = torch.jit.script(r_tokenizer)
         jit_tokens = jit_r_tokenizer(test_sample)
 
         assert not r_tokenizer.is_jitable
-        assert r_tokenizer.to_ivalue().is_jitable
+        assert r_tokenizer.__prepare_scriptable__().is_jitable
```
cpuhrsch, line 156: Same here

test/data/test_functional.py
```diff
@@ -186,7 +186,7 @@ def test_load_and_save(self):
 
         with self.subTest('torchscript'):
             save_path = os.path.join(self.test_dir, 'regex_torchscript.pt')
-            tokenizer = regex_tokenizer(patterns_list).to_ivalue()
+            tokenizer = regex_tokenizer(patterns_list).__prepare_scriptable__()
```
cpuhrsch, line 189: Same here

test/experimental/test_transforms.py
```diff
@@ -74,7 +74,7 @@ def test_sentencepiece_load_and_save(self):
 
         with self.subTest('torchscript'):
             save_path = os.path.join(self.test_dir, 'spm_torchscript.pt')
-            spm = sentencepiece_tokenizer((model_path)).to_ivalue()
+            spm = sentencepiece_tokenizer((model_path)).__prepare_scriptable__()
```
cpuhrsch, line 77: Same here

test/data/test_functional.py
zhangguanheng66: Here we want to check if the building block backed by torchbind is jitable. Although we don't recommend the torchbind version for eager mode, I think we still need a test to cover it. By calling `torch.jit.script`, we are testing the jit mode, which is not the same thing with the torchbind building block.

Same thing for the pickle support.

test/data/test_functional.py
zhangguanheng66: See the comment above.

test/data/test_functional.py
zhangguanheng66: See the comment above.

test/data/test_functional.py
zhangguanheng66: See the comment above. We want to save a torchbind version model here, not exactly the JIT mode.

test/experimental/test_transforms.py
zhangguanheng66: See the comment above.

examples/data_pipeline/transforms.py
zhangguanheng66: This is not necessary any more since the `__prepare_scriptable__` func recursively prepares the components.

examples/data_pipeline/transforms.py
zhangguanheng66: This is not necessary any more since the `__prepare_scriptable__` func recursively prepares the components.

cpuhrsch (approved):

PR #1083, opened 2020-11-17
Update docs for torchtext.nn.InProjContainer

zhangguanheng66:
Update the docs for torchtext.nn.InProjContainer with the purpose of the module. Include an example.


torchtext/nn/modules/multiheadattention.py
```diff
@@ -223,9 +223,19 @@ def forward(self,
...
+                                                    torch.nn.Linear(embed_dim, embed_dim),
+                                                    torch.nn.Linear(embed_dim, embed_dim))
+            >>> q = torch.rand((5, bsz, embed_dim))
+            >>> k = v = torch.rand((6, bsz, embed_dim))
+            >>> q, k, v = in_proj_container(q, k, v)
+
         Shape:
-            - query, key, value: :math:`(S, N, E)`
-            - Output: :math:`(S, N, E)`.
+            - query, key, value: :math:`(S, N, E)`, :math:`(S, N, E)`, :math:`(S, N, E)`
```
cpuhrsch, line 237: I'd consider this misleading, because it entirely depends on the module that the user passes in and by default the linear layer supports broadcasting and any size or shape applicable to matrix multiplication.

cpuhrsch: If this is about the default, i think we should mention that and also refer to the documentation of linear.

torchtext/nn/modules/multiheadattention.py
```diff
@@ -200,13 +200,13 @@ def forward(self, query: torch.Tensor, key: torch.Tensor, value: torch.Tensor,
 
 class InProjContainer(torch.nn.Module):
     def __init__(self, query_proj, key_proj, value_proj):
-        r"""A in-proj container to process inputs.
+        r"""A in-proj container to project query/key/value in MultiheadAttention. This module happens before reshaping
```
cpuhrsch, line 203: A picture could really be useful here or at least a reference to a picture in a paper.

torchtext/nn/modules/multiheadattention.py
cpuhrsch: In fact, let's just write out the code we have as documentation since it's super basic for now. As a follow-up task, maybe it's time to look into using torch.bmm when applicable.

torchtext/nn/modules/multiheadattention.py
```diff
@@ -200,13 +200,14 @@ def forward(self, query: torch.Tensor, key: torch.Tensor, value: torch.Tensor,
...
+        the projected query/key/value into multiple heads. See the linear layers (bottom) of Multi-head Attention in
+        Fig 2 of Attention Is All You Need paper.
```
cpuhrsch, line 205: I'd a reference to MHAContainer that showcases another usesite of this

cpuhrsch (approved):

PR #1036, opened 2020-10-14
Set release version and base PyTorch version

mthrok:

PR #1084, opened 2020-11-20
Bump the release version to 0.8.1 for the point release

zhangguanheng66:


zhangguanheng66:
I think those three files need to update with a new version number. @seemethere let me know if I miss some files.

mthrok:
The version configurations in `packaging` directory are for nightly, you need to change them in `config.yaml.in` and update `config.yaml` like in https://github.com/pytorch/text/pull/1036 .

We can update the version number of domain libraries now but later we also need to update the base PyTorch version when 1.7.1 RC is publicly available.

zhangguanheng66:
> The version configurations in `packaging` directory are for nightly, you need to change them in `config.yaml.in` and update `config.yaml` like in [#1036](https://github.com/pytorch/text/pull/1036/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L39).
> 
> We can update the version number of domain libraries now but later we also need to update the base PyTorch version when 1.7.1 RC is publicly available.

Thanks. Let's update them all together.

PR #1085, opened 2020-11-22
pybind11 pickle support for experimental vocab, vectors, regex_tokenizer

zhangguanheng66:
Add pybind11 pickle support for
- experimental.vocab
- experimental.vectors
- experimental.basic_english_normalize

No need to call `to_ivalue` before saving.


torchtext/csrc/register_bindings.cpp
```diff
@@ -62,7 +92,21 @@ PYBIND11_MODULE(_torchtext, m) {
       .def("lookup_tokens", &Vocab::lookup_tokens)
       .def("lookup_indices", &Vocab::lookup_indices)
       .def("get_stoi", &Vocab::get_stoi)
-      .def("get_itos", &Vocab::get_itos);
+      .def("get_itos", &Vocab::get_itos)
+      .def(py::pickle(
```
cpuhrsch, line 96: Maybe we can use the existing _set_vocab_states etc. functionality

zhangguanheng66: Similar to `def_pickle` func of torchbind, I tried `_set_vectors_states` and `_get_vectors_from_states` in `py::pickle` func of pybind11. However, the intrusive_ptr holder used by `_set_vectors_states` and `_get_vectors_from_states` are not supported in pybind11 pickle mechanism.

```
 error: static assertion failed: pybind11::init(): init function must return a compatible pointer, holder, or value
```

cpuhrsch: I think you could change 
```
_set_vectors_states(const c10::intrusive_ptr<Vectors> &self)
```
to
```
_set_vectors_states(const Vectors &self)
```

and then change the callsite of ```_set_vectors_states(self)``` to ```_set_vectors_states(*self)```.

PR #1102, opened 2020-12-16
Simplify TorchScript registration with TORCH_LIBRARY_FRAGMENT macro

mthrok:
This PR adopts `TORCH_LIBRARY_FRAGMENT` to make the TorchScript registration simpler and similar to the PyBind11.

Also the order of registration is changed a bit to match the one in PyBind11.


cpuhrsch (approved):

mthrok:
thanks!

PR #1104, opened 2020-12-16
Make Regex, RegexTokenizer, Vocab, Vectors, SentencePiece pickle-able

mthrok:
This PR makes Regex, RegexTokenizer, Vocab, Vectors and SentencePiece
pickle-able on both PyBind11 and TorchScript.

closes #1085 

## Approach

1. define `_serialize_XXX` and `_deserialize_XXX` next to these classes
    This is the replacement of `_get_states_XXX` and `_set_states_XXX`.
    I saw the names of the original functions were flipped, and used wrongly in
    `__getstate__` and `__setstate__` so I changed the function names to something
    more descriptive and less confusing.

2. Use `c10::intrusive_ptr` as holder for custom class when using pybind11.
    This allows to use the same serialization/deserialization function for both
    PyBind11 and TorchScript.
    See https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html#smart-pointers
    for the detail of holder.

3. For pickling TorchScript-bound `SentencePiece`, use byte Tensor as bytes container
    The serialized form of `SentencePiece` is byte string and returning `std::string` to 
    Python realm causes decoding error as Python tries to decode it as UTF-8.
    PyBind11 can work around this with `pybind11::bytes` type, but TorchScript does not
    support byte string, this approach uses bytes Tensor as a container/intermediate format
    to pass byte string to Python.

## Problem

* ~[TorchScript does not support `bytes`](https://pytorch.org/docs/stable/jit_language_reference.html#unsupported-typing-constructs), thus SentencePiece bound via TorchScript is not pickle-able.~

## Added I/O round trip tests

| class | PyBind11 | TorchScript |
|:-----:|:---------:|:------------:|
| `Regex` | | |
| `RegexTokenizer` | ✅ | ✅ |
| `BasicEnglishNormalize` | ✅ | ✅ |
| `Vocab` | ✅ | ✅ |
| `Vectors` | ✅ | ✅ |
| `SentencePiece` | ✅ | ✅ |

### DIFF
--- a/test/experimental/test_vectors.py
+++ b/test/experimental/test_vectors.py
@@ -111,6 +111,23 @@ class TestVectors(TorchtextTestCase):
         self.assertEqual(vectors_obj['b'], tensorB)
         self.assertEqual(vectors_obj['not_in_it'], unk_tensor)
 
+    def test_vectors_update(self):
+        tensorA = torch.tensor([1, 0], dtype=torch.float)
+        tensorB = torch.tensor([0, 1], dtype=torch.float)
+        tensorC = torch.tensor([1, 1], dtype=torch.float)
+
+        expected_unk_tensor = torch.tensor([0, 0], dtype=torch.float)
+
+        tokens = ['a', 'b']
+        vecs = torch.stack((tensorA, tensorB), 0)
+        vectors_obj = build_vectors(tokens, vecs)
+
+        vectors_obj['b'] = tensorC
+
+        self.assertEqual(vectors_obj['a'], tensorA)
+        self.assertEqual(vectors_obj['b'], tensorC)
+        self.assertEqual(vectors_obj['not_in_it'], expected_unk_tensor)
+
     def test_vectors_load_and_save(self):
         tensorA = torch.tensor([1, 0], dtype=torch.float)
         tensorB = torch.tensor([0, 1], dtype=torch.float)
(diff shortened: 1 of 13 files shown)


mthrok:
I cannot find the usage of `Regex` class in Python realm. What should I do with that?

test/experimental/test_vectors.py
```diff
@@ -111,25 +111,49 @@ def test_vectors_add_item(self):
         self.assertEqual(vectors_obj['b'], tensorB)
         self.assertEqual(vectors_obj['not_in_it'], unk_tensor)
 
-    def test_vectors_load_and_save(self):
+    def test_vectors_update(self):
```
mthrok, line 114: Splitting the test for updating as I think it should be tested separately from serialization.

torchtext/csrc/register_bindings.cpp
```diff
@@ -129,13 +168,21 @@ TORCH_LIBRARY_FRAGMENT(torchtext, m) {
...
+        // and returing it as std::string cause UTF8 decoding error.
+        // Since TorchScript does not support byte string, we use byte Tensor to
+        // pass around the data.
         // __getstate__
-        [](const c10::intrusive_ptr<SentencePiece> &self) -> std::string {
-          return self->content_;
+        [](const c10::intrusive_ptr<SentencePiece> &self) -> torch::Tensor {
+          auto *data = static_cast<void*>(const_cast<char*>(self->content_.data()));
+          auto numel = static_cast<int64_t>(self->content_.size());
+          return torch::from_blob(data, {numel}, {torch::kUInt8});
```
mthrok, line 179: I think it's okay to return a tensor with borrowed memory. The resulting Tensor should be serialized right away.

cpuhrsch: For now, let's return a copy and then remove that copy as a performance optimization later on when we can be 100% sure about this.

torchtext/csrc/regex_tokenizer.cpp
```diff
@@ -44,4 +44,15 @@ void RegexTokenizer::split_(std::string &str, std::vector<std::string> &tokens,
   }
 }
 
+RegexTokenizerStates _serialize_regex_tokenizer(const c10::intrusive_ptr<RegexTokenizer> &self) {
+  return std::make_tuple(self->patterns_, self->replacements_, self->to_lower_);
+}
+
+c10::intrusive_ptr<RegexTokenizer> _deserialize_regex_tokenizer(RegexTokenizerStates states) {
+  return c10::make_intrusive<RegexTokenizer>(
+      std::move(std::get<0>(states)),
```
cpuhrsch, line 53: I think using move semantics is a good idea, but then this function also should accept an rvalue.

cpuhrsch (approved):

mthrok:
The issue with linux conda build has nothing to do with the changes in this PR

https://github.com/conda/conda-package-handling/issues/71

seemethere:
> The issue with linux conda build has nothing to do with the changes in this PR
> 
> [conda/conda-package-handling#71](https://github.com/conda/conda-package-handling/issues/71)

This should be resolved with https://github.com/pytorch/text/pull/1106

zhangguanheng66 (approved):

PR #1106, opened 2020-12-18
packaging: Specify cpuonly for conda meta.yaml

seemethere:
Package installation for pytorch binaries was failing out due to
https://github.com/conda/conda-package-handling/issues/71

Default to cpuonly since that won't fail out.

Similar to https://github.com/pytorch/audio/pull/1105

This should resolve issues within the nightly pipeline for conda builds

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -237,7 +237,7 @@ jobs:
           command: |
             set -x
             source /usr/local/etc/profile.d/conda.sh && conda activate python${PYTHON_VERSION}
-            conda install -v -y -c pytorch-${UPLOAD_CHANNEL} pytorch
+            conda install -v -y -c pytorch-${UPLOAD_CHANNEL} pytorch cpuonly
             conda install -v -y -c file://$HOME/workspace/conda-bld torchtext
       - run:
           name: smoke test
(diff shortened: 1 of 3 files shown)


xuzhao9:
Sorry I am trying to understand what does "cpuonly" mean in this PR? Does it mean we only upload cpu version of torchtext nightly to conda (and by default `conda install torchtext -c pytorch-nightly` won't have gpu support)?

seemethere:
> Sorry I am trying to understand what does "cpuonly" mean in this PR? Does it mean we only upload cpu version of torchtext nightly to conda (and by default `conda install torchtext -c pytorch-nightly` won't have gpu support)?

`cpuonly` is a metapackage that we add as a dependency for `cpu` versions of `pytorch`

Basically the way the conda dependency resolver works is that if we specify that we explicitly want `cpuonly` installed as a package then it should pick the `pytorch` package that has `cpuonly` as a dependency (which is our `cpu` specific package).

The same works for `cudatoolkit` versions as well, so for example if we have `conda install -y -c pytorch pytorch cudatoolkit=11.0` it should install the `pytorch` package that is compiled with `CUDA` 11.0 support.

seemethere:
Actually, @xuzhao9 made me realize that this might not be the best solution since it'd limit the `pytorch` version you can install with `torchtext` to one that depends on `cpuonly`, meaning it'd only install the `cpu` specific version, let me see if I can resolve this

seemethere:
Okay I made a change that puts in more in line with what `torchaudio` is doing which shouldn't limit what people can install with `torchtext`

xuzhao9 (approved):
Looks good!

PR #1119, opened 2021-01-13
Strip "+cpu" suffix in pip runtime dependency.

xuzhao9:
Currently, torchtext wheels depend on torch versions that have "+cpu" suffix, for example:
    
    $ pkginfo -f requires_dist torchtext-0.9.0.dev20210109-cp38-cp38-linux_x86_64.whl
       torch (==1.8.0.dev20210109+cpu)

In fact, we don't publish torch packages with "+cpu" suffix on PyPI. This PR is to remove this suffix as dependency.


seemethere:
I don't think this will actually have the desired effect that you're looking for since the line you edited is more of a build time dependency and not a runtime dependency.

The code you're looking to modify has more to do with the `PYTORCH_VERSION` environment variable and less to do with the `pip install` that you're currently modifying

xuzhao9:
Superseded by #1120

PR #1124, opened 2021-01-20
[WIP] Add sentencepiece to BERT model

zhangguanheng66:
In the [XLM-R](https://arxiv.org/pdf/1911.02116.pdf) model, SentencePiece is used to tokenize the strings. We enable the sentencepiece processing pipeline here for the BERT workflow.


cpuhrsch (changes requested):
Let's train and get new numbers before merging this change

PR #1133, opened 2021-01-28
DistCrossLingualMLMTask with two shards

pbelevich:
I started with looking into the CrossLingualMLMTask model parameters using torchsummary. According to my measurements the Embedding layers has 195841536 params, then we have 12 x 7087872 = 85054464 params of the Encoder layers and 192842128 params on the end layers. 196M + 85M + 193M ~= 474M. Since the largest layers are the first one and the last one, it was quite obvious to split the model right in the middle to form 2 shards with almost equal number of params. XLMRModel holds about 3/5 of params and prevents from dividing the model into two equal parts. Therefore I had to "inline" XLMRModel layers into CrossLingualMLMTask and split its encoder to two sequential encoders. Then I used the pipeline parallelism tutorial and adopted it to DistCrossLingualMLMTask.
The code seems working, but don't judge me for this very early version. I need to fix grad clip and scheduler. It easily allows to increase most of hyper-params. I was able to use batch size of 32, emsiz of 1024, nhid of 4096, nlayers of 24 and nhead of 16.

python cross_lingual_mlm_task.py --num_lines 3000 --batch_size=32 --split_size=32 --nlayers=24 --nhead=16 --emsize=1024 --nhid=4096

| epoch   1 |    10/10781 batches | lr 00nan | ms/batch 8582.67 | loss 13.76 | ppl 942230.75
| epoch   1 |    20/10781 batches | lr 00nan | ms/batch 8199.55 | loss 12.29 | ppl 218385.44
| epoch   1 |    30/10781 batches | lr 00nan | ms/batch 8059.80 | loss 12.17 | ppl 192853.54
| epoch   1 |    40/10781 batches | lr 00nan | ms/batch 7069.72 | loss 12.01 | ppl 164905.54
| epoch   1 |    50/10781 batches | lr 00nan | ms/batch 8507.27 | loss 12.00 | ppl 162889.12
| epoch   1 |    60/10781 batches | lr 00nan | ms/batch 8764.98 | loss 11.76 | ppl 127593.91
| epoch   1 |    70/10781 batches | lr 00nan | ms/batch 6516.39 | loss 11.69 | ppl 119626.47
| epoch   1 |    80/10781 batches | lr 00nan | ms/batch 8618.86 | loss 11.58 | ppl 106942.64
| epoch   1 |    90/10781 batches | lr 00nan | ms/batch 8512.14 | loss 11.47 | ppl 96242.86
| epoch   1 |   100/10781 batches | lr 00nan | ms/batch 8598.28 | loss 12.19 | ppl 197643.46

### DIFF
--- a/examples/BERT/cross_lingual_mlm_task.py
+++ b/examples/BERT/cross_lingual_mlm_task.py
@@ -1,15 +1,24 @@
 import argparse
-import time
 import math
+import os
+import time
+from typing import List
+
 import torch
+import torch.distributed.autograd as dist_autograd
+import torch.distributed.rpc as rpc
+import torch.multiprocessing as mp
 import torch.nn as nn
-from data import CC100
-from model import CrossLingualMLMTask
+import torch.optim as optim
+from torch.distributed.optim import DistributedOptimizer
+from torch.nn.utils.rnn import pad_sequence
 from torch.utils.data import DataLoader
+
+from data import CC100
+from dist_model import DistCrossLingualMLMTask
+from model import CrossLingualMLMTask
 from torchtext.experimental.transforms import sentencepiece_tokenizer
 from transforms import PretrainedSPVocab
-from torch.nn.utils.rnn import pad_sequence
-from typing import List
 
 
 def collate_batch(batch_data, args, mask_id, pad_id, text_transform):
@@ -47,8 +56,28 @@ def evaluate(data_source, model, mask_id, pad_id, ntokens, criterion, args, devi
     return total_loss / (len(data_source) - 1)  # Set batch # to 1 for inference
 
 
+def step(model, data, targets, criterion, optimizer, ntokens):
+    optimizer.zero_grad()
+    output = model(data)
+    loss = criterion(output.view(-1, ntokens), targets.view(-1))
+    loss.backward()
+    torch.nn.utils.clip_grad_norm_(model.parameters(), args.clip)
+    optimizer.step()
+    return loss
+
+
+def dist_step(model, data, targets, criterion, optimizer, ntokens):
+    with dist_autograd.context() as context_id:
+        output = model(data)
+        loss = criterion(output.view(-1, ntokens), targets.view(-1))
+        dist_autograd.backward(context_id, [loss])
+        # torch.nn.utils.clip_grad_norm_(model.parameters(), args.clip)
+        optimizer.step(context_id)
+        return loss
+
+
 def train(model, mask_id, pad_id, train_loss_log, train_data, text_transform,
-          optimizer, criterion, ntokens, epoch, scheduler, args, device, rank=None):
+          optimizer, criterion, ntokens, epoch, last_lr, args, device, step_impl):
     model.train()
     total_loss = 0.
     start_time = time.time()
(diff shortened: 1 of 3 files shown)


zhangguanheng66:
I just count the number of parameters (for both fairseq XLM-R and torchtext XLM-R):

Model Part | Num
-- | --
Embedding | ~256M
Encoder | ~302M
LM head | ~257M

So it definitely makes sense to split encoder to have two shard :).

```
# func to count the parameters in a model
import torch
def count_parameters(model):
    model_parameters = filter(lambda p: p.requires_grad, model.parameters())
    params = sum([torch.prod(torch.tensor(p.size())) for p in model_parameters])
    return params / 10**6
```

examples/BERT/model.py
```diff
@@ -171,21 +176,137 @@ class CrossLingualMLMTask(nn.Module):
...
+            CrossLingualMLMTaskShard2,
+            args=(get_cuda_if_available(1),) + args,
+            kwargs=kwargs
+        )
+
+    def forward(self, xs):
+        # Split the input batch xs into micro-batches, and collect async RPC
+        # futures into a list
+        out_futures = []
+        for x in iter(xs.split(self.split_size, dim=0)):
```
zhangguanheng66, line 294: Just curious. What's the benefit to split the input batch into micro-batches? Is this the pipeline parallel mechanism to enable different shards to work at the same time?
Without RPC, I run a batch size of 8 for ~250ms (depending on the sequence length of the input batch).

examples/BERT/model.py
```diff
@@ -171,21 +176,137 @@ class CrossLingualMLMTask(nn.Module):
...
+    def _forward(self, src):
+        output = self.transformer_encoder(src)
+        output = self.mlm_span(output)
+        output = self.activation(output)
+        output = self.norm_layer(output)
+        output = self.mlm_head(output)
+        return output
+
+
+class DistCrossLingualMLMTask(nn.Module):
```
zhangguanheng66, line 266: Seem to me that this could be a container which users can pass model shards, workers, arguments to. And the container will handle the forward func (a.k.a. splitting input batch, RPC reference, and concatenating the output.

```
class RPCShardContainer(nn.Module):
    def __init__(self, split_size, model_shard, workers, *args, **kwargs):
        ...
    def forward(self, data_batch):
        ...
model = RPCShardContainer(2, devices, [CrossLingualMLMTaskShard1, CrossLingualMLMTaskShard2], ...)
```

examples/BERT/model.py
```diff
@@ -171,21 +176,137 @@ class CrossLingualMLMTask(nn.Module):
...
 
+def get_cuda_if_available(i):
+    assert i >= 0
+    if torch.cuda.is_available():
+        return f"cuda:{min(i, torch.cuda.device_count() - 1)}"
+    else:
+        return "cpu"
+
+
+class CrossLingualMLMTaskBase(nn.Module):
```
zhangguanheng66, line 214: This module seems not task specific. Maybe we can put it in pytorch distributed later on.

examples/BERT/model.py
pbelevich: Honestly, I don't know, I borrowed this from [the original RPC pipeline example](https://github.com/pytorch/examples/blob/5751edd199f20c17a366c0eed4932539b3697c18/distributed/rpc/pipeline/main.py#L162) and tried to play with different values, but didn't see any performance impact

examples/BERT/cross_lingual_mlm_task.py
```diff
@@ -197,6 +258,15 @@ def _forward(x):
                         help='path to load the reference model for evaluation')
     parser.add_argument('--mask_frac', type=float, default=0.15,
                         help='the fraction of masked tokens')
+    parser.add_argument('--dist', type=bool, default=False,
+                        help='run distributed version')
+    parser.add_argument('--world_size', type=int, default=3,
```
zhangguanheng66, line 263: Just a question to help me understand it. Why do we set the world size to 3 if we shard the model on two device?

zhangguanheng66 (commented):
Just wondering if you can update the context of the PR, especially the value of ms/batch. I want to compare the time to run a single batch of 32. Currently, with a batch size of 8, the runtime is ~ 260ms per batch.

zhangguanheng66 (approved):
LGTM.

nit: you might update the README file with the command to run the case.

PR #1134, opened 2021-01-28
Count the total number of parameters in a model

zhangguanheng66:


torchtext/models/utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+    Examples:
+        >>> import torch
+        >>> import torchtext
+        >>> from torchtext.models.utils import count_model_param
+        >>> model = torch.nn.Embedding(100, 200)
+        >>> count_model_param(model, unit=10**3)
+        >>> 20.
+    """
+    model_parameters = filter(lambda p: p.requires_grad, nn_model.parameters())
```
cpuhrsch, line 20: Double check this definition of parameter. Does it need to require a gradient to be counted?

zhangguanheng66: I double check this by counting fairseq's xlmr.large model and get the total number of parameters - 560.1M. This is consistent with fairseq's value ([link](https://github.com/pytorch/fairseq/tree/master/examples/xlmr#pre-trained-models)).

zhangguanheng66: Check the discussion in the [post](https://discuss.pytorch.org/t/how-do-i-check-the-number-of-parameters-of-a-model/4325/9) and the solution is to count the "trainable" parameters which require grad.

torchtext/models/utils.py
```diff
@@ -0,0 +1,22 @@
+import torch
+
+
+def count_model_param(nn_model, unit=10**6):
+    r"""
+    Count the parameters in model
```
cpuhrsch, line 6: nit: in *a* model

cpuhrsch (approved):
Looks good, just move to experimental please

PR #1136, opened 2021-01-31
[WIP][RFC] Prototype pretrained models in torchtext

zhangguanheng66:


torchtext/experimental/models/roberta_model.py
```diff
@@ -0,0 +1,74 @@
+import os
+import torch
+import torch.nn as nn
+from torchtext.experimental.modules import TransformerEncoder
+from .roberta_transform import RobertaTransform
+from torchtext.utils import download_from_url, extract_archive, files_exist, load_args_from_json
+
+
+class RobertaModel(nn.Module):
```
nairbv, line 9: nit: I don't see "Model" suffixes on torchvision models; I suspect we'll prefer `models.Roberta` instead of `models.RobertaModel`?
Do we care about matching the capitalization/spelling from the paper (I see it as `RoBERTa` here: https://arxiv.org/abs/1907.11692)

zhangguanheng66: This idea is still prototype. My plan is to have XXModel as a wrapper of pretrained models, preprocessing transforms and config. So RoBERTa (without the "model" suffix) will stay in the module folder. But this is still under discussions with the broader audience.

torchtext/experimental/models/roberta_model.py
```diff
@@ -0,0 +1,74 @@
...
+        filepath = os.path.join(directory, checkpoint_file)
+        encoder.load_state_dict(torch.load(filepath))
+        return cls(args, encoder)
+
+    def forward(self, src):
+        src = self.bert_embed(src)
+        output = self.transformer_encoder(src)
+        return output
+
+    def encode(self, input_src: str) -> torch.Tensor:
```
hudeven, line 43: @zhangguanheng66 how about adding a corresponding "def decode(logits: Tensor) -> output", which translate model logits to meaningful output(e.g. [0, 1] to ["False", "True"])

torchtext/experimental/models/roberta_model.py
```diff
@@ -0,0 +1,74 @@
...
+        return cls(args, encoder)
+
+    def load_transform(self, directory='./', tokenizer_file="sentencepiece.bpe.model", vocab_file='vocab.txt'):
+        self.transform = RobertaTransform.from_pretrained(directory=directory,
+                                                          tokenizer_file=tokenizer_file,
+                                                          vocab_file=vocab_file)
+
+    # [TODO] use torch.hub.load_state_dict_from_url
+    @classmethod
+    def from_pretrained(cls, args, directory='./', checkpoint_file="model.pt"):
```
hudeven, line 31: How about adding a "PreTrainedModel" as super class or mixin? Only a "PreTrainedModel" is required to implement this method, the others(e.g. CNN) is not.

torchtext/experimental/models/roberta_model.py
hudeven: @zhangguanheng66 could you add more variations to demo how to reuse the building blocks and the inheritance between them? e.g. RobertaModel, RobertaModelForMaskedLM,  RobertaModelForClassification, XlmrModel*, etc. Do we need define to Head class for each variation?

torchtext/experimental/models/roberta_model.py
zhangguanheng66: The `encode` func here is more like a data processing pipeline. In fairseq, a `decode` func will return the raw string of the "encoded"

torchtext/experimental/models/roberta_model.py
zhangguanheng66: Yup. This func is general enough and we will move it to the base class later.

torchtext/experimental/models/roberta_model.py
```diff
@@ -0,0 +1,74 @@
...
+import torch.nn as nn
+from torchtext.experimental.modules import TransformerEncoder
+from .roberta_transform import RobertaTransform
+from torchtext.utils import download_from_url, extract_archive, files_exist, load_args_from_json
+
+
+class RobertaModel(nn.Module):
+    """Contain a Roberta encoder model."""
+
+    def __init__(self, args, encoder):
```
fmassa, line 12: In general I think it gives a bad user experience to construct models via opaque `args` or `config`.
It makes it much harder to just jump in an interpreter and explore / experiment with the model. It is much simpler if the arguments are just passed in right away.

If we want to have a method for constructing the model via configs, we could have a builder function like `build_model` which takes the config and unpacks it, while still leaving the `__init__` take the unpacked arguments

torchtext/experimental/models/roberta_model.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def load_transform(self, directory='./', tokenizer_file="sentencepiece.bpe.model", vocab_file='vocab.txt'):
+        self.transform = RobertaTransform.from_pretrained(directory=directory,
+                                                          tokenizer_file=tokenizer_file,
+                                                          vocab_file=vocab_file)
+
+    # [TODO] use torch.hub.load_state_dict_from_url
+    @classmethod
+    def from_pretrained(cls, args, directory='./', checkpoint_file="model.pt"):
+        encoder = TransformerEncoder(args.ntoken, args.embed_dim, args.nhead,
```
fmassa, line 32: It's not clear to me what is the benefit of having a `from_pretrained` method.
It's hard to enforce that the `args` correspond to the `checkpoint_file`, and it can give the user the expectation that it can magically work, but the user will just hit opaque error messages (like "sizes don't match for mha1.weight").

If we were to make the `from_pretrained` more robust, one would need to save the `args` inside the `checkpoint_file`, and use it instead to build the model

torchtext/experimental/models/roberta_model.py
```diff
@@ -0,0 +1,74 @@
...
+                                     args.feedforward_dim, args.nlayers, args.dropout)
+        filepath = os.path.join(directory, checkpoint_file)
+        encoder.load_state_dict(torch.load(filepath))
+        return cls(args, encoder)
+
+    def forward(self, src):
+        src = self.bert_embed(src)
```
fmassa, line 39: nit: this hasn't been defined in the code

torchtext/experimental/models/roberta_model.py
```diff
@@ -0,0 +1,74 @@
...
+        output = self.transformer_encoder(src)
```
fmassa, line 40: Is RoBERTa equal to transform + encoder only?


nit: `transformer_encoder` has been defined as `encoder` in the constructor

torchtext/experimental/models/roberta_model.py
```diff
@@ -0,0 +1,74 @@
...
+    @classmethod
+    def build_model(self, args):
+        encoder = TransformerEncoder(args.ntoken, args.embed_dim, args.nhead,
+                                     args.feedforward_dim, args.nlayers, args.dropout)
+        return cls(args, encoder)
+
+    def load_transform(self, directory='./', tokenizer_file="sentencepiece.bpe.model", vocab_file='vocab.txt'):
+        self.transform = RobertaTransform.from_pretrained(directory=directory,
+                                                          tokenizer_file=tokenizer_file,
+                                                          vocab_file=vocab_file)
```
fmassa, line 27: Why have a method for this, instead of just passing the `transform` as an argument to the constructor?

torchtext/experimental/models/roberta_model.py
```diff
@@ -0,0 +1,74 @@
...
+        output = self.transformer_encoder(src)
+        return output
+
+    def encode(self, input_src: str) -> torch.Tensor:
+        return torch.tensor(self.transform(input_src), dtype=torch.long)
+
+
+# [TODO] Add torch.hub support
+# [TODO] Download file from manifold
+def xlmr_base_model(directory='./', checkpoint_file='model.pt', args_file='args.json',
```
fmassa, line 49: Why do we need a `directory` file, instead of just specifying full paths for the `checkpoint` / `args` / `tokenizer` / `vocab`? What if they are not all in the same directory?

torchtext/experimental/modules/transformer.py
```diff
@@ -0,0 +1,91 @@
...
+        self.norm2 = LayerNorm(embed_dim)
+        self.dropout1 = Dropout(dropout)
+        self.dropout2 = Dropout(dropout)
+
+        if activation == "relu":
+            self.activation = F.relu
+        elif activation == "gelu":
+            self.activation = F.gelu
+        else:
+            raise RuntimeError("only relu/gelu are supported, not {}".format(activation))
```
fmassa, line 32: Why pass a string here instead of just accepting a `nn.Module` for the activation?

torchtext/experimental/modules/transformer.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def init_weights(self):
+        self.mha.in_proj_container.query_proj.init_weights()
+        self.mha.in_proj_container.key_proj.init_weights()
+        self.mha.in_proj_container.value_proj.init_weights()
+        self.mha.out_proj.init_weights()
```
fmassa, line 38: Can't you create a `init_weights` for the `MultiheadAttentionContainer` that dispatches to the calling `init_weights` for all it's submodules?

torchtext/experimental/modules/transformer.py
```diff
@@ -0,0 +1,91 @@
...
+
+class TransformerEncoder(nn.Module):
+    """Contain a transformer encoder."""
+
+    def __init__(self, ntoken, embed_dim=768, nhead=12, feedforward_dim=3072, nlayers=12, dropout=0.2):
+        super(TransformerEncoder, self).__init__()
+        self.model_type = 'Transformer'
+        self.transformer_encoder_embedding = self.build_transformer_encoder_embedding(ntoken, embed_dim)
+        encoder_layer = self.build_transformer_encoder_layer(embed_dim, nhead, feedforward_dim, dropout)
+        self.encoder_layers = ModuleList([copy.deepcopy(encoder_layer) for i in range(nlayers)])
```
fmassa, line 64: Do we want to enforce that all `encoder_layers` get initialized with the same set of weights?
Otherwise we can just build `nlayers` encoder layers independently and pass them to `ModuleList`

torchtext/experimental/modules/transformer.py
```diff
@@ -0,0 +1,91 @@
...
+        src = self.norm2(src)
+        return src
+
+
+class TransformerEncoder(nn.Module):
+    """Contain a transformer encoder."""
+
+    def __init__(self, ntoken, embed_dim=768, nhead=12, feedforward_dim=3072, nlayers=12, dropout=0.2):
+        super(TransformerEncoder, self).__init__()
+        self.model_type = 'Transformer'
```
fmassa, line 61: Do we plan to use this for something? If we want to make checks in the future about model types being `Transformer` / `CNN`, having an attribute to the model could be a way, but there might be better ways of achieving this (inheritance for example)

torchtext/experimental/modules/transformer.py
```diff
@@ -0,0 +1,91 @@
...
+        self.transformer_encoder_embedding = self.build_transformer_encoder_embedding(ntoken, embed_dim)
+        encoder_layer = self.build_transformer_encoder_layer(embed_dim, nhead, feedforward_dim, dropout)
+        self.encoder_layers = ModuleList([copy.deepcopy(encoder_layer) for i in range(nlayers)])
+        self.embed_dim = embed_dim
+
+    def forward(self, src):
+        src = self.transformer_encoder_embedding(src)
+        output = self.encoder_layers(src)
```
fmassa, line 69: `encoder_layers` is a `ModuleList`, which doesn't implement a `forward`, so you'll need to implement it yourself, or make it a `nn.Sequential`

torchtext/experimental/modules/transformer.py
```diff
@@ -0,0 +1,91 @@
...
+        src2 = self.linear2(self.dropout(self.activation(self.linear1(src))))
+        src = src + self.dropout2(src2)
+        src = self.norm2(src)
+        return src
+
+
+class TransformerEncoder(nn.Module):
+    """Contain a transformer encoder."""
+
+    def __init__(self, ntoken, embed_dim=768, nhead=12, feedforward_dim=3072, nlayers=12, dropout=0.2):
```
fmassa, line 59: There will need to be a trade-off here between ease-of-use and generality.

I see that you implement a `build_transformer_encoder_embedding` and `build_transformer_encoder_layer` for constructing the embedding and the encoder. This is done so that the user can override those functions and have their own variants.

But the problem that we have with the current approach is that we impose two contracts on the module API: their forward should respect the current API (which is expected), but also its *constructor* can only be defined by the arguments that you specified.

This means that the range of possible extensions that this class can have is limited, and won't support things like stochastic depth, linformer, etc.

One solution for this would be to pass in the constructor a `encoder_layer`, so that `TransformerEncoder` would look like
```python
def __init__(self, ntoken, encoder_layer, nlayers):
    ...
```
but other solutions are also possible

torchtext/experimental/models/roberta_transform.py
```diff
@@ -0,0 +1,28 @@
...
+class RobertaTransform(nn.Module):
+    """Roberta encode transform."""
+
+    def __init__(self, tokenizer, vocab):
+        super(RobertaTransform, self).__init__()
+        self.tokenizer = tokenizer
+        self.vocab = vocab
+
+    @classmethod
+    def from_pretrained(cls, directory='./',
```
fmassa, line 17: same comment here about removing `directory` and having full paths for `tokenizer` and `vocab`

torchtext/experimental/models/roberta_transform.py
```diff
@@ -0,0 +1,28 @@
...
+                        tokenizer_file="sentencepiece.bpe.model",
+                        vocab_file='vocab.txt'):
```
fmassa, line 19: Do you think all `*Transform` will have the same API for `from_pretrained`, and will consist of a `tokenizer_file` and a `vocab_file`, or can other arguments be present for different transforms?

torchtext/experimental/models/roberta_model.py
fmassa: I would vote for removing the `Model` suffix.

I wouldn't have classes for wrapping pre-trained models, but instead builder functions.

So instead of having a `RoBERTa` (generic class` and a `RoBERTaModel` (which loads pre-trained weights and sets up the right hyperparameters), I would instead have a `RoBERTa` for the generic class and a function `roberta_typeXx_ondataset_y` for building the model

nairbv: > I would instead have a `RoBERTa` for the generic class and a function `roberta_typeXx_ondataset_y` for building the model

Would the generic class here possibly just be `BERT`?  It seems BERT is the architecture of the model and RoBERTa is a method for training it?

torchtext/experimental/modules/transformer.py
zhangguanheng66: For now, I work on the interface to load pretrained model so I didn't spend time to implement the detail model.

torchtext/experimental/models/roberta_model.py
fmassa: > Would the generic class here possibly just be BERT? It seems BERT is the architecture of the model and RoBERTa is a method for training it?

Possibly, I don't know the specific details of RoBERTa to be 100% sure, but that's something worth considering if possible

zhangguanheng66: I put the generic `TransformerEncoder` class (similar to torchvision's [ResNet](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L144)) in the `modules` folder, which is composed of embedding layer plus the transformer encoder. Here, I can have a class `RobertaModel` to: 1) create an instance of the `TransformerEncoder` class 2) load pretrained weights 3) load the preprocessing transforms. Of course, this class can be replaced with a factory func, like [resnet18](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L268).

torchtext/experimental/modules/transformer.py
zhangguanheng66: It will be similar to [nn.TransformerEncoder](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/transformer.py#L161).

fmassa: Yes, looks like what I discussed would make it look similar to what we currently have in `nn.TransformerEncoder`

torchtext/experimental/modules/transformer.py
```diff
@@ -0,0 +1,93 @@
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.nn import Linear, Dropout, LayerNorm, ModuleList
+from torchtext.nn import MultiheadAttentionContainer, InProjContainer, ScaledDotProduct
+import copy
+from .embedding import PositionalEmbedding
+
+
+class TransformerEncoderLayer(nn.Module):
```
cpuhrsch, line 9: Is there a reason we can't use [torch.nn.TransformerEncoderLayer](https://pytorch.org/docs/master/generated/torch.nn.TransformerEncoderLayer.html#torch.nn.TransformerEncoderLayer) here?

zhangguanheng66: We could use it for sure.

torchtext/experimental/models/xlmr_model.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def _load_sentence_classifier(checkpoint_file='model.pt', args_file='args.json'):
+    args = load_args_from_json(args_file)
+    classifier = SentenceClassificationHead(args.num_labels, args.embed_dim, args.dropout)
+    classifier.load_state_dict(torch.load(checkpoint_file))
+    return classifier
+
+
+class TransformerEncoderSentenceClassificationTask(nn.Module):
```
hudeven, line 92: Task might cause confusion with STL Task. Call it something like TransformerForSentenceClassification?

zhangguanheng66: I will call it `TransformerSentenceClassification`.

torchtext/experimental/models/xlmr_model.py
```diff
@@ -0,0 +1,126 @@
...
+from .xlmr_transform import XLMRTransform
+from torchtext.utils import download_from_url, extract_archive, load_args_from_json
+from torchtext.experimental.transforms import sentencepiece_tokenizer
+from torchtext.experimental.vocab import load_vocab_from_file
+
+
+# [TODO] Add torch.hub support
+# [TODO] Download file from manifold
+# [TODO] check base model config
+def xlmr_base(root='./.model'):
```
fmassa, line 13: I've commented on this already, why pass a `root` directory, and not instead rely on `torch.hub` caching strategy (which can be overriden by the user)?

torchtext/experimental/models/xlmr_model.py
```diff
@@ -0,0 +1,126 @@
...
+PRETRAINED = {'xlmr.regular': 'https://pytorch.s3.amazonaws.com/models/text/pretrained_models/xlmr.regular.tar.gz',
+              'xlmr.base': 'https://pytorch.s3.amazonaws.com/models/text/pretrained_models/xlmr.base.tar.gz'}
+MD5 = {'xlmr.regular': 'adf75f3d20c8a876533b206ccb3a7cb6',
+       'xlmr.base': 'abc0a28171f883c6e1b2e8cf184c1ce8'}
+
+##################################################################################
+# This part will be moved to stl-text/models folder
+
+
+class SentenceClassificationHead(nn.Module):
```
fmassa, line 66: If we are going to be naming things heads, we should have a contract on what a "head" is.

For vision for example I find the definition of head to be fairly fuzzy (is it the last linear layer? Or can it be something else?).

Also, the implementation here seems like one instantiation of a possible Head for sentence classification, so might be better to name it in a more explicit way (like two mlp head or smth like that)

torchtext/experimental/models/xlmr_transform.py
```diff
@@ -0,0 +1,28 @@
...
+class XLMRTransform(nn.Module):
+    """XLM-R encode transform."""
+
+    def __init__(self, tokenizer, vocab):
+        super(XLMRTransform, self).__init__()
+        self.tokenizer = tokenizer
+        self.vocab = vocab
+
+    @classmethod
+    def from_pretrained(cls, directory='./',
```
fmassa, line 17: Could you please comment on why we might want to keep a `directory` here?

torchtext/experimental/models/xlmr_transform.py
```diff
@@ -0,0 +1,28 @@
...
+                        tokenizer_file="sentencepiece.bpe.model",
+                        vocab_file='vocab.txt'):
+        filepath = os.path.join(directory, tokenizer_file)
+        tokenizer = sentencepiece_tokenizer(filepath)
+        filepath = os.path.join(directory, vocab_file)
+        with open(filepath, 'r') as f:
+            vocab = load_vocab_from_file(f)
```
fmassa, line 24: Will all `pretrained` XLMRTransform` instances that we provide contain a `sentencepiece_tokenizer` and use a vocab loaded via `load_vocab_from_file`? If not, I see two options:
- try to make this code generic on the tokenizer type (which can be annoying to handle and brittle)
- remove this method altogether from `XLMRTransform` and instead let if be a free-function elsewhere.

torchtext/experimental/models/xlmr_model.py
```diff
@@ -0,0 +1,126 @@
...
+    Examples:
+        >>> from torchtext.experimental.models import xlmr_base_sentence_classifier
+        >>> xlmr_sentence_classifier_model, xlmr_base_transform = xlmr_base_sentence_classifier()
+        >>> xlmr_base_transform('this is an example')
+        >>> tensor([  903,    83,   142, 27781])
+    '''
+    # Load pretrained XLM-R
+    tar_file = download_from_url(PRETRAINED['xlmr.base'], root=root,
+                                 hash_value=MD5['xlmr.base'], hash_type='md5')
+    checkpoint_file, tokenizer_file, vocab_file, args_file = extract_archive(tar_file, overwrite=True)
```
fmassa, line 113: We should really use torchhub here for handling this for us. This keeps the logic centralized in a single location. Note that we might not need to save the fizes as zip or tar, but just a pytorch `.pth` file.

torchtext/experimental/models/xlmr_model.py
```diff
@@ -0,0 +1,126 @@
...
+    xlmr_model = _load_xlmr_model(checkpoint_file=checkpoint_file, args_file=args_file)
+    xlmr_transform = _load_xlmr_transform(tokenizer_file=tokenizer_file, vocab_file=vocab_file)
+
+    # Load classifier head
+    tar_file = download_from_url(TASK_PRETRAINED['xlmr.base.sentence.classifier'], root=root,
+                                 hash_value=TASK_MD5['xlmr.base.sentence.classifier'], hash_type='md5')
+    checkpoint_file, args_file = extract_archive(tar_file, overwrite=True)
+    sentence_classifier = _load_sentence_classifier(checkpoint_file=checkpoint_file, args_file=args_file)
```
fmassa, line 121: FYI: a head is tightly coupled with the "base" that comes before it. Mixing a head with the wrong base will lead to completely invalid results.

I suppose the use-case here is for finetuning on different tasks, while keeping the "base" frozen, is that the case?

torchtext/experimental/models/xlmr_model.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+class TransformerEncoderSentenceClassification(nn.Module):
+    def __init__(self, transformer_encoder, classifier_head):
+        super(TransformerEncoderSentenceClassification, self).__init__()
+        self.transformer_encoder = transformer_encoder
+        self.classifier_head = classifier_head
+
+    def forward(src):
+        raise NotImplementedError("forward func has not been implemented yet.")
```
fmassa, line 99: I think it would be a good exercise to show the skeleton (like you have in here) of the abstractions that you are thinking about providing for the other tasks as well. It's not clear to me yet why this `TransformerEncoderSentenceClassification` implementation would be any different for a different task.

torchtext/experimental/models/xlmr_model.py
zhangguanheng66: Yes. This is the use case. You have a pretrained model as the base and fine-tune the head for different task.

torchtext/experimental/models/xlmr_model.py
zhangguanheng66: Will internal users be able to use torchhub? I think we still need some things to handle the files from manifold.

torchtext/experimental/models/xlmr_transform.py
zhangguanheng66: I didn't fix this one here. Good catch.

torchtext/experimental/models/xlmr_transform.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, tokenizer, vocab):
+        super(XLMRTransform, self).__init__()
+        self.tokenizer = tokenizer
+        self.vocab = vocab
+
+    def forward(self, input_src: str) -> List[int]:
+        return self.vocab(self.tokenizer(input_src))
+
+
+def load_xlmr_transform():
```
zhangguanheng66, line 18: @datumbox Here is a pretrained transform case for the text domain.

datumbox: Sounds good. This should be supported fine for presets.

The only question that I have is over the use of `load_model_from_url`. Do you typically store the whole model or only its weights? As far as I've seen the latter it typically preferred but maybe that does not work for you if the definition of the model is on an external lib.

zhangguanheng66: Yes. For the model itself, we store weights. However, for those two transforms, they are not really `nn.Module` so we have to store the whole models. Happy to hear a better solution.

PR #1138, opened 2021-02-02
Fix broken CI tests due to spacy 3.0 release

zhangguanheng66:
In spacy 3.0 release, the shortcut "en" and "de" have been depreciated. The new language packages in spacy 3.0 introduce BC breaking and this propagates to our datasets. 

However, since the Multi30k dataset is in the nightly release, we don't need to add [BC Breaking] tag.

### DIFF
--- a/torchtext/data/field.py
+++ b/torchtext/data/field.py
@@ -143,7 +143,7 @@ class Field(RawField):
     def __init__(self, sequential=True, use_vocab=True, init_token=None,
                  eos_token=None, fix_length=None, dtype=torch.long,
                  preprocessing=None, postprocessing=None, lower=False,
-                 tokenize=None, tokenizer_language='en_core_web_sm', include_lengths=False,
+                 tokenize=None, tokenizer_language='en', include_lengths=False,
                  batch_first=False, pad_token="<pad>", unk_token="<unk>",
                  pad_first=False, truncate_first=False, stop_words=None,
                  is_target=False):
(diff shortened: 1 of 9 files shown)


torchtext/data/utils.py
```diff
@@ -72,7 +72,7 @@ def _basic_english_normalize(line):
     return line.split()
 
 
-def get_tokenizer(tokenizer, language='en'):
+def get_tokenizer(tokenizer, language='en_core_web_sm'):
```
cpuhrsch, line 75: Why is changing the default argument and removing the check ```if language != 'en':``` not BC-breaking?

torchtext/data/utils.py
```diff
@@ -101,8 +101,6 @@ def get_tokenizer(tokenizer, language='en'):
         return _split_tokenizer
 
     if tokenizer == "basic_english":
-        if language != 'en':
```
cpuhrsch, line 104: This change should be isolated to the spacy tokenizer, so we should change the meaning of the ```language``` argument to the spacy part of this function and not change the default value and its meaning for all types of tokenizers.

For BC compatibility we could add a check to the ```spacy``` part of below tokenizer that appends ```_core_web_sm``` is the language is ```en``` and throws a warning that references this PR indicating to users that now need to use this name for the languages.

torchtext/data/utils.py
```diff
@@ -110,6 +110,8 @@ def get_tokenizer(tokenizer, language='en'):
         return tokenizer
 
     if tokenizer == "spacy":
+        if language == 'en':
+            raise RuntimeError("The en package has been deprecated in Spacy 3.0 release. Please switch to en_core_web_sm.")
```
cpuhrsch, line 114: I'd make explicit reference to the "language" keyword argument used above. A user might not know that this translate into a spacy package and what to change about their code to fix this issue.

torchtext/data/field.py
```diff
@@ -143,7 +143,7 @@ class Field(RawField):
     def __init__(self, sequential=True, use_vocab=True, init_token=None,
                  eos_token=None, fix_length=None, dtype=torch.long,
                  preprocessing=None, postprocessing=None, lower=False,
-                 tokenize=None, tokenizer_language='en', include_lengths=False,
```
cpuhrsch, line 146: I'd definitely consider this a BC-breaking change, especially since 'en' is still the correct choice for tokenizers other than spacy.

torchtext/data/utils.py
cpuhrsch: Instead of throwing an error we can detect this and add a warning that we're using "en_core_web_sm" instead. The same applies to "de". That way we also maintain BC compatibility and users don't need to update their API.

cpuhrsch (approved):

PR #1140, opened 2021-02-04
Replace model with full name when spacy load is used

malfet:
Model name shortcuts no longer work in spacCy 3+ or later
Found while looking into the tutorial build failure in https://app.circleci.com/pipelines/github/pytorch/tutorials/3214/workflows/d79dc18a-734e-4a87-ac73-3a687ae99d56/jobs/58297

### DIFF
--- a/torchtext/data/utils.py
+++ b/torchtext/data/utils.py
@@ -132,6 +132,8 @@ def get_tokenizer(tokenizer, language='en'):
                 }
                 if language not in OLD_MODEL_SHORTCUTS:
                     raise
+                import warnings
+                warnings.warn(f'Spacy model "{language}" could not be loaded, trying "{OLD_MODEL_SHORTCUTS[language]}" instead')
                 spacy = spacy.load(OLD_MODEL_SHORTCUTS[language])
             return partial(_spacy_tokenize, spacy=spacy)
         except ImportError:
--- a/torchtext/data/utils.py
+++ b/torchtext/data/utils.py
@@ -117,19 +117,7 @@ def get_tokenizer(tokenizer, language='en'):
             except IOError:
                 # Model shortcuts no longer work in spaCy 3.0+, try using fullnames
                 # List is from https://github.com/explosion/spaCy/blob/b903de3fcb56df2f7247e5b6cfa6b66f4ff02b62/spacy/errors.py#L789
-                OLD_MODEL_SHORTCUTS = {
-                    'en': 'en_core_web_sm',
-                    'de': 'de_core_news_sm',
-                    'es': 'es_core_news_sm',
-                    'pt': 'pt_core_news_sm',
-                    'fr': 'fr_core_news_sm',
-                    'it': 'it_core_news_sm',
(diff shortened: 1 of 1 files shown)


torchtext/data/utils.py
```diff
@@ -112,7 +112,27 @@ def get_tokenizer(tokenizer, language='en'):
...
+                    'pt': 'pt_core_news_sm',
+                    'fr': 'fr_core_news_sm',
+                    'it': 'it_core_news_sm',
+                    'nl': 'nl_core_news_sm',
+                    'el': 'el_core_news_sm',
+                    'nb': 'nb_core_news_sm',
+                    'lt': 'lt_core_news_sm',
+                    'xx': 'xx_ent_wiki_sm'
+                }
+                if language not in OLD_MODEL_SHORTCUTS:
```
cpuhrsch, line 133: Could you also throw a warning that this is happening so that users can migrate?

cpuhrsch (approved):

torchtext/data/utils.py
```diff
@@ -112,7 +112,27 @@ def get_tokenizer(tokenizer, language='en'):
...
+                    'es': 'es_core_news_sm',
+                    'pt': 'pt_core_news_sm',
+                    'fr': 'fr_core_news_sm',
+                    'it': 'it_core_news_sm',
+                    'nl': 'nl_core_news_sm',
+                    'el': 'el_core_news_sm',
+                    'nb': 'nb_core_news_sm',
+                    'lt': 'lt_core_news_sm',
+                    'xx': 'xx_ent_wiki_sm'
+                }
```
malfet, line 132: I wonder whether this would be better (or worse)
```suggestion
                OLD_MODEL_SHORTCUTS = spacy.errors.OLD_MODEL_SHORTCUTS if hasattr(spacy.errors, 'OLD_MODEL_SHORTCUTS') else {}
```

zhangguanheng66:
We still need to update circieci to set up the correct spacy package. https://github.com/pytorch/text/pull/1138/files

torchtext/data/utils.py
cpuhrsch: Just for reference, the code for that is [here](https://github.com/explosion/spaCy/blob/a59f3fcf5dab3acf5570483cc314b47cc5833f39/spacy/errors.py#L789). I actually prefer this.

cpuhrsch:
With this change we don't *have* to update the spacy link in our [environment.yml](https://github.com/pytorch/text/blob/master/.circleci/unittest/linux/scripts/environment.yml#L20-L21), but it's probably better if we do since we're adding code to handle this logic.

zhangguanheng66:
Should we merge this one first and use #1138 to update the tests and environment?

cpuhrsch:
The tests did not run green before being merged, see the relevant failure [here](https://app.circleci.com/pipelines/github/pytorch/text/2232/workflows/c78a537b-8a52-4306-9a90-8db11def20a4/jobs/63664).

PR #1143, opened 2021-02-04
switch data_select in dataset signature to split

zhangguanheng66:


torchtext/experimental/datasets/raw/language_modeling.py
```diff
@@ -55,18 +55,18 @@ def _setup_datasets(dataset_name, root, data_select, year, language):
...
+def WikiText2(root='.data', split=('train', 'valid', 'test')):
     """ Defines WikiText2 datasets.
 
     Create language modeling dataset: WikiText2
     Separately returns the train/test/valid set
 
     Args:
         root: Directory where the datasets are saved. Default: ".data"
-        data_select: a string or tupel for the returned datasets. Default: ('train', 'valid, 'test')
+        split: a string or tupel for the returned datasets. Default: ('train', 'valid, 'test')
```
cpuhrsch, line 69: Just a drive-by comment as a follow-up: Should be "tuple" not "tupel"

cpuhrsch (approved):

PR #1144, opened 2021-02-05
Revert "Replace model with full name when spacy load is used"

cpuhrsch:
Reverts pytorch/text#1140 which was merged with red CI.


zhangguanheng66:
We should merge this one to have the CI test great again. https://github.com/pytorch/text/pull/1138

PR #1145, opened 2021-02-06
Add offset arg in the raw text dataset

zhangguanheng66:
Add `offset`arg to the raw text datasets.
Remove `setup_iter` func from `RawTextIterableDataset` class

### DIFF
--- a/torchtext/experimental/datasets/raw/common.py
+++ b/torchtext/experimental/datasets/raw/common.py
@@ -14,7 +14,7 @@ class RawTextIterableDataset(torch.utils.data.IterableDataset):
     """Defines an abstraction for raw text iterable datasets.
     """
 
-    def __init__(self, name, full_num_lines, iterator):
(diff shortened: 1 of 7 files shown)


torchtext/experimental/datasets/raw/common.py
```diff
@@ -14,16 +14,18 @@ class RawTextIterableDataset(torch.utils.data.IterableDataset):
...
         self.full_num_lines = full_num_lines
         self._iterator = iterator
         self.has_setup = False
-        self.start = 0
+        self.start = offset
         self.num_lines = None
+        self.setup_iter(start=offset, num_lines=full_num_lines - offset)
+        self.stride = stride
 
     def setup_iter(self, start=0, num_lines=None):
```
cpuhrsch, line 30: I think it makes sense to inline this code and remove the setup_iter method until we need it again.

cpuhrsch (changes requested):
Let's handle stride in a separate PR

cpuhrsch (approved):

PR #1147, opened 2021-02-10
Add support for string-typed split values to raw datasets.

cpuhrsch:
Add support for the following behavior

```
train = Multi30k(split='train')
```

currently the user has to do

```
train, = Multi30k(split=('train',))
```

Changes include
- [x] Generic argument checking and return value constructor
- [x] Raw dataset tests
- [x] Regular dataset tests
- [x] Updated documentation
- [x] torchtext.experimental.datasets.raw.URLS dictionary that maps a dataset to the corresponding URL(s)

Follow-up:
- [ ] Use parameterized (as in [torchaudio](https://github.com/pytorch/audio/blob/7ee1c46b7831cdd5a66a46f23e4f32ffda1d0b52/test/torchaudio_unittest/functional/batch_consistency_test.py#L56)) to create one test per dataset.
- [ ] Re-enable tests for datasets hosted on statmt
- [ ] - [ ] Re-enable tests for datasets hosted on google drive

PR #1149, opened 2021-02-10
Add and use parameterized for testing

cpuhrsch:
Introduces the parameterized module to increase resolution of tests and allow one test per dataset written generically.

Changes include
- [x] Parameterization of test_raw_datasets_split_argument
- [x] Parameterization of test_datasets_split_argument
- [x] Add module as dependency to CI system

Follow-up to https://github.com/pytorch/text/pull/1147

### DIFF
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -4,6 +4,8 @@ import os
 import torchtext.data as data
 import torch
 import torchtext
+import unittest
+from parameterized import parameterized
 from ..common.torchtext_test_case import TorchtextTestCase
 from ..common.assets import conditional_remove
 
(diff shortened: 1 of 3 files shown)


test/data/test_builtin_datasets.py
```diff
@@ -143,37 +144,35 @@ def test_num_lines_of_dataset(self):
...
+        train2, = dataset(split=('train',))
+        for d1, d2 in zip(train1, train2):
+            self.assertEqual(d1, d2)
+            # This test only aims to exercise the argument parsing and uses
+            # the first line as a litmus test for correctness.
+            break
+        # Exercise default constructor
+        _ = dataset()
+
+    @parameterized.expand(["AG_NEWS", "WikiText2", "IMDB"])
```
mthrok, line 164: Is this valid? I think you need to use tuple, like
```python
expand([
("AG_NEWS", ),
("WikiText2", ),
...
])
```

mthrok (approved):
~As long as the tests are passing,~ it looks good to me!

test/data/test_builtin_datasets.py
mthrok: obviously, tests are passing.

PR #1150, opened 2021-02-10
Detect when Google drive link exceeds quota and throw an error.

cpuhrsch:
Google drive URLs are flaky, because the process sending the request might run into quota limits. This is currently not obvious to the user. In particular they'll see the following confusing error.

```
$ python -c "import torchtext; torchtext.experimental.datasets.raw.AmazonReviewFull()"
Traceback (most recent call last):
[...]
KeyError: 'content-disposition'
```

This PR attempts to detect this situation and throw a more readable exception instead.

```
$ python -c "import torchtext; torchtext.experimental.datasets.raw.AmazonReviewFull()"
Traceback (most recent call last):
[...]
RuntimeError: Google drive link https://drive.google.com/uc?export=download&id=0Bz8a_Dbh9QhbZVhsUnRWRDhETzA is currently unavailable, because the quota was exceeded.
```

Unfortunately it doesn't actually solve the problem of relying on a flaky host.

### DIFF
--- a/torchtext/utils.py
+++ b/torchtext/utils.py
@@ -63,6 +63,8 @@ def download_from_url(url, path=None, root='.data', overwrite=False, hash_value=
         chunk_size = 16 * 1024
         total_size = int(r.headers.get('Content-length', 0))
         if filename is None:
+            if 'content-disposition' not in r.headers:
+                raise RuntimeError("Internal error: headers don't contain content-disposition.")
             d = r.headers['content-disposition']
             filename = re.findall("filename=\"(.+)\"", d)
             if filename is None:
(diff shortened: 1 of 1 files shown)


torchtext/utils.py
```diff
@@ -123,6 +125,14 @@ def _process_response(r, root, filename):
     for k, v in response.cookies.items():
         if k.startswith("download_warning"):
             confirm_token = v
+    if confirm_token is None:
+        if "Quota exceeded" in str(response.content):
+            raise RuntimeError(
+                "Google drive link {} is currently unavailable, because the quota was exceeded.".format(
+                    url
+                ))
```
mthrok, line 133: nit: use of f-string will make it more readable.

mthrok (approved):
Looks good!

torchtext/utils.py
```diff
@@ -63,6 +63,8 @@ def _process_response(r, root, filename):
         chunk_size = 16 * 1024
         total_size = int(r.headers.get('Content-length', 0))
         if filename is None:
+            if 'content-disposition' not in r.headers:
```
zhangguanheng66, line 66: This one is more common for some google drive link

zhangguanheng66 (approved):
I think you don't mean a drink from Google.

cpuhrsch:
Haha good catch! I'll address Moto's comment in a follow-up.

PR #1151, opened 2021-02-11
[BC-breaking] Standardize raw dataset doc strings and argument order.

cpuhrsch:
This PR comes with a few changes

#### 1. 
Addresses inconsistencies such as documentation indicating the incorrect order of splits (train, valid, test vs. train, test, valid), inconsistent indentation or whitespace and standardizes documentation of common arguments such as split and root across all raw datasets.

#### 2.
Signatures such as 
```
def Multi30k(train_filenames=("train.de", "train.en"),
             valid_filenames=("val.de", "val.en"),
             test_filenames=("test_2016_flickr.de", "test_2016_flickr.en"),
             split=('train', 'valid', 'test'), root='.data', offset=0):
```
and
```
AmazonReviewPolarity(root='.data', split=('train', 'test'), offset=0):
```
are aligned in the order of their arguments.

If a user expects to pass the root folder first and calls Multi30k positionally, i.e. ```Multi30k("my_data_path")```, because the behavior is expected based on the signature of AmazonReviewPolarity, they will inadvertently set the train_filenames argument to a single string. This PR standardizes on the majority by moving the root folder first, followed by split and offset and then by dataset dependent arguments such as language, year or train_filenames.

#### 3.
Commonalities between documentation are centralized and generated. This also includes a programatic signature check to make sure all datasets have a consistent interface. This is achieved via a decorator that sanitizes input arguments and preprends a common documentation header.

#### TODO
- [ ] Verify docstring generation works with website doc generation

#### Follow-up
- [ ] Decorator to append datapoint example including datapoint types
- [ ] Represent dataset selections for translation datasets more compactly


torchtext/experimental/datasets/raw/common.py
```diff
@@ -25,6 +27,87 @@ def wrap_datasets(datasets, split):
     return datasets
 
 
+def dataset_docstring_header(fn):
```
cpuhrsch, line 30: I can still add more documentation here

torchtext/experimental/datasets/raw/common.py
```diff
@@ -25,6 +27,87 @@ def wrap_datasets(datasets, split):
...
+
+    raise ValueError("default_split type expected to be of string or tuple but got {}".format(type(default_split)))
+
+
+def dataset_docstring_header_decorator(fn):
+    fn.__doc__ = dataset_docstring_header(fn)
+    return fn
+
+
+def input_sanitization_decorator(fn):
```
cpuhrsch, line 79: I can still add more documentation here

torchtext/experimental/datasets/raw/common.py
```diff
@@ -25,6 +27,87 @@ def wrap_datasets(datasets, split):
...
+            argspec.defaults[0] == ".data" and
+            argspec.defaults[2] == 0 and
+            argspec.varargs is None and
+            argspec.varkw is None and
+            len(argspec.kwonlyargs) == 0 and
+            argspec.kwonlydefaults is None and
+            len(argspec.annotations) == 0
+            ):
+        raise ValueError("Internal Error: Given function {} did not adhere to standard signature.".format(fn))
+
```
mthrok, line 93: The following part is not intuitively clear what it intends to do, so adding comments will be helpful.

torchtext/experimental/datasets/raw/common.py
mthrok: That will be nice. The second half of the implementation is not intuitive to the first time reader.
nit: Also I would use verb of adjective for decorator name. Even if it is noun, I will not call it as `@XXXdecorator` sounds redundant.

mthrok (approved):
Looks good. Approving to unblock. But docstring can be improved.

zhangguanheng66 (approved):

zhangguanheng66 (approved):
We can monitor the nightly release doc and check if this actually works. http://pytorch.org/text/master/

cpuhrsch:
The master documentation is currently not pushing html for the raw datasets. I'll send a PR to address this.

PR #1154, opened 2021-02-12
Make TorchText installable from direct reference

malfet:
In order to make `pip install https://github.com/pytorch/text` succeed
`setup.py` should have logic to initialize submodules

Add `check_submodules()` routine to setup.py to check if
'third_party/re2/CMakeLists.txt' is present on the filesystem and
attempt to initialize submodules if it is not

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,14 @@
 #!/usr/bin/env python
+from build_tools import setup_helpers
+import distutils.command.clean
 import io
 import os
+from pathlib import Path
+from setuptools import setup, find_packages
 import shutil
 import subprocess
-from pathlib import Path
-import distutils.command.clean
-from setuptools import setup, find_packages
-
-from build_tools import setup_helpers
+import sys
+import time
 
 ROOT_DIR = Path(__file__).parent.resolve()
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -44,6 +45,22 @@ def _export_version(version, sha):
         fileobj.write("git_version = {}\n".format(repr(sha)))
 
 
+def check_submodules():
+    third_party_dir = ROOT_DIR / 'third_party'
+    if not (third_party_dir / 're2' / 'CMakeFiles.txt').exists():
+        print(' --- third_party/re2/CMakeFiles.txt do not exist, trying to init submodules')
+        try:
```
cpuhrsch, line 52: [PyTorch fails and asks the user to do it manually](https://github.com/pytorch/pytorch/blob/master/setup.py#L296-L299). I worry that this might cause issues when someone wants to debug / modify the submodules and then subsequently runs setup.py. Are there similar plans for PyTorch as well?

malfet: @cpuhrsch yes, I plan to make similar change for pytorch as well, although Pytorch is less likely to be built directly from source (building TorchText using direct link takes less than 15 min, whereas for PyTorch, especially with GPU support will likely to take more than an hour)

mthrok: What is the context of this work? Are we going to do the same for torchaudio (which I support)?

mthrok (approved):

PR #1156, opened 2021-02-15
[BC-breaking] Split raw text classification datasets into individual files

cpuhrsch:
To further align the dataset organization of the raw datasets with the other domain libraries (torchaudio and torchvision) we split them into separate files, one each per dataset. This PR introduces this change for all text classification datasets. It is BC-breaking in the sense that the torchtext.experimental.datasets.raw.text_classification module is being removed.

Follow-up:
- Split raw sequence tagging datasets into individual files
- Split raw translation datasets into individual files
- Split raw language modeling datasets into individual files
- Split raw question answer datasets into individual files

TODO:
- [x] Add unit test, one per dataset, that compares first line output and verifies dataset constants.


torchtext/experimental/datasets/raw/amazonreviewpolarity.py
```diff
@@ -0,0 +1,39 @@
...
+    'test': 400000,
+}
+
+_PATH = 'amazon_review_polarity_csv.tar.gz'
+
+
+@wrap_split_argument
+@add_docstring_header()
+def AmazonReviewPolarity(root='.data', split=('train', 'test'), offset=0):
+    def _create_data_from_csv(data_path):
```
zhangguanheng66, line 24: I feel we don't need to repeat `_create_data_from_csv` function for all the datasets. We can have a utility func and keep it private somewhere.

zhangguanheng66 (commented):
I thought we don't use [BC-breaking] for the prototype features.
And users will experience BC breaking if they import the datasets from `torchtext.experimental.datasets.raw.text_classification`, right?

torchtext/experimental/datasets/raw/amazonreviewpolarity.py
cpuhrsch: Yes, I agree. I'll think of some generic form that we can share across all datasets, which is another step after the datasets have been split.

cpuhrsch:
Yes, the datasets in ```torchtext.experimental.datasets.raw.text_classification``` are being replaced by individual files.

Indeed BC-breaking doesn't quite make sense for a prototype, but I added it anyway.

PR #1161, opened 2021-02-16
Make version static text into a link to versions.html

mattip:
Builds on #1160 to turn the static version text into a link to the manually-maintained versions.html page to allow selecting a version of the documentation to view. Once this is merged, the next nightly document build and push will make this work with the `master` documentation


mattip:
@brianjo

zhangguanheng66:
A comment not relevant to this PR - I found the font of torchtext website is not consistent with audio/vision. We use bold everywhere http://pytorch.org/text/stable/index.html while audio/vision don't https://pytorch.org/audio/stable/index.html. I guess we don't use the sphinx correctly.

brianjo:
I’ll check it out when I get a chance. We install the theme through requirements.txt during the doc build so it should be the same. Thanks!


________________________________

PR #1163, opened 2021-02-16
Fix sphinx warnings and turn warnings into errors

NicolasHug:
Similar to what was done in `torchvision` https://github.com/pytorch/vision/pull/3290 and `torchaudio` https://github.com/pytorch/audio/pull/1247

This PR fixes sphinx warnings as well as some other doc layout issues, and sets the 

```py
napoleon_numpy_docstring = False
napoleon_google_docstring = True
```

for sphinx to issue warnings when the wrong parameter syntax is used.

The warnings are now turned into errors. Note: sphinx will now stop compilation at the first warning/error it sees; if we relied on sphinx>=8, we could use the `--keep-going` option to let it finish.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -W
+SPHINXOPTS    = -W  # turn warnings into errors
 SPHINXBUILD   = sphinx-build
 SPHINXPROJ    = torchtext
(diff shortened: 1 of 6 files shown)


zhangguanheng66 (approved):

cpuhrsch:
Haha, jinx https://github.com/pytorch/text/pull/1159 - notice that we can also build docs in parallel using the -j flag.

PR #1167, opened 2021-02-17
DOC Document undocumented parameters and add CI check

NicolasHug:
Similar to what was done in `torchvison` https://github.com/pytorch/vision/pull/3299/ and `torchaudio` https://github.com/pytorch/audio/pull/1248

This PR documents parameters that were so far undocumented and adds a CI check that runs `pydocstyle` so that docstring concistency is enforced going forward.


zhangguanheng66 (approved):
Thanks for working on this. The failed tests are because the google drive links exceed the quota. I will re-run the tests later on before merging this PR.

torchtext/datasets/iwslt2016.py
```diff
@@ -222,7 +222,7 @@ def _construct_filepaths(paths, src_filename, tgt_filename):
 
 
 @wrap_split_argument(('train', 'valid', 'test'))
-def IWSLT2016(root='.data', split=('train', 'valid', 'test'), offset=0, language_pair=('de', 'en'), valid_set='tst2013', test_set='tst2014'):
+def IWSLT2016(root='.data', split=('train', 'valid', 'test'), language_pair=('de', 'en'), valid_set='tst2013', test_set='tst2014'):
```
NicolasHug, line 225: @parmeet the `offset` parameter was introduced in https://github.com/pytorch/text/pull/1191 but was undocumented, which made the CI fail. I removed it as it's not used anywhere, LMK if I should put it back (and if so, what should be the doc)

thanks ;)

zhangguanheng66: This is correct. Thanks for catching it.

PR #1172, opened 2021-02-17
Merge fbsync into master branch - retire legacy code

zhangguanheng66:


test/legacy/data/test_batch.py
```diff
@@ -1,7 +1,7 @@
 import torch
-import torchtext.data as data
+import torchtext.legacy.data as data
 
-from ..common.torchtext_test_case import TorchtextTestCase
+from ...common.torchtext_test_case import TorchtextTestCase
```
cpuhrsch, line 4: Triple "." is probably not intended here

cpuhrsch (approved):

test/legacy/data/test_batch.py
zhangguanheng66: It needs to jump two folders to access `TorchtextTestCase` class

PR #1173, opened 2021-02-17
[BC-breaking] Split raw language modeling datasets into individual files

cpuhrsch:
Follow-up to https://github.com/pytorch/text/pull/1156/files#

To further align the dataset organization of the raw datasets with the other domain libraries (torchaudio and torchvision) we split them into separate files, one each per dataset. This PR introduces this change for all language modeling datasets. It is BC-breaking in the sense that the torchtext.experimental.datasets.raw.language_modeling module is being removed.


zhangguanheng66 (approved):
Do you want to update the hash reference? https://github.com/pytorch/text/pull/1169

cpuhrsch:
I'll include the change and close the PR once this merged.

PR #1175, opened 2021-02-18
[BC-breaking] Split raw translation datasets into individual files

cpuhrsch:
Follow-up to https://github.com/pytorch/text/pull/1156/files#

To further align the dataset organization of the raw datasets with the other domain libraries (torchaudio and torchvision) we split them into separate files, one each per dataset. This PR introduces this change for all translation datasets. It is BC-breaking in the sense that the torchtext.experimental.datasets.raw.translation module is being removed.


zhangguanheng66 (approved):

zhangguanheng66:
Need to fix the CI test.

PR #1177, opened 2021-02-18
SingleProcessPipeline and RPCPipeline with CPU RPC and CUDA RPC

pbelevich:
This PR introduces 3 options to run CrossLingualMLMTask:

1. `--pipeline_mode=sp` A single process pipeline to run the model in a single process sharded on local devices. This is a default option.
```
python cross_lingual_mlm_task.py --num_lines 3000 --gpus=8 --pipeline_mode=sp
```
2. `--pipeline_mode=cpu` CPU-RPC pipeline to spawn master process and `gpu` number of workers, each holding one shard, communication is done via CPU RPC
```
python cross_lingual_mlm_task.py --num_lines 3000 --gpus=8 --pipeline_mode=cpu
```
3. `--pipeline_mode=cuda` CUDa-RPC pipeline to spawn master process and `gpu` number of workers, each holding one shard, communication is done via CUDA RPC
```
python cross_lingual_mlm_task.py --num_lines 3000 --gpus=8 --pipeline_mode=cuda
```


zhangguanheng66 (commented):
I would suggest to update the context with a detailed description:

- what's the pipeline parallel added in this PR? (very briefly).
- LocalPipeline vs DistPipeline.
- commands to run the case.

zhangguanheng66 (approved):
LGTM. Thanks for putting together the PR.

examples/BERT/cross_lingual_mlm_task.py
```diff
@@ -116,40 +121,87 @@ def text_transform(x: str) -> List:
...
+        total += params
+        print(f'shard{i} = {int(params)}M')
+    print(f'total = {int(total)}M')
+
+    print("Allocating memory")
     if not args.dist:
-        device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
-        model = CrossLingualMLMTask(ntokens, args.emsize, args.nhead, args.nhid, args.nlayers, args.dropout)
-        model = model.to(device)
+        model = LocalPipeline(shards, devices)
```
zhangguanheng66, line 166: Is this the situation like DDP? If so, should I set `args.dist` to `False`? But I thought we consider `DDP` under distributed. Just curious.

examples/BERT/cross_lingual_mlm_task.py
```diff
@@ -116,40 +121,87 @@ def text_transform(x: str) -> List:
     pad_id = vocab(['pad'])[0]
     ntokens = len(vocab)
 
+    xlmr = XLMRModelShards(ntokens, args.emsize, args.nhead, args.nhid, args.dropout)
+    mlm = MLMShards(ntokens, args.emsize)
+    devices = [f"cuda:{i}" for i in range(args.gpus)] if torch.cuda.is_available() else ["cpu"]
+
+    if len(devices) == 1:
```
zhangguanheng66, line 128: I would add some comments to explain these three branch.

examples/BERT/shard_model.py
```diff
@@ -0,0 +1,30 @@
+import torch.nn as nn
+from model import TransformerEncoderLayer, XLMREmbedding
+
+class XLMRModelShards():
```
zhangguanheng66, line 4: `XLMRModelShards` and `MLMShards` literally are three utility functions to construct `XLMREmbedding`, `TransformerEncoder`, and `MLM`. Do we really need them here?

zhangguanheng66:
I also update `xlmr_mlm` branch to fix the broken tests.

examples/BERT/cross_lingual_mlm_task.py
pbelevich: Sorry for the wrong naming. Maybe it makes sense to:
1. Rename `LocalPipeline` to `SingleProcessPipeline` and `DistPipeline` to `RPCPipeline`
2. `RPCPipeline` will have two implementations CPU RPC and CUDA RPC
3. Replace `args.dist` and `args.cuda_rpc` with 
a) no argument to use `SingleProcessPipeline`
b) `args.use_cpu_rpc` to use `RPCPipeline` with CPU RPC
c) `args.use_cuda_rpc`  to use `RPCPipeline` with CUDA RPC
4. TBD: `LocalPipeline` can utilize DDP. I can implement it in the next pull request

WDYT?

examples/BERT/shard_model.py
pbelevich: These are just factories to create models' shards, they are not required but make code cleaner

PR #1179, opened 2021-02-18
[WIP][Do not review] enabling tests based on google drive

parmeet:
#1171 

Enabling back all the tests that depend on downloading data from google drive (essentially reversing what was done in #1171)

Our current understanding is that based on:
1) we are caching .data folder (where all the data downloaded goes during testing) which is refreshed every week.
2) Based on merge of #1176, all the google drive data now have defined path in .data folder (defined using _PATH variable) 
3) If data is available in .data folder, it will used during testing (and will not trigger the download: refer https://github.com/pytorch/text/blob/1197514eb8cc33ccff10f588534f405b43908660/torchtext/utils.py#L79) 

Ideally this should ensure that we do not download if the data is already cached by CCI. Any blind spot here?

### DIFF
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -9,19 +9,6 @@ from ..common.torchtext_test_case import TorchtextTestCase
 from ..common.parameterized_utils import load_params
 from ..common.assets import conditional_remove
 
-GOOGLE_DRIVE_BASED_DATASETS = [
-    'AmazonReviewFull',
-    'AmazonReviewPolarity',
-    'DBpedia',
-    'IMDB',
-    'IWSLT',
-    'SogouNews',
-    'WMT14',
-    'YahooAnswers',
-    'YelpReviewFull',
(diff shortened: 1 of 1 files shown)


cpuhrsch:
[One of the jobs](https://app.circleci.com/pipelines/github/pytorch/text/2492/workflows/65fb015b-f07c-47e6-bb86-42f17960986e/jobs/70958) seems to indicate the CI ran out of Google drive quota again.

parmeet:
> [One of the jobs](https://app.circleci.com/pipelines/github/pytorch/text/2492/workflows/65fb015b-f07c-47e6-bb86-42f17960986e/jobs/70958) seems to indicate the CI ran out of Google drive quota again.

right, the challenge really is that all the tests must pass and all the data must be cached at once, otherwise the cache key would just contain partial data that would lead to potential failures in subsequent runs for missing datasets until the key is refreshed again to hopefully contain all the data. The chance of failure of atleast one test is significant based on what I have seen so far.

PR #1180, opened 2021-02-18
[WIP][Do Not Review] enabling tests that depends on downloading from google drive

parmeet:
Enabling back all the tests that depend on downloading data from google drive (essentially reversing what was done in #1171)

Our current understanding is that based on:
1) we are caching .data folder (where all the data downloaded goes during testing) which is refreshed every week.
2) Based on merge of #1176, all the google drive data now have defined path in .data folder (defined using _PATH variable) 
3) If data is available in .data folder, it will used during testing (and will not trigger the download: refer https://github.com/pytorch/text/blob/1197514eb8cc33ccff10f588534f405b43908660/torchtext/utils.py#L79) 

Ideally this should ensure that we do not download if the data is already cached by CCI. Any blind spot here?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -343,7 +343,7 @@ jobs:
       - run:
           name: Generate cache key
           # This will refresh cache on Sundays, nightly build should generate new cache.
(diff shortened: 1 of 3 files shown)


parmeet:
Closing it up, No longer relevant at the moment.

PR #1184, opened 2021-02-18
[BC Breaking] Add EnWiki9 raw dataset

zhangguanheng66:
To have consistent datasets in `torchtext.datasets`, `EnWik9` dataset has been re-writed to yield the raw text data. The old `EnWik9` can be found in `torchtext.legacy.datasets`.


test/data/test_builtin_datasets.py
```diff
@@ -13,6 +13,7 @@
     'AmazonReviewFull',
     'AmazonReviewPolarity',
     'DBpedia',
+    'EnWik9'
```
cpuhrsch, line 16: This dataset is not based on google drive

cpuhrsch (approved):

PR #1187, opened 2021-02-18
Move torchtext.datasets.common to torchtext.data.datasets_utils

cpuhrsch:
This moves common.py out of torchtext/datasets so that the folder then only contains datasets


cpuhrsch:
Adressed by https://github.com/pytorch/text/pull/1188

PR #1189, opened 2021-02-19
Add Stanford Sentiment Treebank (SST) dataset

zhangguanheng66:
The legacy SST was retired in `torchtext.legacy` folder. This one yields the raw text strings.


torchtext/datasets/sst.py
```diff
@@ -0,0 +1,45 @@
...
+from torchtext.datasets.common import RawTextIterableDataset
+from torchtext.datasets.common import wrap_split_argument
+from torchtext.datasets.common import add_docstring_header
+from torchtext.datasets.common import find_match
+import os
+import io
+try:
+    from nltk.tree import Tree
+except ImportError:
+    print("Please install NLTK. See the docs at https://nltk.org for more information.")
```
cpuhrsch, line 11: I'd raise this as an error, so that processing doesn't continue. And further I'd make explicit that this is due to this dataset specifically.

PR #1191, opened 2021-02-19
re-parameterization of IWSLT dataset

parmeet:
Summary:
re-parameterization of IWSLT dataset key-word arguments

Input arguments are broken down into:
- source language
- target language
- year
- validation set
- test set

Follow-up:
- [ ] Generic XML cleanup function shared across both datasets
- [ ] Generic Tags cleanup function shared across both datasets
- [ ] Key number of lines on set instead of tuple to reduce entries
- [ ] Add example data point output and datapoint type to docstring for IWSLT2016/2017
- [ ] Audit other translation datasets under the same light as this change: Multi30k and WMT14
- [ ] Create issue to ask about option to return raw data without removal of tags etc.


torchtext/datasets/iwslt.py
```diff
@@ -70,178 +99,55 @@ def _construct_filepaths(paths, src_filename, tgt_filename):
...
-        train.tags.fr-en.fr
+def IWSLT(root='.data', split=('train', 'valid', 'test'), offset=0, language_pair=('de', 'en'), year=2016, valid_set='dev2010', test_set='tst2010'):
+    """TODO
     """
-    if not isinstance(train_filenames, tuple) and not isinstance(valid_filenames, tuple) \
-            and not isinstance(test_filenames, tuple):
-        raise ValueError("All filenames must be tuples")
+
+    if not isinstance(language_pair, list) and not isinstance(language_pair, tuple):
+        raise ValueError("language_pair must be list or tuple")
```
cpuhrsch, line 107: nit: It's often useful to also print the value you actually got that causes the error.

"language_pair must be list of tuple, but got {} instead".type(language_pair)

torchtext/datasets/iwslt.py
```diff
@@ -7,11 +7,40 @@
...
+            'de': ['ro', 'en', 'nl', 'it'],
+            'it': ['en', 'nl', 'de', 'ro'],
+            'nl': ['de', 'en', 'it', 'ro'],
+        },
+        'year': 17,
+    }
+}
 
-MD5 = 'c393ed3fc2a1b0f004b3331043f615ae'
+URL = [v['URL'] for v in SUPPORTED_DATASETS.values()]
```
cpuhrsch, line 42: You could make this a dictionary that's key'd on the year

torchtext/datasets/iwslt.py
```diff
@@ -70,178 +99,55 @@ def _construct_filepaths(paths, src_filename, tgt_filename):
...
-        train.tags.en-ar.en
-        train.tags.en-cs.cs
-        train.tags.en-cs.en
-        train.tags.en-de.de
-        train.tags.en-de.en
-        train.tags.en-fr.en
-        train.tags.en-fr.fr
-        train.tags.fr-en.en
-        train.tags.fr-en.fr
+def IWSLT(root='.data', split=('train', 'valid', 'test'), offset=0, language_pair=('de', 'en'), year=2016, valid_set='dev2010', test_set='tst2010'):
```
cpuhrsch, line 102: Let's split this into two datasets. One IWSLT2016 and one IWSLT2017. This should make the API easier.

torchtext/datasets/iwslt2016.py
```diff
@@ -0,0 +1,331 @@
...
+def _clean_xml_file(f_xml):
+    f_txt = os.path.splitext(f_xml)[0]
+    with codecs.open(f_txt, mode='w', encoding='utf-8') as fd_txt:
+        root = ET.parse(f_xml).getroot()[0]
+        for doc in root.findall('doc'):
+            for e in doc.findall('seg'):
+                fd_txt.write(e.text.strip() + '\n')
+
+
+def _clean_tags_file(f_orig):
```
cpuhrsch, line 191: As a follow-up this can lead to a generic dataset_utils.py function that can be used in multiple contexts.

torchtext/datasets/iwslt2016.py
```diff
@@ -0,0 +1,331 @@
...
+}
+
+
+def _read_text_iterator(path):
+    with io.open(path, encoding="utf8") as f:
+        for row in f:
+            yield row
+
+
+def _clean_xml_file(f_xml):
```
cpuhrsch, line 182: As a follow-up can this lead to a generic XML cleaning function?

torchtext/datasets/iwslt2016.py
```diff
@@ -0,0 +1,331 @@
...
+
+}
+
+URL = SUPPORTED_DATASETS['URL']
+MD5 = SUPPORTED_DATASETS['MD5']
+
+NUM_LINES = {
+    'train': {
+        'train': {
+            ('en', 'ar'): 224126,
```
cpuhrsch, line 32: If you use sets here as a key you don't need to store both orders of language pairs

cpuhrsch (approved):

PR #1192, opened 2021-02-19
Remove the step to clean the cache for test_iwslt

zhangguanheng66:


test/data/test_builtin_datasets.py
```diff
@@ -267,39 +267,40 @@ def test_imdb(self):
         self._helper_test_func(len(test_iter), 25000, next(test_iter)[1][:25], 'I love sci-fi and am will')
         del train_iter, test_iter
 
-    def test_iwslt(self):
```
cpuhrsch, line 270: To explicitly disable a test like this you can use [unittest.skip](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures)

cpuhrsch (approved):

test/data/test_builtin_datasets.py
zhangguanheng66: Yup. I forget this. Let me re-send it.

PR #1207, opened 2021-02-23
re-parameterization of Multi30K dataset

parmeet:
Summary:

- Refactoring Multi30K translation dataset in accordance with #1191 
- Moved it to experimental folder
- Added example usage 
- Added new doc page for experimental raw datasets
- Updated doc string


torchtext/experimental/datasets/raw/multi30k.py
```diff
@@ -0,0 +1,526 @@
...
+        }
+    }
+}
+
+
+URL = {'train': [], 'valid': [], 'test': []}
+MD5 = {'train': [], 'valid': [], 'test': []}
+NUM_LINES = {'train': [], 'valid': [], 'test': []}
+
+for task in SUPPORTED_DATASETS.keys():
```
cpuhrsch, line 374: nit: the "in" operator should return a key by default, so I don't think you need .keys() here

cpuhrsch (approved):

PR #1208, opened 2021-02-23
convert version text to link

mattip:
Convert the static version text to links in the 0.8 and 0.8.1 documentation

@brianjo

### DIFF
--- a/0.8.1/data.html
+++ b/0.8.1/data.html
@@ -184,7 +184,7 @@
               
               
                 <div class="version">
-                  0.8.1
+                  <a href=https://pytorch.org/text/versions.html>0.8.1  &#x25BC</a>
                 </div>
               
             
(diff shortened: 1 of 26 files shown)


zhangguanheng66 (commented):
Does this mean that we need to manually change this in the future?

mattip:
No, going forward the documentation will be generated with this as part of the format template. 

Speaking of which, I see that v0.9.0-rc1 was tagged, but the tag commit did not trip the CI document upload. I think this is because the [job was canceled](https://app.circleci.com/pipelines/github/pytorch/text?branch=release%2F0.9)? @seemethere thoughts?

mattip:
Do I need to do something more here?

zhangguanheng66:
cc @brianjo Could you give a review?

brianjo (approved):

PR #1211, opened 2021-02-24
re-parameterization of WMT14 with move to experimental folder

parmeet:
Summary:

- Refactored WMT14 translation dataset in accordance with #1191 
- Added example for data pointers
- removing keys() when iterating over dictionary in Multi30k
- update examples to use 'next' with iterators


cpuhrsch:
Given the unresolved complexities, let's move this into experimental and defer release.

cpuhrsch (approved):

PR #1213, opened 2021-02-25
resolving download folder for some raw datasets

parmeet:
Summary:

Following datasets were not taking into account the provided root folder
- AmazonReviewFull
- AmazonReviewPolarity
- SogouNews

Added dedicated WMT14 folder to provided root path as it extract all the files without creating sub-folder


cpuhrsch (approved):

cpuhrsch:
Thanks for catching this! As a follow-up, can we setup a test that discovers this issue?

PR #1216, opened 2021-02-25
replaced data pool with batch_sampler

parmeet:
Summary:

- Replaced pooled data with batch_sample in-build utility in Dataloader to promote standards.
- Potentially more efficient?


cpuhrsch (approved):
A few more inline comments might make it easier to read, otherwise this looks good to me!

PR #1217, opened 2021-02-26
added unit tests to do sanity check on raw data download

parmeet:
We would like to do basic sanity checks on data download

- The download should be in the provided root directory (and it's subdirectories) 
- We want to ensure that root directory does not get polluted by downloaded and extracted files. So any raw data download is not allowed to create more than two top level files. So creators of new raw datasets must ensure that if the downloaded/extracted data creates multiple files, they should be in dedicated subdirectories of the root directory


parmeet:
The properties being tested here are built-in using the decorator.

PR #1219, opened 2021-02-27
Set up Daily Caching mechanism with Master job 

parmeet:
Setting up daily cache with Master job that create data cache and refresh it on daily basis. The other jobs will depend on this to ensure we consume cache from Master job.

- [ ] Provide a summary of missing datasets in cache
- [ ] Provide sanity check for cache between cache creation and PRs
- [ ] Enable assertions for paths in download_extract_validate


test/common/cache_utils.py
```diff
@@ -0,0 +1,84 @@
...
+import json
+import torchtext
+import functools
+import unittest
+from .parameterized_utils import load_params
+
+CACHE_STATUS_FILE = '.data/cache_status_file.json'
+
+
+def check_cache_status(_func=None, *, input_dataset_name=None, input_split=None):
```
cpuhrsch, line 11: Can't these verifications be run once at import time instead of for each test separately again and again? The wrapper can then simply store the information that if the test is run it must fail, else just runs the test.

parmeet: Yes, now we are checking the cache status in setUpClass method, and raise exception if cache was not successfully build for all the raw datasets. This is a strict policy where we let the whole test suit fail if at-least one dataset is missing in the cache.

test/common/cache_utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def check_cache_status():
+    assert os.path.exists(CACHE_STATUS_FILE), "Cache status file does not exists"
+    with open(CACHE_STATUS_FILE, 'r') as f:
+        cache_status = json.load(f)
+        for dataset_name in cache_status:
+            for split in cache_status[dataset_name]:
+                if cache_status[dataset_name][split]['status'] == "fail":
+                    raise FileNotFoundError("Failing all raw dataset unit tests as cache is missing atleast one raw dataset")
```
cpuhrsch, line 16: You could mention which one is missing

torchtext/data/datasets_utils.py
```diff
@@ -165,7 +165,8 @@ def download_extract_validate(root, url, url_md5, downloaded_file, extracted_fil
     dataset_tar = download_from_url(url, path=os.path.join(root, downloaded_file),
                                     hash_value=url_md5, hash_type=hash_type)
     extracted_files = extract_archive(dataset_tar)
-    assert extracted_file == find_match(extracted_file, extracted_files)
```
cpuhrsch, line 168: Follow-up: We probably still want this (since this assert has helped us in the past), but we can do it as a follow-up.

cpuhrsch (approved):

test/common/cache_utils.py
```diff
@@ -0,0 +1,46 @@
+import os
+import json
+import torchtext
+from .parameterized_utils import load_params
+
+CACHE_STATUS_FILE = '.data/cache_status_file.json'
```
cpuhrsch, line 6: Eventually instead of pass/fail we could include the hash of the expected files to prevent a cache corruption between creation and PR consumption

test/common/cache_utils.py
parmeet: Added a follow-up task

torchtext/data/datasets_utils.py
parmeet: Added a follow-up task

test/common/cache_utils.py
parmeet: Added a follow-up task

.circleci/config.yml
```diff
@@ -347,6 +347,71 @@ jobs:
...
+            cat .data/cache_status_file.json
+      - save_cache:
+
+          key: data-linux-v1-{{ checksum ".circleci-daily" }}
+
+          paths:
+            - .data
+      - run:
+          name: Post process
+          command: .circleci/unittest/linux/scripts/post_process.sh
```
mthrok, line 412: @parmeet `Post process` is not necessary and it could alter the unittest coverage report so they should be removed.

parmeet: Thanks @mthrok for pointing this out. Let me do a follow-up on this.

PR #1220, opened 2021-03-01
[WIP] moving translation data tags clean-up functions to datasets_utils

parmeet:
This is follow up task for PR #1191 

Summary:
- Moved translational data tag clean-up functions to datasets_utils

### DIFF
--- a/torchtext/data/datasets_utils.py
+++ b/torchtext/data/datasets_utils.py
@@ -1,15 +1,43 @@
 import functools
 import inspect
 import os
+import io
 import torch
 from torchtext.utils import validate_file
 from torchtext.utils import download_from_url
 from torchtext.utils import extract_archive
-
+import codecs
+import xml.etree.ElementTree as ET
 """
 These functions and classes are meant solely for use in torchtext.datasets and not
 for public consumption yet.
 """
+
+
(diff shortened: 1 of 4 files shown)


torchtext/data/datasets_utils.py
```diff
@@ -1,17 +1,45 @@
...
-
+import codecs
+import xml.etree.ElementTree as ET
 """
 These functions and classes are meant solely for use in torchtext.datasets and not
 for public consumption yet.
 """
 
 
+def _clean_xml_file(f_xml):
```
zhangguanheng66, line 17: Are those building blocks general enough now to be placed in datasets_utils?

parmeet: It's a good question: I am not so sure about it right now. It is possible that they are too specific to IWSLT datasets. we should do some discussion on what we need to change among these two functions to make them more generic, or if not, what should be the better place to put this shared code.

zhangguanheng66: I'm not sure for now. One possible solution could be: add those utils functions in one dataset, and let other datasets import them from there whenever the datasets need.

zhangguanheng66:
Merged by https://github.com/pytorch/text/pull/1224

PR #1226, opened 2021-03-02
feat(docs): navigate with left/right arrow keys

ydcjeff:
Enables docs navigation with left/right arrow keys. It can be useful for the ones who navigate with keyboard a lot.
More info : https://github.com/sphinx-doc/sphinx/pull/2064


zhangguanheng66:
Can we have a preview of this change?

zhangguanheng66:
cc @brianjo

ydcjeff:
@zhangguanheng66 There is no artifacts uploaded in `build_doc` job as far as I found out.
But there is a preview from audio : https://174236-90321822-gh.circle-artifacts.com/0/docs/index.html

zhangguanheng66:
I was wondering if we could avoid to collapse the navigation for "Package Reference".

ydcjeff:
> I was wondering if we could avoid to collapse the navigation for "Package Reference".

That was caused by `pytorch_sphinx_theme`, I have sent a PR here https://github.com/pytorch/pytorch_sphinx_theme/pull/97

zhangguanheng66:
> > I was wondering if we could avoid to collapse the navigation for "Package Reference".
> 
> That was caused by `pytorch_sphinx_theme`, I have sent a PR here [pytorch/pytorch_sphinx_theme#97](https://github.com/pytorch/pytorch_sphinx_theme/pull/97)

Thanks for the quick response. I will merge it once @brianjo approves it.

zhangguanheng66:
I also have a request. Different from [vision](https://pytorch.org/vision/stable/index.html) and [audio](https://pytorch.org/audio/stable/index.html), the font of [text](https://pytorch.org/text/stable/index.html) is bold everywhere. Can we fix this for consistency?

ydcjeff:
> I also have a request. Different from [vision](https://pytorch.org/vision/stable/index.html) and [audio](https://pytorch.org/audio/stable/index.html), the font of [text](https://pytorch.org/text/stable/index.html) is bold everywhere. Can we fix this for consistency?

I think that maybe cause of fonts torchtext using (Lato which has thick line width) while vision and audio is FreightSans which has thin line width. Both of 3 sites are using same font weight.

zhangguanheng66:
> > I also have a request. Different from [vision](https://pytorch.org/vision/stable/index.html) and [audio](https://pytorch.org/audio/stable/index.html), the font of [text](https://pytorch.org/text/stable/index.html) is bold everywhere. Can we fix this for consistency?
> 
> I think that maybe cause of fonts torchtext using (Lato which has thick line width) while vision and audio is FreightSans which has thin line width. Both of 3 sites are using same font weight

Can we fix this? Or point me where to fix this. Thanks a lot.

ydcjeff:
Sure you need to remove this line and `font-family` in [this file](https://github.com/pytorch/text/blob/master/docs/source/_static/css/pytorch_theme.css)

zhangguanheng66:
> Sure you need to remove this line and `font-family` in [this file](https://github.com/pytorch/text/blob/master/docs/source/_static/css/pytorch_theme.css)

Not sure. It seems torchaudio also use Lato. See [here](https://github.com/pytorch/audio/blob/master/docs/source/_static/css/pytorch_theme.css#L1-L8).

ydcjeff:
@zhangguanheng66 Updated the previous comment.
torchaudio doesn't have Lato font in https://github.com/pytorch/audio/blob/master/docs/source/conf.py

PR #1227, opened 2021-03-02
Remove Lato font

zhangguanheng66:
The font of text website is bold. Remove the Lato font and the whole `setup` func is not necessary.


brianjo (approved):
Looks good. Thanks for noticing the problem George!

PR #1236, opened 2021-03-04
workspace to share cache key within jobs

parmeet:
Update config to avoid changing cache key between jobs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -362,6 +362,10 @@ jobs:
           command: |
             echo "$(date +"Year-Week %Y-%U")" > .circleci-weekly
             echo "$(date +"Year-Date %Y-%d")" > .circleci-daily-linux
+      - persist_to_workspace:
+          root: .
+          paths:
+            - .circleci-daily-linux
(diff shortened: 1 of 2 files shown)


cpuhrsch (approved):

.circleci/config.yml
```diff
@@ -545,12 +555,13 @@ jobs:
       - checkout
       - designate_upload_channel
       - load_conda_channel_flags
+      - attach_workspace:
+          at: .
       - run:
           name: Generate cache key
           # This will refresh cache on Sundays, nightly build should generate new cache.
           command: |
             echo "$(date +"Year-Week %Y-%U")" > .circleci-weekly
```
cpuhrsch, line 564: Shouldn't this command then also suffer from the same issue? Do we even need the weekly cache anymore?

parmeet: Not really, because the key unique owned by the job, and does not rely on any other cache building-up job (like the case for daily cache). 

will do a follow-up on removing the weekly cache by consolidating everything into daily cache.

PR #1238, opened 2021-03-04
Clean up CCI cache handling and shorten test data caching jobs when cache exists

### LINKED ISSUE #1237: Shorten the cache job
Currently the cache job detects if the cache exists by retrieving the cache. This adds 5 mins to all the unit tests and GBs of data transferred but not used. We should stop retrieving the cache in cache generation jobs.

Unfortunately CCI does not provide a mechanism to check if a cache exists.

A workaround is to cache a small file, which is easy to fetch, alongside of the actual cache, then restore only this small file.
If the small file exists, nothing needs to be done otherwise, the job can generate cache. 

```
restore_cache:
  key: foobar
  paths:
    - small_file
cache_generation:
  command:
    <generate cache only when small_file does not exist>
save_cache:  # if the small_file existed in `restore_cache` stage, this has no effect
  key: foobar
  paths:
    - small_file
    - the real cache
```

cc @parmeet

mthrok:
1. Add `generate_cachekey` and `fetch_cachekey` command that abstracts away the way cache key is generated.
2. Fix the daily cache key. `date +%D` (the previous `"+%Y-%d"` does not change on daily basis)
3. Remove conda environment cache for simplicity (this is no longer the bottleneck and having cached conda env often causes more trouble when macOS test is enabled.)

TODO: After #1235 is merged, adapt the commands to macOS tests too.

Resolves #1237


mthrok:
Initial run (no cache):

https://app.circleci.com/pipelines/github/pytorch/text/2896/workflows/2c5be66a-77e8-4ca9-a3f1-b2f857b21372

parmeet:
> Resolves #1237

@mthrok  FYI: There was a subtle bug, in the way #1219  implemented the daily cache which is resolved in #1236 by sharing the cache key file between jobs through workspace.

mthrok:
> > Resolves #1237
> 
> @mthrok FYI: There was a subtle bug, in the way #1219 implemented the daily cache which is resolved in #1236 by sharing the cache key file between jobs through workspace.

@parmeet God it. I think this PR will still work complementarily.

parmeet:
@mthrok I just looked more closely at your PR. I think it's great that you are caching cache_status_file.json separately because it just occurred to me that we still want to try to download the missing datasets that might have failed when we tried to build the cache. So basically the workflow could look something like this:

Steps:

1. Check if cache_status_file.json exists
2. If answer to 1 is yes, then check if any of the 'status' is 'fail' in the entries of file.
3. If answer to 2 is yes, or answer to 1 is no, we continue and call the generate_cache.sh otherwise we halt to job. Also if answer to 1 is yes, then we also download the existing cache and try to download the data that didn't succeed earlier.

We also need to modify https://github.com/pytorch/text/blob/408c4cfad9f2ec68c2d462326867787df2dba307/test/common/cache_utils.py#L19 to not return if cache_status_file.json exists, instead we iterate over the entries and retry to download the datatasets that failed.

What do you think?

cc @cpuhrsch

mthrok:
> Steps:
> 
> 1. Check if cache_status_file.json exists
> 2. If answer to 1 is yes, then check if any of the 'status' is 'fail' in the entries of file.
> 3. If answer to 2 is yes, or answer to 1 is no, we continue and call the generate_cache.sh otherwise we halt to job. Also if answer to 1 is yes, then we also download the existing cache and try to download the data that didn't succeed earlier.

@parmeet I am afraid that does not work in the current setting. On CCI, once a cache is stored with a certain key, you cannot update it. To achieve the failure recovery, keys have to be changed. One way to workaround I can think of is splitting restore/store cache job into different steps with different keys. (or maybe using multiple key for restoring might work, but I have never tried)

parmeet:
> > Steps:
> > 
> > 1. Check if cache_status_file.json exists
> > 2. If answer to 1 is yes, then check if any of the 'status' is 'fail' in the entries of file.
> > 3. If answer to 2 is yes, or answer to 1 is no, we continue and call the generate_cache.sh otherwise we halt to job. Also if answer to 1 is yes, then we also download the existing cache and try to download the data that didn't succeed earlier.
> 
> @parmeet I am afraid that does not work in the current setting. On CCI, once a cache is stored with a certain key, you cannot update it. To achieve the failure recovery, keys have to be changed. One way to workaround I can think of is splitting restore/store cache job into different steps with different keys. (or maybe using multiple key for restoring might work, but I have never tried)

oh right, my bad! yes, then I guess we have no escape once we have incomplete cache? This might indicate we should revisit our testing strategy based on incomplete cache.

cpuhrsch (approved):

parmeet (approved):
LGTM!

PR #1244, opened 2021-03-05
fixing benchmark tests

parmeet:
We have modified our datasets and the way we pass splits and the way datasets are wrapped in decorator. In light of this some of the stuffs in benchmark suit need to be adapted.

### DIFF
--- a/benchmark/benchmark_basic_english_normalize.py
+++ b/benchmark/benchmark_basic_english_normalize.py
@@ -18,17 +18,17 @@ def benchmark_basic_english_normalize():
     experimental_jit_basic_english_normalize = torch.jit.script(experimental_basic_english_normalize)
 
     # existing eager lookup
-    train, _ = AG_NEWS()
+    train = AG_NEWS(split='train')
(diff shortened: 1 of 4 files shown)


zhangguanheng66 (approved):
Good catch! LGTM.

parmeet:
> Good catch! LGTM.

In the same light, I think our examples folder is also potentially out-of-date with the latest changes.

zhangguanheng66:
> > Good catch! LGTM.
> 
> In the same light, I think our examples folder is also potentially out-of-date with the latest changes.

We can either pick them up here, or I can send out a PR to fix that.

parmeet:
> > > Good catch! LGTM.
> > 
> > 
> > In the same light, I think our examples folder is also potentially out-of-date with the latest changes.
> 
> We can either pick them up here, or I can send out a PR to fix that.

Sure, either way is fine I guess. I am not sure though if the diff become huge if there are potentially many files to fix? If so, we could go with separate PR as we find them.

PR #1233, opened 2021-03-03
Raw datasets cleanup

parmeet:
Summary:
- collect common file reading functionalities into datasets_utils
- create decorator to add dataset_name sub-folder to the given root folder to avoid cluttering

PR #1247, opened 2021-03-08
fixing stylecheck

parmeet:
Fixing style checks for dataset files that were introduced in #1233


zhangguanheng66 (approved):
Add a stamp. Let's merge it ASAP after the tests pass. Please also refer to the original PR.

Thanks for fixing it.

PR #1248, opened 2021-03-08
c10::string_views to avoid copies during query

parmeet:
Summary:

Update Vocab class to make use of c10::string_view to avoid copy operations during query. 

Changes:
- Encoding PyBind11::str explicitly inside __getitem__ and created string_view on encoded bytes
- Using custom hash and open addressing dictionary implementation from fastText https://github.com/facebookresearch/fastText/blob/a20c0d27cd0ee88a25ea0433b7f03038cd728459/src/dictionary.cc#L163

Gains:
AG_NEWS: Reduction in time: 
- Token-by-token: ~15%-17% 
-  line-by-line look-up: ~50%-55% 

SogouNews: Reduction in time:
- token-by-token loop up: ~15%
- line-by-line look-up: ~80-85%

- [ ] Look into if static vocab size need to be changed or if we should implement dynamic vocab


torchtext/csrc/vocab.cpp
```diff
@@ -196,31 +194,34 @@ void parse_raw_text_file_chunk(const std::string &file_path, size_t offset,
...
   }
 };
 
-std::tuple<IndexDict, StringList>
-_concat_tokens(std::vector<std::shared_ptr<IndexDict>> chunk_counters,
-               const std::string &unk_token, const int64_t min_freq,
-               const int64_t num_lines, const bool sort_tokens) {
+StringList _concat_tokens(
+    std::vector<std::shared_ptr<
+        ska_ordered::order_preserving_flat_hash_map<std::string, uint32_t>>>
```
cpuhrsch, line 208: nit: could you also create a typedef for ```ska_ordered::order_preserving_flat_hash_map<std::string, uint32_t>``` to shorten the signature a bit?

torchtext/csrc/vocab.cpp
```diff
@@ -230,13 +231,13 @@ _concat_tokens(std::vector<std::shared_ptr<IndexDict>> chunk_counters,
       // add to tokens list only if we exceed min_freq for the first time
       if (tokens_freq[item.first] - cur_token_freq < min_freq &&
           tokens_freq[item.first] >= min_freq) {
-        unique_tokens.push_back(item.first);
+        unique_tokens.push_back(std::string{item.first});
```
cpuhrsch, line 234: Shouldn't item.first already be of type std::string?

torchtext/csrc/vocab.cpp
```diff
@@ -264,44 +265,37 @@ _concat_tokens(std::vector<std::shared_ptr<IndexDict>> chunk_counters,
...
-  return std::make_tuple(std::move(stoi), std::move(unique_tokens));
+  return unique_tokens;
 }
 
-constexpr int64_t GRAIN_SIZE = 13107;
+constexpr uint32_t GRAIN_SIZE = 13107;
 Vocab _load_vocab_from_file(const std::string &file_path,
                             const std::string &unk_token,
-                            const int64_t min_freq, const int64_t num_cpus) {
+                            const uint32_t min_freq, const uint32_t num_cpus) {
```
cpuhrsch, line 274: What's the purpose of changing these arguments from int64 to uint32?

cpuhrsch (changes requested):
The PR includes a lot of changes of int64_t to uint32_t. Is this intended?

torchtext/csrc/vocab.cpp
parmeet: Will restore the original state, it's a result of trying out something else and unfortunately didn't get clean up properly.

parmeet:
> The PR includes a lot of changes of int64_t to uint32_t. Is this intended?

Ya, I wanted to change everything to uint32_t, but then ran into compilation issue due to torchbind, then as i reverted it back, some places didn't clean up properly. Let me restore things back to original state. Thanks for catching this.

torchtext/csrc/vocab.h
```diff
@@ -10,29 +10,53 @@ typedef std::tuple<std::string, std::vector<int64_t>, std::vector<std::string>,
...
+      h = h ^ uint32_t(uint8_t(str[i]));
+      h = h * 16777619;
+    }
+    return h;
+  }
+
+  uint32_t _find(const c10::string_view &w) const {
+    uint32_t stoi_size = stoi_.size();
+    uint32_t id = _hash(w) % stoi_size;
+    while (stoi_[id] != -1 && c10::string_view{itos_[stoi_[id]].data(),
```
cpuhrsch, line 46: Is this view construction actually necessary to do the comparison? I'd have expected c10::string_view to be comparable with std::string (i.e. the entries of itos_).

torchtext/csrc/vocab.h
```diff
@@ -10,29 +10,53 @@ typedef std::tuple<std::string, std::vector<int64_t>, std::vector<std::string>,
...
+                                               itos_[stoi_[id]].size()} != w) {
+      id = (id + 1) % stoi_size;
+    }
+    return id;
+  }
+
+  void _add(const std::string &w) {
+    uint32_t h = _find(c10::string_view{w.data(), w.size()});
+    if (stoi_[h] == -1) {
+      itos_.push_back(w);
```
cpuhrsch, line 56: What happens if we reached max size?

cpuhrsch (approved):
Seems fine, see two comments left to resolve before merging.

torchtext/csrc/vocab.h
parmeet: Right, it's not necessary, nice catch :)

torchtext/csrc/vocab.h
parmeet: It's a good question, which I think can be addressed in the context of dynamic v/s static dictionary (follow-up item). As of now, the static size is 30 million, which is probably reasonable given the current vocab size typically is orders of magnitude lower.

PR #1250, opened 2021-03-08
 fixing and re-organizing pipelines

parmeet:
FIxing pipelines according to new features in torchtext and removing pytext dependency


examples/data_pipeline/README.md
```diff
@@ -36,31 +36,7 @@ This pipeline example shows the application with the vocab text file from Huggin
...
-
-* `torchtext.experimental.transforms.BasicEnglishNormalize` backed by `re2` regular expression library
-* `from pytext.torchscript.vocab.ScriptVocabulary`
-
-With the dependency of `pytext` library, the command to run the pipeline:
-
-    python pipelines.py --pipeline pytext
-
-
-## Experimental PyText
```
zhangguanheng66, line 54: why do we want to remove this pipeline?

parmeet: The github repo of pytext is not maintained anymore and is not in good state as of this writing. This would mean that we have code in torchtext that is breaking going forward. I wasn't so sure, if we still want to maintain these code snippets that may sporadically break?

cc: @hudeven

examples/data_pipeline/pipelines.py
```diff
@@ -166,44 +145,36 @@ def collate_fn(data_batch):
...
+}
+
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description='Data procesing pipelines')
     parser.add_argument('--pipeline', type=str, default='sentencepiece',
                         help='The name of pipeline')
     parser.add_argument('--dataset', type=str, default='AG_NEWS',
                         help='Dataset for performance benchmark')
-    parser.add_argument('--spm-filename', type=str, default='m_user.model',
+    parser.add_argument('--spm-filename', type=str, default='text_unigram_25000',
```
cpuhrsch, line 168: Why this change?

parmeet: This is for ease of use in default mode. If the user does not have access to spm model, user can simply run the code in default mode that will download one of the pre-trained spm model (text_unigram_25000). This change does not impact previous behavior, i.e if the use indeed specify m_user.model  (with name other that the name in pre-trained spm models) it will work with the user model.

examples/data_pipeline/README.md
cpuhrsch: Let's remove this in a separate PR, since we're not entirely clear on it and the other changes in this diff can go ahead, plus we might need it for comparison relatively soon.

cpuhrsch (approved):
See the comment before landing

PR #1254, opened 2021-03-14
Helper function for Vocab object

TuckerBMorgan:
I found it hard to understand how to preform a "full loop" of Token -> index -> Token, I think the intended manner is to just access the itos variable of the vocab. This feels to me to be inconsistent with how the API is setup to get a index for a token, which uses the __getitem__ overload. While this is not a perfect solution, I do think it will help other new users to have a guiding hand about how to get at this information.

### DIFF
--- a/torchtext/vocab.py
+++ b/torchtext/vocab.py
@@ -103,6 +103,12 @@ class Vocab(object):
     def _default_unk_index(self):
         return self.unk_index
 
+    def get_token(self, index):
+        result = self.itos[index]
+        if result is None:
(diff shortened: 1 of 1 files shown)


zhangguanheng66:
You can use `itos` attribute in the vocab object. Just FYI, this vocab will be retired and replaced with `torchtext.experimental.vocab`. cc @parmeet

joecummings:
Covered by current functionality, but might be a good idea to include in the documentation how to get a token.

PR #1263, opened 2021-03-23
using std::array in vocab for additional speed-ups

parmeet:
We replace container type of stoi_ from std::vector to std::array. This bring ~15-20% relative improvement in look-up speed


joecummings:
@Nayef211  Is this still relevant? If so, might be worth cleaning up quickly and merging for the speedup.

PR #1267, opened 2021-03-31
Fixing documentation for torchtext nn modules

parmeet:
Changing Documentation of torchchtext nn modules.

Previous doc looks like:
`CLASS torchtext.nn.modules.multiheadattention.MultiheadAttentionContainer`
New doc looks like:
`CLASS torchtext.nn.MultiheadAttentionContainer`


zhangguanheng66 (changes requested):
If we decide to move this out of the `nn` folder, just remember to have a comprehensive check for the `examples` folder.

cpuhrsch (changes requested):
We should still preserve the torchtext.nn.modules.multiheadattention path, because we already have it, so I'd leave the folder structure (torchtext/nn/modules) as is, but add an init file under (torchtext/nn) to enable torchtext.nn.MultiheadAttentionContainer etc. Note that right now the new path would be torchtext.nn.multiheadattention.MultiheadAttentionContainer

zhangguanheng66:
> We should still preserve the torchtext.nn.modules.multiheadattention path, because we already have it, so I'd leave the folder structure (torchtext/nn/modules) as is, but add an init file under (torchtext/nn) to enable torchtext.nn.MultiheadAttentionContainer etc. Note that right now the new path would be torchtext.nn.multiheadattention.MultiheadAttentionContainer

We should leave a warning for the current users and instruct them to switch to the new path.

parmeet:
> We should still preserve the torchtext.nn.modules.multiheadattention path, because we already have it, so I'd leave the folder structure (torchtext/nn/modules) as is, but add an init file under (torchtext/nn) to enable torchtext.nn.MultiheadAttentionContainer etc. Note that right now the new path would be torchtext.nn.multiheadattention.MultiheadAttentionContainer

ahh, got it. Thanks for clarification! Let me do the changes accordingly.

parmeet:
> > We should still preserve the torchtext.nn.modules.multiheadattention path, because we already have it, so I'd leave the folder structure (torchtext/nn/modules) as is, but add an init file under (torchtext/nn) to enable torchtext.nn.MultiheadAttentionContainer etc. Note that right now the new path would be torchtext.nn.multiheadattention.MultiheadAttentionContainer
> 
> We should leave a warning for the current users and instruct them to switch to the new path.

@zhangguanheng66  basically the symbols are already imported in the __init__.py of torchtext.nn and hence are already available both from torchtext.nn as well as torchtext.nn.modules. So do you mean if the user is importing them as 
`from torchtext.nn.modules import MultiheadAttentionContainer ` we should raise warning to switch to torch.nn import?

zhangguanheng66:
> > > We should still preserve the torchtext.nn.modules.multiheadattention path, because we already have it, so I'd leave the folder structure (torchtext/nn/modules) as is, but add an init file under (torchtext/nn) to enable torchtext.nn.MultiheadAttentionContainer etc. Note that right now the new path would be torchtext.nn.multiheadattention.MultiheadAttentionContainer
> > 
> > 
> > We should leave a warning for the current users and instruct them to switch to the new path.
> 
> @zhangguanheng66 basically the symbols are already imported in the **init**.py of torchtext.nn and hence are already available both from torchtext.nn as well as torchtext.nn.modules. So do you mean if the user is importing them as
> `from torchtext.nn.modules import MultiheadAttentionContainer ` we should raise warning to switch to torch.nn import?

For example
```
from torchtext.nn.modules import MultiheadAttentionContainer
```
They should have a warning to remove `modules`.

cpuhrsch (approved):

PR #1280, opened 2021-04-07
Added Windows build instructions

mstfbl:
This PR adds Windows build instructions for the repo. The repo does not compile without using the [MSVC x64 toolset](https://docs.microsoft.com/en-us/cpp/build/how-to-enable-a-64-bit-visual-cpp-toolset-on-the-command-line?view=msvc-160#use-vcvarsallbat-to-set-a-64-bit-hosted-build-architecture). The given instructions in the PR is the same way in which the Windows builds are triggered for PRs in the repo. As the instructions for Windows builds were missing, I figured it would be good to add them in the README.


joecummings:
Important, but this PR is stale. Will do in a follow-up.

Issue #1265, opened 2021-03-30
How to split `_RawTextIterableDataset`


KickItLikeShika:
## ❓ Questions and Help
I am trying to move from using `legacy` and use new provided features, i was doing this:
```
from torchtext import legacy
TEXT = legacy.data.Field(lower=True, batch_first=True)
LABEL = legacy.data.LabelField(dtype=torch.float)
train_data, test_data = legacy.datasets.IMDB.splits(TEXT, LABEL, root='/tmp/imdb/')
train_data, valid_data = train_data.split(split_ratio=0.8, random_state=random.seed(SEED))
```
But now i want to split train_data, how can i do that?
```
from torchtext.datasets import IMDB
train_iter, test_iter = IMDB(split=('train', 'test'))
# I need to split train_iter into train_iter and valid_iter
```

zhangguanheng66:
It's an iterator so I don't think you can split/shuffle it. I think, it's worth an option to set up the offset or the beginning of line. So for the valid set, you can start from a different line. cc @cpuhrsch @parmeet

zhangguanheng66:
As a temporary solution, you can list all the items from the iterator and split them (if the dataset can fit the memory).
```
train_data = list(train_iter)
```

KickItLikeShika:
Thanks! but i think if we did that we can not get it back to `_RawTextIterableDataset`, right?

parmeet:
@KickItLikeShika Thank you for bringing this up. We will discuss this request internally and get back to you shortly!

parmeet:
We have migrated our datasets on top of [torchdata datapipes](https://github.com/pytorch/data). @ejguan  I wonder if there is an elegant way to split datapipe into two random disjoint datapipes?

enigdata:
could you provide code example?  split the train_iter of iMDB into train and validation

PR #1281, opened 2021-04-14
[WIP] adding offset and to map-style dataset options to Raw Text Dataset Iterator

parmeet:
Reference issue: https://github.com/pytorch/text/issues/1265

Summary:

- This PR add offset option to the Raw Text dataset iterator. This will let users split the data and get data at arbitrary offset location.
- This PR also add a helper class member function to convert the iterator to map-style dataset

User can provide offset in dictionary where keys are the split type and values are the corresponding offsets. By default we offset at 0 i.e we return start at line 0.

Usage Examples:

```
    from torchtext.datasets import IMDB

    # Starting iterator at some offset
    train_iter = IMDB(split='train', offset={'train':1000})

    # Converting iterator to map style dataset
    train_map_style_dataset = train_iter.to_map_dataset()
```


torchtext/data/datasets_utils.py
```diff
@@ -279,18 +280,39 @@ def _download_extract_validate(root, url, url_md5, downloaded_file, extracted_fi
...
+        ), "offset {} should be less than the number of lines {} in dataset".format(
+            offset, full_num_lines
+        )
+        assert offset >= 0, "offset must be non-negative but {} is given".format(offset)
         self.description = description
         self.full_num_lines = full_num_lines
         self._iterator = iterator
-        self.num_lines = full_num_lines
+        self.num_lines = full_num_lines - offset
+        for _ in range(offset):
```
zhangguanheng66, line 314: Hi @parmeet, thanks for taking this issue. I tried this method before when training XLM-R with CC100. In practice, it does take time to skip lines and for a very large dataset, it's not efficient. What I really want is the real offset, something similar to the legacy EnWik9 dataset. Here are some [code pointers](https://github.com/pytorch/text/blob/ba0bf52fd0f7981407ec7479918a5d34b00d2211/torchtext/legacy/datasets/unsupervised_learning.py#L112-L115). Let me know if this makes sense to you.

parmeet: > Hi @parmeet, thanks for taking this issue. I tried this method before when training XLM-R with CC100. In practice, it does take time to skip lines and for a very large dataset, it's not efficient. What I really want is the real offset, something similar to the legacy EnWik9 dataset. Here are some [code pointers](https://github.com/pytorch/text/blob/ba0bf52fd0f7981407ec7479918a5d34b00d2211/torchtext/legacy/datasets/unsupervised_learning.py#L112-L115). Let me know if this makes sense to you.
> 
> Technically, we can generate those offsets offline and save them so users can access them efficiently. wdyt? cc @cpuhrsch

Hi @zhangguanheng66 , Thanks for the proposal. I think it's a good idea. Although I am not sure, what should be viable solution for offsets storage. I could think of following:
1) Provide them as package binary.
Pros: Instant availability
Cons: Package size
2) Store it in public cloud storage and download it on-demand.
Pros: Almost Instant availability
Cons: maintenance/dependence on public cloud.
3) Generate it once when the user requests the data and then store it in local cache for later use.
Pros: We could potentially avoid cons of 1 and 2
cons: Need to iterate over dataset at-least once before we return the iterator to user.

What are your thoughts on this?

zhangguanheng66: I think the options here will depends on the size of the datasets. We can do an experiment with WikiText103/AmazonFullReview and see how this will affect the performance.

parmeet: Typically a user would split data at particular offset and most likely stick with it in consecutive uses/epochs. Hence it may not be necessary to provide offsets for all the lines. But we do not know in advance what offset the user would want. Hence we can decide to create this offset on-the-fly when user request the iterator at given offset. 

IMHO, I won't recommend option 1 and 2 above for reasons stated in the cons and for unnecessary storage at all the offsets. Option 3 seems like a good trade-off with efficiency and storage requirements. For option 3 the workflow could look like this:
1) If user do not provide offset, we just return the iterator as we do now (zero-cost)
2) If user do request data with offset, we first check if offset is cached or not. 
    2.1) If yes, we simply read it from cache and return the offset iterator (zero-cost).
    2.2) If not, then we first iterate over dataset (one-time cost) and store the corresponding offset in cache (minimal storage/maintenance requirements) and then go to step 2.1

This way for any consecutive call to data iterator, we do not incur any additional run-time cost and at the same time need not to worry about potential storage for offsets.

torchtext/datasets/sogounews.py
```diff
@@ -36,9 +36,14 @@
 @_add_docstring_header(num_lines=NUM_LINES, num_classes=5)
 @_create_dataset_directory(dataset_name=DATASET_NAME)
 @_wrap_split_argument(('train', 'test'))
-def SogouNews(root, split):
+def SogouNews(root, split, offset=None):
     path = _download_extract_validate(root, URL, MD5, os.path.join(root, _PATH), os.path.join(root, _EXTRACTED_FILES[split]),
                                       _EXTRACTED_FILES_MD5[split], hash_type="md5")
     logging.info('Creating {} data'.format(split))
+    offset_split = 0
```
cpuhrsch, line 43: This logic should be hidden in the _wrap_split_argument decorator

torchtext/data/datasets_utils.py
```diff
@@ -279,18 +280,39 @@ def _download_extract_validate(root, url, url_md5, downloaded_file, extracted_fi
     return extracted_file
 
 
+class _RawTextMapDataset(torch.utils.data.Dataset):
+    """Helper class to convert IterableDatatset to Dataset"""
+
+    def __init__(self, raw_text_iter_data):
+        """Initialized with _RawTextIterableDataset"""
+        self.description = raw_text_iter_data.description
+        self._data = [d for d in raw_text_iter_data]
```
cpuhrsch, line 289: Using ```list(raw_text_iter_data)``` could potentially be more efficient. But large lists of text are difficult in general (e.g. see https://github.com/pytorch/pytorch/issues/13246). As a mitigation we could explore using numpy tensors to store strings or a light custom class that's effectively a wrapper around ```std::vector<std::string>```

docs/source/datasets.rst
```diff
@@ -18,6 +18,12 @@ General use cases are as follows: ::
     for label, line in train_iter:
         tokens += tokenize(label, line)
 
+    # Starting iterator at some offset
+    train_iter = IMDB(split='train', offset={'train':1000})
```
cpuhrsch, line 22: I think it might be more straightforward to only accept a single int for offset for now and broadcast it across all splits.

torchtext/datasets/sogounews.py
parmeet: Right, make sense. Completely missed the decorator. Thanks!

docs/source/datasets.rst
parmeet: But different splits potentially contains different number of data-points?  Perhaps we could simplify by accepting either tuple (in case split is tuple, we can do one to one mapping) or int if user wants to apply same offset to all the splits?

torchtext/data/datasets_utils.py
cpuhrsch: Another option is to store a few offset values (say every 10,000 lines) and their corresponding byte offset. Then we can use the very efficient [seek](https://man7.org/linux/man-pages/man2/lseek.2.html) in cases where the underlying text data is a file. So if a user requests offset 12,345 we'll first seek to line 10,000 based on the stored byte value and then forward the iterator by 2,345. Of course we can vary the values here (store a larger array of offset metadata, etc.).

docs/source/datasets.rst
cpuhrsch: Sounds good to me

torchtext/data/datasets_utils.py
parmeet: Thanks @cpuhrsch for the proposal. This sounds like a pretty decent approach to me!

parmeet:
Need to split the PR into two:
1) converting iterator style data into map-style data (https://github.com/pytorch/text/issues/1296)
2) Adding offsets to Iterator style data

PR #1288, opened 2021-04-20
Store artifacts of rendered docs so that rendered docs can be checked on each PR

NicolasHug:
The `build_docs` CI run builds the doc on every PR but the artifacts aren't saved, so we can't access the rendered html files.

This PR saves the artifacts so that the built docs can be browsed on each PR by going in the "Artifacts" tab of the `build_docs` CI run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -559,6 +559,9 @@ jobs:
           root: ./
           paths:
             - "*"
+      - store_artifacts:
+          path: ./docs/build/html
+          destination: docs
 
   upload_docs:
     <<: *binary_common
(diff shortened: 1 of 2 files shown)


NicolasHug:
The artifacts for this PR are now at https://app.circleci.com/pipelines/github/pytorch/text/3182/workflows/3f4963f4-1ac3-4644-9dbc-4eb96b104daf/jobs/97137/artifacts

vincentqb (approved):
LGTM, same as [torchaudio](https://github.com/pytorch/audio/blob/master/.circleci/config.yml.in#L571) and https://github.com/pytorch/audio/pull/1321

PR #1290, opened 2021-04-22
Function to compare batch look-up for vocab 

parmeet:
This PR added a function to plot graph for speed-up w.r.t number of tokens in batch. Here we benchmark experimental vocab  against existing one which is based on Python dict.

We do have benchmarking code that compares the performance on real dataset, but this function provides capability to do the benchmarking in controlled settings. 

The output looks like below:

speedup


parmeet:
Discarding due to inclusion of some unnecessary changes

Issue #1272, opened 2021-04-02
EnWik9 is XML dataset, not raw text


pbelevich:
According to http://mattmahoney.net/dc/textdata:

> The data is UTF-8 encoded XML consisting primarily of English text. enwik9 contains 243,426 article titles,

```
In [1]: from torchtext.datasets import EnWik9

In [2]: enwik9 = EnWik9()

In [3]: len(enwik9)
Out[3]: 1

In [4]: from torchtext.legacy.datasets import EnWik9

In [5]: enwik9 = EnWik9()

In [6]: len(enwik9)
Out[6]: 133220996
```

zhangguanheng66:
The new dataset now is an iterator.

pbelevich:
That's super cool, but the new dataset doesn't preprocess XML:
```
In [1]: from torchtext.legacy.datasets import EnWik9

In [2]: enwik9 = EnWik9()

In [3]: enwik9[:30]
Out[3]:
['redirect',
 'applied',
 'ethics',
 'anarchism',
 'originated',
 'as',
 'a',
 'term',
 'of',
 'abuse',
 'first',
 'used',
 'against',
 'early',
 'working',
 'class',
 'radicals',
 'including',
 'the',
 'diggers',
 'of',
 'the',
 'english',
 'revolution',
 'and',
 'the',
 'sans',
 'culottes',
 'of',
 'the']

In [4]: from torchtext.datasets import EnWik9

In [5]: enwik9 = EnWik9()

In [6]: [next(enwik9[0]) for x in range(30)]
Out[6]:
['<mediawiki xmlns="http://www.mediawiki.org/xml/export-0.3/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.mediawiki.org/xml/export-0.3/ http://www.mediawiki.org/xml/export-0.3.xsd" version="0.3" xml:lang="en">\n',
 '  <siteinfo>\n',
 '    <sitename>Wikipedia</sitename>\n',
 '    <base>http://en.wikipedia.org/wiki/Main_Page</base>\n',
 '    <generator>MediaWiki 1.6alpha</generator>\n',
 '    <case>first-letter</case>\n',
 '      <namespaces>\n',
 '      <namespace key="-2">Media</namespace>\n',
 '      <namespace key="-1">Special</namespace>\n',
 '      <namespace key="0" />\n',
 '      <namespace key="1">Talk</namespace>\n',
[...]

PR #1292, opened 2021-04-23
Processing XML for enwik9 data 

parmeet:
https://github.com/pytorch/text/issues/1272

This PR adds functional to process/filter enwik9 data as was done in the legacy dataset. The processing is done similar to [this](https://github.com/facebookresearch/fastText/blob/master/wikifil.pl) script.

It also fixes following processing issues in the legacy data:
1) It removes lines which contains \#redirect as also done by [this](https://github.com/facebookresearch/fastText/blob/master/wikifil.pl) script.
2) It return raw lines in the iterator instead of tokens.

Example Usage:
```
from torchtext.data.functional import filter_wikipedia_xml_from_iterator
from torchtext.datasets import EnWik9
data_iter = EnWik9(split='train')
filter_data_iter =filter_wikipedia_xml_from_iterator(data_iter)
```

- [ ] Verify that the output is same as produced by https://github.com/facebookresearch/fastText/blob/master/word-vector-example.sh#L20
- [ ] Make the function Scriptable


torchtext/data/datasets_utils.py
```diff
@@ -19,6 +20,61 @@
...
+             (r'\s+', ' '),
+             (r'\n\s*\n', r'\n')
+             ]
+
+
+def _create_data_from_wikipedia_xml_dumps(input_filename):
+    # Clean wikipedia xml dumps according to https://github.com/facebookresearch/fastText/blob/master/wikifil.pl
+    norm_transform = custom_replace(_patterns)
+    with open(input_filename, 'r', encoding='utf-8') as f:
+        while True:
```
cpuhrsch, line 67: Instead of
```
while True:
    line = f.readline()
```
you should be able to just do
```
for line in f:
```

torchtext/datasets/enwik9.py
```diff
@@ -26,9 +26,9 @@
 @_create_dataset_directory(dataset_name=DATASET_NAME)
 @_wrap_split_argument(('train',))
 def EnWik9(root, split):
+    logging.info('Creating {} data'.format(split))
     dataset_tar = download_from_url(URL, root=root, hash_value=MD5, hash_type='md5')
     extracted_files = extract_archive(dataset_tar)
     path = extracted_files[0]
-    logging.info('Creating {} data'.format(split))
     return _RawTextIterableDataset(DATASET_NAME,
-                                   NUM_LINES[split], _read_text_iterator(path))
```
cpuhrsch, line 34: Since this is BC-breaking and removes this data irrevocably and by default, should we offer the transform separately as a functional and to be optionally applied by the user via e.g.
```
data = EnWik9(split='train')
data = _create_data_from_wikipedia_xml_dumps(data)
```
(we might want to pick a different name for the functional then)

parmeet: Yes, I like the idea of functional. This way we can also avoid putting additional options if the user wanted the raw data without filtering.

torchtext/data/functional.py
```diff
@@ -180,3 +180,65 @@ def numericalize_tokens_from_iterator(vocab, iterator, removed_tokens=None):
...
+        >>> data_iter = EnWik9(split='train')
+        >>> filter_data_iter =filter_wikipedia_xml_from_iterator(data_iter) 
+    """
+
+    norm_transform = custom_replace(_patterns)
+    for line in raw_text_iterator:
+        if '#redirect' in line or '#REDIRECT' in line:
+            continue
+        line = list(norm_transform([line]))[0]
+        if line != ' ' and line != '':
```
cpuhrsch, line 243: nit: You could also do [line.strip()](https://docs.python.org/3/library/stdtypes.html?highlight=strip#str.strip), which will cover any line compromised only of white-space
```
line, = list(norm_transform([line]))
line = line.strip()
if line:
    yield line.strip()
```

torchtext/data/functional.py
```diff
@@ -180,3 +180,65 @@ def numericalize_tokens_from_iterator(vocab, iterator, removed_tokens=None):
...
+             (r'6', ' six '), (r'7', ' seven '), (r'8', ' eight '),
+             (r'9', ' nine '),
+             (r'[^a-z\n]+', ' '),
+             (r'\n ', ''),
+             (r'\s+', ' '),
+             (r'\n\s*\n', r'\n')
+             ]
+
+
+def filter_wikipedia_xml_from_iterator(raw_text_iterator):
```
cpuhrsch, line 225: I'd suggest we remove ```_from_iterator```, it actually works for iterables. So you can also feed lists of strings into this, or a regular iterator or a dataset iterator or even just plain an open file 

```
list(filter_wikipedia_xml(open('my_xml_file.txt')))
```

parmeet: Sounds good! I like the idea. Let me modify the doc and name accordingly.

Let's also have a separate discussion on torchtext standards so that we can include them on landing page (should be useful for contributing guidelines as well).

cpuhrsch (approved):
Looks good! Follow-up work can include making this scriptable using our C++ regex implementation (once we move it into master).

PR #1293, opened 2021-04-23
adding version.txt file

parmeet:
Required for FBcode.

Reference: https://github.com/pytorch/vision/pull/3202


setup.py
```diff
@@ -18,7 +18,8 @@ def read(*names, **kwargs):
 
 
 def _get_version():
```
mthrok, line 20:
```python
if 'BUILD_VERSION' in os.environ:
    version = os.environ('BUILD_VERSION')
else:
    with open(...) as f:
        version = f.readline().strip()
```

datumbox: I believe if you want to do what you propose, you will need to handle the extra corner-cases of adding or not the `sha` which is done on line 35. So you might end up requiring a general code refactoring of this method.

mthrok: more concretely

```python
def _get_version():
    try:
        cmd = ['git', 'rev-parse', 'HEAD']
        sha = subprocess.check_output(cmd, cwd=str(ROOT_DIR)).decode('ascii').strip()
    except Exception:
        sha = None

    if 'BUILD_VERSION' in os.environ:
        version = os.environ('BUILD_VERSION')
    else:
        with open(...) as f:
            version = f.readline().strip()
        if sha is not None:
            version += '+' + sha[:7]

    if sha is None:
        sha = 'Unknown'
    return version, sha
```

parmeet: @mthrok Thanks for suggestion, just to make sure this is just a refactoring of the code right? the end result is same except in once case we always end up reading the file?

datumbox (approved):
LGTM either as-is or with @mthrok's proposal.

setup.py
mthrok: Correct, it is more for the readability / maintainability for the release process.
In fact, introducing `version.txt` forces us to define the version number in multiple sources, (env var / version.txt) but the release process is expecting environment variables, so IMHO the code logic better reflects the intention.

cpuhrsch (approved):

Issue #1296, opened 2021-04-29
Convert iterator-style raw datasets to map-style raw datasets


parmeet:
## 🚀 Feature

**Motivation**

torchtext provide several open source nlp datasets in raw form. These datasets are provide as [Iterables](https://pytorch.org/docs/stable/data.html#iterable-style-datasets). Although there are times when user may prefer [map-style](https://pytorch.org/docs/stable/data.html#map-style-datasets) datasets.

**Pitch**

We would like to implement a functionality that would convert iterable datasets into map-style datasets. This functionality can either be implemented as functional where-in the input would be raw iterable dataset. An alternative implementation would be to directly provide it as a member function for [raw dataset iterable class](https://github.com/pytorch/text/blob/master/torchtext/data/datasets_utils.py#L282).

**Alternatives**

A naive approach would be to simply materialize iterator into list as follows:

```
from torchtext.datasets import IMDB
train_iter = IMDB(split='train')
train_dataset = list(train_iter)
```

Unfortunately, passing list to PyTorch DataLoader would have memory regressions for multi-processing. More details below.

**Additional context**

Note that the solution stated in Alternatives would have issues when doing multi-process data loading. This issue is discussd in details here (https://github.com/pytorch/pytorch/issues/13246). Other options would be to explore numpy tensors, but seems like it would suffer from same issue as lists, refer this this comment (https://github.com/pytorch/pytorch/issues/13246#issuecomment-445770039).

**Potential Solution**

One idea would be to create a thin C++ wrapper where in the data is stored in std::array<std::string> data-structure. This wrapper can then be binded in Python using [pybind11](https://github.com/pybind/pybind11). One potential caveat though is performance regression due to overhead of querying through binding functions. Hopefully this cost is negligible compared to downstream processing.

cc: @cpuhrsch

cpuhrsch:
I think we should also consider a functional that will do this operation for arbitrary iterables, so effectively a factory function for this kind of Dataset. Next to memory issues with lists of strings we should also keep in mind the shared memory semantics. Can we make this datastructure share-able across workers? Something that might help here (and is new in Python 3.8) is [shared_memory](https://docs.python.org/3/library/multiprocessing.shared_memory.html) and could maybe make this easier.

Kousik-Sasmal:
@parmeet @cpuhrsch  Is this issue still open? If yes, can you please assign this to me?

PR #1299, opened 2021-05-05
functional to convert iterables into map-style dataset

parmeet:
With reference to issue #1296, this is a naive implementation to convert to map-style datasets.

Follow-up:
- [ ] Move away from lists and implement specialized containers. Reference: #1296

### DIFF
--- a/torchtext/data/functional.py
+++ b/torchtext/data/functional.py
@@ -7,6 +7,7 @@ __all__ = [
     "sentencepiece_numericalizer", "sentencepiece_tokenizer",
     "numericalize_tokens_from_iterator",
     "filter_wikipedia_xml",
+    "to_map_style_dataset",
 ]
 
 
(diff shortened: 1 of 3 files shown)


torchtext/data/functional.py
```diff
@@ -249,3 +250,35 @@ def filter_wikipedia_xml(text_iterator):
...
+        >>> from torchtext.datasets import IMDB
+        >>> from torchtext.data import to_map_style_dataset
+        >>> train_iter = IMDB(split='train')
+        >>> train_dataset = to_map_style_dataset(train_iter)
+        >>> file_name = '.data/EnWik9/enwik9'
+        >>> data_iter = to_map_style_dataset(open(file_name,'r'))
+    """
+
+    # Inner class to convert iterable-style to map-style dataset
+    class map_style_dataset(torch.utils.data.Dataset):
```
cpuhrsch, line 272: nit: classes in Python should use [CamelCase](https://www.python.org/dev/peps/pep-0008/#id41) and we might want to prefix it with an underscore because it's private.

cpuhrsch (approved):

PR #1301, opened 2021-05-07
unit testing for experimental vocab

parmeet:
Modified unit tests to check the Exception being raised if trying to insert or append token that already exist in vocab


parmeet:
These changes are part of #1302

PR #1304, opened 2021-05-12
Modifying and moving Vocab factory functions

parmeet:
This PR moves two of the factory functions for creating Vocab object out of experimental. Furthermore, it also removes unknown token argument. Issue #1305 discusses the behavior in further details.


test/experimental/test_with_asset.py
```diff
@@ -178,23 +178,22 @@ def test_glove_different_dims(self):
...
         asset_path = get_asset_path(asset_name)
-        tokenizer = basic_english_normalize()
-        jit_tokenizer = torch.jit.script(tokenizer)
-        v = build_vocab_from_text_file(asset_path, jit_tokenizer, unk_token='<new_unk>')
-        expected_itos = ['<new_unk>', "'", 'after', 'talks', '.', 'are', 'at', 'disappointed',
-                         'fears', 'federal', 'firm', 'for', 'mogul', 'n', 'newall', 'parent',
-                         'pension', 'representing', 'say', 'stricken', 't', 'they', 'turner',
-                         'unions', 'with', 'workers']
+        v = build_vocab_from_text_file(asset_path)
+        expected_itos = ['after', 'talks', "'disappointed'", 'Fears', 'Federal',
```
cpuhrsch, line 191: What happened to the  "'" in the previous expected_itos?

torchtext/csrc/vocab.cpp
```diff
@@ -356,6 +383,49 @@ Vocab _build_vocab_from_text_file(const std::string &file_path,
...
+  std::condition_variable cv;
+  std::atomic<int> thread_count(0);
+
+  // create threads
+  int64_t j = 0;
+  for (int64_t i = 0; i < num_lines; i += chunk_size) {
+    auto counter_ptr = std::make_shared<IndexDict>();
+    thread_count++;
+    at::launch([&, file_path, num_lines, chunk_size, j, i, counter_ptr]() {
+      parse_raw_text_file_chunk_using_python_tokenizer(
```
cpuhrsch, line 409: This will likely cause hefty issues because of the global interpreter lock. You can't parallelize this without multiple processes.

torchtext/vocab.py
```diff
@@ -187,6 +196,81 @@ def __prepare_scriptable__(self):
...
+        min_freq: The minimum frequency needed to include a token in the vocabulary.
+
+    Returns:
+        torchtext.vocab.Vocab: A `Vocab` object
+
+    Examples:
+        >>> #generating vocab from text file
+        >>> import io
+        >>> from torchtext.vocab import build_vocab_from_iterator
+        >>> def yield_tokens_batch(file_path):
```
cpuhrsch, line 258: Why "batch"?

test/experimental/test_with_asset.py
parmeet: Actually we changed the tokenizer to python split() function to get away with the dependency on jited tokenizer as they are only available in experimental. This changed the order and expected words for the new vocab constructed using split() function tokenizer.

torchtext/vocab.py
parmeet: We are expecting it to yield list/iterator of tokens instead of just tokens. We could potentially use yield_tokens_list if 'batch' is may be misleading?

torchtext/csrc/vocab.cpp
parmeet: Ahh great catch!

test/experimental/test_with_asset.py
parmeet: The basic_english_normalize() tokenizer/normalizer broke the 'disappointed' into "'", "disappointed" and "'" which is not the case with split()

torchtext/vocab.py
cpuhrsch: Yes, I only worry that someone might think this means "batch of sentences" whereas it's just a list of tokens for a single line.

test/experimental/test_with_asset.py
cpuhrsch: Hm, but using _build_vocab_from_text_file_using_python_tokenizer below this should work now even if it's not JIT'd right?

torchtext/experimental/vocab_factory.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def build_vocab_from_text_file(file_path: str, tokenizer: Optional[Callable] = None, min_freq: int = 1, num_cpus: int = 4) -> Vocab:
+    r"""Create a `Vocab` object from a raw text file.
+    The `file_path` can contain any raw text. This function applies a generic JITed tokenizer in
+    parallel to the text.
+
+    Args:
+        file_object: A file object to read data from.
+        tokenizer: A python callable to split input sentence into tokens. It can also be a Jited Module. By default, the function will do tokenization based on python split() function.
```
cpuhrsch, line 23: I'd add an example of this to the docstring

cpuhrsch (approved):

torchtext/csrc/vocab.cpp
```diff
@@ -356,6 +357,54 @@ Vocab _build_vocab_from_text_file(const std::string &file_path,
   return Vocab(std::move(tokens));
 }
 
+Vocab _build_vocab_from_text_file_using_python_tokenizer(
+    const std::string &file_path, const int64_t min_freq,
+    const int64_t num_cpus, py::object tokenizer) {
```
cpuhrsch, line 362: num_cpus is unused?

torchtext/experimental/vocab_factory.py
```diff
@@ -0,0 +1,68 @@
...
+        min_freq: The minimum frequency needed to include a token in the vocabulary.
+        num_cpus: the number of cpus to use when loading the vectors from file.
```
cpuhrsch, line 25: I'd add that num_cpus only applies when the tokenizer is JIT'd

test/experimental/test_with_asset.py
parmeet: Yes, now we support both JIT'd and pure Python object as tokenizers. Added corresponding tests to check it works with both the cases.

torchtext/vocab.py
parmeet: right, got you!! I just named it to yield_tokens, hopefully the doc should provide the clarify for users what we expect as input to build_vocab_from_iterator.

torchtext/csrc/vocab.cpp
parmeet: ahh my bad! will fix it before merge

PR #1319, opened 2021-06-01
Streamlining download functionality using iopath

parmeet:
This PR introduces usage of iopath in download_from_url and cleans up the function by removing redundant logic flow.


hudeven (approved):
LGTM

PR #1331, opened 2021-06-15
update docs for 0.10.0

parmeet:
Updating docs for the release candidate


versions.html
```diff
@@ -20,6 +20,9 @@ <h1> PyTorch Documentation </h1>
         <li class="toctree-l1">
           <a class="reference internal" href="master/">master (unstable)</a>
         </li>
+        <li class="toctree-l1">
+          <a class="reference internal" href="0.10.0/">v0.10.0 (release candidate)</a>
```
mthrok, line 24: Since we are very close to release, you might want to remove `stable release` label from `0.9.0` and add it to `0.10.0`. Otherwise you will have to make a separate PR.

mthrok (approved):
I could not check the content so stamping.

Note: if you use `gh-pages` as branch name (not `gh_pages`) then GitHub will set up the page and you can review it at `parmeet.github.io/text`

brianjo (approved):
This looks good. We'll check this this morning and I'll get it all set. Thanks!

brianjo:
@mattip Parmeet built out the docs for text 0.10.0 so we'll just need to fix the versions file and the symlink at launch.

PR #1332, opened 2021-06-15
fix tag build so it triggers prerequisites

mattip:
Port pytorch/vision#3998 and pytorch/vision#4053 so that adding a tag will trigger a documentation build-and-upload. The problem was that the upload depends on both the wheel and the build_docs builds, but those did not have the correct filter so they were not running.


mattip:
It would be nice to see this merged before the next tag event, to make sure the docs are now building (the last few tags did not generate docs)

mattip:
I am pretty sure this needs to be backported to the 0.10 branch

mattip:
hmm, binary_linux_wheel_py3.8 did not trigger?

malfet (approved):

parmeet:
Looks like this merge is causing build_docs to fail? https://app.circleci.com/pipelines/github/pytorch/text/3664/workflows/4a92d4fa-4cbc-46da-af5b-b5d75176acc9/jobs/116759 @malfet @mattip

malfet:
I don't think this PR could have broken anything, it just docs build has not been running for non-tagged builds. Looking into the failures now.

PR #1336, opened 2021-06-16
Update sentencepiece to v0.1.95

malfet:
Makes it compileable on M1


mthrok (approved):
Token approval. macOS bibary jobs are failing though.

parmeet:
> Token approval. macOS bibary jobs are failing though.

macOS binary on Master are failing even before (here is previous fail: https://github.com/pytorch/text/pull/1330)

PR #1338, opened 2021-06-17
AG NEWS iterable dataset offsets

salilsdesai:
Added offsets for AG NEWS iterable


test/data/test_builtin_datasets.py
```diff
@@ -76,6 +76,29 @@ def test_raw_text_classification(self, info):
...
         del data_iter
 
+    @parameterized.expand(
+        load_params('raw_datasets.jsonl'),
+        name_func=_raw_text_custom_name_func)
+    def test_dataset_iterator_offsets(self, info):
+        dataset_name = info['dataset_name']
+        split = info['split']
+
+        # TODO: Change to this
```
salilsdesai, line 86: I'm going to change this, but left it like this for now to make sure the test is getting run

test/asset/raw_datasets.jsonl
```diff
@@ -1,7 +1,7 @@
...
 {"dataset_name": "IMDB", "split": "test", "NUM_LINES": 25000, "MD5": "7c2ac02c03563afcf9b574c7e56c153a", "URL": "http://ai.stanford.edu/~amaas/data/sentiment/aclImdb_v1.tar.gz", "first_line": "80dd1b651d74220e41f7c7c4d0be1d7c"}
-{"dataset_name": "AG_NEWS", "split": "train", "NUM_LINES": 120000, "MD5": "b1a00f826fdfbd249f79597b59e1dc12", "URL": "https://raw.githubusercontent.com/mhjabreel/CharCnn_Keras/master/data/ag_news_csv/train.csv", "first_line": "b34631d550696857784c780e75d1019b"}
-{"dataset_name": "AG_NEWS", "split": "test", "NUM_LINES": 7600, "MD5": "d52ea96a97a2d943681189a97654912d", "URL": "https://raw.githubusercontent.com/mhjabreel/CharCnn_Keras/master/data/ag_news_csv/test.csv", "first_line": "4ec431dd3276af3ddaa26a18687e134f"}
+{"dataset_name": "AG_NEWS", "split": "train", "NUM_LINES": 120000, "MD5": "b1a00f826fdfbd249f79597b59e1dc12", "URL": "https://raw.githubusercontent.com/mhjabreel/CharCnn_Keras/master/data/ag_news_csv/train.csv", "first_line": "b34631d550696857784c780e75d1019b", "offset": 11010, "offset_line": "b7923bb47f29e45f94cfc02c2c92e7cd"}
```
cpuhrsch, line 3: I'd suggest to double the lines and make offset=0 the default when not specified for all other test. So you'll have one line with offset 0 and then another with an explicit offset set instead of additional flags. Otherwise as we add flags we might end up with offset_line, offset and chunk_line, chunk, etc.

torchtext/datasets/ag_news.py
```diff
@@ -25,16 +25,21 @@
     'test': 7600,
 }
 
+OFFSETS = {
+    'train': [0, 2513230, 4989675, 7448280, 9917610, 12373034, 14841577, 17310009, 19755260, 22180771, 24603053, 27027049],
```
cpuhrsch, line 29: I'd suggest to use a dictionary that maps line number to bytes.

test/asset/raw_datasets.jsonl
parmeet: I think this much better to avoid testing complexity @salilsdesai

torchtext/datasets/ag_news.py
parmeet: I was curious on reasoning behind using dict over list of bytes. One thing is it would change the look-up logic but performance-wise it would be same right?

test/asset/raw_datasets.jsonl
salilsdesai: @cpuhrsch @parmeet. I'm not sure that I entirely understand what you mean. Does this mean that I add offset=0 and offset_line=... to each line in the original file, then for AG_NEWS train and AG_NEWS split, I add duplicate lines but with the non-zero offset and corresponding offset line?

parmeet: yes, I think this would be the idea. And in our unit test code, since we do not have offsets argument available for other datasets, we may just ignore the test cases with non-zero offsets for datasets other than AG_NEWS.

salilsdesai: > we may just ignore the test cases with non-zero offsets for datasets other than AG_NEWS

Do you mean "ignore the test cases with **zero** offsets"? Since none of the test cases will have non-zero offsets except AG_NEWS.

parmeet: After giving it a second thought, we can just include the lines with non-zero offsets for the datasets that supports offset (AG_NEWS for now) and rest of the lines with offset=0. The unit test [here](https://github.com/pytorch/text/blob/master/test/data/test_builtin_datasets.py#L58) will iterate over each line provided in the raw_datasets.jsonl file. When offset=0, we simply run the test without passing the offset to datasets (this is valid behaviour for datasets both that support and that doesn't support offset), and when offset is non-zero (only AG_NEWS for now), we call the dataset with given offset.

salilsdesai: Yeah I think that makes more sense too. The pull request is updated with this now.

test/data/test_builtin_datasets.py
```diff
@@ -76,6 +76,26 @@ def test_raw_text_classification(self, info):
             self.assertEqual(torchtext.datasets.MD5[dataset_name], info['MD5'])
         del data_iter
 
+    @parameterized.expand(
+        load_params('raw_datasets.jsonl'),
+        name_func=_raw_text_custom_name_func)
+    def test_dataset_iterator_offsets(self, info):
```
parmeet, line 82: Do we need a new unit test or can we modify the existing one [here](https://github.com/pytorch/text/blob/master/test/data/test_builtin_datasets.py#L58)?

salilsdesai: I thought about this too, but I felt it would be it good to not cram many checks into one unit test. If you think it's better to just modify the existing one, let me know.

parmeet: As such after introducing new modifications in raw_datasets.jsonl w.r.t offsets, I do not see any difference between the two unit tests for the part where we validate hash. I think it's ok to keep this part in a separate unit test. If so, we probably want to remove line verification in the other function.

salilsdesai: The line verification in the other function is specifically for the first line though. Once we make the offsets non-zero for the rest of the datasets wouldn't we want to keep the first line verification as is?

parmeet: isn't it testing at offset=0 is same as what's being tested in other function?

salilsdesai: yeah. but when we make the offsets non-zero in the future the test for the first line won't happen anymore if we remove it from the existing test.

parmeet: I'd keep tests for all the datasets/splits with offset=0, and keep adding lines in raws_datasets.jsonl with non-zero offsets for datasets that support offset.

torchtext/datasets/ag_news.py
cpuhrsch: Yes, it's likely to have the same performance, lists might even be faster, but it's unlikely to show up. For now I'd optimize for correctness.

salilsdesai:
> # [Codecov](https://codecov.io/gh/pytorch/text/pull/1338?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pytorch) Report
> > Merging [#1338](https://codecov.io/gh/pytorch/text/pull/1338?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pytorch) ([0e7966b](https://github.com/pytorch/text/commit/0e7966bfd88cf0254bc67b89af71b904c416b158)) into [master](https://codecov.io/gh/pytorch/text/commit/0435df13924fd4582d67e5b17bc09f6ded18be8b?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pytorch) ([0435df1](https://github.com/pytorch/text/commit/0435df13924fd4582d67e5b17bc09f6ded18be8b)) will **decrease** coverage by `0.27%`.
> > The diff coverage is `46.42%`.
> 
> [Impacted file tree graph](https://codecov.io/gh/pytorch/text/pull/1338?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pytorch)
> 
> ```diff
> @@            Coverage Diff             @@
> ##           master    #1338      +/-   ##
> ==========================================
> - Coverage   76.85%   76.58%   -0.28%     
> ==========================================
>   Files          63       63              
>   Lines        3275     3301      +26     
> ==========================================
> + Hits         2517     2528      +11     
> - Misses        758      773      +15     
> ```
> 
> [Impacted Files](https://codecov.io/gh/pytorch/text/pull/1338?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pytorch)	Coverage Δ	
> [torchtext/data/datasets_utils.py](https://codecov.io/gh/pytorch/text/pull/1338/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pytorch#diff-dG9yY2h0ZXh0L2RhdGEvZGF0YXNldHNfdXRpbHMucHk=)	`83.63% <42.30%> (-5.60%)`	⬇️
> [torchtext/datasets/ag_news.py](https://codecov.io/gh/pytorch/text/pull/1338/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pytorch#diff-dG9yY2h0ZXh0L2RhdGFzZXRzL2FnX25ld3MucHk=)	`100.00% <100.00%> (ø)`	
> [Continue to review full report at Codecov](https://codecov.io/gh/pytorch/text/pull/1338?src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pytorch).
> 
> > **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pytorch)
> > `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> > Powered by [Codecov](https://codecov.io/gh/pytorch/text/pull/1338?src=pr&el=footer&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pytorch). Last update [0435df1...0e7966b](https://codecov.io/gh/pytorch/text/pull/1338?src=pr&el=lastupdated&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pytorch). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pytorch).

This is because I used this function to generate the offsets and hardcoded them into the AG NEWS dataset file, but the function isn't actually run otherwise.  Should I remove it from the code, or add a test for it, or do something else so it is used?

joecummings:
Stale.

PR #1329, opened 2021-06-14
remove experimental features from release branch

parmeet:
Removing experimental features from release:
- Deleted experimental folder
- Removed corresponding tests
- updated docs

PR #1339, opened 2021-06-18
fixing benchmark code

parmeet:
Fixing broken benchmark due to removal of experimental datasets (#1329) and renaming experimental vocab to new vocab as we the vocab module is already in beta phase


NicolasHug (approved):
stamping!

PR #1340, opened 2021-06-21
up the priority of numpy array comparisons in self.assertEqual (#59067)

heitorschueroff:
Summary:
Fixes https://github.com/pytorch/pytorch/issues/58988.

Pull Request resolved: https://github.com/pytorch/pytorch/pull/59067

Reviewed By: jbschlosser

Differential Revision: D28986642

Pulled By: heitorschueroff

PR #1341, opened 2021-06-22
Fbsync: up the priority of numpy array comparisons in self.assertEqual

parmeet:
https://github.com/pytorch/text/pull/1340


mthrok (approved):
Looks good!

PR #1354, opened 2021-07-06
Fix IMDB: all samples are positive

tomassosorio:
Currently, if the path of the working space as a 'pos' in its name, e.g. '/Users/user/Documents/re**pos**itories/<something> since repositories contains 'pos' all samples will be positive and some files that are not samples will be considered as samples .e.g 'labeledBow.feat'

By using this approach this will no longer be an issue.


parmeet (approved):
Great, Thank you @tomassosorio for the catch and providing the clear description of the problem!

PR #1361, opened 2021-07-21
[WIP] changing default root for datasets

parmeet:
We would like to modify the default location for datasets downloads. Currently the dataset get downloaded in the working directory which may not always be the desirable behavior. For instance, it causes issue if working directory happens to be within some form source-control, the downloaded files show-up as untracked files which could be annoying.


torchtext/data/datasets_utils.py
```diff
@@ -213,7 +213,7 @@ def _wrap_split_argument_with_fn(fn, splits):
         raise ValueError("Internal Error: Given function {} did not adhere to standard signature.".format(fn))
 
     @functools.wraps(fn)
-    def new_fn(root='.data', split=splits, **kwargs):
+    def new_fn(root=os.path.expanduser('~/.torchtext/cache'), split=splits, **kwargs):
```
mthrok, line 216: Is this BC-breaking change?

parmeet: It is a good question. This is the default path where datasets will be downloaded. In worst case, the data would be downloaded again for users relying on default behavior instead of providing their own root folder. Any think else that could potentially make it BC-breaking?

mthrok: I see. I think it's on a boarder-line but it is okay to it as backward compatible.
Yet, we probably should emphasize in the upcoming release note.

.circleci/config.yml
```diff
@@ -41,7 +41,7 @@ commands:
     steps:
       - run:
           name: Generate CCI cache key
-          command:
+          command: |
```
mthrok, line 44: 🧐 Is this cosmetic or was it broken before?

mthrok (approved):
LGTM.

PR #1374, opened 2021-08-09
Update torchtext.legacy.data doc for 0.9 release

mthrok:
cc @parmeet 

Looks like `experimental`, is not supposed to be in the release, but the code is still there. Can you advise?


parmeet:
> cc @parmeet
> 
> Looks like `experimental`, is not supposed to be in the release, but the code is still there. Can you advise?

Yes, we removed the experimental code from torchtext 0.10.0 release branch. For some reason (or not), it wasn't removed from 0.9.0 release branch. Do we need to update the branch?

mthrok:
> > cc @parmeet
> > Looks like `experimental`, is not supposed to be in the release, but the code is still there. Can you advise?
> 
> Yes, we removed the experimental code from torchtext 0.10.0 release branch. For some reason (or not), it wasn't removed from 0.9.0 release branch. Do we need to update the branch?

I think the release branch code should reflect the release package. If the `experimental` module was included in the release binary, I think it's okay, but otherwise, we should remove it.

cc @msaroufim I think this will apply to LTS branch as well.

parmeet:
> TS branch as well.

Yes, experimental was part of release binary for 0.9.0

PR #1376, opened 2021-08-11
Amazonfull datapipes

HamidShojanazeri:
Migrating AmazonFull to data pipes.


erip:
I think this can be closed with #1499

PR #1383, opened 2021-08-30
Fix examples in get_vecs_by_tokens documentation.

iyedbennour:
Fix examples in `get_vecs_by_tokens` documentation.


parmeet (approved):
Thanks @iyedbennour for fixing the example. Changes LGTM!

PR #1389, opened 2021-09-14
Added PCRE library to torchtext

Nayef211:
## Description
- Replacing [`RE2`](https://github.com/google/re2) with [`PCRE`](https://github.com/vmg/pcre) library in `regex.cpp` to support negative look aheads in regex expressions
- Added new `find_all()` function similar to Python's `re.findall()`
- Added unit tests for new function

## Followup
- Followup discussion needed around [regex library performance](https://rust-leipzig.github.io/regex/2017/03/28/comparison-of-regex-engines/#timings) (PCRE vs re2)

### DIFF
--- a/build_tools/setup_helpers/extension.py
+++ b/build_tools/setup_helpers/extension.py
@@ -103,11 +103,13 @@ def _build_third_party(debug):
     if platform.system() == 'Windows':
         extra_args = [
             '-GNinja',
+            '-DCMAKE_C_FLAGS=-fPIC'
         ]
         build_env.setdefault('CC', 'cl')
         build_env.setdefault('CXX', 'cl')
     else:
-        extra_args = ['-DCMAKE_CXX_FLAGS=-fPIC ' + _get_cxx11_abi(), "-DCMAKE_C_FLAGS=-fPIC"]
+        extra_args = ['-DCMAKE_CXX_FLAGS=-fPIC ' + _get_cxx11_abi(),
+                      '-DCMAKE_C_FLAGS=-fPIC']
     subprocess.run(
         args=[
             'cmake',
--- a/build_tools/setup_helpers/extension.py
+++ b/build_tools/setup_helpers/extension.py
@@ -79,8 +79,9 @@ def _get_libraries():
     return [
         'sentencepiece_train',
         'sentencepiece',
-        're2',
+        'pcrecpp',
         'pcre',
+        're2',
         'double-conversion',
     ]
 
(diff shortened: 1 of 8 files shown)


build_tools/setup_helpers/extension.py
```diff
@@ -80,7 +80,8 @@ def _get_libraries():
         'sentencepiece_train',
         'sentencepiece',
         're2',
-        'double-conversion'
+        'pcre',
```
mthrok, line 83: PCRE generates three different libraries, and this causes `undefined symbol: _ZN7pcrecpp3Arg12parse_stringEPKciPv`.

Add the rest;

```
        'pcre',
	'pcrecpp',
        'pcreposix',
```

but it still cause an issue `undefined symbol: pcre_free`, which I cannot find a definition.

```
$ nm third_party/build/lib/libpcrecpp.a  | grep pcre_free
                 U pcre_free
$ nm third_party/build/lib/libpcre.a  | grep pcre_free
                 U pcre_free
                 U pcre_free
                 U pcre_free
0000000000000950 T pcre_free_substring
0000000000000770 T pcre_free_substring_list
0000000000000010 D pcre_free
                 U pcre_free
0000000000001080 T pcre_free_study

$ nm third_party/build/lib/libpcreposix.a  | grep pcre_free
                 U pcre_free

Comment (mthrok) [reply build_tools/setup_helpers/extension.py:83]: Okay, I think I found a solution.

RE2 has an implementation of `pcre_free` (which is actually an empty function)

https://github.com/google/re2/blob/5aec8b553863350e32d13bbf745eae81adcdce93/util/pcre.cc#L60-L61

so the order has to be

```
pcrecpp
pcre
re2
```

Comment (mthrok) [reply build_tools/setup_helpers/extension.py:83]: You might get an error like

```
/fsx/users/moto/conda/compiler_compat/ld: /fsx/users/moto/text/third_party/build/lib/libpcre.a(pcre_compile.c.o): relocation R_X86_64_PC32 against symbol `_pcre_OP_lengths' can not be used when making a shared object; recompile with -fPIC
```

You need to pass `-fPIC` to pcre compilation. 

```
set(CMAKE_C_FLAGS ${CMAKE_C_FLAGS} -fPIC)
add_subdirectory(pcre)
```

is one way but there should be a better way.

mthrok: Hmm looking into the detail of the RE2 code, this might not be the right way to work around...

https://github.com/google/re2/blob/5aec8b553863350e32d13bbf745eae81adcdce93/util/pcre.cc#L41-L75

abhinavarora:
@Nayef211 Should we just close this PR now that PCRE is not needed?

Nayef211:
> @Nayef211 Should we just close this PR now that PCRE is not needed?

Yup you're right. Closing the PR

PR #1399, opened 2021-09-29
 Replace issue templates with new issue forms

NicolasHug:
This PR introduces the new issue forms that replace issue templates.

This is similar to what was done in torchvision https://github.com/pytorch/vision/pull/4299, you can see the end result here: https://github.com/pytorch/vision/issues/new/choose (click on one of the "Get Started" buttons)

The main new thing is that we can enforce some of the fields to be filled, especially for bug reports. It's also a much cleaner GUI for users IMHO, and we can provide better examples and instructions.

There is still a "blank" template available.

I removed the "Questions" form: we pretend we close these issues but we actually don't. I replaced it with a direct link to https://discuss.pytorch.org. Since we still have a "blank" template, I think this  covers all previous use-cases properly.

We don't need to have the exact same templates across all projects, so **please feel free to directly push to this PR to address your own comments and to tweak the templates as you please** :)


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,60 @@
...
+      # All necessary imports at the beginning
+      import torch
+      import torchvision
+      from torchvision.ops import nms
+
+      # A succinct reproducing example trimmed down to the essential parts:
+      N = 5
+      boxes = torch.rand(N, 4)  # Note: the bug is here, we should enforce that x1 < x2 and y1 < y2!
+      scores = torch.rand(N)
+      nms(boxes, scores, iou_threshold=.9)
```
NicolasHug, line 27: This is still vision-specific, I guess it'd be better to have a torchaudio snippet here. Perhaps @parmeet  will be more inspired than I am for this?

joecummings:
This has already been implemented.

NicolasHug:
I don't think it has been implemented @joecummings .

You can see the difference of what the [current issue template looks like for torchtext](https://github.com/pytorch/text/issues/new?assignees=&labels=&template=bug-report.md)  vs [what it would look like if this PR was merged](https://github.com/pytorch/vision/issues/new?assignees=&labels=&template=bug-report.yml)

Nayef211:
@joecummings reopening the issue as per @NicolasHug's comment

Issue #1401, opened 2021-10-06
ModuleNotFoundError: No module named 'torchtext.experimental' for "examples/BERT/qa_task.py"


batermj:
## 🐛 Bug
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Environment**

Please copy and paste the output from our
[environment collection script](https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py)
(or fill out the checklist below manually).

You can get the script and run it with:
```
wget https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py
# For security purposes, please check the contents of collect_env.py before running it.
python collect_env.py
python -c "import torchtext; print(\"torchtext version is \", torchtext.__version__)"
```

 - PyTorch Version (e.g., 1.0):
 - OS (e.g., Linux):
 - How you installed PyTorch (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version:
 - GPU models and configuration:
 - Any other relevant information:

**Additional context**

parmeet:
Hi @batermj, The experimental package is not part of release binaries. It is available through nightly releases. Also on side note, the examples folder is broken as we have removed some of the experimental functionalities in this PR https://github.com/pytorch/text/pull/1326 on which they depend. We are in process of fixing our examples directory.

parmeet:
closing the issue as BERT example is removed from the repo

PR #1326, opened 2021-06-10
Remove experimental datasets

dgenzel2:
Tested via `pytest -v -s test`

PR #1402, opened 2021-10-06
remove BERT example

parmeet:
The BERT example is based on experimental version of datasets which were removed in this PR (https://github.com/pytorch/text/pull/1326). Hence they are no longer functional and It causes confusion for the users (https://github.com/pytorch/text/issues/1401). 

The fix is non-trivial. Hence we would like to remove this example. Going forward, we will provide complementary examples ones the pre-trained models are added to the repo.


Nayef211 (approved):
Looks good to me. I would just make sure test failures aren't related to these changes before merging.

batermj:
As a result, where is the working BERT example provided in pytorch repo?

PR #1430, opened 2021-11-03
Enable optional return of all states in transformer encoder

parmeet:
Adding option to return all layers or only the last layer in transformer encoder. This would allow downstream models to only request what's necessary out of encoder layer.

### DIFF
--- a/torchtext/models/roberta/modules.py
+++ b/torchtext/models/roberta/modules.py
@@ -1,4 +1,4 @@
-from typing import Optional, List
+from typing import Optional, List, Union
 
 import torch
 from torch import nn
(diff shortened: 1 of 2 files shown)


torchtext/models/roberta/modules.py
```diff
@@ -282,15 +284,25 @@ def forward(self, tokens: torch.Tensor) -> List[torch.Tensor]:
...
 
-        states = [encoded]
+        if self.return_all_layers:
+            states = [encoded]
 
-        for layer in self.layers:
-            encoded = layer(encoded, padding_mask)
-            states.append(encoded)
+            for layer in self.layers:
+                encoded = layer(encoded, padding_mask)
```
ebsmothers, line 291: One nit for readability is to consider moving this for loop outside the if statements, since it is the same in both cases. Either way is fine though :)

ebsmothers (approved):
Thanks so much @parmeet! The changes look good to me

torchtext/models/roberta/modules.py
parmeet: Actually we are also storing all the layer states when return_all_layers is set True. If for loop is moved outside, then I would need to bring the if loop inside the for loop :).

PR #1432, opened 2021-11-05
Allow inferred scaling in MultiheadSelfAttention for head_dim != 64

ebsmothers:
Rather than raise an exception whenever head_dim != 64, we can just infer the scaling value and continue to provide a warning.

Also add an assertion in case embed_dim is not a multiple of num_heads (in which case forward will break).


parmeet (approved):
LGTM! Thanks.

PR #1435, opened 2021-11-08
add attention mask to transformer encoder modules

parmeet:
Add Attention mask to transform encoder modules. The implementation is based on [Encoder layer from fairseq](https://github.com/pytorch/fairseq/blob/0b21875e45f332bedbcc0617dcf9379d3c03855f/fairseq/modules/transformer_layer.py#L20) 

Follow-up:

- [ ] Add Documentation
- [x] Add unit-test

### DIFF
--- a/torchtext/models/roberta/modules.py
+++ b/torchtext/models/roberta/modules.py
@@ -224,13 +224,13 @@ class TransformerEncoderLayer(Module):
 
         if self.normalize_before:
             x = self.attention_layer_norm(input)
-            attention = self.attention(x, key_padding_mask)
+            attention = self.attention(x, key_padding_mask, attn_mask)
             attention = self.dropout(attention)
             biased_input = input + attention
             x = self.final_layer_norm(biased_input)
             return self.residual_mlp(x) + biased_input
         else:
-            attention = self.attention(input, key_padding_mask)
+            attention = self.attention(input, key_padding_mask, attn_mask)
             attention = self.dropout(attention)
             biased_input = input + attention
             biased_input = self.attention_layer_norm(biased_input)
--- a/torchtext/models/roberta/modules.py
+++ b/torchtext/models/roberta/modules.py
@@ -99,7 +99,7 @@ class MultiheadSelfAttention(Module):
         self.input_projection = nn.Linear(embed_dim, 3 * embed_dim)
         self.output_projection = nn.Linear(embed_dim, embed_dim)
 
-    def forward(self, query, key_padding_mask):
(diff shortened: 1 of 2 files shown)


torchtext/models/roberta/modules.py
```diff
@@ -124,6 +124,10 @@ def forward(self, query, key_padding_mask):
         )
 
         attn_weights = torch.bmm(q, k.transpose(1, 2))
+        if attn_mask is not None:
+            torch._assert(attn_mask.dim() == 2, "Expected attn_mask of dim 2 but got {}".format(attn_mask.dim()))
```
ebsmothers, line 128: At this point I wonder if it makes sense to explicitly check that the sizes of attn_mask and attn_weights match exactly (or right after attn_mask.unsqueeze)

torchtext/models/roberta/modules.py
```diff
@@ -209,18 +213,27 @@ def __init__(
...
 
-    def forward(self, input, key_padding_mask):
+    def forward(self, input: torch.Tensor, key_padding_mask: torch.Tensor, attn_mask: Optional[torch.Tensor] = None):
+        if attn_mask is not None:
+            torch._assert(attn_mask.dtype == torch.bool, "Expected attn_mask dtype as `torch.bool` but got {}".format(attn_mask.dtype))
+            torch._assert(attn_mask.dim() == 2, "Expected attn_mask of dim 2 but got {}".format(attn_mask.dim()))
+            attn_mask = attn_mask.masked_fill(
+                attn_mask.to(torch.bool),
+                -1e8 if input.dtype == torch.float32 else -1e4
+            )
```
ebsmothers, line 223: Any particular reason for performing the masked fill in the layer forward as opposed to the encoder or self-attention forwards? (I am not advocating for one in particular over the others, mostly I am just curious)

torchtext/models/roberta/modules.py
parmeet: yupp make sense. Added the checks!

torchtext/models/roberta/modules.py
parmeet: It's a good question. 

Regarding Encoder v/s Encoder layer: I think Encoder is client of Encoder layer. We could potentially user Encoder layer in other places too. So this choice is mainly about code re-usabilty.

Regarding Encoder layer v/s Multi head self attention: I think the choice here is less obvious. I think both are viable candidates to do this conversion. So based on your question, I actually made the conversion on both the modules and added support both for floating point and boolean mask. Let me know if this look better now.

ebsmothers (approved):
Looks great, thanks for implementing this! (And for adding the unit test :) )

PR #1436, opened 2021-11-10
[Vocab] Refactor vocab factory method to accept special tokens as a keyword argument

abhinavarora:
This PR changes the signature of the `vocab` method from  `vocab_factory.py` to also pass in special characters to the vocab. Prior to this PR, the expectation was that the users are responsible for adding the special characters to the `OrderedDict`. However, this is a pattern not familiar to NLP library users. 

Local Testing Instructions

`pytest -v -s test`

### DIFF
--- a/torchtext/vocab/vocab_factory.py
+++ b/torchtext/vocab/vocab_factory.py
@@ -6,7 +6,9 @@ from torchtext._torchtext import (
 )
 
 
-def vocab(ordered_dict: Dict, min_freq: int = 1) -> Vocab:
+def vocab(ordered_dict: Dict, min_freq: int = 1,
+          specials: Optional[List[str]] = None,
+          special_first: bool = True) -> Vocab:
     r"""Factory method for creating a vocab object which maps tokens to indices.
 
     Note that the ordering in which key value pairs were inserted in the `ordered_dict` will be respected when building the vocab.
@@ -15,6 +17,8 @@ def vocab(ordered_dict: Dict, min_freq: int = 1) -> Vocab:
     Args:
         ordered_dict: Ordered Dictionary mapping tokens to their corresponding occurance frequencies.
         min_freq: The minimum frequency needed to include a token in the vocabulary.
+        specials: Special symbols to add. The order of supplied tokens will be preserved.
+        special_first: Indicates whether to insert symbols at the beginning or at the end.
 
     Returns:
         torchtext.vocab.Vocab: A `Vocab` object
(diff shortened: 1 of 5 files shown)


torchtext/vocab/vocab_factory.py
```diff
@@ -6,7 +6,7 @@
 )
 
 
-def vocab(ordered_dict: Dict, min_freq: int = 1) -> Vocab:
+def vocab(ordered_dict: Dict, min_freq: int = 1, specials: List[str] = [], special_first: bool = True) -> Vocab:
```
parmeet, line 9: I think we may want to use None as default value of mutable arguments per the [docs](https://docs.python.org/3/reference/compound_stmts.html#function-definitions). So instead we could do:
```
specials: Optional[List[str]] = None
```

torchtext/vocab/vocab_factory.py
```diff
@@ -34,19 +36,27 @@ def vocab(ordered_dict: Dict, min_freq: int = 1) -> Vocab:
         >>> unk_token = '<unk>'
         >>> default_index = -1
```
parmeet, line 37: we could move this close to V3 since it is no longer used in V2

torchtext/vocab/vocab_factory.py
```diff
@@ -34,19 +36,27 @@ def vocab(ordered_dict: Dict, min_freq: int = 1) -> Vocab:
...
+    for token in specials:
+        ordered_dict.pop(token, None)
 
     tokens = []
     for token, freq in ordered_dict.items():
         if freq >= min_freq:
             tokens.append(token)
 
+    if special_first:
+        tokens[0:0] = specials
```
Nayef211, line 56: Just out of curiosity, what does `tokens[0:0]` do? I've never seen this syntax before, but I'm assuming this is prepending `specials` to the `tokens` list. Time-complexity wise, is there any benefit to doing this as opposed to list concatenation i.e. `tokens = specials + tokens`?

torchtext/vocab/vocab_factory.py
abhinavarora: Sounds reasonable. I'll refactor this!

torchtext/vocab/vocab_factory.py
abhinavarora: You are right that this syntax is used to prepend specials to the tokens. Time complexity wise there is no difference. However, space-complexity wise this is better because it prepends in-place while list concatenation creates a new list.

torchtext/vocab/vocab_factory.py
abhinavarora: Done!

torchtext/vocab/vocab_factory.py
abhinavarora: Done. Saw that v2 is not needed. So made v3 as v2

parmeet (approved):
Thanks! LGTM.

PR #1437, opened 2021-11-10
Remove TorchText legacy folder and its associated tests

abhinavarora:
This PR does the following:

1. Removes legacy folder from TorchText
2. Removes legacy tests
3. Removes legacy migration notebooks
4. Refactors `test_build.py` to use new Vectors class instead of legacy vocab

**Testing Instruction**

`pytest -v -s test`


parmeet (approved):
This is great! Thank you for cleaning things up, LGTM!

JaniC-WPI:
Helloo, I have been trying to work with Field object from a pytorch book, and as I can see wigth the removal of legacy most of the objects are also removed. Is there any documentation for the way of replacing Field object. Please help.

Nayef211:
> Helloo, I have been trying to work with Field object from a pytorch book, and as I can see wigth the removal of legacy most of the objects are also removed. Is there any documentation for the way of replacing Field object. Please help.

Hi @JaniC-WPI. If you would like access to Field object you can downgrade your version of torchtext before these files were deleted. We no longer maintain legacy classes in torchtext!

PR #1439, opened 2021-11-11
fix attention mask testing

parmeet:
The non-zero output values are likely affected by platform. To make the test safer, we only like to ensure the output is all zeros for first token.

Follow-up:

- [ ] test for non-zero tokens


Nayef211 (approved):
LGTM. Do we have plans on testing for the non-zero outputs in the future?

parmeet:
> LGTM. Do we have plans on testing for the non-zero outputs in the future?

certainly, we should try to make the test more robust.

PR #1441, opened 2021-11-12
Fix CircleCI failures on Windows for XLM-R unit tests

abhinavarora:
This PR fixes the Circle CI model test failures on Windows.

## Cause
 
To generate download URLs we used `os.path.join(_TEXT_BUCKET, "xlmr.vocab.pt")`. `os.path.join` is not the right way to join URL paths and must only be used for local OS directory paths. In our case, since `_TEXT_BUCKET` has forward slashes (because it is a URL), `os` library acts smart and on Windows it produces the URL: `https://download.pytorch.org/models/text\xlmr.vocab.pt` which gives a 403 on CURL.

## Fix
Instead of `os.path.join`, we must use `urllib.parse.urljoin`

### DIFF
--- a/torchtext/__init__.py
+++ b/torchtext/__init__.py
@@ -1,5 +1,5 @@
 import os
-_TEXT_BUCKET = 'https://download.pytorch.org/models/text'
+_TEXT_BUCKET = 'https://download.pytorch.org/models/text/'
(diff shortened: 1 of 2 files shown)


parmeet (approved):
LGTM! thanks for fixing this :)

PR #1442, opened 2021-11-12
Add a class method in Model Bundler to facilitate model creation with user-defined configuration and checkpoint

parmeet:
This PR attempts to propose a simplified mechanism for users to instantiate models with user-defined configuration and checkpoints.
 
Up-till now, users can do so with following:
```python
# Alternative 1: By creating user-specific bundle
from torchtext.models import RobertaEncoderConf, RobertaModelBundle
model_weights_path = "https://download.pytorch.org/models/text/xlmr.base.encoder.pt"
roberta_encoder_conf = RobertaEncoderConf(vocab_size=250002)
roberta_bundle = RobertaModelBundle(_encoder_conf=roberta_encoder_conf, _path=model_weights_path)
encoder = roberta_bundle.get_model()
 
# Alternative 2: By instantiating model and explicitly loading the checkpoint
from torchtext.models import RobertaEncoderConf, RobertaModel
roberta_encoder_conf = RobertaEncoderConf(vocab_size=250002)
encoder = RobertaModel(config=roberta_encoder_conf)
model_weights_path = "https://download.pytorch.org/models/text/xlmr.base.encoder.pt"
from torch.hub import load_state_dict_from_url
state_dict = load_state_dict_from_url(model_weights_path)
encoder.load_state_dict(state_dict)
```
 
With proposed API, we simplify the process:
1) For alternative 1: Avoiding to explicitly create a bundle just to eventually call the get_model method to get the desired model.
2) For alternative 2: Un-burdening the user from explicitly loading the checkpoint from user-specified path and assigning the weights to the model
 
The workflow would now be as follows:
```python
from torchtext.models import RobertaEncoderConf, RobertaModelBundle
model_weights_path = "https://download.pytorch.org/models/text/xlmr.base.encoder.pt"
roberta_encoder_conf = RobertaEncoderConf(vocab_size=250002)
encoder = RobertaModelBundle.from_config(config=roberta_encoder_conf, checkpoint=model_weights_path)
# create classifier
from torchtext.models import RobertaClassificationHead
classifier_head = RobertaClassificationHead(num_classes=2, input_dim=768)
classifier = RobertaModelBundle.from_config(encoder_conf=roberta_encoder_conf, head=classifier_head, checkpoint=model_weights_path)
```
 
### Behavior around checkpoint and head
 
The `from_config` API accepts (along with other arguments):
1)  encoder configuration (via RobertaEncoderConf dataclass)
2) head module
3) checkpoint
4) Whether or not to override checkpoint head parameters (if available) when head module is given
 
The checkpoint may or may not contain the head parameters. The default behavior is that head parameters are taken from the checkpoint, unless the user explicitly asks for overriding the parameters with the user provided `head` module.  If the checkpoint only contains encoder parameters, the head module preserves its state. 
 
## Additional Notes
This PR also updates the `get_model` function. Note that in this case, when the user provides `head`, it will replace the default head of the model. 

### Why the behavior around `head` is different for `get_model` and `from_config`:
The reason is driven based on the use case. The `from_config` class method provides a functional API for model construction. The use-case it supports is to provide a mechanism for model compositionality i.e composing encoder with task specific head. In that case, it is expected that the user would provide head to create a model and not to replace something from the provided checkpoint.
 
On the other hand, the `get_model` API is called after the RobertaModelBundler class is created with all the necessary ingredients (including head if any). Calling `get_model` would simply return the model based on the constructed class. But if the user explicitly provides `head` then the intention here would be to replace any head already present in the class.


test/models/test_models.py
```diff
@@ -91,3 +93,15 @@ def test_xlmr_transform_jit(self):
         actual = transform_jit([test_text])
         expected = [[0, 43523, 52005, 3647, 13293, 113307, 40514, 2]]
         torch.testing.assert_close(actual, expected)
+
+    def test_roberta_bundler_from_config(self):
+        from torchtext.models import RobertaEncoderConf
+        asset_name = "xlmr.base.output.pt"
+        asset_path = get_asset_path(asset_name)
+        model_path = os.path.join(_TEXT_BUCKET, "xlmr.base.encoder.pt")
```
abhinavarora, line 101: Let's use `urllib.parse.urljoin` similar to #1441

abhinavarora (changes requested):
Lets use `urllib.parse.urljoin` instead of `os.path.join`

abhinavarora (approved):

torchtext/models/roberta/bundler.py
```diff
@@ -97,6 +105,40 @@ def get_model(self, head: Optional[Module] = None, load_weights: bool = True, fr
...
+        if checkpoint is not None:
+            if torch.jit.isinstance(checkpoint, Dict[str, torch.Tensor]):
+                state_dict = checkpoint
+            elif isinstance(checkpoint, str):
+                dl_kwargs = {} if dl_kwargs is None else dl_kwargs
+                state_dict = load_state_dict_from_url(checkpoint, **dl_kwargs)
+            else:
+                raise TypeError("checkpoint must be of type `str` or `Dict[str, torch.Tensor]` but got {}".format(type(checkpoint)))
+            if head is not None:
+                model.load_state_dict(state_dict, strict=False)
```
mthrok, line 137: I feel like there is an ambiguity in the logic flow here.
Is checkpoint expected to contain parameter for head or does not it matter?

mthrok:
> The workflow would now be as follows:
> 
> ```python
> from torchtext.models import RobertaEncoderConf, RobertaModelBundle
> model_weights_path = "https://download.pytorch.org/models/text/xlmr.base.encoder.pt"
> roberta_encoder_conf = RobertaEncoderConf(vocab_size=250002)
> encoder = RobertaModelBundle.from_config(config=roberta_encoder_conf, checkpoint=model_weights_path)
> from torchtext.models import RobertaClassificationHead
> classifier_head = RobertaClassificationHead(num_classes=2, input_dim=768)
> classifier = RobertaModelBundle.from_config(config=roberta_encoder_conf, head=classifier_head, checkpoint=model_weights_path)
> ```

In the second use case, say I train classifier, how do I save the trained classifier and later load from different session?

torchtext/models/roberta/bundler.py
parmeet: Related thread https://github.com/pytorch/text/pull/1424#discussion_r736932480

The use-case I am interested in is:
1) User can optionally start from pre-trained encoder weights provided by the the library (or otherwise)
2) User can optionally start from checkpoint saved by the user
or
3) User can just start from scratch (although in this case Alternative 2 in summary above is also equivalent)

The checkpoint could optionally contain the weights of head (for example to cover point 2 above), or only contain the encoder weights (for example to cover point 1 above). 

Now if head is not None, and we are in scenario 1, then setting strict=True will cause key error. I wonder if there are better ways to deal with this? (Perhaps augment the state_dict with head state_dict if head state_dict is not present and then always set strict=True?) 


BTW, I understand that this is bit unconventional the way we treat the model and check-points in the context of pre-trained models (that typically have one-2-one relationship with model and their checkpoints). I am open to other suggestions to cover the use-case above, but I felt that putting such an API in bundler is potentially a safer choice than creating explicit builder functions to deal with such compositional requirements.

torchtext/models/roberta/bundler.py
```diff
@@ -97,6 +105,45 @@ def get_model(self, head: Optional[Module] = None, load_weights: bool = True, fr
...
+                dl_kwargs = {} if dl_kwargs is None else dl_kwargs
+                state_dict = load_state_dict_from_url(checkpoint, **dl_kwargs)
+            else:
+                raise TypeError("checkpoint must be of type `str` or `Dict[str, torch.Tensor]` but got {}".format(type(checkpoint)))
+            if head is not None:
+                regex = re.compile(r"^head\.")
+                head_state_dict = {k: v for k, v in model.state_dict().items() if regex.findall(k)}
+                # if not all the head keys are present in checkpoint then we shall update the state_dict with the provided head state_dict
+                if not all(key in state_dict.keys() for key in head_state_dict.keys()):
+                    state_dict.update(head_state_dict)
```
parmeet, line 141: @mthrok I have updated the logic how to handle the case when checkpoint contains the head parameters or not.

Case1: Does not contains head parameters: In this case, we would update the state with supplied head state dict. This would mean that we are keeping intact the head module whatever is provided by the user. This is inline with what we discussed here https://github.com/pytorch/text/pull/1424#discussion_r737018029

Case2: Does contain head parameters: In this case we would use the checkpoint as such.

Note that in both the cases, now we have made `strict=True` to ensure all key matches. Let me know if this sounds reasonable to contain various edge cases?

mthrok:
Okay, I gave some thoughts about it.
I think treating the head argument as overwrite makes things simpler.

So that the code logic I suggest is
1. that model is built according to the provided config.
2. If checkpoint is provided, parameters are loaded
3. If head is provided, head is replaced. (parameters loaded to head in step2 is discarded.)

The following table summarizes what the resulting model looks like and I believe the use cases you are describing are still covered.

> 1. User can optionally start from pre-trained encoder weights provided by the the library (or otherwise)
> 2. User can optionally start from checkpoint saved by the user
> 3. User can just start from scratch (although in this case Alternative 2 in summary above is also equivalent)

and now that I think of it, in any use cases, at the end of the training, newly saved parameter files would contain head. (because I think users would simply do `torch.save(model.state_dic())`, which is safer than separately saving head and the rest.) This assumption can merge the use cases of library-provided model and user-provided models, aren't they?
And, the code should be able to assume that head is always contained in parameter files.

<table class="tg">
<thead>
  <tr>
    <th class="tg-0pky">Pattern</th>
    <th class="tg-0pky">Head</th>
    <th class="tg-0pky">Checkpoint</th>
    <th class="tg-0pky">Returned Model</th>
    <th class="tg-0pky">Note</th>
    <th class="tg-0pky"></th>
  </tr>
</thead>
<tbody>
  <tr>
    <td class="tg-0pky">i</td>
    <td class="tg-0pky">None</td>
    <td class="tg-0pky">None</td>
    <td class="tg-0pky">Untrained</td>
    <td class="tg-0pky">- Usage full scratch training</td>
    <td class="tg-0pky"></td>
  </tr>
  <tr>
    <td class="tg-0pky">ii</td>
    <td class="tg-0pky">Given (untrained)</td>
    <td class="tg-0pky">None</td>
    <td class="tg-0pky">Untrained</td>
    <td class="tg-0pky">- Practically same as pattern (i).</td>
    <td class="tg-0pky"></td>
  </tr>
  <tr>
    <td class="tg-0pky">iii</td>
    <td class="tg-0pky">Given (trained)</td>
    <td class="tg-0pky">None</td>
    <td class="tg-0pky">Only head is trained</td>
    <td class="tg-0pky">- Practically same as (ii) (so as (i))<br>- not sure if there is such use case</td>
    <td class="tg-0pky"></td>
  </tr>
  <tr>
    <td class="tg-0pky">iv</td>
    <td class="tg-0pky">None</td>
    <td class="tg-0pky">Given</td>
    <td class="tg-0pky">Trained head + body (both from checkpoint)</td>
    <td class="tg-0pky">- Config and checkpoint have to match.<br>- Usage:<br>&nbsp;&nbsp;- inference with library-provided models<br>&nbsp;&nbsp;- inference with user-trained models (partially or full-scratch)</td>
    <td class="tg-0pky"></td>
  </tr>
  <tr>
    <td class="tg-0pky">v</td>
    <td class="tg-0pky">Given (untrained)</td>
    <td class="tg-0pky">Given</td>
    <td class="tg-0pky">Untrained head + trained body (from checkpoint)</td>
    <td class="tg-0pky">- Usage: transfer learning</td>
    <td class="tg-0pky"></td>
  </tr>
  <tr>
    <td class="tg-0pky">vi</td>
    <td class="tg-0pky">Given (trained)</td>
    <td class="tg-0pky">Given</td>
    <td class="tg-0pky">Trained head + trained body (from checkpoint)</td>
    <td class="tg-0pky">- Usage: quick model surgery? <br>- (not sure if there is such use case)</td>
    <td class="tg-0pky"></td>
  </tr>
</tbody>
</table>

parmeet:
Perfect! Thanks for capturing everything (and more :) ) in the table what I originally intent to cover.

Just one change, I think head need to be given by the user in Patter iV as well (detail in answer below for point 1)


> So that the code logic I suggest is
> 
> 1. that model is built according to the provided config.

Unfortunately, we could only afford to prepare encoder through config. The reason for this is we do not know in advance what head implementation user want to attach. The head may not even be available as part of library, so it is may not be feasible to combine complete model configuration in advance. Which is the reason, I am proposing composable API where-in head is explicitly provided by the user that they wanted to attach.  Note that we are passing this burden to user, only when they want to use our APIs for their own specific needs (encoder+head combo and checkpoints). They do not have to worry for any of these when they are going to work with our pre-trained bundler objects which will come with pre-trained heads (if any) like we have for [ASR models](https://github.com/pytorch/audio/blob/0cd25093626d067e008e1f81ad76e072bd4a1edd/torchaudio/pipelines/_wav2vec2/impl.py#L214)  and shall use get_model API to get the model.

> 2. If checkpoint is provided, parameters are loaded
> 3. If head is provided, head is replaced. (parameters loaded to head in step2 is discarded.)

I think this valid, when user uses the get_model API (for eg: we provide pre-trained classification model, and user called get_model with their own head, this would lead to head replacement). But for from_config API, where we expect the head to be explicitly provided by the user, we should not discard the head from checkpoint to support pattern iV.
 
> and now that I think of it, in any use cases, at the end of the training, newly saved parameter files would contain head. (because I think users would simply do `torch.save(model.state_dic())`,

Yes, this we can be safely assumed to always be the case.
 
> This assumption can merge the use cases of library-provided model and user-provided models, aren't they? 

Correct!

> And, the code should be able to assume that head is always contained in parameter files.

Well, not always :). For transfer learning cases (Pattern V), which only have weights for the body.

PR #1449, opened 2021-11-23
add unit tests for testing model training

parmeet:
Add Unit tests to test model training:
1. when encoder is not freezed
2. when encoder is freezed

Follow-up:

- [ ] Organize current model unit tests into integration/smoke tests, since they use multiple components from the library.

### DIFF
--- a/test/models/test_models.py
+++ b/test/models/test_models.py
@@ -1,5 +1,7 @@
 import torchtext
 import torch
+from torch.nn import functional as torch_F
+import copy
 from ..common.torchtext_test_case import TorchtextTestCase
 from ..common.assets import get_asset_path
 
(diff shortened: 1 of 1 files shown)


test/models/test_models.py
```diff
@@ -126,3 +128,49 @@ def test_roberta_bundler_from_config(self):
             encoder_state_dict['encoder.' + k] = v
         model = torchtext.models.RobertaModelBundle.from_config(encoder_conf=dummy_encoder_conf, head=dummy_classifier_head, checkpoint=encoder_state_dict)
         self.assertEqual(model.state_dict(), dummy_classifier.state_dict())
+
+    def test_roberta_bundler_train(self):
```
mthrok, line 132: Since this is a rather complicated for unit test, can you add doc string?
There are many style for writing docstring for tests, but one should be able to describe a test like "Given A, B should be met."

If the purpose of the test is just ensuring no error happens, then I call that kind of tests as "smoke test". (Could be different from what the "smoke test" means in industry though)

test/models/test_models.py
```diff
@@ -126,3 +128,49 @@ def test_roberta_bundler_from_config(self):
...
+
+        encoder_current_state_dict = copy.deepcopy(model.encoder.state_dict())
+        head_current_state_dict = copy.deepcopy(model.head.state_dict())
+
+        _train(model)
+
+        self.assertNotEqual(model.encoder.state_dict(), encoder_current_state_dict)
+        self.assertNotEqual(model.head.state_dict(), head_current_state_dict)
+
+        # freeze encoder
```
mthrok, line 162: I would split the test into two separate tests.

mthrok (approved):
The test looks good in the sense it does not look suspicious or strange.
The logic of the test could be improved depending on how much you want to pursue.
(But it is always arguable what is the correct model. So I won't argue here.)

nit: IMHO, this test utilizes multiple features from the library, so I would call it integration test and place it separately from unit test. You might want to consider that in the future.

FYI: we put tests that use multiple modules from library, or external resource in integration tests directory. https://github.com/pytorch/audio/tree/main/test

test/models/test_models.py
parmeet: Thanks @mthrok, I agree on all the points mentioned. This would go in integration/smoke tests. Let me organize all the model tests in follow-up PR. There are other model test (not introduced in this PR) too that accordingly need to be separated.

PR #1451, opened 2021-11-24
add projection layer to roberta encoder

parmeet:
User may want to additionally project the features from encoder. This PR add support for projecting features to different dimensional space.

### DIFF
--- a/torchtext/models/roberta/model.py
+++ b/torchtext/models/roberta/model.py
@@ -10,6 +10,7 @@ import torch.nn as nn
 
 from .modules import (
     TransformerEncoder,
+    ProjectionLayer,
 )
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


Nayef211:
Overall LGTM. It seems like we're getting a lot of code coverage warnings. Should we also add unit tests for the `ProjectionLayer` class?

parmeet:
> Overall LGTM. It seems like we're getting a lot of code coverage warnings. Should we also add unit tests for the `ProjectionLayer` class?

Yes, I think it's a good idea. The models tests didn't really touch this part of the code since we are not projecting the encoder output for standard architectures. Though, it would be more like integration/smoke tests instead of unit-test. We also do not have unit tests for other modules as well, but they are covered well through our end-2-end model tests (we call them unit-tests for now, but as @mthrok  commented in [here](https://github.com/pytorch/text/pull/1449#pullrequestreview-813170457), we are going to move them to integration tests as well).

Nayef211 (approved):

PR #1452, opened 2021-11-28
change from_config to build_model and other minor clean-ups

parmeet:
This PR changes the Bundle API `from_config` class method name to `build_model` to better reflect the job of this method. Additionally, `strict` option is exposed to the user in order to work with under(over)-specified checkpoint keys.
 
Other changes (mostly clean-ups):
- Removed the `from_config` class method from the `RobertaEncoder` class. This function doesn't add any value.
- Removed private builder function `_get_model`. This is not really needed since the `build_model` function is doing what's required to build the model.
- Added the `freeze` option to `RobertaEncoder` class. This enables freezing encoder parameters directly inside the class instead of doing it explicitly inside `RobertaModel`.
- Renaming: `override_head` -> `override_checkpoint_head` for `build_model` method
- Renaming: `mask` -> `masked_tokens` for model forward function
- Improved doc-string examples

### DIFF
--- a/test/models/test_models.py
+++ b/test/models/test_models.py
@@ -93,13 +93,13 @@ class TestModels(TorchtextTestCase):
         expected = [[0, 43523, 52005, 3647, 13293, 113307, 40514, 2]]
         torch.testing.assert_close(actual, expected)
 
-    def test_roberta_bundler_from_config(self):
+    def test_roberta_bundler_build_model(self):
         from torchtext.models import RobertaEncoderConf, RobertaClassificationHead, RobertaModel, RobertaModelBundle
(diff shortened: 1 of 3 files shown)


mthrok (approved):
Quickly skimmed and things look okay. [stamp]

It would be easier to review if the cleanup is separated. https://google.github.io/eng-practices/review/developer/small-cls.html#refactoring

parmeet:
> It would be easier to review if the cleanup is separated. https://google.github.io/eng-practices/review/developer/small-cls.html#refactoring

Thanks @mthrok for quick review and sharing best-practices on code cleanups, certainly a valuable advice :).

PR #1453, opened 2021-12-04
Update annotation types of transforms and functionals

parmeet:
## Summary:
* This PR updates the annotation types of transforms to `Any`
* Add `Truncate` transform
 
## Problem when transforms input and outputs are type annotated
The transforms and functionals support multiple styles of inputs (both batched and non-batched are supported) as well as sometimes they may support multiple types (refer to Truncate transform added in this PR that can work with both `int` and `str` primitive types). Composability of these transforms and support for torch scriptability may not go hand-in-hand. Below code snippets shows couple of these issues.
 
 ```python
import torch
from typing import List, Union

#primitive transform
def foo(input: Union[int, List[int]])->Union[int, List[int]]:
    if torch.jit.isinstance(input, List[int]):
        return input
    elif torch.jit.isinstance(input, int):
       return input
    else:
        raise TypeError

#primitive transform
def bar(input: List[int])-> List[int]:
    if torch.jit.isinstance(input, List[int]):    
        return input
    else:
        raise TypeError

#composite transform
def goo(input: List[int]):
    x = bar(foo(input))
    return x

goo_jit = torch.jit.script(goo)

# gives following error:
RuntimeError: 

bar(int[] input) -> (int[]):
Expected a value of type 'int' for argument '<varargs>' but instead found type 'Union[List[int], int]'.
:
  File "<ipython-input-17-9a79ce8ea4b4>", line 13
def goo(input: List[int]):
    x = bar(foo(input))
        ~~~ <--- HERE
    return x


def foofoo(input: List[int], use_foo:bool = True):
    output = bar(input)
    if use_foo:
        output = foo(output)
 
    return output

foofoo_jit = torch.jit.script(foofoo)

#gives following error:
RuntimeError: 
Variable 'output' previously had type List[int] but is now being assigned to a value of type Union[List[int], int]
:
  File "<ipython-input-18-10b17e1c770a>", line 4
    output = bar(input)
    if use_foo:
        output = foo(output)
        ~~~~~~ <--- HERE
 
    return output
```
 
## Solution
By changing the input and output type to `Any` and using [type refinement (using torch.jit.instance)](https://pytorch.org/docs/stable/generated/torch.jit.isinstance.html) in primitive transforms, the above problem can be alleviated. As a nice side-benefit, this would force the developer to always check the input types while implementing primitive transforms. 

```python
import torch
from typing import Any, List

#primitive transform
def foo(input: Any)->Any:
    if torch.jit.isinstance(input, List[int]):
        return input
    elif torch.jit.isinstance(input, int):
       return input
    else:
        raise TypeError

#primitive transform
def bar(input: Any)->Any:
    if torch.jit.isinstance(input, List[int]):    
        return input
    else:
        raise TypeError

#composite transform
def goo(input: List[int]):
    x = bar(foo(input))
    return x

goo_jit = torch.jit.script(goo)

#composite transform
def foofoo(input: List[int], use_foo:bool = True):
    output = bar(input)
    if use_foo:
        output = foo(output)
 
    return output

foofoo_jit = torch.jit.script(foofoo)
```
 
## Other concerns
* As we remove the type annotations, we would add the doc-strings to ensure that the user is aware of supported input types.

### DIFF
--- a/docs/source/transforms.rst
+++ b/docs/source/transforms.rst
@@ -37,3 +37,10 @@ LabelToIndex
 .. autoclass:: LabelToIndex
 
    .. automethod:: forward
+
+Truncate
+------------
+
+.. autoclass:: Truncate
+
+   .. automethod:: forward
--- a/docs/source/transforms.rst
+++ b/docs/source/transforms.rst
@@ -37,10 +37,3 @@ LabelToIndex
 .. autoclass:: LabelToIndex
 
    .. automethod:: forward
-
-Truncate
-------------
-
-.. autoclass:: Truncate
-
-   .. automethod:: forward
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -14,6 +14,7 @@ class TestTransforms(TorchtextTestCase):
         transform = transforms.SentencePieceTokenizer(asset_path)
         if test_scripting:
             transform = torch.jit.script(transform)
+
         actual = transform(["Hello World!, how are you?"])
         expected = [['▁Hello', '▁World', '!', ',', '▁how', '▁are', '▁you', '?']]
         self.assertEqual(actual, expected)
@@ -23,10 +24,11 @@ class TestTransforms(TorchtextTestCase):
         self.assertEqual(actual, expected)
 
     def test_spmtokenizer(self):
-
+        """test tokenization on single sentence input as well as batch on sentences"""
         self._spmtokenizer(test_scripting=False)
 
     def test_spmtokenizer_jit(self):
+        """test tokenization with scripting on single sentence input as well as batch on sentences"""
         self._spmtokenizer(test_scripting=True)
 
     def _vocab_transform(self, test_scripting):
@@ -43,9 +45,11 @@ class TestTransforms(TorchtextTestCase):
         self.assertEqual(actual, expected)
 
     def test_vocab_transform(self):
+        """test token to indices on both sequence of input tokens as well as batch of sequence"""
         self._vocab_transform(test_scripting=False)
 
     def test_vocab_transform_jit(self):
+        """test token to indices with scripting on both sequence of input tokens as well as batch of sequence"""
         self._vocab_transform(test_scripting=True)
 
     def _totensor(self, test_scripting):
@@ -65,9 +69,11 @@ class TestTransforms(TorchtextTestCase):
         torch.testing.assert_close(actual, expected)
 
     def test_totensor(self):
+        """test tensorization on both single sequence and batch of sequence"""
         self._totensor(test_scripting=False)
 
     def test_totensor_jit(self):
+        """test tensorization with scripting on both single sequence and batch of sequence"""
         self._totensor(test_scripting=True)
 
     def _labeltoindex(self, test_scripting):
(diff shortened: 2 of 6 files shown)


docs/source/transforms.rst
```diff
@@ -37,3 +37,10 @@ LabelToIndex
 .. autoclass:: LabelToIndex
 
    .. automethod:: forward
+
+Truncate
+------------
```
mthrok, line 42:
```suggestion
--------
```

ReST expects the length of line to match the title.

torchtext/models/roberta/transforms.py
```diff
@@ -32,7 +32,7 @@ def __init__(
         self.sep_token = sep_token
         self.max_seq_len = max_seq_len
 
-        self.token_transform = transforms.SentencePieceTokenizer(spm_model_path)
+        self.tokenizer = transforms.SentencePieceTokenizer(spm_model_path)
```
mthrok, line 35: Technically speaking, I think, this is BC-breaking change. The `dict_state` dumped from the previous version will stop working. Even if the BC-breaking is not an issue, as a good practice I recommend to split name change into a different PR, considering the possibility such as cherry-picking at minor release or something.

torchtext/transforms.py
```diff
@@ -133,12 +138,33 @@ def __init__(
...
         else:
-            return self._label_vocab.__getitem__(labels)
+            raise TypeError("Input type not supported")
 
     @property
     def label_names(self) -> List[str]:
         return self._label_names
+
+
+class Truncate(Module):
```
mthrok, line 154: The implementation LGTM. 

However, since this new transform is not used by the other change, I think that the addition of `Truncate` can be in a separate, self-contained PR. Splitting it will give better UX when users reading release note and looking into the commit.

test/test_transforms.py
```diff
@@ -121,3 +121,31 @@ def test_labeltoindex_jit(self):
         actual = transform_jit("test")
         expected = 0
         self.assertEqual(actual, expected)
+
+    def test_truncate(self):
```
mthrok, line 125: It is very important for unit tests to have doctoring so that future maintainers can get the idea of what this test is **supposed to** do.

torchtext/functional.py
```diff
@@ -10,10 +10,10 @@
 ]
 
 
-def to_tensor(input: Union[List[int], List[List[int]]], padding_value: Optional[int] = None, dtype: Optional[torch.dtype] = torch.long) -> Tensor:
+def to_tensor(input: Any, padding_value: Optional[int] = None, dtype: Optional[torch.dtype] = torch.long) -> Tensor:
```
mthrok, line 13: IIRC, this function is going to be public API in the next release. I am aware adding doctoring is in todo list of the PR, but I think it's better done in the same PR, rather than late PR.

mthrok (approved):
Looks good. I learned that TS compiler can handle `Any`. Thanks.

parmeet:
> Looks good. I learned that TS compiler can handle `Any`. Thanks.

Apparently yes. I was also pleasantly surprise when I closely looked at their example https://pytorch.org/docs/stable/generated/torch.jit.isinstance.html :)

torchtext/models/roberta/transforms.py
parmeet: Thanks for raising this up. I will keep change it back to original and do the naming change in separate PR.

torchtext/transforms.py
parmeet: Again good suggestion. Actually this whole idea of changing to `Any` became suddenly apparent after I started adding this transform. And then I just kept all the changes :). I will do another PR to add `Truncate` and remove it here.

parmeet:
Thanks @mthrok for the quick review and valuable feedback. I am going to break this PR in few. More importantly, I will keep this PR to only update the annotation types of existing transforms and add doc-strings. I will update the summary accordingly.

test/test_transforms.py
parmeet: Done!

PR #1454, opened 2021-12-08
Fix bool attention mask in transformer encoder

ebsmothers:
The bool attention mask was being cast to float incorrectly. This diff fixes the cast and creates an additional test case for bool masks


torchtext/models/roberta/modules.py
```diff
@@ -291,9 +289,11 @@ def __init__(
...
     def forward(self, tokens: torch.Tensor, attn_mask: Optional[torch.Tensor] = None) -> Union[torch.Tensor, List[torch.Tensor]]:
         if attn_mask is not None:
-            torch._assert(attn_mask.dtype == torch.bool, "Expected attn_mask dtype as `torch.bool` but got {}".format(attn_mask.dtype))
-            torch._assert(attn_mask.dim() == 2, "Expected attn_mask of dim 2 but got {}".format(attn_mask.dim()))
-
+            torch._assert(attn_mask.is_floating_point() or attn_mask.dtype == torch.bool, f"Only float or bool types are supported for attn_mask not {attn_mask.dtype}")
+            if attn_mask.dtype == torch.bool:
+                new_attn_mask = torch.zeros_like(attn_mask, dtype=input.dtype)
+                new_attn_mask.masked_fill_(attn_mask, -1e8 if input.dtype == torch.float32 else -1e4)
+                attn_mask = new_attn_mask
```
parmeet, line 296: I wonder if it is necessary to add the conversion here, since we have already added inside TransformerEncoderLayer? Following the same argument, perhaps we could also remove it from TransformerEncoderLayer, since we know that it is going to be passed to MultiHeadSelfAttention that would do this conversion. This was also discussed a bit in here https://github.com/pytorch/text/pull/1435#discussion_r745046146. I feel, we could avoid the redundancy here. wdyt?

ebsmothers: Good question. I think it makes sense to perform the bool-to-float conversion only in MultiHeadSelfAttention, since that is where the mask is actually used. Then we can just keep the checks in TransformerEncoder and TransformerEncoderLayer to ensure the attention mask is either bool or float.

parmeet (approved):
LGTM, thanks for fixing the issue!

PR #1456, opened 2021-12-08
[fbsync] Fix SST2Dataset test iterator

parmeet:
Summary:
## Summary
- Modified SST2 dataset implementation to only return text for test split (since label_ids are not available)
- Updated doc classification datamodule to temporarily use `val_dataset` instead of `test_dataset`
- Updated first line md5 hash for SST2 test split

## Followup Items
- Update doc classification module to work with test splits with and without labels

Reviewed By: parmeet

Differential Revision: D32661112

fbshipit-source-id: ef86aea0ce587c5d5282f2caa943b4b0cdf6f54a


Nayef211 (approved):
LGTM

PR #1457, opened 2021-12-08
remove experimental documentation

parmeet:
This PR removed documentation for the experimental code. The experimental code mostly includes:
* transformations (note that some of them has already been added to main branch in slightly modified form)
* some raw datasets
* experimental vectors

The path to productization is not yet clear on the existing experimental features and we are not seeking feedback on them. As documentation is first source of truth for users, we would like to avoid any confusions that might arise owning to the experimental docs available in the main branch.


Nayef211:
I am wondering if it's necessary to remove these docs altogether rather than keeping them hidden using "hidden" directives? If I understand correctly, using hidden directives will prevent the experimental sections from showing when we publish our docs meaning we can make them visible once we've decided how we want to productionize these components.

I only bring this up because it would create additional work for us when we want to re-add these docs. What do you think?

parmeet:
> I am wondering if it's necessary to remove these docs altogether rather than keeping them hidden using "hidden" directives? If I understand correctly, using hidden directives will prevent the experimental sections from showing when we publish our docs meaning we can make them visible once we've decided how we want to productionize these components.
> 
> I only bring this up because it would create additional work for us when we want to re-add these docs. What do you think?

ahh, that's a good point. I was not aware of it. Let me check that. I agree, we can keep the files as long as it doesn't show up in docs.

mthrok:
> > I am wondering if it's necessary to remove these docs altogether rather than keeping them hidden using "hidden" directives? If I understand correctly, using hidden directives will prevent the experimental sections from showing when we publish our docs meaning we can make them visible once we've decided how we want to productionize these components.
> > I only bring this up because it would create additional work for us when we want to re-add these docs. What do you think?
> 
> ahh, that's a good point. I was not aware of it. Let me check that. I agree, we can keep the files as long as it doesn't show up in docs.

I am skeptic that `hidden` directive would hide the content. I tried to figure out what it is meant but I did not get it. (let me know if I am wrong). There is a configuration to provide exclusion. 

https://github.com/pytorch/text/blob/a074cb2fac0abdc557f7a17043e440a315e15822/docs/source/conf.py#L87-L89

parmeet:
> 



> 

Thanks @mthrok , exclude_pattern was straightforward to work with.

mthrok (approved):
The change looks good.

Please double check that the resulting doc is what you intend it to be.
https://141313-76210069-gh.circle-artifacts.com/0/docs/index.html

PR #1458, opened 2021-12-08
Add Truncate Transform

parmeet:
Add truncate transform

### DIFF
--- a/torchtext/transforms.py
+++ b/torchtext/transforms.py
@@ -183,7 +183,7 @@ class Truncate(Module):
 
     def forward(self, input: Any) -> Any:
         """
-        :param input: Input sequence or batch to be truncated
+        :param input: Input sequence or batch of sequence to be truncated
         :type input: Union[List[Union[str, int]], List[List[Union[str, int]]]]
         :return: Truncated sequence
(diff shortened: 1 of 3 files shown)


torchtext/transforms.py
```diff
@@ -168,3 +168,24 @@ def forward(self, input: Any) -> Any:
...
+    """
+
+    def __init__(self, max_seq_len: int) -> None:
+        super().__init__()
+        self.max_seq_len = max_seq_len
+
+    def forward(self, input: Any) -> Any:
+        """
+        :param input: Input sequence or batch of sequence to be truncated
+        :type input: Union[List[Union[str, int]], List[List[Union[str, int]]]]
```
mthrok, line 187: I think it some users could get confused why the annotation in code is different from docstring. Explaining it will help resolve such confusion. Also it will give opportunity to explain that there are connections between input types and output types.

mthrok (approved):

torchtext/transforms.py
parmeet: I think it's a good point. I will think through it and update the docs accordingly.

PR #1459, opened 2021-12-09
Add GPT-2 BPE tokenizer operator leveraging re2

abhinavarora:


torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -0,0 +1,51 @@
...
+    bool is_whitespace(const std::string &input) {
+        for (std::string::const_iterator it = input.begin(); it != input.end(); ++it) {
+            if (!isspace(*it)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    std::vector<std::string> gpt2_bpe_tokenizer(std::string input) {
```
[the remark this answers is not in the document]
abhinavarora: That's a fair point. I had actually used the name we had internally, but will be happy to change this.

torchtext/data/functional.py
```diff
@@ -282,3 +283,19 @@ def __getitem__(self, idx):
             return self._data[idx]
 
     return _MapStyleDataset(iter_data)
+
+
+def gpt2_bpe_tokenizer(input: str) -> List[str]:
```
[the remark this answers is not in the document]
abhinavarora: Yes, I was thinking that too. However, for the completeness of this PR, I wanted to make sure that we have tests and I observed that functionality exposed via `csrc` did not have its dedicated tests. What's your recommendation on that?

torchtext/csrc/gpt2_bpe_tokenizer.cpp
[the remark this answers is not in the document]
abhinavarora: I think to be consistent with literature, lets call it `gpt2_bpe_pre_tokenizer`.

torchtext/data/functional.py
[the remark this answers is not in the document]
parmeet: It's a valid point, we do not have dedicated testing for operators and the testing is mainly done through python functions using them. let's have dedicated unit-testing for csrc operators. We can start by adding test for this operator. @mthrok I am curious, if there are dedicated unit-tests for operators in audio or is the testing is done indirectly via the APIs that are only exposed to the users?

torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <regex.h> // @manual
+
+namespace torchtext {
+    const Regex kGPT2Regex(
+        "(\\'s|\\'t|\\'re|\\'ve|\\'m|\\'ll|\\'d| ?\\pL+|"
+        " ?\\pN+| ?[^\\s\\v\\pL\\pN]+|[\\s\\v]+)"
+    );
+
+    bool is_whitespace(const std::string &input) {
+        for (std::string::const_iterator it = input.begin(); it != input.end(); ++it) {
```
mthrok, line 12: nit: Range-based for loop might improve the readability.

https://stackoverflow.com/a/9438329
https://en.cppreference.com/w/cpp/language/range-for

torchtext/data/functional.py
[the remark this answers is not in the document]
mthrok: We do not have tests for C++ yet in audio, but we are in desperate need to setup. So we will add one soon.

However, since this `gpt2_bpe_tokenizer` is just a fall through to `torch.ops.torchtext`, so I think you can test it directly, without this API.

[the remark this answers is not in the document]
abhinavarora: Yes, I just added `tests/csrc` . We can use this directory to add any tests for csrc operators. I think to maintain a good standard of unit test coverage, we should add unit tests for most of the operators that we expose through `csrc`.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
abhinavarora: Done

parmeet (approved):
LGTM, thanks! It would be interesting to do the benchmark against python regex module.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -0,0 +1,70 @@
...
+        //     - If token length is >1, add token[0:len(token) - 1] to return list
+        //     - IF token[-1] is space (ascii 32), then carry it over for next token, set append_space = True
+        //     - ELSE make token[-1] its own token and add to return list
+        // - ELSE IF append_space == True, prepend a space to the token and add to return list
+        // - ELSE, add token to return list
+        std::string token;
+        std::vector<std::string> tokens;
+        re2::StringPiece inp(input);
+        bool append_space = false;
+        while (kGPT2Regex.FindAndConsume(&inp, &token)) {
```
mthrok, line 44: note: nit: I feel that the cyclomatic complexity of this main logic, almost hitting the threshold. (though it's not like I calculated or anything), so if, in the future, any update is made, I suggest to consider refactoring. (I understand it's not straight forward though.)

torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -0,0 +1,70 @@
...
+            // tokens.push_back(token);
+            // Check if whitespace
```
mthrok, line 46: I think this comment is redundant, because of the name of the helper function, `is_whitespace`.
Rather comment on the logic in L52-59 would be helpful for future maintainer.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
abhinavarora: I agree. I don't expect this to change in future. If any changes occur in logic. I will try to abstract the logic out in an auxiliary self-contained method.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
abhinavarora: good catch!

abhinavarora: Addressed in latest commit!

PR #1462, opened 2021-12-14
Add Torchscriptable GPT-2 BPE Tokenizer for RoBERTa models

abhinavarora:
## Description

This PR implements the GPT-2 BPE tokenizer that is used by RoBERTa models. The implemented tokenizer is an initial working version that is scriptable. This PR also add test cases to test out the tokenizer.

## Testing
`pytest -k test_gpt2_bpe_tokenizer test/test_transforms.py`

## Future Work
1. Implement Caching 
2. Move implementation to C++
3. Implement support for character indices with tokenization

## Follow Ups to this PR

- [ ] Refactor Tokenizer organization in TorchText and set up tokenization specific constant and util functions.
- [ ] Once integration testing is set up for the repo, create a new hand-crafted unit test and move "shipped BPE model" tests to integration testing
- [ ] Set up time with @mthrok to have an in-depth discussion on the question of accepting file paths v/s file-like objects.


torchtext/transforms.py
```diff
@@ -189,3 +193,350 @@ def forward(self, input: Any) -> Any:
...
+            )
+        bpe_merge_ranks = GPT2BPETokenizer.load_bpe_vocab(vocab_bpe_local_path)
+        return GPT2BPETokenizer(
+            bpe_encoder,
+            bpe_merge_ranks,
+            GPT2BPETokenizer.load_byte_encoder()
+        )
+
+    @classmethod
+    def load_bpe_encoder(cls, bpe_encoder_path: str) -> Dict[str, int]:
```
Nayef211, line 268: Any reason to make this and the methods below `@classmethod` instead of just private methods? Do we expect the user to want to use these methods outside the context of creating a `GPT2BPETokenizer` object?

torchtext/transforms.py
```diff
@@ -189,3 +193,350 @@ def forward(self, input: Any) -> Any:
...
+    @classmethod
+    def load_byte_decoder(cls, byte_encoder: Dict[int, str]) -> Dict[str, int]:
+        byte_decoder: Dict[str, int] = {}
+        for byte_t, unicode_t in byte_encoder.items():
+            byte_decoder[unicode_t] = byte_t
+        return byte_decoder
+
+    # @classmethod
+    # def build_bpe_merge_ranks(cls, bpe_merge_ranks) -> Dict[str, int]:
+    #     return {
```
Nayef211, line 339: nit: remove commented code

Nayef211 (commented):
Overall LGTM. Just had a question and a nit comment

torchtext/transforms.py
```diff
@@ -189,3 +193,350 @@ def forward(self, input: Any) -> Any:
...
+    """
+    SEPERATOR: str = "\u0001"
+    LSTRIP_PATTERN: str = "\u0120"
+
+    def __init__(
+        self,
+        bpe_encoder: Dict[str, int],
+        bpe_merge_ranks: Dict[str, int],
+        byte_encoder: Dict[int, str],
+    ):
```
parmeet, line 224: +1 to @Nayef211  comments below. We should probably just accept encoder_json_path and vocab_bpe_path inside the `__init__` and keep all other aspects of implementation as private. Since we are not yet have design for tokenizer. From Transform perspective, it would probably be just enough to have forward method as public. Also nit: we should probably give more intuitive names to these variables.

test/test_transforms.py
```diff
@@ -149,3 +149,46 @@ def test_truncate(self):
...
         self._truncate(test_scripting=True)
+
+
+class TestGPT2BPETokenizer(TorchtextTestCase):
+    def _gpt2_bpe_tokenizer(self, test_scripting):
+        encoder_json = "gpt2_bpe_encoder.json"
+        bpe_vocab = "gpt2_bpe_vocab.bpe"
+        tokenizer = transforms.GPT2BPETokenizer.get_gpt2_bpe_tokenizer(
+            encoder_json_path=get_asset_path(encoder_json),
+            vocab_bpe_path=get_asset_path(bpe_vocab),
```
parmeet, line 160: we should add them to pytorch/text aws folder as artifacts so that we could use them with pre-trained models.

torchtext/transforms.py
```diff
@@ -189,3 +193,350 @@ def forward(self, input: Any) -> Any:
...
+    :param bpe_merge_ranks: A dict mapping a pair of bpe tokens to rank.
+            Torchscript doesn't support tuple as dict key type, concat
+            a pair of string with SEPERATOR (e.g \u0001) as workaround.
+            For example: {"ing\u0001ly": 210}
+    :type bpe_merge_ranks: Dict[str, int]
+    :param byte_encoder: A dict mapping a byte to an unicode character.
+    :type byte_encoder: Dict[int, str]
+    """
+    SEPERATOR: str = "\u0001"
+    LSTRIP_PATTERN: str = "\u0120"
```
mthrok, line 217: Class attributes are tricky here. If anyone needs to use custom separator/lstip_pattern, then one needs to modify them on class.

```python
GPT2BPETokenizer.SEPARATOR = my_separator
GPT2BPETokenizer.LSTRIP_PATTERN = my_lstrip_pattern
foo = GPT2BPETokenizer.load_xxx(...)
```

It looks like these separator and strip_pattern are logically connected with the data to be loaded (json file and bpe file), which means that the customization should best happen at loading time.
Making them into class attribute means, that if I have two BPE tokenizers, one with default and the other with custom separator/lstrip_pattern, I cannot use them at the same time.

Looking at the implementation of `get_pairs` and `bpe` method, `self.merge_seperator` are used. This means that they are better coupled with instance, instead of class.

Also this approach is forcing all the `load_XXX` functions to be `classmethod`. In general in Python, there is not much advantage of using `classmethod`/`staticmethod` over plain module level function (except logical grouping of functions and the very specific case where the method is intended to be subclassed and overridden from child class).

Having class method could be troublesome if the instance is going to be used in subprocess (i.e. the instances are pickled and unpicked when passed from main process to subprocess. I am not sure that's the case for this class, but in general, it's better not to have one).

Therefore, if these attributes are often customized, then I recommend to make them as optional constructor argument, else, putting them as a global variable wouldn't hurt much.

torchtext/transforms.py
mthrok: > +1 to @Nayef211 comments below. We should probably just accept encoder_json_path and vocab_bpe_path inside the `__init__` and keep all other aspects of implementation as private. Since we do not yet have design for tokenizer. From Transform perspective, it would probably be just enough to have forward method as public. Also nit: we should probably give more intuitive names to these variables.

Note: In general, rather than taking a path object, taking a file-like object or the data with proper structure makes the interface more flexible as you can test with dummy data without need of dumping data to disk. IMO making a file path a public interface would this class a bit restricted, especially the actual data structure this class requires is just a dictionary. If I want to hide the loading mechanism I would use a uion of 1file-like object` (not a path object) and the proper structure, `Dict[std, int]]`. (but it's my spontaneous thinking so, there should be a better approach.)

torchtext/transforms.py
```diff
@@ -189,3 +193,350 @@ def forward(self, input: Any) -> Any:
...
+
+    @classmethod
+    def load_bpe_decoder(cls, bpe_encoder: Dict[str, int]) -> Dict[int, str]:
+        bpe_decoder: Dict[int, str] = {}
+        for sub_word, bpe_token_id in bpe_encoder.items():
+            bpe_decoder[bpe_token_id] = sub_word
+        return bpe_decoder
+
+    @classmethod
+    def load_byte_decoder(cls, byte_encoder: Dict[int, str]) -> Dict[str, int]:
```
mthrok, line 331: The logic inside of `load_byte_decoder` and `load_bpe_decoder` looks identical. I think they are better extracted to a helper function. (I recommend a plain helper function rather than helper method)

torchtext/transforms.py
abhinavarora: That's a fair point. Let me refactor this one. For now, I was using the internal implementation as it is with limited modifications.

torchtext/transforms.py
```diff
@@ -189,3 +193,350 @@ def forward(self, input: Any) -> Any:
...
+    def load_bpe_vocab(cls, bpe_vocab_path: str) -> Dict[str, int]:
+        if os.path.exists(bpe_vocab_path):
+            local_path = bpe_vocab_path
+        else:
+            local_path = download_from_url(url=bpe_vocab_path, root=_CACHE_DIR)
+        with open(local_path, "r", encoding="utf-8") as f:
+            bpe_vocab = f.read()
+
+        return {
+            cls.SEPERATOR.join(merge_pair.split()): i
```
mthrok, line 286: Curious, why BPE file cannot be formatted as JSON? 
If bpe_vocab and bpe_encoder are in same format, then they can share the same I/O logic.

torchtext/transforms.py
```diff
@@ -189,3 +193,350 @@ def forward(self, input: Any) -> Any:
...
+        """
+        if torch.jit.isinstance(input, List[str]):
+            tokens: List[List[str]] = []
+            for text in input:
+                tokens.append(self.tokenize(text))
+            return tokens
+        elif torch.jit.isinstance(input, str):
+            return self.tokenize(input)
+        else:
+            raise TypeError("Input type not supported")
```
mthrok, line 542: nit: It is more common to perform the input validation before proceeding to the actual logic, but this method seems to be jit-compatible, performance-sensitive and simple, so I guess it's okay.

torchtext/transforms.py
```diff
@@ -189,3 +193,350 @@ def forward(self, input: Any) -> Any:
...
+            token_list = list(new_token_list)
+            if len(token_list) == 1:
+                break
+            else:
+                pairs = self.get_pairs(token_list)
+
+        return token_list
+
+    @torch.jit.export
+    def byte_encode(self, token: str) -> List[str]:
```
mthrok, line 458: Is there a test to verify that the unicode input/output patterns are sam for eager mode execution and scripted execution?

I guess that the logic branch at `torch.jit.is_scripting` is to handle the difference of string handling between Python and C++, but it would be more assuring to have that in test.

In the past, audio had an issue where string I/O causes different results in eager and scripted executions. The root cause is the comparison and it does not apply to this method, but I think it
s safer to have such test.

https://github.com/pytorch/audio/pull/1601#issuecomment-866972482

mthrok (commented):
I recommend to document how the new assets are generated. Ideally, if the asset format allows comment, there, otherwise either in README of unit test or the test class.

https://github.com/pytorch/audio/pull/1104/files#diff-c8c2b1f2318cddbb6e9602b7615608204bd21d175922f7661466d77dd846c693R258

torchtext/transforms.py
parmeet: Sure, I think the suggestions make sense. We could keep the interface same. The loading can be handled by the client code, given it is just a plain python dictionary.

mthrok:
I think it's better to hand-craft the test assets. I cannot check the file size of the 50000 lines of text since I'm on mobile, but it would increase the repo size, git history size, memory consumption during the unit tests, and time consumed on test preparation. Also, the vast majority of data in the assets are not used during the test and with this size, no one will ever inspect the content, meaning that they obfuscate the logic of tests. I think curating a minimum set of test assets would help better the future maintainers. 

I think the test with full-blown BPE data can go to integration test where the BPE data are retrieved from the internet on-the-fly (and maybe cached)

torchtext/transforms.py
abhinavarora: @mthrok Thanks for the detailed suggestions. Your points make a lot of sense and I will try to get rid of the unnecessary class methods. 

 One thing, I would like to emphasize that these class attributes are not designed to be customizable. The LSTRIP_PATTERN is how GPT-2 encodes spaces. We don't need users to change this unless they want to try out an idea that is different from the original implementation. As far as SEPERATOR is concerned, that is our workaround because Torchscript does not support Tuple as Dict keys. SEPERATOR in that way is closely toed to our implementation and we don't expect any user to modify this. 

> Also this approach is forcing all the load_XXX functions to be classmethod. In general in Python, there is not much advantage of using classmethod/staticmethod over plain module level function (except logical grouping of functions and the very specific case where the method is intended to be subclassed and overridden from child class).

I am not sure I understand why this forces us to make class methods. Let me refactor these to be ScriptModule constants as described at the bottom of https://pytorch.org/docs/stable/jit_language_reference.html

torchtext/transforms.py
abhinavarora: By plain helper function do you mean a util function outside the class?

torchtext/transforms.py
abhinavarora: Technically, nothings stops us from formatting them as a JSON. However, if I think from a user's perspective, we should align with the file formats that are released by AllenAI, HuggingFace, etc. If people train their own GPT-2 tokenizers using the original implementation of any of these libraries, it is likely that the files produced by them will be in the same format. Hence, I suggest keeping the same file format and letting the implementation deal with Torchscript-specific nuances. Hope that makes sense?

torchtext/transforms.py
abhinavarora: @mthrok Yes, that was my concern as well. However, for now I was sticking to the same interface as the SentencePiece tokenizer. We have been talking about standardizing the Tokenizer interface and we shall bring this point up in that discussion.

parmeet: These type checks are special once that also do type refinement. We could do the input validation, but I guess that would be redundant in this special case of torch-script implementation where we are doing it via type-refinement when we actually need it.

torchtext/transforms.py
parmeet: Actually I just realized that bpe_merge_ranks is a special dictionary that we construct internally (using the separator) and we cannot expect this to be constructed by the client code. It is very much specific to our implementation as also noted by @abhinavarora in the [comment](https://github.com/pytorch/text/pull/1462#discussion_r769210112). We are better off sticking to the standard path or file-like objects pointing to the encoder_json_path  and vocab_bpe_path 
 instead of accepting dictionary. This way we can handle all the the complexity internally and the interface would accept the standard files generated by other frameworks as noted in the above comment.

mthrok: okay, that makes sense. I would still argue that loading mechanism can stay outside of the class so that class is only responsible for the processing logic but that would introduce one more public API so I am okay either way.

I wonder if it would be valuable in long term to implement a dictionary in C++ that takes tuple as key in TorchScript-compatible manner.

torchtext/transforms.py
abhinavarora: Good catch.

torchtext/transforms.py
```diff
@@ -189,3 +193,350 @@ def forward(self, input: Any) -> Any:
...
+    def load_bpe_encoder(cls, bpe_encoder_path: str) -> Dict[str, int]:
+        if os.path.exists(bpe_encoder_path):
+            local_path = bpe_encoder_path
+        else:
+            local_path = download_from_url(url=bpe_encoder_path, root=_CACHE_DIR)
+        with open(local_path, "r", encoding="utf-8") as f:
+            return json.load(f)
+
+    @classmethod
+    def load_bpe_vocab(cls, bpe_vocab_path: str) -> Dict[str, int]:
```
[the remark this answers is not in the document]
abhinavarora: Yep sounds reasonable! Will take a look at this!

torchtext/transforms.py
abhinavarora: The currently added unit tests actually test this. If you see the test, we run one version of tests with the nn.Module while the other is run after scripting the module. The test has input cases with special unicode characters as well for completeness. We expect same output in both cases.  Furthermore, the scripted implementation is also something that is well tested internally.

abhinavarora:
> I think it's better to hand-craft the test assets. I cannot check the file size of the 50000 lines of text since I'm on mobile, but it would increase the repo size, git history size, memory consumption during the unit tests, and time consumed on test preparation. Also, the vast majority of data in the assets are not used during the test and with this size, no one will ever inspect the content, meaning that they obfuscate the logic of tests. I think curating a minimum set of test assets would help better the future maintainers.
> 
> I think the test with full-blown BPE data can go to integration test where the BPE data are retrieved from the internet on-the-fly (and maybe cached)


The reason I added the full BPE vocab was because I wanted to make sure that tests capture some of the edge cases with unicode special characters that we have seen internally. The 2 files are 500kB and 1MB respectively. Wondering if this will be a bigger issue. Since most users will mostly use the full-blown GPT2-BPE from Allen AI, I thought that adding tests with the full one may be better.

abhinavarora:
> I recommend to document how the new assets are generated. Ideally, if the asset format allows comment, there, otherwise either in README of unit test or the test class.
> 
> https://github.com/pytorch/audio/pull/1104/files#diff-c8c2b1f2318cddbb6e9602b7615608204bd21d175922f7661466d77dd846c693R258

Assets are usually generated by developers of pre-trained models and currently they will be generated via other frameworks. I can add some asset descriptions in the test because we cannot rely on the comments in the asset file as they can get inconsistent.

torchtext/transforms.py
abhinavarora: I think we plan to refactor tokenizers once this goes out so that each tokenizer be its own module. Once we get to that directory design, we can certainly create helper functions for this class. I further plan to move the logic to C++ operators, so that refactor will also help.  In the meanwhile, I am planning to keep the current methods as it is to prevent duplicate effort.

torchtext/transforms.py
abhinavarora: Done! Once we refactor for Tokenizers components, we can take some of these functions and move them module level util functions.

torchtext/transforms.py
abhinavarora: Marking this as resolved for now!

torchtext/transforms.py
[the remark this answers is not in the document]
abhinavarora: Done!

torchtext/transforms.py
abhinavarora: Marking this as resolved. We will revisit this as we design the Tokenizers API.

parmeet:
> > I recommend to document how the new assets are generated. Ideally, if the asset format allows comment, there, otherwise either in README of unit test or the test class.
> > https://github.com/pytorch/audio/pull/1104/files#diff-c8c2b1f2318cddbb6e9602b7615608204bd21d175922f7661466d77dd846c693R258
> 
> Assets are usually generated by developers of pre-trained models and currently they will be generated via other frameworks. I can add some asset descriptions in the test because we cannot rely on the comments in the asset file as they can get inconsistent.

Sounds good!

parmeet:
> > I think it's better to hand-craft the test assets. I cannot check the file size of the 50000 lines of text since I'm on mobile, but it would increase the repo size, git history size, memory consumption during the unit tests, and time consumed on test preparation. Also, the vast majority of data in the assets are not used during the test and with this size, no one will ever inspect the content, meaning that they obfuscate the logic of tests. I think curating a minimum set of test assets would help better the future maintainers.
> > I think the test with full-blown BPE data can go to integration test where the BPE data are retrieved from the internet on-the-fly (and maybe cached)
> 
> The reason I added the full BPE vocab was because I wanted to make sure that tests capture some of the edge cases with unicode special characters that we have seen internally. The 2 files are 500kB and 1MB respectively. Wondering if this will be a bigger issue. Since most users will mostly use the full-blown GPT2-BPE from Allen AI, I thought that adding tests with the full one may be better.

I think this is reasonable. If the size is really a concern, we could use the one directly from AWS server (since we are going to upload these files there for usage with pre-trained models).

mthrok:
> The reason I added the full BPE vocab was because I wanted to make sure that tests capture some of the edge cases with unicode special characters that we have seen internally.

This sounds to me that the issue is more of coverage of unicodes than edge cases. In the current structure, this coverage is defined by the BPE data originated not from library code but from the external data, the signal we will get from the test does not necessarily align with the end-user experience.
I think it makes sense to test if these specific characters are handled correctly when it's tested on **the shipped BPE data**, instead of the separate assets in tests. (alternatively, how is it guaranteed that the BPE data in test assets and the one you are going to publish on S3 as pre-trained data are in-sync? If the BPE data in test assets become obsolete, then we do not get the right signal.)

abhinavarora:
> > The reason I added the full BPE vocab was because I wanted to make sure that tests capture some of the edge cases with unicode special characters that we have seen internally.
> 
> This sounds to me that the issue is more of coverage of unicodes than edge cases. In the current structure, this coverage is defined by the BPE data originated not from library code but from the external data, the signal we will get from the test does not necessarily align with the end-user experience. I think it makes sense to test if these specific characters are handled correctly when it's tested on **the shipped BPE data**, instead of the separate assets in tests. (alternatively, how is it guaranteed that the BPE data in test assets and the one you are going to publish on S3 as pre-trained data are in-sync? If the BPE data in test assets become obsolete, then we do not get the right signal.)

Hey @mthrok, I think I did not explain myself correctly in the previous comment :-) The test is not concerned with specific unicode characters. The unicode tests mainly test that the behavior on out of vocab tokens containing multi-byte characters is correct and is consistent between the scripted and the non-scripted versions. Also, I do expect the test data to be the same as "the shipped BPE data". I don't expect the shipped data in OSS to change for almost all users, unless someone wants to use their own BPE vocab or there is a new pre-trained model released in OSS that leverages a much larger dataset to learn the vocab. Hope that makes sense. Most SOTA text research generally uses the data as it is in OSS. The RoBERTa models open-sourced by FAIR are a great example. They reused the tokenizer and the data from what Allen AI released for GPT.

If you still feel that the test is not right, I could try to handcraft my custom data, which can be a small subset of the shipped data.

torchtext/transforms.py
```diff
@@ -189,3 +186,294 @@ def forward(self, input: Any) -> Any:
...
+    :param bpe_encoder: A dict mapping from bpe tokens(e.g sub-words) to
+            bpe token ids. For example: {"ingly": 4420}
+    :type bpe_encoder: Dict[str, int]
+    :param bpe_merge_ranks: A dict mapping a pair of bpe tokens to rank.
+            Torchscript doesn't support tuple as dict key type, concat
+            a pair of string with SEPERATOR (e.g \u0001) as workaround.
+            For example: {"ing\u0001ly": 210}
+    :type bpe_merge_ranks: Dict[str, int]
+    :param byte_encoder: A dict mapping a byte to an unicode character.
+    :type byte_encoder: Dict[int, str]
```
parmeet, line 207: could you please update doc-strings accordingly. Also could please update [transforms.rst](https://raw.githubusercontent.com/pytorch/text/main/docs/source/transforms.rst) so that it is included in the documentation.

torchtext/transforms.py
```diff
@@ -189,3 +186,294 @@ def forward(self, input: Any) -> Any:
...
+    LSTRIP_PATTERN: str = "\u0120"
+
+    def __init__(
+        self,
+        encoder_json_path: str,
+        vocab_bpe_path: str,
+    ):
+        super().__init__()
+        self.merge_seperator = self.SEPERATOR
+        self.lstrip_pattern = self.LSTRIP_PATTERN
```
parmeet, line 219: nit: can we instead directly call self.SEPERATOR and self.LSTRIP_PATTERN? not sure if it is necessary to assign to another class variable?

torchtext/transforms.py
```diff
@@ -189,3 +186,294 @@ def forward(self, input: Any) -> Any:
...
+            return json.load(f)
+
+    def _load_bpe_vocab(self, bpe_vocab_path: str) -> Dict[str, int]:
+        with open(get_asset_local_path(bpe_vocab_path), "r", encoding="utf-8") as f:
+            bpe_vocab = f.read()
+
+        return {
+            self.merge_seperator.join(merge_pair.split()): i
+            for i, merge_pair in enumerate(bpe_vocab.split("\n")[1:-1])
+        }
```
parmeet, line 239: nit: These functions seems to be called internally inside constructor only. can we instead directly implement the logic of them inside constructor instead of making a function call?

parmeet (approved):
LGTM! Please make sure to update the summary (potentially checklist) for any relevant outstanding issues discussed in this PR, so that we can try to cover them up in follow-up PRs.

mthrok:
@abhinavarora @parmeet I do not want to block the PR, so you folks can go ahead, but once a file is checked-in, that action cannot be undone, so I present my thoughts one more time.

> Hey @mthrok, I think I did not explain myself correctly in the previous comment :-) The test is not concerned with specific unicode characters. The unicode tests mainly test that the behavior on out of vocab tokens containing multi-byte characters is correct and is consistent between the scripted and the non-scripted versions. 

Thanks for the explanation. I think both correctness checking and TS consistency checking can be done without 50k data. In fact correctness checking is easier to verify with eyes when data are small and hand-crafted. So I still wonder what is absolute necessity to use and check-in the 50k data. 

> Also, I do expect the test data to be the same as "the shipped BPE data". I don't expect the shipped data in OSS to change for almost all users, unless someone wants to use their own BPE vocab or there is a new pre-trained model released in OSS that leverages a much larger dataset to learn the vocab. 

I understand that shipped data are unlikely changed. But if these data files are checked-in, now there are two different sources of truth for testing and shipped package. I am not against testing on full-brown data, but if such tests are done, I think it should be done on shipped package, i.e. in the form of integration test, fetching the shipped data on-the-fly.

mthrok:
BTW, reading the documentation on `Vocab`/`vocab` https://pytorch.org/text/main/vocab.html, I feel that having a factory function that loads data from external file, (and have the GPT2 class accept the plain data type like dictionaries) feels more consistent with the other library design.

torchtext/transforms.py
```diff
@@ -189,3 +186,277 @@ def forward(self, input: Any) -> Any:
...
+        best_rank: int = self.bpe_merge_ranks.get(best_pair, self.inf)
+
+        for pair in pairs[1:]:
+            rank: int = self.bpe_merge_ranks.get(pair, self.inf)
+            if rank < best_rank:
+                best_pair = pair
+                best_rank = rank
+        return best_pair
+
+    @ torch.jit.export
```
mthrok, line 282:
```suggestion
    @torch.jit.export
```

torchtext/transforms.py
```diff
@@ -189,3 +186,277 @@ def forward(self, input: Any) -> Any:
...
+        """
+        pairs: Dict[str, int] = {}
+        prev_token: str = token_list[0]
+        for token in token_list[1:]:
+            pair: str = prev_token + self.SEPERATOR + token
+            pairs[pair] = 0
+            prev_token = token
+        return list(pairs.keys())
+
+    @ torch.jit.export
```
mthrok, line 265:
```suggestion
    @torch.jit.export
```

torchtext/transforms.py
```diff
@@ -189,3 +186,277 @@ def forward(self, input: Any) -> Any:
...
+    def _list_str_index(self, list_: List[str], element: str, start: int) -> int:
+        """
+        Equivalent to: list.index(v, start)
+        """
+        for i, t in enumerate(list_[start:]):
+            if t == element:
+                return start + i
+        return -1
+
+    @ torch.jit.export
```
mthrok, line 246:
```suggestion
    @torch.jit.export
```

torchtext/transforms.py
```diff
@@ -189,3 +186,277 @@ def forward(self, input: Any) -> Any:
...
+        self.inf = len(self.bpe_merge_ranks) + 1
+
+        # load byte encoder and decoder
+        self.byte_encoder = bytes_to_unicode()
+        self.byte_decoder = {
+            unicode_t: byte_t
+            for byte_t, unicode_t in self.byte_encoder.items()
+        }
+
+    @ torch.jit.export
```
mthrok, line 236:
```suggestion
    @torch.jit.export
```

torchtext/transforms.py
```diff
@@ -189,3 +186,277 @@ def forward(self, input: Any) -> Any:
...
+    """
+    SEPERATOR: torch.jit.Final[str]
+
+    def __init__(
+        self,
+        encoder_json_path: str,
+        vocab_bpe_path: str,
+    ):
+        super().__init__()
+        self.SEPERATOR = "\u0001"
```
mthrok, line 211:
```suggestion
        self._separator = "\u0001"
```

abhinavarora:
@mthrok Thank you for the feedback. `SEPERATOR: torch.jit.Final[str]` is no more a Class attribute. It is now an instance level contant. the declaration there is just to let torchscript know that this is a Final attribute and torchscript can do some optimizations during scripting for this. With the latest commits, I have gotten rid of all class level attributes.

abhinavarora:
> @abhinavarora @parmeet I do not want to block the PR, so you folks can go ahead, but once a file is checked-in, that action cannot be undone, so I present my thoughts one more time.
> 
> > Hey @mthrok, I think I did not explain myself correctly in the previous comment :-) The test is not concerned with specific unicode characters. The unicode tests mainly test that the behavior on out of vocab tokens containing multi-byte characters is correct and is consistent between the scripted and the non-scripted versions.
> 
> Thanks for the explanation. I think both correctness checking and TS consistency checking can be done without 50k data. In fact correctness checking is easier to verify with eyes when data are small and hand-crafted. So I still wonder what is absolute necessity to use and check-in the 50k data.
> 
> > Also, I do expect the test data to be the same as "the shipped BPE data". I don't expect the shipped data in OSS to change for almost all users, unless someone wants to use their own BPE vocab or there is a new pre-trained model released in OSS that leverages a much larger dataset to learn the vocab.
> 
> I understand that shipped data are unlikely changed. But if these data files are checked-in, now there are two different sources of truth for testing and shipped package. I am not against testing on full-brown data, but if such tests are done, I think it should be done on shipped package, i.e. in the form of integration test, fetching the shipped data on-the-fly.

Hey @mthrok Your points make total sense. @parmeet and I have been talking about integration tests and we plan to create a new task for us to set up integration testing for this library. Let me add an action item on me to replace this test with a dummy hand-crafted test once integration testing is ready. I want to make sure that this testing is in place because I plan to implement some of the components from this PR in C++. As a result having this testing is of high importance. 

One thing I did not understand from your comment was why can't a file check-in not be undone? We can always replace the test files, right? Wondering if there is any additional perspective I am missing? 

Thank you again for all your awesome feedback. I am updating the summary of this PR with action items so that we can follow up on all your suggestions before closing out this workstream.

abhinavarora:
> BTW, reading the documentation on `Vocab`/`vocab` https://pytorch.org/text/main/vocab.html, I feel that having a factory function that loads data from external file, (and have the GPT2 class accept the plain data type like dictionaries) feels more consistent with the other library design.

I think @parmeet was forced to design Vocab this way because `Vocab` was an API that was already being used by users and he wanted  to keep the API consistent. Not sure if we would have still followed the same practice if this was not an interface we inherited.

abhinavarora:
@mthrok and @parmeet I have added follow up items in the PR summary so that we can address them in the near future and sync up together to find a common solution. Will land this PR for now so that we can unblock RoBERTa model for now.

mthrok:
> One thing I did not understand from your comment was why can't a file check-in not be undone? We can always replace the test files, right? Wondering if there is any additional perspective I am missing?

Even if the file is updated later, the original file stays in the git history, so the file size transferred with `git clone` will stay the same. (unless one uses partial clone or shallow clone)

PR #1463, opened 2021-12-16
add AddToken transform

parmeet:
This PR add the AddToken transform. This is a common operation that adds tokens at the start or end of input sequence to indicate begin or end of sentence.

Follow-up Items

- [ ] Organize transforms tests using parameterization


test/test_transforms.py
```diff
@@ -149,3 +149,56 @@ def test_truncate(self):
     def test_truncate_jit(self):
         """test truncation with scripting on both sequence and batch of sequence with both str and int types"""
         self._truncate(test_scripting=True)
+
+    def _add_token(self, test_scripting):
```
abhinavarora, line 153: nit: Could we create a separate test class for this transform? Within that class we could have several different test cases such as script v/s non-script and int vs str tokens.

abhinavarora (commented):
Looks good overall. Left a nit comment.

test/test_transforms.py
parmeet: hmm, I am curious if that reasoning also applies to other transform tests as well? So basically within one function we are testing multiple input types. should they belong to different unit-tests? I am not sure what's the right answer here, but one thing is this potentially increase the number of unit tests by a lot.

abhinavarora: I think there are 2 different items we are referring to:
1. Separate out tests for different transforms into their own classes.
2. Break the tests for different input types into multiple tests.

Regarding the 1st one, wouldn't it make more sense from an organizational perspective? 

For the 2nd one, we can discuss more but do you see any concern with the increased number of test cases? The amount of code executed is the same, just organized granularly. Also, `pytest` does allow multiple workers to parallelize.

parmeet: I am thinking for point 2, we could either potentially  use @parameterized decorator like already done for some of our tests, for eg https://github.com/pytorch/text/blob/da914e6935bfb88b27dfd3e42d8cad470e5afff3/test/data/test_builtin_datasets.py#L55  or make use of pytest parameterization https://docs.pytest.org/en/6.2.x/parametrize.html. I know vision had recently migrated to pytest, @NicolasHug  could provide any additional context here. Regarding point 1, I do not have a good answer. I wonder if @NicolasHug , @mthrok  have additional suggestions. 

I would suggest we have a follow-up here and revamp all our transforms tests together. Would that be OK @abhinavarora ?

NicolasHug: > I know vision had recently migrated to pytest, @NicolasHug could provide any additional context here

pytest comes with many additional benefits, but parametrization is IMHO the main one. I wrote about our experience in torchvision [here](https://fb.workplace.com/groups/3512737345441020/permalink/4721646271216782/) (internal link). Switching to pytest has been greatly beneficial to us and I would be happy to assist torchtext if you want. **if** you already have a dependency on the `parameterize` package, maybe sticking to it for now is enough.

> Regarding point 1, I do not have a good answer. I wonder if @NicolasHug , @mthrok have additional suggestions.

In torchvision we group related tests in a single class on a case-by-base basis. Sometimes we do, sometimes we don't. You can see e.g. https://github.com/pytorch/vision/blob/main/test/test_transforms.py.

mthrok: > I think there are 2 different items we are referring to:
> 
> 1. Separate out tests for different transforms into their own classes.
> 2. Break the tests for different input types into multiple tests.
> 
> Regarding the 1st one, wouldn't it make more sense from an organizational perspective?
> 
> For the 2nd one, we can discuss more but do you see any concern with the increased number of test cases? The amount of code executed is the same, just organized granularly. Also, `pytest` does allow multiple workers to parallelize.

@parmeet @abhinavarora 

In audio, to reduce the boilerplate code for individual testing, we took an approach similar to 2. We break tests into cases based on kinds of test and dtype/devices. In https://github.com/pytorch/audio/tree/main/test/torchaudio_unittest/transforms, we have autograd tests, batch consistency tests, third party consistency tests, torch script consistency tests, and transform-specific correctness tests. Most of which are run on the combination of cpu/gpu x float32/float64/complex64/complex128. 

Both 1 and 2 has pros and cons, but the reason we did not take approach 1 is that we think that it's much easier to see what kind of tests are there and how to add from the perspective of new contributor. With the current organization of tests (and potentially thanks to [README](https://github.com/pytorch/audio/tree/main/test/torchaudio_unittest) and [CONTRIBUTING](https://github.com/pytorch/audio/blob/main/CONTRIBUTING.md)) as far as I recall, we did not receive any question from new contributor on how to add tests.

abhinavarora: @mthrok @parmeet  @NicolasHug Thanks for the discussion. Actually I was not comparing both 1) and 2). I was counting them as 2 separate items to work on. My proposal was that we do both things. Today if you see our tests, then we have a single Test class that tests *all* transforms. This currently works fine as number of transforms is low. However, it will start becoming messy as number  of transforms increases. Secondly for each Transform, today we write one monolithic test method, that covers all the cases. I think there is opportunity to refactor this use parametrization and also create more methods for non-parameterizable test cases. Hope this makes sense?

parmeet: Sounds good @abhinavarora. Let's work on revamping the transforms tests. My suggestion would be to do it together for all transforms in a separate PR to maintain consistency in test suit. I have added this item as a follow-up item in the summary.

abhinavarora: Sounds like a plan!

abhinavarora (approved):
LGTM! As discussed we can take up on test organization later!

PR #1466, opened 2021-12-20
replace string typing with pkg_resources.parse_version

erip:
Instead of using string methods to parse version information, we can use tested and standard tools in the standard library.

### DIFF
--- a/test/data/test_utils.py
+++ b/test/data/test_utils.py
@@ -1,5 +1,6 @@
 import io
 from torchtext.data import get_tokenizer
+from torchtext.data.utils import get_torch_version
 from torchtext.utils import unicode_csv_reader
 from ..common.torchtext_test_case import TorchtextTestCase
 from ..common.assets import get_asset_path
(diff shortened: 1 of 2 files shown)


mthrok:
> Instead of using string methods to parse version information, we can use tested and standard tools in the standard library.

Technically `pkg_resources` is not a standard library. It is distributed with `setuptools`, which is highly available, yet it's questionable to me if it should be a runtime dependency of torchtext.

test/data/test_utils.py
```diff
@@ -48,3 +49,12 @@ def test_text_nomalize_function(self):
                 ref_lines.append(line)
 
         self.assertEqual(ref_lines, test_lines)
+
+    def test_version_parsing(self):
+        import torch
+        # Assign a torch version arbitrarily
+        version = (1, 10, 1)
+        major, minor, patch = version
+        torch.__version__ = ".".join(map(str, version))
```
mthrok, line 58: This effect persists after the test is done. Either perform the clean up or change the signature of `get_torch_version` to accept the value to parse.

erip: I guess this begs the question: where is this function used? I did a quick scan of the repo and couldn't find it anywhere. Is it an artifact of legacy API?

mthrok: Indeed I got the same question and looks like the it's not really torchtext's responsibility or interest to provide this function.

@parmeet @abhinavarora what do you think?

parmeet: Yes, We are in process of cleaning up stale-code and this is potential candidate. I agree with @mthrok that it's not in torchtext's interest to maintain this function.

erip: If it's on the chopping block, shall I repurpose this PR to remove it? It's technically part of the public API, but I can at least deprecate it if desired.

mthrok: I think it makes sense to repurpose this PR for deprecation, but let's hear from @parmeet .

PR #1468, opened 2021-12-20
non-distributed training example for SST2 dataset using XLM-Roberta model

parmeet:
Adding example for SST-2 binary text classification using XLM-Roberta


examples/tutorials/sst2_classification_non_distributed.py
```diff
@@ -0,0 +1,156 @@
...
+# TODO
+
+
+######################################################################
+# Common imports
+# --------------
+
+import torch
+import torch.nn as nn
+DEVICE = torch.DEVICE("cuda") if torch.cuda.is_available() else "cpu"
```
mthrok, line 23: Did you use some sort of find and replace?
I do not think this is what you intended.

examples/tutorials/sst2_classification_non_distributed.py
```diff
@@ -0,0 +1,156 @@
...
+
+#######################################################################
+# Data Transformation
+# -------------------
+# TODO
+
+import torchtext.transforms as T
+from torch.hub import load_state_dict_from_url
+
+PADDING_IDX = 1
```
mthrok, line 34: Since this file is now in the form more appropriate to tutorial, I think you can lowercase the most of variables.

docs/requirements.txt
```diff
@@ -1,2 +1,3 @@
 sphinx==3.5.4
--e git+git://github.com/pytorch/pytorch_sphinx_theme.git#egg=pytorch_sphinx_theme 
+-e git+https://github.com/pytorch/pytorch_sphinx_theme.git@b4d0005#egg=pytorch_sphinx_theme
```
mthrok, line 2: The reason why this is pinned in audio is that we customize the behavior of HTML so that tutorials have links like `Run in Google Colab`, `Download Notebook` and `View on GitHub`.

To achieve the same effect, you might also want to check-in https://github.com/pytorch/audio/blob/main/docs/source/_templates/layout.html with modification of `audio` -> `text`.

examples/tutorials/sst2_classification_non_distributed.py
```diff
@@ -0,0 +1,156 @@
...
+    loss, accuracy = evaluate()
+    print("Epoch = [{}], loss = [{}], accuracy = [{}]".format(e, loss, accuracy))
+    MODEL.train()
+    for batch in TRAIN_DATAPIPE:
+        input = F.to_tensor(batch['token_ids'], padding_value=PADDING_IDX).to(DEVICE)
+        target = torch.tensor(batch['target']).to(DEVICE)
+        train_step(input, target)
+
+loss, accuracy = evaluate()
+print("Epoch = [{}], loss = [{}], accuracy = [{}]".format(NUM_EPOCHS), loss, accuracy)
```
mthrok, line 156: It might be a good idea to show the result of the training. like running an inference on couple of examples, if that can be done without adding too much.

mthrok (commented):
Looks good so far.

mthrok:
I also recommend adding this customization to `conf.py` so that link from documentation to tutorial automatically shows up without the use of explicit `.. minigallery::`.

https://github.com/pytorch/audio/blob/0a072f9a5cc1c8b23a4aed44873237404f4386c1/docs/source/conf.py#L320-L338

https://github.com/pytorch/audio/pull/2101

examples/tutorials/sst2_classification_non_distributed.py
parmeet: Thanks for catching (since I was testing on CPU, this code never executed ;) )

docs/source/conf.py
```diff
@@ -107,7 +169,7 @@
 # documentation.
 #
 html_theme_options = {
-    'pytorch_project': 'docs',
```
mthrok, line 110: I think this should be `'text'`, though I am not sure what difference it makes.

docs/requirements.txt
parmeet: I made the changes though I don't see those link in my local doc builds. Not sure if I am doing something wrong?

docs/source/_templates/layout.html
```diff
@@ -6,3 +6,69 @@
...
+{#
+    ################################################################################
+    # Adding Colab / notebook header like tutorials repo
+    # Based off of
+    # https://github.com/pytorch/pytorch_sphinx_theme/blob/fe1f3d5b9233497d81d04f55f5750ccad92500be/pytorch_sphinx_theme/layout.html#L275-L319
+    ################################################################################
+#}
+
+{%- block content %}
+    {% if pagename.endswith('tutorial') %}
```
mthrok, line 19:
```suggestion
    {% if 'tutorial' in pagename %}
```

Should do the trick.

Screen Shot 2022-01-11 at 20 11 15

docs/requirements.txt
mthrok: okay, I did not test the tutorials without `_tutorial` suffix.

See https://github.com/pytorch/text/pull/1468#discussion_r782639168 for the fix.

parmeet: > See [#1468 (comment)](https://github.com/pytorch/text/pull/1468#discussion_r782639168) for the fix.

cool, this worked. thanks!

mthrok:
> this error `intersphinx inventory 'http://docs.scipy.org/doc/numpy/objects.inv' not readable due to ValueError` ?

The URLs were deprecated a while ago, probably now it's removed. 

They need to be updated, like https://github.com/pytorch/audio/pull/1893/files

I thought there was an open PR for that, but I can't find it.

docs/source/_templates/layout.html
```diff
@@ -6,3 +6,69 @@
...
+		  notebookLink = $(".reference.download")[1].href,
+		  notebookDownloadPath = notebookLink.split('_downloads')[1],
+		  colabLink = "https://colab.research.google.com/github/pytorch/text/blob/gh-pages/main/_downloads" + notebookDownloadPath;
+
+	      $(".pytorch-call-to-action-links a[data-response='Run in Google Colab']").attr("href", colabLink);
+	      $(".pytorch-call-to-action-links a[data-response='View on Github']").attr("href", githubLink);
+	  }
+      });
+    </script>
+{% endblock %}
```
mthrok, line 74: It's probably better to have the new line at the end, otherwise it might fail the lint job that is being added in the other PR.

mthrok (approved):
It looks good. But I would like to double check the built tutorial after the doc job is fixed.

mthrok (approved):
LGTM

PR #1469, opened 2021-12-22
Migrate GPT-2 BPE Encoder logic to C++

abhinavarora:
## Description
This PR  does the following: 
1. Migrates the GPT-2 BPE encoder to C++ and exposes it via Torchbind and PyBind. 
2. Modifies the Python implementation of `GPT2BPETokenizer` to use the C++ encoder.
3. Add support for caching in eager mode
4. Add pickle support for tokenizer in both eager and torchscript mode

## Testing

```
pytest -k test_gpt2_bpe_tokenizer test/test_transforms.py
```

## Follow ups

- [ ] Benchmark performance against fairseq tokenizer
- [ ] Pass c10::string_views instead of strings from PyBind


torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -1,68 +1,242 @@
...
+      }
+    } else if (prepend_space) {
+      tokens.push_back(" " + token);
+      prepend_space = false;
+    } else {
+      tokens.push_back(token);
+    }
+  }
+  return tokens;
+}
```
mthrok, line 80: @abhinavarora Up to this point, there is only style changes and no functional change, is that correct?

abhinavarora: @mthrok That's correct. I found the style to be inconsistent across our files, so for the time being I used the Google formatter in these recently added files.

torchtext/transforms.py
```diff
@@ -236,187 +236,16 @@ def __init__(
...
-        bpe_token_ids: List[int] = []
-        for token in self._regex(text):
-            for bpe_token in self._bpe(self._byte_encode(token)):
-                bpe_token_ids.append(self.bpe_encoder[bpe_token])
-        return bpe_token_ids
+        self.bpe = torch.classes.torchtext.GPT2BPEEncoder(bpe_encoder, bpe_merge_ranks,
+                                                          self._seperator, bytes_to_unicode())
 
     @torch.jit.export
     def tokenize(self, text: str) -> List[str]:
```
parmeet, line 251: Perhaps we can also remove this method and call encode method directly in the forward function?

parmeet, line 251: Also i missed this in previous PR, but I think converting token IDs into string is something we are doing internally to take care of vocab. I am not sure if this is a standard practice. To keep it consistent with Fairseq and OpenAI implementation, we should probably just return integer IDs. WDYT?

abhinavarora: I would suggest let's keep the ids as string for now till we standardize the tokenization API. My understanding is that the models that use this downstream expect strings as they build the vocab. Even the GPT-2 tokenizer for the model in HF transformers repo creates string ids.  [Reference code](https://github.com/huggingface/transformers/blob/master/src/transformers/models/gpt2/tokenization_gpt2.py#L252)

abhinavarora: Updated the PR to make the `tokenize` method private.

parmeet (approved):
LGTM! Would be great to add some benchmarks against the python implementation in fairseq as a followup :).

torchtext/csrc/register_torchbindings.cpp
```diff
@@ -122,10 +123,27 @@ TORCH_LIBRARY_FRAGMENT(torchtext, m) {
...
+          // __setstate__
+          [](GPT2BPEEncoderStatesTorchbind states)
+              -> c10::intrusive_ptr<GPT2BPEEncoder> {
+            return _deserialize_gpt2_bpe_encoder_torchbind(states);
+          });
+
   m.def("torchtext::generate_sp_model", &generate_sp_model);
   m.def("torchtext::load_sp_model", &load_sp_model);
   m.def("torchtext::load_sp_model_string", &load_sp_model_string);
   m.def("torchtext::gpt2_bpe_pre_tokenizer", &gpt2_bpe_pre_tokenizer);
```
parmeet, line 146: I guess we no longer need to bind this method right?

torchtext/csrc/register_pybindings.cpp
```diff
@@ -153,6 +155,29 @@ PYBIND11_MODULE(_torchtext, m) {
...
+      .def(py::pickle(
+          // __getstate__
+          [](const c10::intrusive_ptr<GPT2BPEEncoder> &self)
+              -> GPT2BPEEncoderStatesPybind {
+            return _serialize_gpt2_bpe_encoder_pybind(self);
+          },
+          // __setstate__
+          [](GPT2BPEEncoderStatesPybind states)
+              -> c10::intrusive_ptr<GPT2BPEEncoder> {
+            return _deserialize_gpt2_bpe_encoder_pybind(states);
```
parmeet, line 178: I am not sure if we have added tests for loading and saving the tokenizer, Could you please check? We have similar tests for [Vocab](https://github.com/pytorch/text/blob/be7bc66e9d720850850ce8551af021c957cadf03/test/test_vocab.py#L193).

torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -1,68 +1,318 @@
...
+    } else {
+      pairs = _get_pairs(tok_list, seperator_);
+    }
+  }
+
+  if (caching_enabled_) cache_.insert(concatenated, tok_list);
+  return tok_list;
+}
+
+std::vector<int64_t> GPT2BPEEncoder::Encode(const std::string &text) {
```
parmeet, line 255: We could instead use C10::string_view to avoid copying the strings. [Here](https://github.com/pytorch/text/blob/be7bc66e9d720850850ce8551af021c957cadf03/torchtext/csrc/vocab.cpp#L35) is how it is done for vocab. Then we can register like [here](https://github.com/pytorch/text/blob/be7bc66e9d720850850ce8551af021c957cadf03/torchtext/csrc/register_pybindings.cpp#L119) for pybind to potentially gain performance.

torchtext/csrc/register_torchbindings.cpp
abhinavarora: We need to keep this method bound for internal GPT2 implementation that currently calls this operator. We can remove it later after internal dependency is removed and everything internally is migrated to the OSS tokenizer.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
abhinavarora: That sounds good. I can take this in a follow up PR as there will be few code changes in some private methods as well, including the interface of the `gpt2_bpe_pre_tokenizer ` method. Adding a follow up item for the same!

torchtext/csrc/register_pybindings.cpp
abhinavarora: Added in latest commit.

parmeet:
LGTM! Thanks for addressing the comments and adding the follow-up items. Feel free to merge, already approved before :)

PR #1472, opened 2021-12-22
[WIP] add sentencepiece from_iterator function.

### LINKED ISSUE #1471: Generate sentencepiece model from iterable
## 🚀 Feature

**Motivation**

The current method of training a sentencepiece model requires a file to be passed. It would be nice if this was not required.

**Pitch**

Like other data-related functions in torchtext, there should be a `_from_iterator` function to train a sentencepiece model.

**Alternatives**

For some datasets, it may be prohibitively expensive to continually replenish the iterator on each pass. A middle-of-the-road approach would be to allow passing an iterator, write the contents to a temp file, and call the "classic" function using the temp file which is then cleaned up.

**Additional context**

N/A

erip:
Fixes #1471.

### DIFF
--- a/torchtext/data/functional.py
+++ b/torchtext/data/functional.py
@@ -10,7 +10,6 @@ __all__ = [
     "to_map_style_dataset",
 ]
 
-
 """
 This file contains experimental functionality.
 All of these are experimental, unstable, and subject to change or deletion.
@@ -38,6 +37,29 @@ def generate_sp_model(filename, vocab_size=20000,
         >>> generate_sp_model('test.csv', vocab_size=23456, model_prefix='spm_user')
     """
     torch.ops.torchtext.generate_sp_model(filename, vocab_size, model_type, model_prefix)
+
+
+def generate_sp_model_from_iterator(lines, vocab_size=20000,
+                                    model_type="unigram",
+                                    model_prefix='m_user'):
+    r"""Train a SentencePiece tokenizer.
+
+    Args:
+        lines: an iterator containing data for training SentencePiece model.
+        vocab_size: the size of vocabulary (Default: 20,000).
+        model_type: the type of SentencePiece model, including unigram,
+            bpe, char, word.
+        model_prefix: the prefix of the files saving model and vocab.
+
+    Outputs:
+        The model and vocab are saved in two separate files with
+            model_prefix.
+
+    Examples:
+        >>> from torchtext.data.functional import generate_sp_model_from_iterator
+        >>> generate_sp_model_from_iterator(lines, vocab_size=23456, model_prefix='spm_user')
+    """
+    torch.ops.torchtext.generate_sp_model_from_iterator(lines, vocab_size, model_type, model_prefix)
 
 
 def load_sp_model(spm):
@@ -90,6 +112,7 @@ def sentencepiece_numericalizer(sp_model):
     def _internal_func(txt_iter):
         for line in txt_iter:
             yield sp_model.EncodeAsIds(line)
+
     return _internal_func
 
 
@@ -116,6 +139,7 @@ def sentencepiece_tokenizer(sp_model):
     def _internal_func(txt_iter):
         for line in txt_iter:
             yield sp_model.EncodeAsPieces(line)
+
     return _internal_func
 
 
@@ -138,6 +162,7 @@ def custom_replace(replace_pattern):
             for pattern_re, replaced_str in _patterns:
                 line = pattern_re.sub(replaced_str, line)
             yield line
+
     return _internal_func
 
 
@@ -180,7 +205,7 @@ def numericalize_tokens_from_iterator(vocab, iterator, removed_tokens=None):
             yield iter(vocab[token] for token in tokens)
         else:
             yield iter(map(lambda x: vocab[x],
-                       filter(lambda x: x not in removed_tokens, tokens)))
+                           filter(lambda x: x not in removed_tokens, tokens)))
 
 
 _patterns = [(r'<.*>', ''),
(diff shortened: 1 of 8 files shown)


erip:
One issue here is that the `generate_sp_from_iterator` is not JIT'able because it relies on `Python.h`. This is not a good limitation, but maybe there's a way to move it so it's only exposed to the Python frontend. I'm not so familiar with the organization of torchtext, so if a maintain has thoughts about how to realize this, I'm happy to make the change.

parmeet:
> One issue here is that the `generate_sp_from_iterator` is not JIT'able because it relies on `Python.h`. This is not a good limitation and is likely a non-starter, but maybe there's a way to move it so it's only exposed to the Python frontend. I'm not so familiar with the organization of torchtext, so if a maintain has thoughts about how to realize this, I'm happy to make the change.

We could implement the function in a separate file and register it through Pybind only. [_build_vocab_from_text_file_using_python_tokenizer](https://github.com/pytorch/text/blob/52d38e8de628384c8adf4461baa7430de854b731/torchtext/csrc/vocab_factory.h#L8) is an example that is registered only through pybind11 [here](https://github.com/pytorch/text/blob/52d38e8de628384c8adf4461baa7430de854b731/torchtext/csrc/register_pybindings.cpp#L161).

> Otherwise if there's a "nicer" way to have a JITted version of this by cutting the reliance on the Python header, that's a better option.

I am not so sure if Iterators are supported through torchscript. One alternative would be to materialize the iterator and pass `List[str]`, but I guess that would defeat the purpose of iterator in the first place.

erip:
Thanks for the pointers, @parmeet! I think following that separate pybind recipe should do nicely. Regarding iterators in torchscript: I doubt JIT will give us anything in this one-off training anyway, so maybe not a big deal. I can get something working and we can revisit as needed.

parmeet:
> Regarding iterators in torchscript: I doubt JIT will give us anything in this one-off training anyway, so maybe not a big deal. I can get something working and we can revisit as needed.

Right, i do not expect any speed-ups and such using JIT. Also torch-scriptability is something that's concerned with Inference in non-pythonic environments. I wonder if there are use-cases when we need/want to train SPM in non-pythonic environments. I am not yet aware of such requirements, so I guess it's a low-risk to drop support for scripting in this particular case.

erip:
Looks like the error is [here](https://github.com/pytorch/text/pull/1472/files#diff-2d0b522718ce0d246b154a2d68180ac4fbd151ba4ebb1e1b33ca4a06948f5bf8R44). I'm not positive what's going on but I'll dig into it.

torchtext/data/functional.py
```diff
@@ -40,6 +41,29 @@ def generate_sp_model(filename, vocab_size=20000,
...
+
+    Outputs:
+        The model and vocab are saved in two separate files with
+            model_prefix.
+
+    Examples:
+        >>> from torchtext.data.functional import generate_sp_model_from_iterator
+        >>> generate_sp_model_from_iterator(lines, vocab_size=23456, model_prefix='spm_user')
+    """
+    return _generate_sp_model_from_iterator(lines, vocab_size, model_type, model_prefix)
```
mthrok, line 64: I think this can be combined with the existing `generate_sp_model`.

```python
def generate_sp_model(...):
    if not torch.jit.is_scripting():
        if not <the first arg is path-like object>
            return torchtext._torchtext._generate_sp_model_from_iterator(...)
    return torch.ops.torchtext.generate_sp_model(...)
```

erip:
It looks like the one remaining issue is surrounding the vocab size of the trained model. The test is failing with

```error
RuntimeError: Failed to train SentencePiece model. Error: Internal: .../text/third_party/sentencepiece/src/trainer_interface.cc(579) [(trainer_spec_.vocab_size()) == (model_proto->pieces_size())] Vocabulary size too high (23456). Please set it to a value <= 9575.
```

torchtext/data/functional.py
erip: I like this approach, but one concern I have is that the first arg of `generate_sp_model` is `filename` -- it would be misleading to users to have an iterable passed with name "filename" and changing it is a breaking change (albeit a silly one since it's a positional arg). WDYT?

erip:
At present the code doesn't error out, but the test fails. I've opened an question issue [here](https://github.com/google/sentencepiece/issues/717).

erip:
I was informed that this is not guaranteed to generate the same form of model as `spm-train` which is a bit disappointing. Because the feature is undocumented, I'll close this for now...

nateanl:
@erip Which part is not guaranteed to generate the same form of model as ``spm-train``? In torchaudio we train the sentencepiece model using sentence_iterator like this:
```python
def train_spm(input):
    model_writer = io.BytesIO()
    spm.SentencePieceTrainer.train(
        sentence_iterator=iter(input),
        model_writer=model_writer,
        vocab_size=4096,
        model_type="bpe",
        input_sentence_size=-1,
        character_coverage=1.0,
        bos_id=0,
        pad_id=1,
        eos_id=2,
        unk_id=3,
    )
    return model_writer.getvalue()
```
Does that generate the same model as ``spm-train``? 

cc @hwangjeff

erip:
I'm sort of reading between the lines from [this](https://github.com/google/sentencepiece/issues/717#issuecomment-1041056278), but your approach may work. I'm not sure if there are differences between the bpe path and the unigram path but I can try soon.

erip:
Yes, it seems like using this code I still cannot generate the same model:

> RuntimeError: Internal: /Users/travis/build/google/sentencepiece/src/trainer_interface.cc(590) [(trainer_spec_.vocab_size()) == (model_proto->pieces_size())] Vocabulary size too high (23456). Please set it to a value <= 10426.

Whether it's good to build a model with sentencepiece vocab size 23456 on such small data is one question, but whether we can technically do it seems to be another. 😓

PR #1473, opened 2021-12-23
Fix flake8 errors in PyTorch Text repo

abhinavarora:


parmeet (approved):
LGTM! Curious why it didn't show up in our CI testing? Perhaps the linter only covers whats in main torchtext and test repo? Should we add examples directly for style-check?

abhinavarora:
@parmeet That's a good question. I have no idea. I can check out what's the issue. Could you elaborate more on what you mean by `Should we add examples directly for style-check?`

PR #1475, opened 2021-12-31
fix optionality of default arg

erip:


erip:
Failure is unrelated

Nayef211 (approved):
Thanks for correcting this @erip. Approved!

PR #1479, opened 2022-01-02
add AddSpecialTokens transform

erip:
Adds an analog to the `AddToken` transform which prepends BOS and appends EOS.

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -203,6 +203,45 @@ class TestTransforms(TorchtextTestCase):
     def test_add_token_jit(self):
         self._add_token(test_scripting=True)
 
+    def _add_special_tokens(self, test_scripting):
+        bos_id = 0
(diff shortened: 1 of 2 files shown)


parmeet:
> Adds an analog to the `AddToken` transform which prepends BOS and appends EOS.

Thanks @erip for the PR. I am curious if it is necessary to add this transforms given that we can already accomplish this using AddToken? Typically we try to to add transforms that are primitive (although there is no hard rule or such), we should be deliberate when adding new transforms if they can be already done by composing existing transforms. Could you please provide additional arguments on why this should be a useful addition and worth the cost of maintaining?

erip:
My thought was fundamentally that it's such a common op that including it in core made sense so consumers don't necessarily need to reinvent it. If you think it doesn't make sense to add it, I can close this PR.

PR #1481, opened 2022-01-04
add scriptable sequential transform

parmeet:
This PR adds a scriptable version of [torch.nn.Sequential](https://pytorch.org/docs/stable/generated/torch.nn.Sequential.html) in the context of text transforms.

The transforms composed using torch.nn.Sequential cannot be scripted, since the expected input (inferred) for torch.nn.Sequential is torch Tensor. By modifying the forward method to accept `Any`, the inferred type is no longer Tensor, and hence the resultant transform can be torch-scripted.


torchtext/experimental/transforms.py
```diff
@@ -164,25 +163,6 @@ def __prepare_scriptable__(self):
         return RegexTokenizer(regex_tokenizer)
 
 
-class TextSequentialTransforms(nn.Sequential):
```
abhinavarora, line 167: Could you also clean up the tests associated with this Transform. The unit tests are currently failing for this.

test/test_transforms.py
```diff
@@ -203,6 +203,31 @@ def test_add_token(self):
     def test_add_token_jit(self):
         self._add_token(test_scripting=True)
 
+    def _sequential(self, test_scripting):
```
abhinavarora, line 206: Could we put this in a separate test class to isolate the tests for each transform. I know we discussed this before, don't remember the outcome of our discussion :-)

Nayef211: @abhinavarora could you clarify what you mean by this? Are you suggesting to have tests foreach transform in a completely separate test class?

parmeet: yes we wanted to revamp our test-suite ( for reference here is our previous conversation on this https://github.com/pytorch/text/pull/1463#discussion_r770988886). Should we take it up as a separate discussion/PR item? Separating out sequential in this PR would leave out TestTransforms class inconsistent.

abhinavarora: @Nayef211 your understanding is correct. The discussion was about making Test classes smaller and responsible for specific transforms. I see this practice in fbcode and it makes diving into the tests much easier. Many users from PyText would look at tests to better understand the API and this just makes it easier for them to isolate the code they need to check.

@parmeet Would it make sense if all new transforms like this one and the GPT2 be self-contained and that revamp task can apply only to the existing transforms. The idea is that new code can follow the new practice, so that the revamp task doesn't grow with time. Wdyt?

parmeet: hmm, I see your point! Sure, I don't think it would hurt to do so, and it will get us started toward this effort :). Let me push the code with these changes.

abhinavarora (approved):
LGTM! Thanks for making the changes :-)

PR #1482, opened 2022-01-04
remove xlmr transform class and instead use sequential for model transforms composition

parmeet:
## Background
During adding support for pre-trained models in #1406, we also added[ XLMR model transform class](https://github.com/pytorch/text/blob/be7bc66e9d720850850ce8551af021c957cadf03/torchtext/models/roberta/transforms.py#L11) to facilitate Model Bundle provide complete pre-processing (transformation) for input text. As such, the value of this additional class is not proven yet, and it was added only to cover for the lack of equivalent transforms that can be composed together in scriptable way, using some form of container class.

## Proposal
As we added necessary primitive transforms, along with support of [Scriptable sequential container](https://github.com/pytorch/text/pull/1481) for text transforms, we can easily avoid overhead of maintaining a separate class for model transform. Instead in this PR, we argue that this can be accomplished in much more sustainable way by directly constructing sequential transform as part of Model Bundle object.

## BC breakage

The above class is used in [text recipes](https://github.com/facebookresearch/recipes/blob/880451f1b86481f5639d4ac02f0f6f27598d9f58/torchrecipes/text/doc_classification/transform/doc_classification_text_transform.py#L16). To resolve it, we could do a similar composition directly inside doc classification transform.  CC: @Nayef211


torchtext/transforms.py
```diff
@@ -485,3 +485,17 @@ def bytes_to_unicode():
             n += 1
     cs = [chr(n) for n in cs]
     return dict(zip(bs, cs))
+
+
+class Sequential(torch.nn.Sequential):
```
erip, line 491: Does it make sense to break this PR into two separate commits -- one which brings `Sequential` out of experimental and another which migrates `XLMRTransform`?

parmeet: Indeed yes :). I have another PR to cover `Sequential` https://github.com/pytorch/text/pull/1481. Since that PR hasn't landed yet, I just created this PR out of the feature branch. I wonder if there are better ways to handle this? @mthrok any thoughts on creating new PR on existing PR(s)?

erip: One idea would be to branch off the feature branch where Sequential is un-experimentalized so you have the same history from this PR. I think if you do that, once that is merged you can just rebase main onto this branch for linear history.

mthrok: > Indeed yes :). I have another PR to cover `Sequential` #1481. Since that PR hasn't landed yet, I just created this PR out of the feature branch. I wonder if there are better ways to handle this? @mthrok any thoughts on creating new PR on existing PR(s)?

If the question is what I think it is, then I do not have a good solution. I got used to deal with multiple of PRs and commits. (at the moment I am juggling about 5 to 6 branches covering 3 different, independent aspects of one project.) Maybe gh-stack might help, but I have not tried yet. https://github.com/ezyang/ghstack

torchtext/models/roberta/bundler.py
```diff
@@ -156,10 +155,13 @@ def encoderConf(self) -> RobertaEncoderConf:
...
     _path=urljoin(_TEXT_BUCKET, "xlmr.base.encoder.pt"),
     _encoder_conf=RobertaEncoderConf(vocab_size=250002),
-    transform=partial(get_xlmr_transform,
-                      vocab_path=urljoin(_TEXT_BUCKET, "xlmr.vocab.pt"),
-                      spm_model_path=urljoin(_TEXT_BUCKET, "xlmr.sentencepiece.bpe.model"),
-                      )
+    transform=lambda: T.Sequential(
+        T.SentencePieceTokenizer(urljoin(_TEXT_BUCKET, "xlmr.sentencepiece.bpe.model")),
+        T.VocabTransform(load_state_dict_from_url(urljoin(_TEXT_BUCKET, "xlmr.vocab.pt"))),
+        T.Truncate(510),
```
Nayef211, line 161: I am curious as to why we are truncating to a length of 510 exactly? In the XLMR transform that is being removed, I noticed that `max_seq_len` was an input arg with a default value of 514 and we would truncate to a length of `self.max_seq_len - 2` which would be 512 instead of 510 right?

torchtext/models/roberta/bundler.py
```diff
@@ -156,10 +155,13 @@ def encoderConf(self) -> RobertaEncoderConf:
...
+        T.AddToken(token=0, begin=True),
+        T.AddToken(token=2, begin=False),
```
Nayef211, line 163: Is it guaranteed that the `bos` and `eos` tokens will always have an index of 0 and 2 respectively? Could there be a usecase where the user uses their own vocab instead of loading from `xlmr.vocab.pt`?

torchtext/models/roberta/bundler.py
erip: I think 510 + the BOS and EOS is right. The 514 seems like an off-by-two.

torchtext/models/roberta/bundler.py
erip: Certainly plausible that they'd add, say, additional special tokens. They would be appended to the vocab though, so 0 and 2 should remain as BOS and EOS respectively. If they didn't, you'd break the "contract" with how RoBERTa was trained.

parmeet: These are pre-trained models, and the transform comply with what the model was originally trained with. Certainly users can bring in their own flavors or Sentencepiece models, vocab etc, but then that transform is specific to user's model and won't apply to the pre-trained weights.

torchtext/models/roberta/bundler.py
```diff
@@ -156,10 +155,13 @@ def encoderConf(self) -> RobertaEncoderConf:
 XLMR_BASE_ENCODER = RobertaModelBundle(
     _path=urljoin(_TEXT_BUCKET, "xlmr.base.encoder.pt"),
     _encoder_conf=RobertaEncoderConf(vocab_size=250002),
-    transform=partial(get_xlmr_transform,
-                      vocab_path=urljoin(_TEXT_BUCKET, "xlmr.vocab.pt"),
-                      spm_model_path=urljoin(_TEXT_BUCKET, "xlmr.sentencepiece.bpe.model"),
-                      )
+    transform=lambda: T.Sequential(
```
mthrok, line 158: Assigning lambda is an anti-pattern.

https://docs.quantifiedcode.com/python-anti-patterns/correctness/assigning_a_lambda_to_a_variable.html

torchtext/models/roberta/bundler.py
mthrok: Sounds like some tests would help.

torchtext/models/roberta/bundler.py
Nayef211: Got it thanks for the clarification!

torchtext/models/roberta/bundler.py
mthrok: hmm, reading this, it might not

https://stackoverflow.com/a/27928036

mthrok: yet I feel overwhelmed by the fact that this lambda spans multiple lines.

parmeet: hmm, interesting I wasn't aware of this around lambda usage. thanks for sharing. I honestly thought it looks nicer being able to define a lazy transform in-place :). The alternative would be to define a private function that returns the composite transform.

mthrok: yeah, I am looking at all the possible reasoning, but I am also being convinced that `lambda` is the simplest here.

erip: Another way would be to compose a no-op lambda with the rest of the transform with `partial`, though it isn't as easy to read.

torchtext/transforms.py
abhinavarora: Here is a nice blog post by the author of ghstack: https://0xc0d1.com/blog/git-stack/

abhinavarora (approved):
Thank you for the PR @parmeet . I found Sequential to be much more intuitive than the previous transform. This certainly makes the code easier to read for a user.

LGTM!

PR #1484, opened 2022-01-05
Remove Python 3.6 support as it has reached EOL

abhinavarora:
Python 3.6 reached [EOL](https://endoflife.date/python) last month. PyTorch core is going to do the same in pytorch/pytorch#69005. Other libraries doing the same are:
- Torchvision pytorch/vision#5152
- Torchdata pytorch/data#156
- Torchaudio pytorch/audio#2051

This should also fix the CircleCI failures in 3_6 conda builds

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -614,9 +614,6 @@ workflows:
     jobs:
       - circleci_consistency
       - binary_linux_wheel:
-          name: binary_linux_wheel_py3.6
-          python_version: '3.6'
-      - binary_linux_wheel:
           name: binary_linux_wheel_py3.7
           python_version: '3.7'
       - binary_linux_wheel:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -613,9 +613,6 @@ workflows:
   build:
```
mthrok, line 613: FYI: You can use `.gitattribute` file to tell git that this is a generated file. That way, it will be hidden in PR by default.

https://github.com/pytorch/audio/blob/main/.gitattributes

mthrok (approved):

Nayef211 (approved):

.circleci/config.yml
abhinavarora: That's a good callout . Will send you a PR shortly!

PR #1488, opened 2022-01-05
Switching to use FileOpener from FileLoader

NivekT:
TorchData is changing its API, renaming `FileLoader` to `FileOpener`.  See https://github.com/pytorch/pytorch/pull/70367

This PR changes the usage in TorchText experimental to match the new API.


Nayef211 (approved):
Lgtm, thanks for fixing this! I will import this PR to phabricator after you land to ensure that this fixes the failing tests for torchrecipes we saw earlier :)

PR #1490, opened 2022-01-06
Migrating AmazonReviewPolarity to datapipes

parmeet:
Reference Issue: https://github.com/pytorch/text/issues/1494

Summary
* Migrate AmazonReviewPolarity
* Importing torchdata conditional and raising exception during function call if package is not available
* Update CI for datasets testing

### DIFF
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -41,20 +41,6 @@ class TestDataset(TorchtextTestCase):
     @parameterized.expand(
         load_params('raw_datasets.jsonl'),
         name_func=_raw_text_custom_name_func)
-    def test_raw_text_name_property(self, info):
-        dataset_name = info['dataset_name']
-        split = info['split']
-
-        if dataset_name == 'WMT14':
-            return
-        else:
-            data_iter = torchtext.datasets.DATASETS[dataset_name](split=split)
-
-        self.assertEqual(str(data_iter), dataset_name)
-
-    @parameterized.expand(
-        load_params('raw_datasets.jsonl'),
-        name_func=_raw_text_custom_name_func)
     def test_raw_text_classification(self, info):
         dataset_name = info['dataset_name']
         split = info['split']
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -63,8 +63,7 @@ class TestDataset(TorchtextTestCase):
             return
         else:
             data_iter = torchtext.datasets.DATASETS[dataset_name](split=split)
-        self.assertEqual(len(data_iter), info['NUM_LINES'])
-        self.assertEqual(hashlib.md5(json.dumps(next(data_iter), sort_keys=True).encode('utf-8')).hexdigest(), info['first_line'])
+        self.assertEqual(hashlib.md5(json.dumps(next(iter(data_iter)), sort_keys=True).encode('utf-8')).hexdigest(), info['first_line'])
         if dataset_name == "AG_NEWS":
             self.assertEqual(torchtext.datasets.URLS[dataset_name][split], info['URL'])
             self.assertEqual(torchtext.datasets.MD5[dataset_name][split], info['MD5'])
--- a/torchtext/data/datasets_utils.py
+++ b/torchtext/data/datasets_utils.py
@@ -209,8 +209,7 @@ def _wrap_split_argument_with_fn(fn, splits):
             argspec.args[1] == "split" and
             argspec.varargs is None and
             argspec.varkw is None and
-            len(argspec.kwonlyargs) == 0 and
-            len(argspec.annotations) == 0
+            len(argspec.kwonlyargs) == 0
             ):
         raise ValueError("Internal Error: Given function {} did not adhere to standard signature.".format(fn))
 
@@ -246,10 +245,9 @@ def _create_dataset_directory(dataset_name):
                 argspec.args[1] == "split" and
                 argspec.varargs is None and
                 argspec.varkw is None and
-                len(argspec.kwonlyargs) == 0 and
-                len(argspec.annotations) == 0
+                len(argspec.kwonlyargs) == 0
                 ):
-            raise ValueError("Internal Error: Given function {} did not adhere to standard signature.".format(fn))
+            raise ValueError("Internal Error: Given function {} did not adhere to standard signature.".format(func))
 
         @functools.wraps(func)
         def wrapper(root=_CACHE_DIR, *args, **kwargs):
--- a/torchtext/data/datasets_utils.py
+++ b/torchtext/data/datasets_utils.py
@@ -239,22 +239,22 @@ def _wrap_split_argument(splits):
 
 
 def _create_dataset_directory(dataset_name):
-    def decorator(func):
-        argspec = inspect.getfullargspec(func)
+    def decorator(fn):
+        argspec = inspect.getfullargspec(fn)
         if not (argspec.args[0] == "root" and
                 argspec.args[1] == "split" and
                 argspec.varargs is None and
                 argspec.varkw is None and
                 len(argspec.kwonlyargs) == 0
                 ):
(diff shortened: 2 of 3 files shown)


erip:
Is this ready for review, @parmeet? I think this LGTM

parmeet:
> Is this ready for review, @parmeet? I think this LGTM

I think it is now. My only concern was around the conditional dependency on torchdata. For now I am raising ModuleNotFoundError, just in case a user is working with the source directly.

erip (commented):
This seems like a better strategy, IMO. I think before this the user would experience a `NameError` if they didn't have torchdata installed but tried to use this dataset. This seems more friendly.

torchtext/datasets/amazonreviewpolarity.py
```diff
@@ -25,20 +27,33 @@
...
-}
 
 DATASET_NAME = "AmazonReviewPolarity"
 
 
 @_add_docstring_header(num_lines=NUM_LINES, num_classes=2)
 @_create_dataset_directory(dataset_name=DATASET_NAME)
-@_wrap_split_argument(('train', 'test'))
+@_wrap_split_argument(("train", "test"))
 def AmazonReviewPolarity(root, split):
```
erip, line 37: Nit: do we want to take this as an opportunity to add typing info here?

abhinavarora: Agree on this one. Typing info and maybe comments on how the class is used would be quite useful and also consistent with our codebase.

test/data/test_builtin_datasets.py
```diff
@@ -38,20 +38,6 @@ def test_raw_ag_news(self):
         self._helper_test_func(len(test_iter), 7600, next(test_iter)[1][:25], 'Fears for T N pension aft')
         del train_iter, test_iter
 
-    @parameterized.expand(
```
abhinavarora, line 41: A n00b question. Why did we remove this and how this change relate to migration of AmazonReviewPolarity to datapipes ?

torchtext/datasets/amazonreviewpolarity.py
```diff
@@ -25,20 +27,24 @@
...
-    path = _download_extract_validate(root, URL, MD5, os.path.join(root, _PATH), os.path.join(root, _EXTRACTED_FILES[split]),
-                                      _EXTRACTED_FILES_MD5[split], hash_type="md5")
-    logging.info('Creating {} data'.format(split))
-    return _RawTextIterableDataset(DATASET_NAME, NUM_LINES[split],
-                                   _create_data_from_csv(path))
+    # TODO Remove this after removing conditional dependency
+    if not is_module_available("torchdata"):
+        raise ModuleNotFoundError("Package `torchdata` not found. Please install following instructions at `https://github.com/pytorch/data`")
```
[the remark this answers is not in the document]
erip: I think once there's a stable release of torchdata published that this check will disappear. Is it still worth the test since it'll be removed soon anyway?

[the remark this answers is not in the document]
parmeet: I agree, this is temporary. We can probably skip adding test.

torchtext/datasets/amazonreviewpolarity.py
```diff
@@ -1,13 +1,15 @@
+from torchtext._internal.module_utils import is_module_available
+
+if is_module_available("torchdata"):
+    from torchdata.datapipes.iter import FileOpener, GDriveReader, IterableWrapper
+
```
Nayef211, line 5: I wonder if we also need to import `HttpReader` so that we can use this dataset internally within FB by swapping out public URLs with internal URLs. This is what I'm doing for the SST2 dataset implementation (https://github.com/Nayef211/text/blob/main/torchtext/experimental/datasets/sst2.py#L15-L17)

parmeet: That's a good catch @Nayef211. Note that in this case we need GDriveReader. Since this is not yet implementation internally, let fix this part once we import this PR into fb-code.

torchtext/datasets/amazonreviewpolarity.py
parmeet: Added the annotations. The usage of datasets is common across all. As of now, we have added generic usage (https://pytorch.org/text/stable/datasets.html#torchtext-datasets, this need to be updated once we migrate to data pipes) for datasets that have only `root` and `split` as arguments. There are other datasets that have additional `kwargs` (example: https://pytorch.org/text/stable/datasets.html#iwslt2016) for which we may add additional doc-string examples.

torchtext/data/datasets_utils.py
```diff
@@ -246,10 +245,9 @@ def decorator(func):
                 argspec.args[1] == "split" and
                 argspec.varargs is None and
                 argspec.varkw is None and
-                len(argspec.kwonlyargs) == 0 and
-                len(argspec.annotations) == 0
+                len(argspec.kwonlyargs) == 0
                 ):
-            raise ValueError("Internal Error: Given function {} did not adhere to standard signature.".format(fn))
+            raise ValueError("Internal Error: Given function {} did not adhere to standard signature.".format(func))
```
Nayef211, line 250: NIT: wondering if we want to rename `func` to be `fn` here to be consistent with the naming in the [`_wrap_split_argument_with_fn` method](https://github.com/pytorch/text/blob/97ee0c13db433bb0cb8a0876476618ffe642916b/torchtext/data/datasets_utils.py#L195)

Nayef211 (approved):
LGTM!

torchtext/datasets/amazonreviewpolarity.py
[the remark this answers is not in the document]
abhinavarora: Sounds reasonable to me.

mthrok:
I am curious how you see and deal with backward compatibility.

Assuming that the API is not changed, the new implementation will require the new dependency, torchdata. So exciting installation will fail without installing torchdata. How do you communicate the change to users?

erip:
@mthrok I think releasing torchtext 0.12.0 will be blocked until a stable version of torchdata is cut, after which it becomes a required dependency. For most end users it should be a fairly transparent migration since there really isn't any breaking change.

torchtext/datasets/amazonreviewpolarity.py
```diff
@@ -25,20 +27,24 @@
...
+        raise ModuleNotFoundError("Package `torchdata` not found. Please install following instructions at `https://github.com/pytorch/data`")
+
+    url_dp = IterableWrapper([URL])
+    cache_dp = url_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, _PATH), hash_dict={os.path.join(root, _PATH): MD5}, hash_type="md5"
+    )
+    cache_dp = GDriveReader(cache_dp).end_caching(mode="wb", same_filepath_fn=True)
+    cache_dp = FileOpener(cache_dp, mode="b")
+    extracted_files = cache_dp.read_from_tar()
+    filter_extracted_files = extracted_files.filter(lambda x: split in x[0])
```
Nayef211, line 49: @parmeet I didn't catch this earlier, but I strongly recommend updating this logic to include the extension of the split (i.e. `.csv` or `.tsv`). This caused bugs in the SST2 implemenation in the past. Look at https://github.com/pytorch/text/pull/1444 for reference

Nayef211: An alternative would be to do `lambda x: _EXTRACTED_FILES[split] in x[0]` which is what @erip is doing in https://github.com/pytorch/text/pull/1499

parmeet: Thanks @Nayef211 for catching this. Yes, I remember this breakage earlier, and completely forgot to take care of this. Will do PR to fix it.

parmeet:
> So existing installation will fail without installing torchdata. 

I guess by existing you mean torchtext <=0.11, In that case, torchdata is not required right?

This change only affect torchtext 0.12 onwards, when we shall make torchdata required dependency. Not sure if i miss-understood the concern here?

PR #1495, opened 2022-01-07
fix max sequence length for xlmr transform

parmeet:
The max sequence length for base architecture is 256. This PR fixes it.


Nayef211 (approved):
LGTM!

Issue #1494, opened 2022-01-07
Migrate datasets to build on top of torchdata datapipes


parmeet:
## 🚀 Feature

**Motivation**

https://github.com/pytorch/data#why-composable-data-loading

**_user-experience:_** TorchData datasets enable new functional API, auto-sharding, and snapshotting support out-of-the-box. They also enable standard flow-control like batching, collation, shuffling, bucketing, and mapping/transformation using user-defined functions and transforms (UDFs).

**_Maintenance:_** By relying on TorchData, we no longer have to maintain low level functionality like downloading, extracting, caching, file/steam parsing, etc.

**Reference** 
Examples: https://github.com/facebookexternal/torchdata/tree/main/examples/text
TorchData: https://github.com/facebookexternal/torchdata

## Backlog of datasets

- [x] AG_NEWS https://github.com/pytorch/text/pull/1498
- [x] AmazonReviewFull https://github.com/pytorch/text/pull/1499
- [x] AmazonReviewPolarity https://github.com/pytorch/text/pull/1490
- [x] DBpedia https://github.com/pytorch/text/pull/1500
- [x] SogouNews https://github.com/pytorch/text/pull/1503
- [x] YelpReviewFull https://github.com/pytorch/text/pull/1507
- [x] YelpReviewPolarity https://github.com/pytorch/text/pull/1509
- [x] YahooAnswers https://github.com/pytorch/text/pull/1508
- [x] CoNLL2000Chunking https://github.com/pytorch/text/pull/1515
- [x] UDPOS https://github.com/pytorch/text/pull/1535
- [x] IWSLT2016 https://github.com/pytorch/text/pull/1545
- [x] IWSLT2017 https://github.com/pytorch/text/pull/1547
- [x] Multi30K https://github.com/pytorch/text/pull/1536
- [x] SQuAD1 https://github.com/pytorch/text/pull/1513
- [x] SQuAD2 https://github.com/pytorch/text/pull/1514
- [x] PennTreebank https://github.com/pytorch/text/pull/1511
- [x] WikiText103 https://github.com/pytorch/text/pull/1518
- [x] WikiText2 https://github.com/pytorch/text/pull/1519
- [x] EnWik9 https://github.com/pytorch/text/pull/1512
- [x] IMDB https://github.com/pytorch/text/pull/1531
- [x] SST2 https://github.com/pytorch/text/pull/1538
- [ ] CC-100 https://github.com/pytorch/text/pull/1562

## Contributing
[...]

Issue #1493, opened 2022-01-07
Revamp TorchText Dataset Testing Strategy


Nayef211:
## 🚀 Feature
Revamp our dataset testing strategy in order to reduce the amount of time we spend waiting on tests to complete before merging our PRs in torchtext.

**Motivation**

TorchText dataset testing currently relies on downloading and caching the datasets daily and then running CircleCI tests on the cached data. This can be slow and unreliable for the first PR that kicks off the dataset download and caching. In addition, dataset extraction can be time consuming for some of the larger datasets within torchtext and this extraction process occurs each time the dataset tests are run on a PR. Due to these reasons, tests on CircleCI can take up to an hour to run for each PR whereas vision/audio tests run in mere minutes. We want to revamp our dataset testing strategy in order to reduce the amount of time we spend waiting on tests to complete before merging our PRs in torchtext.

**Pitch**
We need to update the legacy dataset tests within torchtext. Currently we test for things including: 
- URL link
- MD5 hash of the entire dataset
- dataset name
- number of lines in dataset

Going forward it doesn’t make sense to test the MD5 hash or the number of lines in the dataset. Instead we 
- Use mocking to test the implementation of our dataset
- Use smoke tests for URLs and integrity of data (potentially with Github Actions)

**Alternatives**

- Download and cache the dataset daily before running tests (current state of testing)
- Create mock data for each dataset (used by torchaudio, and torchvision)
   - Requires us to understand the structure of the datasets before creating tests
- Store a small portion of the dataset (10 lines) in an assets folder
   - Might run into legal problems since we aren’t allowed to host datasets

### Additional Context
We explored how other domains implemented testing for datasets and summarize them below. We will implement our new testing strategy by taking inspiration from TorchAudio and TorchVision

**TorchAudio Approach**
- Each test lives in its own file
- Plan to add integration tests in the future to check dataset URLs
- Each test class extends `TestBaseMixin` and `PytorchTestCase` ([link](https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/common_utils/case_utils.py#L93))
   - `TestBaseMixin` base class provide consistent way to define device/dtype/backend aware TestCase
- Each test file contains a `get_mock_dataset()` method which is responsible for creating the mocked data and saving it to a file in a temp dir ([link](https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/datasets/cmuarctic_test.py#L14-L52))
   - This method gets called in the setUp classmethod within each test class
[...]

PR #1498, opened 2022-01-08
migrate AG_NEWS to datapipes.

erip:
- [x] Migrate dataset in `torchtext/datasets/`
- [x] Update tests

### DIFF
--- a/torchtext/datasets/ag_news.py
+++ b/torchtext/datasets/ag_news.py
@@ -37,12 +37,12 @@ def AG_NEWS(root: str, split: Union[Tuple[str], str]):
         raise ModuleNotFoundError("Package `torchdata` not found. Please install following instructions at `https://github.com/pytorch/data`")
 
     url_dp = IterableWrapper([URL[split]])
-    http_dp = HttpReader(url_dp)
-    cache_dp = http_dp.on_disk_cache(
+    cache_dp = url_dp.on_disk_cache(
         filepath_fn=lambda x: os.path.join(root, split + ".csv"),
         hash_dict={os.path.join(root, split + ".csv"): MD5[split]},
         hash_type="md5"
-    ).end_caching(mode="w", same_filepath_fn=True)
-
+    )
+    cache_dp = HttpReader(cache_dp)
+    cache_dp = cache_dp.end_caching(mode="w", same_filepath_fn=True)
     cache_dp = FileOpener(cache_dp, mode="r")
     return cache_dp.parse_csv().map(fn=lambda t: (int(t[0]), " ".join(t[1:])))
--- a/torchtext/datasets/ag_news.py
+++ b/torchtext/datasets/ag_news.py
@@ -1,12 +1,13 @@
-from torchtext.utils import (
-    download_from_url,
-)
+from torchtext._internal.module_utils import is_module_available
+from typing import Union, Tuple
+
+if is_module_available("torchdata"):
+    from torchdata.datapipes.iter import FileOpener, HttpReader, IterableWrapper
+
 from torchtext.data.datasets_utils import (
-    _RawTextIterableDataset,
     _wrap_split_argument,
     _add_docstring_header,
     _create_dataset_directory,
-    _create_data_from_csv,
 )
 import os
 
@@ -30,11 +31,18 @@ DATASET_NAME = "AG_NEWS"
 
 @_add_docstring_header(num_lines=NUM_LINES, num_classes=4)
 @_create_dataset_directory(dataset_name=DATASET_NAME)
-@_wrap_split_argument(('train', 'test'))
-def AG_NEWS(root, split):
-    path = download_from_url(URL[split], root=root,
-                             path=os.path.join(root, split + ".csv"),
-                             hash_value=MD5[split],
-                             hash_type='md5')
-    return _RawTextIterableDataset(DATASET_NAME, NUM_LINES[split],
-                                   _create_data_from_csv(path))
+@_wrap_split_argument(("train", "test"))
+def AG_NEWS(root: str, split: Union[Tuple[str], str]):
+    if not is_module_available("torchdata"):
+        raise ModuleNotFoundError("Package `torchdata` not found. Please install following instructions at `https://github.com/pytorch/data`")
+
+    url_dp = IterableWrapper([URL[split]])
+    http_dp = HttpReader(url_dp)
+    cache_dp = http_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, split + ".csv"),
+        hash_dict={os.path.join(root, split + ".csv"): MD5[split]},
+        hash_type="md5"
+    ).end_caching(mode="w", same_filepath_fn=True)
+
+    cache_dp = FileOpener(cache_dp, mode="r")
+    return cache_dp.parse_csv().map(fn=lambda t: (int(t[0]), " ".join(t[1:])))
--- a/test/data/test_builtin_datasets.py
+++ b/test/data/test_builtin_datasets.py
@@ -1,6 +1,5 @@
 #!/user/bin/env python3
 # Note that all the tests in this module require dataset (either network access or cached)
-import torch
 import torchtext
 import json
 import hashlib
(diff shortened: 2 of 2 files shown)


erip:
This error is a bit mysterious. 😄  I'll take a closer look at this, but local tests aren't even running as `pytest test/data/test_builtin_datasets.py` results in a segfault. Quite a mysterious situation for such a small change!

erip:
I pulled in newest changes and it seems like CI isn't outright failing everywhere. Tests still aren't running locally, but I can debug from CI for now...

erip:
These errors are "real" errors insomuch as these `IterDataPipe`s are actually not iterators since they don't define `__next__`. I'm not sure if this is something we want to push back to pytorch core team (as I don't think it's a torchdata issue) or if it's something we can live with. I think wrapping the dataset in an `iter` is an ugly solution... not sure. I welcome any thoughts.

parmeet:
> I rebuilt torchtext from source and pulled down the cache and tests are now running locally. Maybe this is worth adding to `CONTRIBUTING.md` or similar?

Thanks @erip for raising this. Yes, you are right, it's not obvious that the datasets cache need to build locally (only when you look into CI that you know this). Certainly this need to be made more obvious in Contribution guidelines. At the same time, we plan to move away from caching datasets at all and rather rely on mocking. Please refer to #1493 .

parmeet:
> These errors are "real" errors insomuch as these `IterDataPipe`s are actually not iterators since they don't define `__next__`. I'm not sure if this is something we want to push back to pytorch core team (as I don't think it's a torchdata issue) or if it's something we can live with. I think wrapping the dataset in an `iter` is an ugly solution... not sure. I welcome any thoughts.



> These errors are "real" errors insomuch as these `IterDataPipe`s are actually not iterators since they don't define `__next__`. I'm not sure if this is something we want to push back to pytorch core team (as I don't think it's a torchdata issue) or if it's something we can live with. I think wrapping the dataset in an `iter` is an ugly solution... not sure. I welcome any thoughts.

Yes, datapipes are iterable but not Iterators by design. not sure if it impose any limitation ,or if there use-cases where we want them to be Iterators? (except that this change would be BC-breaking for existing datasets, but in a good way because current datasets exhaust after one iteration ;) )

erip:
Thanks, @parmeet. Mostly it breaks the "next test" but that can be removed if this is desirable.

parmeet:
> Thanks, @parmeet. Mostly it breaks the "next test" but that can be removed if this is desirable.

right for testing, it's easy fix, I guess `next(iter(datapipe))` should be sufficient :)

parmeet:
> > Thanks, @parmeet. Mostly it breaks the "next test" but that can be removed if this is desirable.
> 
> right for testing, it's easy fix, I guess `next(iter(datapipe))` should be sufficient :)

Perhaps we could completely avoid this, once we migrate the testing strategy to mocking.

erip:
Oops, I meant [this one] (https://github.com/pytorch/text/blob/main/test/data/test_builtin_datasets.py#L93)

parmeet:
> Oops, I meant [this one](https://github.com/pytorch/text/blob/main/test/data/test_builtin_datasets.py#L93). Adding the `next(iter(...))` here wouldn't actually be testing anything anymore, right?

ahh, I see. Got it! Yes, I think we can remove this test.

erip:
OK, I think this one is also ready for review.

erip:
I think this is ready for review -- it follows the cookie cutter from the Amazon Polarity PR. cc @Nayef211  @abhinavarora @parmeet.

Nayef211 (approved):
LGTM! For all the dataset migrations, can you add the following line to the top of the summary "Reference Issue: #1494" so we can keep track of all the PRs in issue #1494

torchtext/datasets/ag_news.py
```diff
@@ -30,11 +31,18 @@
...
-@_wrap_split_argument(('train', 'test'))
-def AG_NEWS(root, split):
-    path = download_from_url(URL[split], root=root,
-                             path=os.path.join(root, split + ".csv"),
-                             hash_value=MD5[split],
-                             hash_type='md5')
-    return _RawTextIterableDataset(DATASET_NAME, NUM_LINES[split],
-                                   _create_data_from_csv(path))
+@_wrap_split_argument(("train", "test"))
+def AG_NEWS(root: str, split: Union[Tuple[str], str]):
```
Nayef211, line 35: NIT: will the type for `split` ever be `Tuple[str]`? Because as far as I understand the `@_wrap_split_argument` decorator will make `n` number of calls to this method where `n` is the number of splits in the `split` argument.

cc @parmeet

erip: You could ask for the train and test splits for instance, right?

parmeet: Yes, the split can be either tuple or just string. A user can pass something like `(train,dev,test)` which returns all the splits or just 'train' which would return only train split. The decorator is precisely taking care of this i.e to handle both the scenarios.

torchtext/datasets/ag_news.py
```diff
@@ -30,11 +31,18 @@
...
+    if not is_module_available("torchdata"):
+        raise ModuleNotFoundError("Package `torchdata` not found. Please install following instructions at `https://github.com/pytorch/data`")
+
+    url_dp = IterableWrapper([URL[split]])
+    http_dp = HttpReader(url_dp)
+    cache_dp = http_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, split + ".csv"),
+        hash_dict={os.path.join(root, split + ".csv"): MD5[split]},
+        hash_type="md5"
+    ).end_caching(mode="w", same_filepath_fn=True)
```
parmeet, line 45: nit: can we update this to follow the same sequence of datapipes url->cache->download->end_caching as oppose to url->download->cache->end_caching as demonstrated in example [here](https://github.com/pytorch/data/blob/ec32ee42983760640280c937868611355e6ed281/torchdata/datapipes/iter/util/cacheholder.py#L130) and done for AmazonReviewPolarity/Full datasets? Not sure if there are any (performance) advantage by following one pattern over another. @ejguan could you please share any insights here?

ejguan: Yes and you should do so. The following pattern would re-download the file that violates the hash.
```suggestion
    cache_dp = url_dp.on_disk_cache(
        filepath_fn=lambda x: os.path.join(root, split + ".csv"),  # Here may need to change this function
        hash_dict={os.path.join(root, split + ".csv"): MD5[split]},
        hash_type="md5"
    )
    cache_dp = HttpReader(cache_dp)
    cache_dp = cache_dp.end_caching(mode="w", same_filepath_fn=True)
```

erip: Sounds good -- I'll make the fix in a moment!

torchtext/datasets/ag_news.py
```diff
@@ -30,11 +31,18 @@
...
+        raise ModuleNotFoundError("Package `torchdata` not found. Please install following instructions at `https://github.com/pytorch/data`")
+
+    url_dp = IterableWrapper([URL[split]])
+    cache_dp = url_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, split + ".csv"),
+        hash_dict={os.path.join(root, split + ".csv"): MD5[split]},
+        hash_type="md5"
+    )
+    cache_dp = HttpReader(cache_dp)
+    cache_dp = cache_dp.end_caching(mode="w", same_filepath_fn=True)
```
Nayef211, line 46: NIT: can we combine these two lines like we did for [AmazonReviewFull](https://github.com/pytorch/text/pull/1499/files#diff-61c4706569928ac5352feafe9003545f246296ecc93c10e20120747f3707ccbdR52) and [AmazonReviewPolarity](https://github.com/pytorch/text/pull/1490/files#diff-daeb7a9b451f1cb20240d2fda55074b6f03d48849646c931c513050af8eff1a5R46)

We want to have as much consistency between the dataset implementations as possible

parmeet (approved):
LGTM!

PR #1499, opened 2022-01-09
migrate Amazon Review Full to datapipes.

erip:
- [x] Migrate dataset in `torchtext/datasets/`
- [x] Update tests

Reference Issue: #1494

PR #1500, opened 2022-01-10
migrate DBPedia to datapipes.

erip:
- [x] Migrate dataset in `torchtext/datasets/`
- [x] Update tests

I'll wait for #1498 or #1499 to land before marking this ready so I can rebase the test changes. 

Reference Issue: #1494

### DIFF
--- a/torchtext/datasets/dbpedia.py
+++ b/torchtext/datasets/dbpedia.py
@@ -23,6 +23,11 @@ NUM_LINES = {
 
 _PATH = 'dbpedia_csv.tar.gz'
 
+_EXTRACTED_FILES = {
+    "train": os.path.join("dbpedia_csv", "train.csv"),
+    "test": os.path.join("dbpedia_csv", "test.csv")
+}
+
 DATASET_NAME = "DBpedia"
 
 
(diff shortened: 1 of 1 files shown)


torchtext/datasets/dbpedia.py
```diff
@@ -28,13 +28,23 @@
...
+    cache_dp = url_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, _PATH),
+        hash_dict={os.path.join(root, _PATH): MD5}, hash_type="md5"
+    )
+    cache_dp = GDriveReader(cache_dp).end_caching(mode="wb", same_filepath_fn=True)
+    cache_dp = FileOpener(cache_dp, mode="b")
+
+    extracted_files = cache_dp.read_from_tar()
+
+    filter_extracted_files = extracted_files.filter(lambda x: split + ".csv" in x[0])
```
Nayef211, line 48: Same comment as in the other PR. Let's make this consistent with the logic [here](https://github.com/pytorch/text/pull/1503/files#diff-8342c3e1b0563ee8e6983216c4660087d8c94d6b641e668780cac56c15227388R53)

erip: Note that the "expected files" map doesn't really exist here. I can make it exist -- should I do that in this PR or a separate one?

Nayef211: I think it should be fine to do that in this PR. 

Let's verify with @parmeet first to see if there's reason that `_EXTRACTED_FILES` and `_EXTRACTED_FILES_MD5` weren't created for this dataset.

Nayef211 (approved):
LGTM!

PR #1501, opened 2022-01-10
Remove redundant get asset functions from parameterized_utils file

Nayef211:
## Summary
- Both `test/common/assets.py` and `test/common/parameterized_utils.py` files have a constant for the path to the assets directory and a `get_asset_path` method
- We remove these redundant constants/methods from `test/common/parameterized_utils.py` and instead use the implementation from `test/common/assets.py`

## Test
- `pytest test/data/test_builtin_datasets.py`
   - Note: the above tests only run on circleci and fail when run locally since the datasets are not cached

## Followup

- [ ] Update `common/__init__.py` file with `__all__` variable to list the public objects within the `test/common` module

### DIFF
--- a/test/common/parameterized_utils.py
+++ b/test/common/parameterized_utils.py
@@ -1,17 +1,10 @@
 import json
+
 from parameterized import param
-import os.path
 
-
-_TEST_DIR_PATH = os.path.realpath(
-    os.path.join(os.path.dirname(__file__), '..'))
-
-
-def get_asset_path(*paths):
-    """Return full path of a test asset"""
-    return os.path.join(_TEST_DIR_PATH, 'asset', *paths)
+from .assets import get_asset_path
 
 
 def load_params(*paths):
-    with open(get_asset_path(*paths), 'r') as file:
+    with open(get_asset_path(*paths), "r") as file:
         return [param(json.loads(line)) for line in file]


mthrok (approved):
Code-wise it looks good.

You may want to alias it inside of `parameterized_utils` by prefixing it with underscore, because it looks as if `get_asset_path` is (re-)defined in `parameetrized_utils` and it is possible for test code to do `from ..common.parameterized_utils import get_sset_path`. 

Another way to disambiguate will be to let `common/__init__.py` imports the utilities used by the tests, i.e. https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/common_utils/__init__.py

Nayef211:
> Code-wise it looks good.
> 
> You may want to alias it inside of `parameterized_utils` by prefixing it with underscore, because it looks as if `get_asset_path` is (re-)defined in `parameetrized_utils` and it is possible for test code to do `from ..common.parameterized_utils import get_sset_path`.
> 
> Another way to disambiguate will be to let `common/__init__.py` imports the utilities used by the tests, i.e. https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/common_utils/__init__.py

Should it be the user's responsibility to verify that they are using the `get_asset_path` from the file it is defined? 

If we end up prefixing with an underscore, we would need to do this in several parts of the codebase. For example `text/test/common/cache_utils.py` does the following import `from .parameterized_utils import load_params`. In this case, the user can technically also import  `load_params` method from cache_utils, but we would expect them to import it from the source where it was defined. 

Is this a valid expectation? If not as a followup, I can update `common/__init__.py` to import test utilities.

mthrok:
> Should it be the user's responsibility to verify that they are using the `get_asset_path` from the file it is defined?
> ... we would expect them to import it from the source where it was defined.
> Is this a valid expectation? If not as a followup, I can update `common/__init__.py` to import test utilities.

Yes and no (but more leaning towards no). Users can use the library in the way it is designed, but at the same time, users are free to ignore them.
If the library provides API that is not confusing, then consensus will occur naturally and it will not be about the responsibility anymore. If a code is huge, it is often case that users will look at how an existing code is using the library and expect that is the right way. torchtext is not huge so it is easy to go check where it is defined.

Another thing you can do is to define `__all__` to tell people looking at the code what is your *intent* on what is public and what not.

Nayef211:
> > Should it be the user's responsibility to verify that they are using the `get_asset_path` from the file it is defined?
> > ... we would expect them to import it from the source where it was defined.
> > Is this a valid expectation? If not as a followup, I can update `common/__init__.py` to import test utilities.
> 
> Yes and no (but more leaning towards no). Users can use the library in the way it is designed, but at the same time, users are free to ignore them. If the library provides API that is not confusing, then consensus will occur naturally and it will not be about the responsibility anymore. If a code is huge, it is often case that users will look at how an existing code is using the library and expect that is the right way. torchtext is not huge so it is easy to go check where it is defined.
> 
> Another thing you can do is to define `__all__` to tell people looking at the code what is your _intent_ on what is public and what not.

I agree with your points. I added a followup item to update `common/__init__.py` file with `__all__` variable

PR #1496, opened 2022-01-07
Parameterized XLMR and Roberta model integration tests

Nayef211:
## Summary
- Refactored model  unit tests to integration tests as a followup from #1491
- Parameterized integration tests to reduce redundant test methods 

## Test
`pytest test/integration_tests/test_models.py`

## Followup
- [ ] Decide the best place to put remaining tests in `test_models.py`
- [ ] Add [`nested_params` helper method](https://fburl.com/xbpvu1sq) from torchaudio to parameterize jit and non-jit model tests

PR #1502, opened 2022-01-10
Parameterize jit and non-jit model integration tests

Nayef211:
## Summary
- Add `nested_params` [helper method from torchaudio](https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/common_utils/parameterized_utils.py#L26) to parameterize jit and non-jit model tests as a followup from https://github.com/pytorch/text/pull/1496

## Test
`pytest test/integration_tests/test_models.py`


test/integration_tests/test_models.py
```diff
@@ -1,67 +1,48 @@
...
+            ),
+            (
+                "roberta.large.output.pt",
+                "Roberta base Model Comparison",
+                ROBERTA_LARGE_ENCODER,
+            ),
+        ],
+        [True, False],
+    )
+    def test_model(self, model_args, is_jit):
```
mthrok, line 32: nit: This test is rather self-explanatory, so I think it's okay, but I recommend writing a short docstrings for tests so that future maintainers can tell what was my *intention* and see if the implementation makes sense.

Writing docstring helps defining what consists of a correct test, what's the definition, when there is no intuitive test criteria for a new feature.

mthrok (approved):

test/integration_tests/test_models.py
Nayef211: Thanks for the feedback. Just added a short docstring to describing what's being tested

PR #1506, opened 2022-01-11
use os.path.join for consistency.

erip:
Using an f string populated with a string literal is just that string, but we can just use `os.path.join` instead. 😄


parmeet (approved):
Thanks for this @erip, LGTM :)

PR #1509, opened 2022-01-12
migrate YelpReviewPolarity to datapipes.

erip:

Reference Issue: #1494

### DIFF
--- a/torchtext/datasets/yelpreviewpolarity.py
+++ b/torchtext/datasets/yelpreviewpolarity.py
@@ -1,13 +1,16 @@
-import os
-from torchtext.utils import download_from_url, extract_archive
+from torchtext._internal.module_utils import is_module_available
+from typing import Union, Tuple
+
+if is_module_available("torchdata"):
+    from torchdata.datapipes.iter import FileOpener, GDriveReader, IterableWrapper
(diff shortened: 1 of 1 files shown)


erip:
These failures are interesting. Am I understanding correctly that the cache is pulled down daily in CI? Maybe these errors are related to the cache not being pulled yet? They seem unrelated as tests pass locally.

parmeet:
> These failures are interesting. Am I understanding correctly that the cache is pulled down daily in CI? Maybe these errors are related to the cache not being pulled yet? They seem unrelated as tests pass locally.

Yes, you got it right. The cache is generated on daily cadence and sometimes fails to populate fully leading to failure of all datasets tests. Test reliability (among others) is one of the reasons we want to move away from caching datasets to mocking based testing in this issue #1493. Can you confirm that the tests related to this dataset are passing locally? I don't mind merging it into main as I know these errors are not related to your PR.

erip:
Seems OK. This is for the other branch, but the changes are effectively the same. 

Passing tests locally

erip:
Just confirmed this is OK.

Nayef211 (approved):
LGTM

PR #1510, opened 2022-01-12
run tests in parallel with pytest-xdist.

erip:
This is an improvement for status-quo test speed performance. Eventually we'll want to rework the caching situation, but we can get decent perf by running in parallel.

### DIFF
--- a/.circleci/unittest/linux/scripts/run_test.sh
+++ b/.circleci/unittest/linux/scripts/run_test.sh
@@ -6,4 +6,4 @@ eval "$(./conda/bin/conda shell.bash hook)"
 conda activate ./env
 
 python -m torch.utils.collect_env
-pytest --cov=torchtext --junitxml=test-results/junit.xml -v --durations 20 test
+pytest --tx "3*popen//python=python" --cov=torchtext --junitxml=test-results/junit.xml -v --durations 20 test
--- a/.circleci/unittest/linux/scripts/environment.yml
+++ b/.circleci/unittest/linux/scripts/environment.yml
@@ -11,6 +11,7 @@ dependencies:
     - revtok
     - pytest
     - pytest-cov
+    - pytest-xdist
     - pytest-pythonpath
     - sacremoses
     - spacy
--- a/.circleci/unittest/windows/scripts/environment.yml
+++ b/.circleci/unittest/windows/scripts/environment.yml
@@ -11,6 +11,7 @@ dependencies:
     - revtok
     - pytest
     - pytest-cov
+    - pytest-xdist
     - pytest-pythonpath
     - sacremoses
     - spacy
--- a/.circleci/unittest/windows/scripts/run_test.sh
+++ b/.circleci/unittest/windows/scripts/run_test.sh
@@ -6,4 +6,4 @@ eval "$(./conda/Scripts/conda.exe 'shell.bash' 'hook')"
 conda activate ./env
 
 python -m torch.utils.collect_env
-pytest --cov=torchtext --junitxml=test-results/junit.xml -v --durations 20 test
+pytest --tx "3*popen//python=python" --cov=torchtext --junitxml=test-results/junit.xml -v --durations 20 test


.circleci/unittest/linux/scripts/run_test.sh
```diff
@@ -6,4 +6,4 @@ eval "$(./conda/bin/conda shell.bash hook)"
 conda activate ./env
 
 python -m torch.utils.collect_env
-pytest --cov=torchtext --junitxml=test-results/junit.xml -v --durations 20 test
+pytest --tx "3*popen//python=python" --cov=torchtext --junitxml=test-results/junit.xml -v --durations 20 test
```
Nayef211, line 9: Can you mention in your summary that this command is creating three subprocesses to run tests in parallel. 

Also, I think it's worth benchmarking what the performance speedups look like for the test suite. Could you run the test suite locally and paste the results before and after parallelizing tests?

erip: Yes, absolutely. I can take a crack at something resembling a benchmark.

As an aside, I actually don't like this `--tx ...` idiom very much as it feels very low level and would prefer to use `-n auto` but ran into some errors about how tests get collected.

Nayef211: Yeah the `-n auto` flag does seem cleaner since it scales to the number of available CPUs on the host machine. What did your errors look like? Also I would prefer to pass in 4 instead of 3 to the `--tx` command since the number of cores on a machine is usually a multiple of 2.

erip: The errors were somewhat non-descript, but generally looked like:

```
Different tests were collected between gw0 and gw1. The difference is: 

--- gw0 
+++ gw1
```

Regarding the multiples-of-2: I believe (but may be mistaken) that the main test process constitutes the 4th process, hence the somewhat ugly 3 processes. It may be a moot point if I can get the `-n auto` working... More info to follow.

mthrok:
Hi @erip 

Thanks for the proposal but I really don't recommend using xdist in CI. 

I understand that time reduction is important but from my own experience, pytest-xdist is not good enough to use in CI. 

It complicates the analysis of test failure. If two tests running together hit the OOM, the entire test process is killed without log. Similarly if there is a build issue, say because of a newly introduced dependency or an issue in PyTorch upstream segfault happens, then again it does not leave any log and the first action is to remove xdist integration.

It adds unpredictabilitg to the tests which makes it very difficultt to debug.

I used xdist in torchaudio for a while and it was surprisingly error prune. I got frustrated at the cost it adds when there is an issue in tests.

.circleci/unittest/linux/scripts/run_test.sh
Nayef211: Ahh gotcha thanks for letting me know! Btw @mthrok strongly suggested not to use `pytest-xdist` and mentioned he would follow up with the reasons later today. Let's try to hold off on this effort until he get's back to us.

erip:
Thanks, @mthrok. We were using it successfully (or, without additional pain) in ignite so I thought I'd see where it gets us. I'm happy to scrap the idea.

mthrok:
> Thanks, @mthrok. We were using it successfully (or, without additional pain) in ignite so I thought I'd see where it gets us. I'm happy to scrap the idea.

Oh I did not know about that. Quickly skimming the ignite repo, ignite seems to be a pure python library, which might be one reason it works well. (just a random guess though)

Here is the PRs I added https://github.com/pytorch/audio/pull/807 and removed https://github.com/pytorch/audio/pull/915 after one month. The observation I had was, the test got longer on Linux (so I did not apply), and it completely stopped working on macOS later.

Nayef211:
@erip @mthrok I'm going to close this PR since we'll be starting the efforts to revamp testing for the existing datasets next week which should fix the slow tests. Lmk if you guys still think we should add `pytest-xdist` after the previous discussion.

PR #1511, opened 2022-01-12
Migrating PennTreebank to datapipes

Nayef211:
Reference Issue: #1494

## Summary

- Migrate PennTreebank
- Importing torchdata conditional and raising exception during function call if package is not available

PR #1512, opened 2022-01-13
Migrating EnWik9 to datapipes #1511

Nayef211:
Reference Issue: #1494

## Summary
- Migrate EnWik9
- Importing torchdata conditional and raising exception during function call if package is not available

### DIFF
--- a/torchtext/datasets/enwik9.py
+++ b/torchtext/datasets/enwik9.py
@@ -38,6 +38,6 @@ def EnWik9(root: str, split: Union[Tuple[str], str]):
         hash_type="md5",
     )
     cache_dp = HttpReader(cache_dp).end_caching(mode="wb", same_filepath_fn=True)
-    cache_dp = FileOpener(cache_dp, mode="b")
-    extracted_files = cache_dp.read_from_zip()
-    return extracted_files.readlines().map(lambda t: t[1].decode())
+    data_dp = FileOpener(cache_dp, mode="b")
+    extracted_files = data_dp.read_from_zip()
+    return extracted_files.readlines(decode=True, return_path=False)
--- a/torchtext/datasets/enwik9.py
+++ b/torchtext/datasets/enwik9.py
@@ -40,12 +40,11 @@ def EnWik9(root: str, split: Union[Tuple[str], str]):
     cache_compressed_dp = HttpReader(cache_compressed_dp).end_caching(
         mode="wb", same_filepath_fn=True
     )
-    cache_compressed_dp = FileOpener(cache_compressed_dp, mode="b")
 
     cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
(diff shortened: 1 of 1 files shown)


erip (changes requested):

torchtext/datasets/enwik9.py
```diff
@@ -1,34 +1,42 @@
...
+    url_dp = IterableWrapper([URL])
+    cache_dp = url_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, _PATH),
+        hash_dict={os.path.join(root, _PATH): MD5},
+        hash_type="md5",
+    )
+    cache_dp = HttpReader(cache_dp).end_caching(mode="wb", same_filepath_fn=True)
+    cache_dp = FileOpener(cache_dp, mode="b")
+    extracted_files = cache_dp.read_from_zip()
+    return extracted_files.readlines().map(lambda t: t[1].decode())
```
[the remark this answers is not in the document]
Nayef211: @erip @parmeet just fixed lmk if it looks good to ship!

erip:
I think this looks OK. I don't think we need to specifically cache the contents of the zip, but it might be worth checking. @parmeet wdyt?

Nayef211:
> I think this looks OK. I don't think we need to specifically cache the contents of the zip, but it might be worth checking. @parmeet wdyt?

Actually it might be worth doing since we're doing it for all the other datasets. Let me quickly add this!

torchtext/datasets/enwik9.py
```diff
@@ -1,34 +1,54 @@
...
+        hash_type="md5",
+    )
+    cache_compressed_dp = HttpReader(cache_compressed_dp).end_caching(
+        mode="wb", same_filepath_fn=True
+    )
+    cache_compressed_dp = FileOpener(cache_compressed_dp, mode="b")
+
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, os.path.splitext(_PATH)[0])
+    )
```
parmeet, line 47: can we move FileOpener after on_disk_cache as suggested here https://github.com/pytorch/text/pull/1530#discussion_r788931995

parmeet: Also please test by removing local cache completely. With presence of local cache, could avoid catching unexpected errors (Something we can improve when mock test comes into play :) )

Nayef211: Just fixed this. And in terms of testing, I'm using Jupyter notebook and a simple script to make sure I can load and iterate through the dataset. But if I don't download the dataset manually, this doesn't work since we need to use proxies to access data in development servers. Do you have any suggestions for enabling proxies programmatically on Jupyter notebooks? 

And this issue should definitely be fixed once we have mock tests for all the datasets!

parmeet (approved):
LGTM!

PR #1513, opened 2022-01-13
migrate SQUAD1 to datapipes.

erip:
Migrates dataset in `torchtext/datasets/` per reference impl [here](https://github.com/pytorch/data/blob/main/examples/text/squad1.py)

Reference Issue: #1494


erip:
Hmm, the failure seems unrelated. I've seen this error before in CI, but I'm not precisely sure how to solve it... it may just be an intermittent thing and this just needs to be rerun later.

erip:
cc @ejguan do [these errors](https://app.circleci.com/pipelines/github/pytorch/text/4501/workflows/152a9c52-bc4e-4a81-b338-5d6e71fa34f4/jobs/151108/tests#failed-test-0) look familiar? These tests were passing before -- are you tracking anything upstream that might introduce a breakage in nightly?

ejguan:
> cc @ejguan do [these errors](https://app.circleci.com/pipelines/github/pytorch/text/4501/workflows/152a9c52-bc4e-4a81-b338-5d6e71fa34f4/jobs/151108/tests#failed-test-0) look familiar? These tests were passing before -- are you tracking anything upstream that might introduce a breakage in nightly? They might be unrelated, but thought I'd check.

Thanks for reaching out. The Error points to TorchData. https://github.com/pytorch/data/blob/ec32ee42983760640280c937868611355e6ed281/torchdata/datapipes/iter/util/cacheholder.py#L257
I will dig into it later today.

BTW, do we want to replace the old implementation directly? I think we may want to keep the original implementation at least one release cycle.

ejguan:
> Our goal had been to hold off on releasing torchtext==0.12.0 until torchdata is "mature enough" (presumably when it's stable enough to be made available in pypi). Ultimately it's @parmeet and co.'s call.

Yes, understood. I mean normally users should have the option to switch between two different implementations at the first release to provide a smooth transition. For the deprecated implementations, we can add a warning for users to pivot them using the new implementation.

Besides, there is DataLoader2 coming after pytorch 1.12-ish, which is actually the targeted DataPipe scheduler (better performance and more features), even though we do support backward compatibility using DataLoader (existing one).

ejguan (commented):
This is super weird to me. I have tested both implementations in TorchData and your PR. And, I couldn't produce such error on either python3.9 or 3.8

erip:
Thanks for confirming, @ejguan -- I also couldn't repro locally. I wonder if there's something wrong with CI specifically... We have a weird caching mechanism in CI but that seems somewhat unrelated... Not positive.

ejguan:
> Thanks for confirming, @ejguan -- I also couldn't repro locally. I wonder if there's something wrong with CI specifically... We have a weird caching mechanism in CI but that seems somewhat unrelated... Not positive.

My main concern is based on the implementation of SQuAD1, the culprit part of code should not even be executed. As we explicitly set mode to `b` or `wb`, it should not read file handle as string stream.

Nayef211:
> > Thanks for confirming, @ejguan -- I also couldn't repro locally. I wonder if there's something wrong with CI specifically... We have a weird caching mechanism in CI but that seems somewhat unrelated... Not positive.
> 
> My main concern is based on the implementation of SQuAD1, the culprit part of code should not even be executed. As we explicitly set mode to `b` or `wb`, it should not read file handle as string stream.

@erip @ejguan Sorry I missed your messages in this thread yesterday. I ended up investigating this error seperately and coming to the same conclusion. Here is the PR that fixes the issue https://github.com/pytorch/text/pull/1517. Can I get a stamp so I can merge it in?

ejguan:
> > > Thanks for confirming, @ejguan -- I also couldn't repro locally. I wonder if there's something wrong with CI specifically... We have a weird caching mechanism in CI but that seems somewhat unrelated... Not positive.
> > 
> > 
> > My main concern is based on the implementation of SQuAD1, the culprit part of code should not even be executed. As we explicitly set mode to `b` or `wb`, it should not read file handle as string stream.
> 
> @erip @ejguan Sorry I missed your messages in this thread yesterday. I ended up investigating this error seperately and coming to the same conclusion. Here is the PR that fixes the issue #1517. Can I get a stamp so I can merge it in?

Aha, I didn't notice the Error comes from AG_NEWS. Stamped. 😆

parmeet:
> BTW, do we want to replace the old implementation directly? I think we may want to keep the original implementation at least one release cycle.

Thanks @ejguan for this call out. I think you made a valid point. Perhaps we could support both the implementation with the ability to switch to new implementation by setting flags or something. Let's take this discussion on the issue (#1494 ) directly.

erip:
@Nayef211 I think this is OK once it passes. It looked good locally.

parmeet (approved):
LGTM! let's make sure the test is passing locally if CI caching is still an issue.

erip:
Yep -- confirmed it's good with a fresh cache locally this morning, so it should be ok!

torchtext/data/datasets_utils.py
```diff
@@ -318,3 +319,26 @@ def pos(self):
 
     def __str__(self):
         return self.description
+
+
+@functional_datapipe("read_squad")
+class _ParseSQuADQAData(IterDataPipe):
```
parmeet, line 325: nit: we could move this to dataset_utils?  (as this one is also used for SQuaD2)

erip: That's where it already lives, right?

parmeet: Oops, my bad :). Thanks!

PR #1514, opened 2022-01-13
Migrate squad2 to datapipes

erip:
Migrates dataset in `torchtext/datasets/` per reference impl [here](https://github.com/pytorch/data/blob/main/examples/text/squad2.py). ~Do not merge until #1513 lands~

Reference Issue: #1494

### DIFF
--- a/torchtext/datasets/squad2.py
+++ b/torchtext/datasets/squad2.py
@@ -1,11 +1,17 @@
-from torchtext.utils import download_from_url
+from torchtext._internal.module_utils import is_module_available
+from typing import Union, Tuple
+
+if is_module_available("torchdata"):
+    from torchdata.datapipes.iter import FileOpener, HttpReader, IterableWrapper
+
 from torchtext.data.datasets_utils import (
-    _RawTextIterableDataset,
     _wrap_split_argument,
     _add_docstring_header,
     _create_dataset_directory,
-    _create_data_from_json,
 )
+
+import os
+
 URL = {
     'train': "https://rajpurkar.github.io/SQuAD-explorer/dataset/train-v2.0.json",
     'dev': "https://rajpurkar.github.io/SQuAD-explorer/dataset/dev-v2.0.json",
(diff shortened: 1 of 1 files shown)


torchtext/data/datasets_utils.py
```diff
@@ -318,3 +319,22 @@ def pos(self):
 
     def __str__(self):
         return self.description
+
+
+class _ParseSQuADQAData(IterDataPipe):
```
Nayef211, line 324: Can we add a short docstring with an explanation of what this function does and why it's put here?  It would probably be good to mention that this util function is relevant to both SQuAD1 and SQuAD2 datasets.

Nayef211, line 324: Also just as an fyi, we can register this datapipe using the `@functional_datapipe("parse_squad_qa_data")` decorator  like you did in this [PR](https://github.com/pytorch/text/pull/1515/files#diff-ab3ee9237fbb0465f1d7fd4ea04808ab303af375b934a0c205c3ebaa1810b220R324). This would allow us to do something like `return cache_dp.parse_json_files().parse_squad_qa_data()` in [this line](https://github.com/pytorch/text/pull/1514/files#diff-76882df4733860930a3fb296d47cf0439b68e0494bacb13549186b86bbbd0df9R54)

erip: Thanks, pushed!

torchtext/data/datasets_utils.py
```diff
@@ -318,3 +319,26 @@ def pos(self):
 
     def __str__(self):
         return self.description
+
+
+@functional_datapipe("read_squad")
```
parmeet, line 324: I guess we could remove this once #1513  lands right?

erip: That's right. These are the same commits so I'll just rebase once squad1 lands.

erip:
This has been rebased. Running tests locally now.

parmeet (approved):
LGTM, thanks! (Finally with green CI :) )

PR #1515, opened 2022-01-13
migrate CONLL 2000  to datapipes.

erip:

Reference Issue: #1494

### DIFF
--- a/test/data/test_dataset_utils.py
+++ b/test/data/test_dataset_utils.py
@@ -47,7 +47,7 @@ class TestDatasetUtils(TorchtextTestCase):
         iterable = [("ignored.txt", e) for e in iob]
         iterable = IterableWrapper(iterable)
         iob_dp = list(_ParseIOBData(iterable, sep=" "))
-        # There's only one example in this dataset
+        # There are two examples in this dataset
         self.assertEqual(len(iob_dp), 2)
         # The length of the first list of surface forms is the length of everything before the empty line.
         # The length of the first labels is the length of everything before the empty line.
--- a/test/data/test_dataset_utils.py
+++ b/test/data/test_dataset_utils.py
@@ -1,0 +1,110 @@
+from ..common.torchtext_test_case import TorchtextTestCase
+
+from torchtext.data.datasets_utils import _ParseIOBData
+from torch.utils.data.datapipes.iter import IterableWrapper
+
+
+class TestDatasetUtils(TorchtextTestCase):
+    def test_iob_datapipe_basic(self):
+        iob = [
+            "Alex I-PER",
+            "is O",
+            "going O",
+            "to O",
+            "Los I-LOC",
+            "Angeles I-LOC",
+            "in O",
+            "California I-LOC"
+        ]
+        iterable = [("ignored.txt", e) for e in iob]
+        iterable = IterableWrapper(iterable)
+        iob_dp = list(_ParseIOBData(iterable, sep=" "))
+        # There's only one example in this dataset
+        self.assertEqual(len(iob_dp), 1)
+        # The length of the list of surface forms is the number of lines in the example
+        self.assertEqual(len(iob_dp[0][0]), len(iob))
+        # The length of the list labels is the number of lines in the example
+        self.assertEqual(len(iob_dp[0][1]), len(iob))
+        iob = [
+            "Alex I-PER",
+            "is O",
+            "going O",
+            "to O",
+            "Los I-LOC",
+            "Angeles I-LOC",
+            "in O",
+            "California I-LOC",
+            "",
+            "Alex I-PER",
+            "is O",
+            "going O",
+            "to O",
+            "Los I-LOC",
+            "Angeles I-LOC",
+            "in O",
+            "California I-LOC",
+        ]
+        iterable = [("ignored.txt", e) for e in iob]
+        iterable = IterableWrapper(iterable)
+        iob_dp = list(_ParseIOBData(iterable, sep=" "))
+        # There's only one example in this dataset
+        self.assertEqual(len(iob_dp), 2)
+        # The length of the first list of surface forms is the length of everything before the empty line.
+        # The length of the first labels is the length of everything before the empty line.
+        self.assertEqual(len(iob_dp[0][0]), iob.index(""))
+        self.assertEqual(len(iob_dp[0][1]), iob.index(""))
+        # The length of the second list of surface forms is the length of everything after the empty line.
+        # The length of the second labels is the length of everything after the empty line.
+        self.assertEqual(len(iob_dp[1][0]), len(iob) - iob.index("") - 1)
+        self.assertEqual(len(iob_dp[1][1]), len(iob) - iob.index("") - 1)
+
+    def test_iob_datapipe_functional(self):
+        iob = [
+            "Alex I-PER",
+            "is O",
+            "going O",
+            "to O",
+            "Los I-LOC",
+            "Angeles I-LOC",
+            "in O",
+            "California I-LOC"
+        ]
+        iterable = [("ignored.txt", e) for e in iob]
+        iob_dp = list(IterableWrapper(iterable).read_iob(sep=" "))
+        # There's only one example in this dataset
+        self.assertEqual(len(iob_dp), 1)
+        # The length of the list of surface forms is the number of lines in the example
+        self.assertEqual(len(iob_dp[0][0]), len(iob))
+        # The length of the list labels is the number of lines in the example
+        self.assertEqual(len(iob_dp[0][1]), len(iob))
+        iob = [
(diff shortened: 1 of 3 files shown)


torchtext/data/datasets_utils.py
```diff
@@ -318,3 +319,27 @@ def pos(self):
 
     def __str__(self):
         return self.description
+
+
+@functional_datapipe("read_iob")
+class IOBDataPipe(IterDataPipe):
```
abhinavarora, line 325: Could we add some comments on what is the purpose of this datapipe?

torchtext/data/datasets_utils.py
```diff
@@ -318,3 +319,29 @@ def pos(self):
...
```
parmeet, line 325: should we make this class private? I guess it is only meant as internal utility for two specific datasets.

erip: That's a good question. I could imagine someone using it externally for a different IOB dataset, but I don't have strong opinions about making it private. Happy to do that.

erip: Made it private and standardized the name with the pipe in SQuAD.

erip:
This is ready for review, @parmeet @abhinavarora @Nayef211

torchtext/datasets/conll2000chunking.py
```diff
@@ -41,12 +44,17 @@
...
+
+    url_dp = IterableWrapper([URL[split]])
+    cache_dp = url_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, "conll2000chunking", os.path.basename(URL[split])),
+        hash_dict={os.path.join(root, "conll2000chunking", os.path.basename(URL[split])): MD5[split]},
+        hash_type="md5"
+    )
+    cache_dp = HttpReader(cache_dp).end_caching(mode="wb", same_filepath_fn=True)
+    cache_dp = FileOpener(cache_dp, mode="b")
+    cache_dp = cache_dp.extract(file_type="gzip").readlines(decode=True)
```
parmeet, line 59: nit: can we try adding caching for extraction as well? I am trying to do for AmazonReviewPolarity here (https://github.com/pytorch/data/pull/169) though I haven't successfully done it yet. Feel free to close this PR without this change and we can follow-up later in another PR. (If you do plan to try it out, make sure to push changes only when you succeed locally otherwise it would fail the cache :) )

test/data/test_dataset_utils.py
```diff
@@ -0,0 +1,110 @@
...
+        # The length of the first list of surface forms is the length of everything before the empty line.
+        # The length of the first labels is the length of everything before the empty line.
+        self.assertEqual(len(iob_dp[0][0]), iob.index(""))
+        self.assertEqual(len(iob_dp[0][1]), iob.index(""))
+        # The length of the second list of surface forms is the length of everything after the empty line.
+        # The length of the second labels is the length of everything after the empty line.
+        self.assertEqual(len(iob_dp[1][0]), len(iob) - iob.index("") - 1)
+        self.assertEqual(len(iob_dp[1][1]), len(iob) - iob.index("") - 1)
+
+    def test_iob_datapipe_functional(self):
```
parmeet, line 61: I am curious if this is a standard practice at pytorch/data as well to write testing both in basic and functional form?

test/data/test_dataset_utils.py
```diff
@@ -0,0 +1,110 @@
...
+            "to O",
+            "Los I-LOC",
+            "Angeles I-LOC",
+            "in O",
+            "California I-LOC",
+        ]
+        iterable = [("ignored.txt", e) for e in iob]
+        iterable = IterableWrapper(iterable)
+        iob_dp = list(_ParseIOBData(iterable, sep=" "))
+        # There's only one example in this dataset
```
parmeet, line 50: there are actually two examples :)

parmeet:
Can we also get remove https://github.com/pytorch/text/blob/main/torchtext/datasets/conll2000chunking.py#L32, I think this is no longer relevant.

test/data/test_dataset_utils.py
erip: 😸  good catch.

test/data/test_dataset_utils.py
erip: I had taken inspiration from the pytorch/data tests. See [here](https://github.com/pytorch/data/blob/main/test/test_local_io.py#L466-L481) for a slight modification on this approach.

erip:
> Can we also get remove
> 
> https://github.com/pytorch/text/blob/a2ab9741415b2cff026d158a5a54b62b993571d9/torchtext/datasets/conll2000chunking.py#L27
> 
> and
> https://github.com/pytorch/text/blob/a2ab9741415b2cff026d158a5a54b62b993571d9/torchtext/datasets/conll2000chunking.py#L32
> 
> , I think this is no longer relevant.

Don't we still need those for  [this](https://github.com/pytorch/text/pull/1515#discussion_r786395354)?

torchtext/datasets/conll2000chunking.py
erip: I have this working locally (it seems) -- will push soon once tests pass for sure.

erip: Looks OK. Will push

Tests with double caching

erip:
cc @ejguan does this "double caching" approach look reasonable? This is an analog to pytorch/data#169

torchtext/datasets/conll2000chunking.py
```diff
@@ -41,12 +44,29 @@
...
+        hash_dict={os.path.join(root, "conll2000chunking", os.path.basename(URL[split])): MD5[split]},
+        hash_type="md5"
+    )
+    cache_dp = HttpReader(cache_dp).end_caching(mode="wb", same_filepath_fn=True)
+    cache_dp = FileOpener(cache_dp, mode="b")
+
+    # Cache and check the gzip extraction for relevant split
+    cache_dp = cache_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, "conll2000chunking", _EXTRACTED_FILES[split]),
+        hash_dict={os.path.join(root, "conll2000chunking", _EXTRACTED_FILES[split]): _EXTRACTED_FILES_MD5[split]},
```
parmeet, line 65: I wonder if it is necessary to re-evaluate hash for the extracted files? Since we have already done it for the compressed package, I am guessing that should be sufficient?

parmeet: Other consideration might be maintainability. For instance, in this case we only have single compressed file, so it's OK I guess. For other cases like IMDB, there are many files that are compressed together (each example is a separate file), so it would be almost impractical to store MD5 for all the content.

test/data/test_dataset_utils.py
erip: Done!

torchtext/datasets/conll2000chunking.py
erip: It's probably OK to remove this. I wonder if there's some mechanism then to add the "raw" source's metadata to the MD5 so we know it hasn't changed. Something like hashing the raw bytes with last change timestamp or similar? Maybe I'm overthinking it... 😅

ejguan:
@erip Double caching definitely makes sense to me. There can be (rare) case that the decompressed files or downloaded files are deleted during iteration.

torchtext/datasets/conll2000chunking.py
parmeet: Ya, I wouldn't be worrying so much here :)

erip: OK, I'll remove this hash check and push.

parmeet (approved):
Thank @erip for all the nice work in this PR :)  LGTM!

parmeet:
@erip can you try once removing the local cache for CoNLL2000Chunking and see if this works from scratch? Unfortunately, it's failing on my side. I suspect it worked on CI (or locally on your side) because the extracted files were already dumped on the disk due to previous implementation.

erip:
When I blow away `$HOME/.torchtext/cache` the tests fail, but that's because of the call to `check_cache_status()`. Is there a "better" way to test this?

parmeet:
> When I blow away `$HOME/.torchtext/cache` the tests fail, but that's because of the call to `check_cache_status()`. Is there a "better" way to test this?

I am trying directly in i-python instead of going via testing route.

erip:
It seems OK for me from the interpreter:

local test

`root=.test` doesn't exist before running this.

parmeet:
> It seems OK for me from the interpreter:
> 
> local test
> 
> `root=.test` doesn't exist before running this.

Ahh cool, thanks for confirming. Then there must be something going on my side, will take a look. Glad it's working :)

PR #1517, opened 2022-01-14
Fixing dataset test failures due to incorrect caching mode in AG_NEWS

Nayef211:
## Summary
- The `EndOnDiskCacheHolderIterDataPipe` expects the mode to contain `b` otherwise it tries to read a string. The `HTTPReader` datapipe returns an IOStream of bytes from a given URL.
- By adding `b` to the `mode` param, we can fix the failing CI tests

### DIFF
--- a/torchtext/datasets/ag_news.py
+++ b/torchtext/datasets/ag_news.py
@@ -43,6 +43,6 @@ def AG_NEWS(root: str, split: Union[Tuple[str], str]):
         hash_type="md5"
     )
     cache_dp = HttpReader(cache_dp)
-    cache_dp = cache_dp.end_caching(mode="w", same_filepath_fn=True)
+    cache_dp = cache_dp.end_caching(mode="wb", same_filepath_fn=True)
     cache_dp = FileOpener(cache_dp, mode="r")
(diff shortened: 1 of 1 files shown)


erip:
Good catch, @Nayef211. Presumably this was passing before because we had a local cache (per CI) so it was not downloading the asset and going straight into reading in text mode?

erip:
It looks like the remaining failures are due to rate limiting from GDrive. This puts a bit more urgency on working through testing revamp, but this PR LGTM. I _think_ this is the only place that this issue occurs...

ejguan (approved):
Thanks!

Nayef211:
> Good catch, @Nayef211. Presumably this was passing before because we had a local cache (per CI) so it was not downloading the asset and going straight into reading in text mode?

Yup that sounds about right!



> It looks like the remaining failures are due to rate limiting from GDrive. This puts a bit more urgency on working through testing revamp, but this PR LGTM. I _think_ this is the only place that this issue occurs...

I'm planning to start on the testing work early next week and hope to have the design for the new tests done ASAP. It's definitely a little scary to add some of these datasets without having proper tests in place. In the meantime, I've been running a jupyter notebook while developing locally and I make sure that I can load and view each of the datasets after migrating to datapipes as a sanity check. I'd urge you to do something similar.

PR #1520, opened 2022-01-14
Attempting to fix version conflict in CI

erip:
- Replaces pip installs with conda installs where applicable
- Follows torchvision pattern for handling pytorch-mutex which conflicts with cpuonly

### DIFF
--- a/packaging/torchtext/meta.yaml
+++ b/packaging/torchtext/meta.yaml
@@ -1,3 +1,4 @@
+{% set build_variant = environ.get('CONDA_BUILD_VARIANT', 'cpu') %}
 package:
   name: torchtext
   version: "{{ environ.get('BUILD_VERSION') }}"
@@ -14,14 +15,22 @@ requirements:
   host:
     - python
     - setuptools
-    - cpuonly
     {{ environ.get('CONDA_PYTORCH_BUILD_CONSTRAINT') }}
 
   run:
     - python
     - requests
     - tqdm
+    - pytorch-mutex 1.0 {{ build_variant }}  # [not osx ]
     {{ environ.get('CONDA_PYTORCH_CONSTRAINT') }}
+
+  {% if build_variant == 'cpu' %}
+  run_constrained:
+    - cpuonly
+  {% elif not osx %}
+  run_constrained:
+     - cpuonly <0
+  {% endif %}
 
 build:
   string: py{{py}}
@@ -40,7 +49,6 @@ test:
 
   requires:
     - pytest
-    - cpuonly
 
 about:
   home: https://github.com/pytorch/text
--- a/.circleci/unittest/linux/scripts/environment.yml
+++ b/.circleci/unittest/linux/scripts/environment.yml
@@ -5,7 +5,6 @@ dependencies:
   - codecov
   - pip
   - pip:
-    - dataclasses
     - nltk
     - requests
     - revtok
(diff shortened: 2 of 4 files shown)


packaging/torchtext/meta.yaml
```diff
@@ -15,12 +15,14 @@ requirements:
     - python
     - setuptools
     - cpuonly
+    - pytorch-mutex 1.0 cpu  # [not osx ]
```
Nayef211, line 18: Can you elaborate a bit more as to why we are adding this to the host?

erip: It may not be required for host. I was mostly following the template in torchvision and [this PR](https://github.com/pytorch/builder/pull/823). The addition to host was from torchvision and I'm throwing things at the wall -- certainly not as principled as I'd like...

mthrok:
Regarding the `environment.yml` in. unit tests (not `meta.yml` in packaging), I originally added them thinking that it will give a good reproducibility, but later it turned out that it is not as robust when macOS CI is involved (this does not apply to torchtext though), so you are free to remove them and change them to regular pip/conda commands in the shell scripts.

erip:
OK, this is looking better. The changes are inspired by the torchvision `packaging/` scheme which sets some build variant (cpu vs. cuda) environment and conditionally includes the cpuonly if needed. This solves the issue with conflicts for pytorch-mutex and pytorch in the nightly anaconda distribution. It can probably be further cleaned up (e.g., do we even _need_ to specify pytorch-mutex?)

erip:
These errors are all related to quota issues on GDrive, FWIW.

parmeet:
> These errors are all related to quota issues on GDrive, FWIW.

This PR https://github.com/pytorch/text/pull/1522 might help fixing the issue. If we are seeing cache related issues on other PRs, hopefully this should fix it. cc: @Nayef211 , @abhinavarora

Nayef211:
> OK, this is looking better. The changes are inspired by the torchvision `packaging/` scheme which sets some build variant (cpu vs. cuda) environment variable and conditionally includes the cpuonly if needed. This solves the issue with conflicts for pytorch-mutex and pytorch in the nightly anaconda distribution. It can probably be further cleaned up (e.g., do we even _need_ to specify pytorch-mutex as a dep?) Relevant discussion [here](https://github.com/pytorch/builder/pull/823)

Thanks for helping to fix our CI build @erip! I don't want to approve these changes because I don't have a full grasp of what these changes are doing. @parmeet @abhinavarora @mthrok would any of you be able to verify if these changes look okay and if there's any additions in this PR that could be removed?

erip:
@Nayef211 I've tried reducing the surface area of this PR. For even smaller diffs we could leave the `*/unittest/*.yml` alone, but it's 6 of one, half a dozen of the other in terms of impact.

Nayef211:
> @Nayef211 I've tried reducing the surface area of this PR. For even smaller diffs we could leave the `*/unittest/*.yml` alone, but it's 6 of one, half a dozen of the other in terms of impact.

@erip I think this diff already looks a lot cleaner than before. I will merge this to make our CI builds green. If any more changes are needed, we can just create a followup diff!

Nayef211 (approved):

Nayef211:
> This change forces all `conda install torchtext` to downgrade both pytorch and torchvision to cpu only

@malfet do you have any ideas as to what changes we could make to resolve the versioning conflicts while preventing the cpu only installs? I'm not super familiar with our CI environment setup.

PR #1521, opened 2022-01-14
Update python_requires in setup.py to reflect support for non-EOL python versions

erip:


Nayef211 (approved):
Thanks for catching this. I guess this was missed in https://github.com/pytorch/text/pull/1484/files

PR #1522, opened 2022-01-14
fixing cache logic to work with datapipes

parmeet:
This PR fixes the caching issue on CI. The move to datapipes broken the cache build script since datapipes are lazy and does not download data unless being iterated upon.


erip:
LGTM

parmeet:
The cache is still failing due to quota exceed error. My guess is many PR are actively in parallel querying from gdrive causing the quota exceed issue (the very reason why we build cache in first place).,

erip:
Yes, I suspect the earlier PRs caused these to be throttled. 😖

parmeet:
> Yes, I suspect the earlier PRs caused these to be throttled. 😖

OK, we can merge this PR and while we get a healthy cache, let's continue testing locally and leave message in comment/summary when local testing is done on PR, so that we can get idea if it is safe to merge.. CC: @Nayef211 , @abhinavarora

Nayef211:
> > Yes, I suspect the earlier PRs caused these to be throttled. 😖
> 
> OK, we can merge this PR and while we get a healthy cache, let's continue testing locally and leave message in comment/summary when local testing is done on PR, so that we can get idea if it is safe to merge.. CC: @Nayef211 , @abhinavarora

SGTM. @ejguan  do you know if it's possible to update the maximum quota when submitting a download request programmatically? Is there any changes we can make to the `HTTPReaderIterDataPipe` or is the max queries parameter specified by the owner of the dataset?

Nayef211 (approved):
LGTM!

ejguan:
> SGTM. @ejguan do you know if it's possible to update the maximum quota when submitting a download request programmatically? Is there any changes we can make to the HTTPReaderIterDataPipe or is the max queries parameter specified by the owner of the dataset?

Sorry I don't know there is option to do so via a simple HTTP request. A more complicated header is probably needed for the HTTP request like embedding verification key. We may be able to use the API from google https://developers.google.com/drive/api/v3/about-sdk later to see if this can be bypassed.

Nayef211:
> Sorry I don't know there is option to do so via a simple HTTP request. A more complicated header is probably needed for the HTTP request like embedding verification key. We may be able to use the API from google https://developers.google.com/drive/api/v3/about-sdk later to see if this can be bypassed.

Gotcha. I've had some experience using the Google Cloud API and I remember being able to pass in a verification token which allows the account associated with that token to set the max number of queries per hour/day. It might be worth looking into if we want to have better support for GCP in the future. Perhaps having a datapipe specific to GPC could be a potential solution?

ejguan:
> > Sorry I don't know there is option to do so via a simple HTTP request. A more complicated header is probably needed for the HTTP request like embedding verification key. We may be able to use the API from google https://developers.google.com/drive/api/v3/about-sdk later to see if this can be bypassed.
> 
> Gotcha. I've had some experience using the Google Cloud API and I remember being able to pass in a verification token which allows the account associated with that token to set the max number of queries per hour/day. It might be worth looking into if we want to have better support for GCP in the future. Perhaps having a datapipe specific to GPC could be a potential solution?

Yeah. @NivekT also opened an issue to track it https://github.com/pytorch/data/issues/167. Like you said, a specific GDrive DataPipe using their API may needed. Note: depending on the design and API from google, we may want to keep the token privately hidden from external users.

Nayef211:
> Yeah. @NivekT also opened an issue to track it [pytorch/data#167](https://github.com/pytorch/data/issues/167). Like you said, a specific GDrive DataPipe using their API may needed. Note: depending on the design and API from google, we may want to keep the token privately hidden from external users.

It's great to hear you guys are already working on this! I'll continue to follow the discussions on the linked thread. GCP actually has a [Cloud Key Management service](https://cloud.google.com/security-key-management) that we could leverage so that we don't have to hardcode the tokens in code.

PR #1525, opened 2022-01-16
add max_tokens kwarg to vocab factory.

### LINKED ISSUE #1523: Add a `max_words` argument to `build_vocab_from_iterator`
## 🚀 Feature

[Link to the docs](https://pytorch.org/text/stable/vocab.html?highlight=build%20vocab#torchtext.vocab.build_vocab_from_iterator)

I believe it would be beneficial to limit the number of words you want in your vocabulary with an argument like `max_words`, e.g.:
```
vocab = build_vocab_from_iterator(yield_tokens_batch(file_path), specials=["<unk>"], max_words=50000)
```

**Motivation**

This allows a controllable-sized `nn.Embedding`, with rare words being mapped to `<unk>`. Otherwise, it would not be practical to use `build_vocab_from_iterator` for larger datasets.

**Alternatives**

Keras and Huggingface's tokenizers would be viable alternatives, but do not nicely integrate with the torchtext ecosystem.

erip:
Fixes #1523

### DIFF
--- a/torchtext/vocab/vocab_factory.py
+++ b/torchtext/vocab/vocab_factory.py
@@ -93,11 +93,15 @@ def build_vocab_from_iterator(iterator: Iterable, min_freq: int = 1, specials: O
         counter.update(tokens)
 
     specials = specials or []
+
+    # First sort by descending frequency, then lexicographically
+    sorted_by_freq_tuples = sorted(counter.items(), key=lambda x: (-x[1], x[0]))
+
     if max_tokens is None:
-        ordered_dict = OrderedDict(counter.most_common())
+        ordered_dict = OrderedDict(sorted_by_freq_tuples)
     else:
         assert len(specials) < max_tokens, "len(specials) >= max_tokens, so the vocab will be entirely special tokens."
-        ordered_dict = OrderedDict(counter.most_common(max_tokens - len(specials)))
+        ordered_dict = OrderedDict(sorted_by_freq_tuples[:max_tokens - len(specials)])
 
     word_vocab = vocab(ordered_dict, min_freq=min_freq, specials=specials, special_first=special_first)
     return word_vocab
--- a/torchtext/vocab/vocab_factory.py
+++ b/torchtext/vocab/vocab_factory.py
@@ -8,8 +8,7 @@ from torchtext._torchtext import (
 
 def vocab(ordered_dict: Dict, min_freq: int = 1,
           specials: Optional[List[str]] = None,
-          special_first: bool = True,
-          max_tokens: Optional[int] = None) -> Vocab:
+          special_first: bool = True) -> Vocab:
     r"""Factory method for creating a vocab object which maps tokens to indices.
 
     Note that the ordering in which key value pairs were inserted in the `ordered_dict` will be respected when building the vocab.
@@ -20,7 +19,6 @@ def vocab(ordered_dict: Dict, min_freq: int = 1,
         min_freq: The minimum frequency needed to include a token in the vocabulary.
         specials: Special symbols to add. The order of supplied tokens will be preserved.
         special_first: Indicates whether to insert symbols at the beginning or at the end.
-        max_tokens: If provided, creates the vocab from the `max_tokens - len(specials)` most frequent tokens.
 
     Returns:
         torchtext.vocab.Vocab: A `Vocab` object
@@ -52,11 +50,7 @@ def vocab(ordered_dict: Dict, min_freq: int = 1,
 
     tokens = []
     # Save room for special tokens
-    max_tokens = (max_tokens or float('inf')) - len(specials)
-    for i, (token, freq) in enumerate(ordered_dict.items()):
-        # Save room
-        if i >= max_tokens:
-            break
+    for token, freq in ordered_dict.items():
         if freq >= min_freq:
             tokens.append(token)
 
@@ -98,10 +92,12 @@ def build_vocab_from_iterator(iterator: Iterable, min_freq: int = 1, specials: O
     for tokens in iterator:
         counter.update(tokens)
 
-    sorted_by_freq_tuples = sorted(counter.items(), key=lambda x: x[0])
-    sorted_by_freq_tuples.sort(key=lambda x: x[1], reverse=True)
-    ordered_dict = OrderedDict(sorted_by_freq_tuples)
+    specials = specials or []
+    if max_tokens is None:
+        ordered_dict = OrderedDict(counter.most_common())
+    else:
+        assert len(specials) < max_tokens, "len(specials) >= max_tokens, so the vocab will be entirely special tokens."
+        ordered_dict = OrderedDict(counter.most_common(max_tokens - len(specials)))
 
-    word_vocab = vocab(ordered_dict, min_freq=min_freq, specials=specials or [],
-                       special_first=special_first, max_tokens=max_tokens)
+    word_vocab = vocab(ordered_dict, min_freq=min_freq, specials=specials, special_first=special_first)
     return word_vocab
--- a/torchtext/vocab/vocab_factory.py
+++ b/torchtext/vocab/vocab_factory.py
@@ -102,7 +102,6 @@ def build_vocab_from_iterator(iterator: Iterable, min_freq: int = 1, specials: O
     sorted_by_freq_tuples.sort(key=lambda x: x[1], reverse=True)
     ordered_dict = OrderedDict(sorted_by_freq_tuples)
 
-
     word_vocab = vocab(ordered_dict, min_freq=min_freq, specials=specials or [],
                        special_first=special_first, max_tokens=max_tokens)
     return word_vocab
--- a/torchtext/vocab/vocab_factory.py
+++ b/torchtext/vocab/vocab_factory.py
@@ -8,7 +8,8 @@ from torchtext._torchtext import (
 
 def vocab(ordered_dict: Dict, min_freq: int = 1,
           specials: Optional[List[str]] = None,
-          special_first: bool = True) -> Vocab:
+          special_first: bool = True,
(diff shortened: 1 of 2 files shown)


xhlulu:
Thanks for making this PR!

torchtext/vocab/vocab_factory.py
```diff
@@ -49,7 +51,12 @@ def vocab(ordered_dict: Dict, min_freq: int = 1,
         ordered_dict.pop(token, None)
 
     tokens = []
-    for token, freq in ordered_dict.items():
+    # Save room for special tokens
+    max_tokens = (max_tokens or float('inf')) - len(specials)
+    for i, (token, freq) in enumerate(ordered_dict.items()):
+        # Save room
+        if i >= max_tokens:
+            break
```
Nayef211, line 59: One issue I see with this approach is if the `ordered_dict` is not sorted in ascending order or sorted by frequency at all. For example, if the dict is sorted in ascending order, and `max_tokens` equals `10`, it could be possible that the first 10 tokens do not meet the min frequency to be included in the `tokens` list. But because we are using the count from `enumerate()` we would break out of the for-loop even after skipping the first 10 tokens in the `ordered_dict`. In the baseline implementation, we would iterate through all tokens regardless of how they were sorted so this wouldn't be a problem. 

I guess if we assume that the `ordered_dict` will always be sorted by frequency, we can avoid solving for this edge case. @parmeet do you have any thoughts on this?

erip: This is a good point. The ordering is enforced when building it using factory functions, but not strictly enforced if someone were to construct their own by calling `vocab` directly. One idea might be to push the sorting to `vocab`, another might be to assert order in `vocab`. I'm interested in your thoughts.

torchtext/vocab/vocab_factory.py
```diff
@@ -19,6 +20,7 @@ def vocab(ordered_dict: Dict, min_freq: int = 1,
         min_freq: The minimum frequency needed to include a token in the vocabulary.
```
xhlulu, line 20: How will `min_freq` interact with `max_tokens`? Does one take precedence over the other? i.e. if `max_tokens` is 50000, but `min_freq` causes the number of tokens to be kept to be 40000, do we ignore one or the other parameter (or give a user warning that both cannot be used at the same time)?

erip: I think it makes most sense to apply both.

torchtext/vocab/vocab_factory.py
```diff
@@ -95,5 +103,5 @@ def build_vocab_from_iterator(iterator: Iterable, min_freq: int = 1, specials: O
     ordered_dict = OrderedDict(sorted_by_freq_tuples)
```
xhlulu, line 103: SInce we already using `Counter`, would it be more computationally efficient if we used the built-in method `counter.most_common(max_words)` before passing to the `vocab()` function?

torchtext/vocab/vocab_factory.py
xhlulu: Makes sense! So I guess it's the `min(max_tokens, num_tokens_after_filtering_min_freq)`?

torchtext/vocab/vocab_factory.py
erip: I was actually thinking the same thing as I made this PR. I think the `OrderedDict` is a vestige of pre-python 3.7 dictionary ordering. 

More to the point, the question is ultimately whether `vocab` should handle `max_tokens` or the `build_from_*`. Since `vocab` also handles the `min_freq` constraint, it made sense to push `max_tokens` there as well. Whether that's ideal is another question.

xhlulu: Yeah totally makes sense. If `vocab` would accept a `max_tokens` then it would make sense to not duplicate the truncation in `build_vocab-from_iterator`.

torchtext/vocab/vocab_factory.py
xhlulu: Sorting would take `O(n log n)`, but seems like `Counter.most_common` takes `O(n log k)` according to [this thread](https://stackoverflow.com/questions/29240807/python-collections-counter-most-common-complexity). Wonder if it might be more computationally efficient to convert to `Counter` first (assuming the constant time of conversion does not exceed the theoretical gain when k << n).

parmeet: I guess the principle behind vocab factory API is to avoid any assumptions on how user want to construct their vocab since this is the most primitive and closest to the underlying C++ layer. I would be bit cautious of addition more options or  assumptions on the ordering here. For instance, what if there is a pre-existing vocab/order with the corresponding embedding table? We can certainly create higher order factory functions (build_vocab_from_iterator is one example) that can provide more functionality out-of-the-box, and uses vocab factory function to return the actual object.

erip: So does it make sense to pull the max_tokens up a layer to the build function and not let that trickle down to vocab? That's an easy enough change to make.

torchtext/vocab/vocab_factory.py
parmeet: > More to the point, the question is ultimately whether `vocab` should handle `max_tokens` or the `build_from_*`. Since `vocab` also handles the `min_freq` constraint, it made sense to push `max_tokens` there as well. Whether that's ideal is another question.

some insights here https://github.com/pytorch/text/pull/1525#discussion_r786256295.

torchtext/vocab/vocab_factory.py
parmeet: yupp, that would be my recommendation, this would help us avoid making assumptions in the primitive factory function :).

erip: Awesome -- will fix quickly.

erip:
In digging into tests, it seems like this complicates the implementation a bit. The special tokens are inserted into the vocabulary unconditionally (which is good), but the interplay with the `max_tokens` can be a bit tricky as it changes the semantics of `special_first` a bit. 

If we consider this silly example, I'd say that this passes is a bit unexpected:

```python
    it = [["hello", "world"], ["goodbye", "world"]]
    vocab = build_vocab_from_iterator(it, max_tokens=1, specials=["<unk>"])
    self.assertEqual(len(vocab), 1)
    # '<unk>' is added first
    self.assertEqual(vocab["<unk>"], 0)

    it = [["hello", "world"], ["goodbye", "world"]]
    vocab = build_vocab_from_iterator(it, max_tokens=1, specials=["<unk>"], special_first=False)
    self.assertEqual(len(vocab), 1)
    # '<unk>' is added last, but non-specials get evicted. Should this be the case?
    self.assertEqual(vocab["<unk>"], 0)
```

The change to `build_vocab_from_iterator is fairly small...

```patch

    sorted_by_freq_tuples = sorted(counter.items(), key=lambda x: x[0])
    sorted_by_freq_tuples.sort(key=lambda x: x[1], reverse=True)

+    if max_tokens is not None:
+        num_specials = len(specials or [])
+        sorted_by_freq_tuples = sorted_by_freq_tuples[:max_tokens-num_specials]

    ordered_dict = OrderedDict(sorted_by_freq_tuples)
````

**Comment by parmeet** [comment]
> but the interplay with the `max_tokens` can be a bit tricky as it changes the semantics of `special_first` a bit.

I would argue that at the concept level, max_tokens should always at-least accommodate special tokens. As long as we document it well to make it clear to the user, I think this should be left to the user to decide whether their vocab size includes special tokens or not, and they would set the max size accordingly. From library perspective, we could through an error if special tokens are more than max_tokens.

**Comment by Nayef211** [comment]
@erip @parmeet just want to double check if we would be okay to merge this PR? Seems like most of the conversations have been resolved

**Comment by erip** [comment]
I'd like to take another stab at poking holes at this. Hopefully I can get it in a good place this evening.

**Comment by Nayef211** [comment]
> I'd like to take another stab at poking holes in this. Hopefully I can get it in a good place this evening.

Gotcha lmk once it's ready to be reviewed!

**Comment by parmeet** [comment]
> @erip @parmeet just want to double check if we would be okay to merge this PR? Seems like most of the conversations have been resolved

I guess, we still need to undo vocab changes and keep the handling of max_tokens in client code based on previous conversations :)

**Comment by erip** [comment]
OK, I think this is ready.

File: torchtext/vocab/vocab_factory.py [hunk 4]
```diff
@@ -90,10 +92,12 @@ def build_vocab_from_iterator(iterator: Iterable, min_freq: int = 1, specials: O
     for tokens in iterator:
         counter.update(tokens)
 
-    sorted_by_freq_tuples = sorted(counter.items(), key=lambda x: x[0])
```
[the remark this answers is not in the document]
erip: Hmm, good catch. I'll revert and add a test.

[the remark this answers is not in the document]
erip: OK, I've added a test and simplified the sorting logic so we're not doing two full sorts. The tests look OK. Thanks again for catching that!

torchtext/vocab/vocab_factory.py
```diff
@@ -90,10 +92,16 @@ def build_vocab_from_iterator(iterator: Iterable, min_freq: int = 1, specials: O
...
-    sorted_by_freq_tuples.sort(key=lambda x: x[1], reverse=True)
-    ordered_dict = OrderedDict(sorted_by_freq_tuples)
+    specials = specials or []
+
+    # First sort by descending frequency, then lexicographically
+    sorted_by_freq_tuples = sorted(counter.items(), key=lambda x: (-x[1], x[0]))
```
parmeet, line 98: nice, so you combined the two separate sorting operations in 1 :).

parmeet (approved):
LGTM, thanks @erip  for adding the feature!

Issue #1526, opened 2022-01-18
Add secondary caching to newly-migrated datapipe-based datasets


erip:
This is mostly so things don't fall through the cracks...

## 🚀 Feature

**Motivation**

CI spends too much time unpacking cached datasets. We can also cache uncompressed and extracted files for relevant splits.

**Pitch**

Add secondary caching to datasets which have been migrated per #1494.

**Alternatives**

N/A

**Additional context**

See discussion in review of #1515

parmeet:
I would also add to Motivation that it helps improve workflow efficiency. With extracted files being cached on disk in the first iteration, datapipe would avoid doing the same in the consecutive iterations. Also the alternative would be [In Memory caching](https://github.com/pytorch/data/blob/c06066ae360fc6054fb826ae041b1cb0c09b2f3b/torchdata/datapipes/iter/util/cacheholder.py#L19) in which case the files won't be dumped to the disk. @ejguan just to confirm this is indeed the right understanding?

ejguan:
> I would also add to Motivation that it helps improve workflow efficiency. With extracted files being cached on disk in the first iteration, datapipe would avoid doing the extraction in the consecutive iterations. Also the alternative would be [In Memory caching](https://github.com/pytorch/data/blob/c06066ae360fc6054fb826ae041b1cb0c09b2f3b/torchdata/datapipes/iter/util/cacheholder.py#L19) in which case the files won't be dumped to the disk. @ejguan just to confirm this is indeed the right understanding?

Correct. But, it requires more careful design. If we do in memory caching over file handlers, it won't actually work unless you explicitly rewind them.

erip:
I think this can be closed since the previously-migrated datasets now have double caching and future migrations will require it as part of code review.

parmeet:
> I think this can be closed since the previously-migrated datasets now have double caching and future migrations will require it as part of code review. Feel free to reopen as necessary.

Thanks @erip for helping keep track of progress through this issue!

PR #1527, opened 2022-01-18
Cache extraction for AmazonReviewPolarity

parmeet:
This PR introduces caching for extracted files. It will help avoid repetitive extractions with every iteration. 

Reference issue: #1526 
### Comparison
command : 
```python

%timeit next(iter(AmazonReviewPolarity(split='train')))
```

Time take without caching: `2.17 s ± 52.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
Time taken with caching: `933 ms ± 17.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

Reducing time by almost 50% :)

### DIFF
--- a/torchtext/datasets/amazonreviewpolarity.py
+++ b/torchtext/datasets/amazonreviewpolarity.py
@@ -23,8 +23,8 @@ NUM_LINES = {
 _PATH = 'amazon_review_polarity_csv.tar.gz'
 
 _EXTRACTED_FILES = {
-    'train': os.path.join('amazon_review_polarity_csv', 'train.csv'),
-    'test': os.path.join('amazon_review_polarity_csv', 'test.csv'),
+    'train': os.path.join(_PATH, 'amazon_review_polarity_csv', 'train.csv'),
+    'test': os.path.join(_PATH, 'amazon_review_polarity_csv', 'test.csv'),
 }
 
 
@@ -40,11 +40,13 @@ def AmazonReviewPolarity(root: str, split: Union[Tuple[str], str]):
         raise ModuleNotFoundError("Package `torchdata` not found. Please install following instructions at `https://github.com/pytorch/data`")
 
     url_dp = IterableWrapper([URL])
-    cache_dp = url_dp.on_disk_cache(
+    cache_compressed_dp = url_dp.on_disk_cache(
         filepath_fn=lambda x: os.path.join(root, _PATH), hash_dict={os.path.join(root, _PATH): MD5}, hash_type="md5"
     )
-    cache_dp = GDriveReader(cache_dp).end_caching(mode="wb", same_filepath_fn=True)
-    cache_dp = FileOpener(cache_dp, mode="b")
-    extracted_files = cache_dp.read_from_tar()
-    filter_extracted_files = extracted_files.filter(lambda x: _EXTRACTED_FILES[split] in x[0])
-    return filter_extracted_files.parse_csv().map(fn=lambda t: (int(t[0]), ' '.join(t[1:])))
+    cache_compressed_dp = GDriveReader(cache_compressed_dp).end_caching(mode="wb", same_filepath_fn=True)
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, os.path.dirname(_EXTRACTED_FILES[split]), os.path.basename(x)))
+    cache_decompressed_dp = FileOpener(cache_decompressed_dp, mode="b").read_from_tar()
+    cache_compressed_dp = cache_decompressed_dp.end_caching(mode="wb", same_filepath_fn=True)
+    data_dp = FileOpener(cache_decompressed_dp.filter(lambda x: _EXTRACTED_FILES[split] in x[0]).map(lambda x: x[0]), mode='b')
+    return data_dp.parse_csv().map(fn=lambda t: (int(t[0]), ' '.join(t[1:])))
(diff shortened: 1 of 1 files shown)


torchtext/datasets/amazonreviewpolarity.py
```diff
@@ -40,11 +40,25 @@ def AmazonReviewPolarity(root: str, split: Union[Tuple[str], str]):
...
+        if not os.path.exists(dir_path):
+            os.makedirs(dir_path)
+        return file_path
+
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
+        filepath_fn=extracted_filepath_fn)
+    cache_decompressed_dp = FileOpener(cache_decompressed_dp, mode="b").\
+        read_from_tar().\
+        filter(lambda x: _EXTRACTED_FILES[split] in x[0]).\
+        map(lambda x: (x[0].replace('_PATH' + '/', ''), x[1]))
```
Nayef211, line 60: Should this be `f'{_PATH}'` instead of just `'_PATH'`

torchtext/datasets/amazonreviewpolarity.py
```diff
@@ -40,11 +40,25 @@ def AmazonReviewPolarity(root: str, split: Union[Tuple[str], str]):
...
         filepath_fn=lambda x: os.path.join(root, _PATH), hash_dict={os.path.join(root, _PATH): MD5}, hash_type="md5"
     )
-    cache_dp = GDriveReader(cache_dp).end_caching(mode="wb", same_filepath_fn=True)
-    cache_dp = FileOpener(cache_dp, mode="b")
-    extracted_files = cache_dp.read_from_tar()
-    filter_extracted_files = extracted_files.filter(lambda x: _EXTRACTED_FILES[split] in x[0])
-    return filter_extracted_files.parse_csv().map(fn=lambda t: (int(t[0]), ' '.join(t[1:])))
+    cache_compressed_dp = GDriveReader(cache_compressed_dp).end_caching(mode="wb", same_filepath_fn=True)
+
+    def extracted_filepath_fn(x):
```
Nayef211, line 48: Should we move this to `torchtext/data/datasets_utils.py` since we would likely want to reuse this across all of our datasets?

torchtext/datasets/amazonreviewpolarity.py
```diff
@@ -40,11 +40,25 @@ def AmazonReviewPolarity(root: str, split: Union[Tuple[str], str]):
...
+            os.makedirs(dir_path)
+        return file_path
+
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
+        filepath_fn=extracted_filepath_fn)
+    cache_decompressed_dp = FileOpener(cache_decompressed_dp, mode="b").\
+        read_from_tar().\
+        filter(lambda x: _EXTRACTED_FILES[split] in x[0]).\
+        map(lambda x: (x[0].replace('_PATH' + '/', ''), x[1]))
+    cache_decompressed_dp = cache_decompressed_dp.end_caching(mode="wb", same_filepath_fn=True)
```
Nayef211, line 61: nit: let's try to be consistent with our usage of single vs double quotes for strings in the torchtext repo. Maybe it's worth adding coding style preferences to the contribution guidelines.

torchtext/datasets/amazonreviewpolarity.py
parmeet: Ahh, that's a good catch, and I just realized it's not necessary to do this since I used `same_filepath_fn=True` in end_caching. So I guess we can remove this mapping of path altogether.

torchtext/datasets/amazonreviewpolarity.py
parmeet: I am not so sure, we might need to use different functions for different datasets depending on how to compression is set.

torchtext/datasets/amazonreviewpolarity.py
parmeet: It's a good point. I do not have a specific preference. Not sure, if there is a convention across PyTorch board. @mthrok Any thoughts/suggestions here?

mthrok: We rely on black to do the formatting, which prefers double quote. Recently @jdsgomes has set up a lint job for audio.

erip:
> Time take without caching: 2.17 s ± 52.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
Time taken with caching: 933 ms ± 17.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Huge improvement. Nice! Overall this LGTM.

torchtext/datasets/amazonreviewpolarity.py
Nayef211: Gotcha let's figure out where to put this function as we start updating the caching across our other datasets.

Nayef211 (approved):
Overall LGTM!

torchtext/datasets/amazonreviewpolarity.py
erip: I can confirm I've used this identical function in a few of my own double-caching migrations so it might be worth pulling out.

parmeet: Actually, While working in parallel here https://github.com/pytorch/data/pull/169 I realize we do not need to explicitly create the directory. I guess, we can get rid of this function all together. Refer to this comment thread (https://github.com/pytorch/data/pull/169#discussion_r787228019)

parmeet: BTW, the main idea behind this function is to look for whether the file already exist or not (refer here https://github.com/pytorch/data/pull/169#discussion_r787229179). So I guess, which files to check would depend and vary for different datasets.

PR #1531, opened 2022-01-19
Migrate IMDB to datapipes

kevinchn:
Migrate IMDB dataset to use torchdata datapipes

Reference Issue: #1494

Test:
`pytest test/data/test_builtin_datasets.py -k IMDB`

### DIFF
--- a/torchtext/datasets/imdb.py
+++ b/torchtext/datasets/imdb.py
@@ -1,10 +1,15 @@
-from torchtext.utils import download_from_url, extract_archive
-from torchtext.data.datasets_utils import _RawTextIterableDataset
-from torchtext.data.datasets_utils import _wrap_split_argument
+import os
+from pathlib import Path
+from typing import Union, Tuple
+
+from torchtext._internal.module_utils import is_module_available
 from torchtext.data.datasets_utils import _add_docstring_header
 from torchtext.data.datasets_utils import _create_dataset_directory
-import io
-from pathlib import Path
+from torchtext.data.datasets_utils import _wrap_split_argument
+
+if is_module_available("torchdata"):
+    from torchdata.datapipes.iter import FileOpener, IterableWrapper, HttpReader
+
 
 URL = 'http://ai.stanford.edu/~amaas/data/sentiment/aclImdb_v1.tar.gz'
 
@@ -23,16 +28,25 @@ DATASET_NAME = "IMDB"
 @_add_docstring_header(num_lines=NUM_LINES, num_classes=2)
 @_create_dataset_directory(dataset_name=DATASET_NAME)
 @_wrap_split_argument(('train', 'test'))
-def IMDB(root, split):
-    def generate_imdb_data(key, extracted_files):
-        for fname in extracted_files:
-            *_, split, label, file = Path(fname).parts
+def IMDB(root: str, split: Union[Tuple[str], str]):
+    if not is_module_available("torchdata"):
+        raise ModuleNotFoundError("Package `torchdata` not found. Please install following instructions at `https://github.com/pytorch/data`")
 
-            if key == split and (label in ['pos', 'neg']):
-                with io.open(fname, encoding="utf8") as f:
-                    yield label, f.read()
-    dataset_tar = download_from_url(URL, root=root,
-                                    hash_value=MD5, hash_type='md5')
-    extracted_files = extract_archive(dataset_tar)
-    iterator = generate_imdb_data(split, extracted_files)
-    return _RawTextIterableDataset(DATASET_NAME, NUM_LINES[split], iterator)
+    url_dp = IterableWrapper([URL])
+
+    cache_dp = url_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, _PATH),
+        hash_dict={os.path.join(root, _PATH): MD5}, hash_type="md5"
+    )
+    cache_dp = HttpReader(cache_dp).end_caching(mode="wb", same_filepath_fn=True)
+    cache_dp = FileOpener(cache_dp, mode="b")
+
+    extracted_files = cache_dp.read_from_tar()
+
+    def filter_imdb_data(key, fname):
+        *_, split, label, file = Path(fname).parts
+        return key == split and (label in ['pos', 'neg'])
+
+    filter_extracted_files = extracted_files.filter(lambda t: filter_imdb_data(split, t[0]))
+
+    return filter_extracted_files.readlines(decode=True).map(lambda t: (Path(t[0]).parts[-2], t[1]))
--- a/torchtext/datasets/imdb.py
+++ b/torchtext/datasets/imdb.py
@@ -41,7 +41,10 @@ def IMDB(root: str, split: Union[Tuple[str], str]):
     cache_compressed_dp = HttpReader(cache_compressed_dp).end_caching(mode="wb", same_filepath_fn=True)
 
     labels = {"neg", "pos"}
-    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(filepath_fn=lambda x: [os.path.join(root, "aclImdb", split, label) for label in labels])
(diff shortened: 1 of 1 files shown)


torchtext/datasets/imdb.py
```diff
@@ -23,16 +28,25 @@
...
+        hash_dict={os.path.join(root, _PATH): MD5}, hash_type="md5"
+    )
+    cache_dp = HttpReader(cache_dp).end_caching(mode="wb", same_filepath_fn=True)
+    cache_dp = FileOpener(cache_dp, mode="b")
+
+    extracted_files = cache_dp.read_from_tar()
+
+    def filter_imdb_data(key, fname):
+        *_, split, label, file = Path(fname).parts
+        return key == split and (label in ['pos', 'neg'])
```
Nayef211, line 48: In the current implementation, we construct the `['pos', 'neg']` list for every file contained within `extracted_files` datapipe. 

Instead, can we create a `labels` set outside the `filter_imdb_data()` function and then do a contains check to see if the label exists within the `labels` set? This would make it so the set is only created once regardless of the number of files within `extracted_files`. Also, `contains` operations are faster on a set than on a list in Python.

torchtext/datasets/imdb.py
```diff
@@ -23,16 +28,25 @@
...
+
+    filter_extracted_files = extracted_files.filter(lambda t: filter_imdb_data(split, t[0]))
+
+    return filter_extracted_files.readlines(decode=True).map(lambda t: (Path(t[0]).parts[-2], t[1]))
```
Nayef211, line 52: Can we also add double caching similar to what we're doing in the [YelpFullReview dataset PR](https://github.com/pytorch/text/pull/1529/files/1ff59b781c5ee72529f742382e0ad3d0b7871582..61cad246874ba2ac7f6e03f8ab93838f89287540#diff-8056dddf0ce233418c7df18c6c8e9d71e5f81498c0c5b2fe9dcfb793144e1be8R43-R54)?

Nayef211 (commented):
Thanks for helping with the dataset migration @kevinchn! Just left a couple of suggestions but otherwise this looks great!

torchtext/datasets/imdb.py
erip: Sets are asymptotically faster with `__contains__`, but with a len-2 iterable, they're about the same. Interestingly, I suspected checking `(label == "pos") or (label == "neg")` would be faster, but it's slower on average...

```python
In [1]: def f1(label):
   ...:     return label in ["pos", "neg"]

In [2]: def f2(label):
   ...:     return label in {"pos", "neg"}

In [3]: def f3(label):
   ...:     return (label == "pos") or (label == "neg")

In [4]: %timeit f1("pos")
90.9 ns ± 1.99 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit f1("neg")
109 ns ± 1.85 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [6]: %timeit f2("pos")
96 ns ± 2.38 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [7]: %timeit f2("neg")
94.2 ns ± 2.55 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [8]: %timeit f3("pos")
100 ns ± 3.85 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [9]: %timeit f3("neg")
120 ns ± 2.84 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

kevinchn:
Updated. Please take a look again!

torchtext/datasets/imdb.py
Nayef211: Thanks @erip for helping verify this! You're right that there is a negligible difference in performance when the length of the iterable is small. I just thought it would be better practice but I guess this would be more of a nit comment 😃

torchtext/datasets/imdb.py
```diff
@@ -23,16 +28,35 @@
...
+    url_dp = IterableWrapper([URL])
+
+    cache_compressed_dp = url_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, _PATH),
+        hash_dict={os.path.join(root, _PATH): MD5}, hash_type="md5"
+    )
+    cache_compressed_dp = HttpReader(cache_compressed_dp).end_caching(mode="wb", same_filepath_fn=True)
+
+    labels = {"neg", "pos"}
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(filepath_fn=lambda x: [os.path.join(root, "aclImdb", split, label) for label in labels])
```
Nayef211, line 44: nit: can we change `"aclImdb"` to `"aclImdb_v1"`. We have been trying to keep the names of the uncompressed folder and the tar archive the same for our datasets.

torchtext/datasets/imdb.py
```diff
@@ -23,16 +28,35 @@
...
+        # eg. fname = "aclImdb/train/neg/12416_3.txt"
+        *_, split, label, file = Path(fname).parts
+        return key == split and label in labels
+
+    cache_decompressed_dp = cache_decompressed_dp.filter(lambda t: filter_imdb_data(split, t[0]))
+
+    cache_decompressed_dp = cache_decompressed_dp.map(lambda t: (Path(t[0]).parts[-2], t[1]))  # eg. "aclImdb/train/neg/12416_3.txt" -> "neg"
+    cache_decompressed_dp = cache_decompressed_dp.readlines(decode=True)
+    cache_decompressed_dp = cache_decompressed_dp.lines_to_paragraphs()  # group by label in cache file
+    cache_decompressed_dp = cache_decompressed_dp.end_caching(mode="wt", filepath_fn=lambda x: os.path.join(root, "aclImdb", split, x))
```
Nayef211, line 58: Same comment as above

torchtext/datasets/imdb.py
```diff
@@ -23,16 +28,35 @@
...
+    cache_decompressed_dp = cache_decompressed_dp.read_from_tar()
+
+    def filter_imdb_data(key, fname):
+        # eg. fname = "aclImdb/train/neg/12416_3.txt"
+        *_, split, label, file = Path(fname).parts
+        return key == split and label in labels
+
+    cache_decompressed_dp = cache_decompressed_dp.filter(lambda t: filter_imdb_data(split, t[0]))
+
+    cache_decompressed_dp = cache_decompressed_dp.map(lambda t: (Path(t[0]).parts[-2], t[1]))  # eg. "aclImdb/train/neg/12416_3.txt" -> "neg"
```
Nayef211, line 55: nit: update comment to say `"aclImdb_v1"`

torchtext/datasets/imdb.py
```diff
@@ -23,16 +28,35 @@
...
+    cache_decompressed_dp = cache_decompressed_dp.readlines(decode=True)
+    cache_decompressed_dp = cache_decompressed_dp.lines_to_paragraphs()  # group by label in cache file
+    cache_decompressed_dp = cache_decompressed_dp.end_caching(mode="wt", filepath_fn=lambda x: os.path.join(root, "aclImdb", split, x))
+
+    data_dp = FileOpener(cache_decompressed_dp, mode="t")
+    # get label from cache file, eg. "aclImdb/train/neg" -> "neg"
```
Nayef211, line 61: nit: update comment to say `"aclImdb_v1"`

Nayef211 (approved):
This LGTM. Let's merge this in after fixing the nit comments!

torchtext/datasets/imdb.py
kevinchn: For this one, `aclImdb` refers to the actual folder name inside the tar. I'll keep it as is if there's no objections

Nayef211: Gotcha that makes sense

PR #1438, opened 2021-11-10
Updated sst2 dataset to accept `validate_hash` parameter

Nayef211:
Summary:
## Description
- Updated sst2 dataset to accept a `validate_hash` parameter
- This allows for testing using partial datasets since downloading the entire dataset takes much longer

Reviewed By: parmeet

Differential Revision: D32250435

fbshipit-source-id: 9b5e7183f62df69638e1a3af2107273daa6f4ac5

PR #1532, opened 2022-01-20
Add AmazonReviewPolarity Mocked Unit Test

Nayef211:
Reference Issue: #1493

## Summary
- Added mocked unit test for AmazonReviewPolarity to test
  - Correctness of loaded data using datapipes
  - Passing string or tuple to split arguments results in the same iterable dataset
- Creating `TempDirMixin` class to provide easy access to temp dir when testing (similar to the [implementation in torchaudio](https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/common_utils/case_utils.py#L17-L44))
- Patching `torchdata.datapipes.iter.util.cacheholder._hash_check` function to return `True`
   - Without this, the `OnDiskCacheHolderIterDataPipe` attempts to re-download the dataset since the hash of the full dataset doesn't match the hash of the mocked dataset

## Test
` pytest test/datasets/amazonreviewpolarity_test.py`

## Follow Up
- [ ] Patch GDriveReader to return tar archive to get more complete test coverage

### DIFF
--- a/test/datasets/amazonreviewpolarity_test.py
+++ b/test/datasets/amazonreviewpolarity_test.py
@@ -1,54 +1,48 @@
-#!/user/bin/env python3
-# Note that all the tests in this module require dataset (either network access or cached)
-import torchtext
-import json
-from parameterized import parameterized
-from ..common.torchtext_test_case import TorchtextTestCase
-from ..common.parameterized_utils import load_params
-from ..common.case_utils import TempDirMixin
 import os.path
+import random
+import string
+import tarfile
+from collections import defaultdict
 
+from parameterized import parameterized
 from torchtext.datasets.amazonreviewpolarity import AmazonReviewPolarity
 
+from ..common.case_utils import TempDirMixin
+from ..common.torchtext_test_case import TorchtextTestCase
 
 
 def get_mock_dataset(root_dir):
     """
     root_dir: directory to the mocked dataset
     """
-    mocked_data = []
-    sample_rate = 16000
-    transcript = "This is a test transcript."
+    base_dir = os.path.join(root_dir, "AmazonReviewPolarity")
+    compressed_dataset_path = os.path.join(
+        base_dir, "amazon_review_polarity_csv.tar.gz"
+    )
+    uncompressed_dataset_dir = os.path.join(base_dir, "amazon_review_polarity_csv")
+    os.makedirs(uncompressed_dataset_dir, exist_ok=True)
 
-    base_dir = os.path.join(root_dir, "ARCTIC", "cmu_us_aew_arctic")
-    txt_dir = os.path.join(base_dir, "etc")
-    os.makedirs(txt_dir, exist_ok=True)
-    txt_file = os.path.join(txt_dir, "txt.done.data")
-    audio_dir = os.path.join(base_dir, "wav")
-    os.makedirs(audio_dir, exist_ok=True)
+    # create empty tar file to skip dataset download
+    with tarfile.open(compressed_dataset_path, "w:gz") as tar:
+        dummy_file_path = os.path.join(base_dir, "dummy_file.txt")
+        with open(dummy_file_path, "w") as f:
+            pass
+        tar.add(dummy_file_path)
 
-    seed = 42
-    with open(txt_file, "w") as txt:
-        for c in ["a", "b"]:
+    seed = 1
+    mocked_data = defaultdict(list)
+    for file_name in ("train.csv", "test.csv"):
+        txt_file = os.path.join(uncompressed_dataset_dir, file_name)
+        with open(txt_file, "w") as f:
             for i in range(5):
-                utterance_id = f"arctic_{c}{i:04d}"
-                path = os.path.join(audio_dir, f"{utterance_id}.wav")
-                data = get_whitenoise(
-                    sample_rate=sample_rate,
-                    duration=3,
-                    n_channels=1,
-                    dtype="int16",
-                    seed=seed,
+                label = seed % 2 + 1
+                rand_string = " ".join(
+                    random.choice(string.ascii_letters) for i in range(seed)
                 )
-                save_wav(path, data, sample_rate)
-                sample = (
-                    normalize_wav(data),
-                    sample_rate,
-                    transcript,
-                    utterance_id.split("_")[1],
-                )
-                mocked_data.append(sample)
-                txt.write(f'( {utterance_id} "{transcript}" )\n')
+                dataset_line = (label, f"{rand_string} {rand_string}")
+                # append line to correct dataset split
+                mocked_data[os.path.splitext(file_name)[0]].append(dataset_line)
+                f.write(f'"{label}","{rand_string}","{rand_string}"\n')
                 seed += 1
     return mocked_data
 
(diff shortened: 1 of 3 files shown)


torchtext/datasets/amazonreviewpolarity.py
```diff
@@ -34,14 +34,18 @@
 @_add_docstring_header(num_lines=NUM_LINES, num_classes=2)
 @_create_dataset_directory(dataset_name=DATASET_NAME)
 @_wrap_split_argument(("train", "test"))
-def AmazonReviewPolarity(root: str, split: Union[Tuple[str], str]):
+def AmazonReviewPolarity(root: str, split: Union[Tuple[str], str], validate_hash: bool = True):
```
mthrok, line 37: You might want to be careful on adding a public constructor arguments for the sake of the tests (though this looks like an argument with valid usecases).

What do you anticipate the future extension of this dataset? If new options related to dataset are added, which will be used more often, then having this argument before it might not look elegant in the end. At the same time, if all the dataset is going to have the same argument, it looks consistent.

If this is intended only for testing, it is worth considering if it is possible to inject the hash in a different way. For example you can use `patch` to overwrite the value of `MD5` for the duration of the test.

https://stackoverflow.com/a/51567124

mthrok (approved):
overall it looks good.

nit: it is advised to split the changes that is not semantically related to the PR. (i.e. linting)

torchtext/datasets/amazonreviewpolarity.py
parmeet: I think patching approach make sense to me. I would be careful about exposing `validate_hash` to users.

test/datasets/amazonreviewpolarity_test.py
```diff
@@ -0,0 +1,82 @@
...
+    root_dir: directory to the mocked dataset
+    """
+    base_dir = os.path.join(root_dir, "AmazonReviewPolarity")
+    compressed_dataset_path = os.path.join(
+        base_dir, "amazon_review_polarity_csv.tar.gz"
+    )
+    uncompressed_dataset_dir = os.path.join(base_dir, "amazon_review_polarity_csv")
+    os.makedirs(uncompressed_dataset_dir, exist_ok=True)
+
+    # create empty tar file to skip dataset download
```
parmeet, line 25: I think we probably want to add non-empty tar-file with all the content structure as in original compressed file and avoid creating uncompressed directory. This way we would get code-coverage for extraction parts of the datapipes.

test/datasets/amazonreviewpolarity_test.py
```diff
@@ -0,0 +1,82 @@
...
+from collections import defaultdict
+
+from parameterized import parameterized
+from torchtext.datasets.amazonreviewpolarity import AmazonReviewPolarity
+
+from ..common.case_utils import TempDirMixin
+from ..common.torchtext_test_case import TorchtextTestCase
+
+
+def get_mock_dataset(root_dir):
```
parmeet, line 14: I wonder if it is technically feasible to Mock the HTTP link as well?

torchtext/datasets/amazonreviewpolarity.py
Nayef211: I think I ended up taking this approach because we used a `validate_hash` arg for testing the SST2 dataset (https://github.com/pytorch/text/pull/1438). I am trying to figure out how to mock the value of `MD5` and the `hash_dict` being passed into `on_disk_cache` but can't seem to figure out a good way to do this. I was thinking of following the suggestions in [this thread](https://stackoverflow.com/questions/18191275/using-pythons-mock-patch-object-to-change-the-return-value-of-a-method-called-w) but I'm not sure if we can mock out the `on_disk_cache` like this (especially since we call it twice and pass in different args to it each time). Do you guys have any tips or code pointers that might help?

cc @ejguan  @mthrok @parmeet

test/datasets/amazonreviewpolarity_test.py
Nayef211: I can look into this. I think it would be a matter of mocking the `GDriveReader` to return the tar-file

test/datasets/amazonreviewpolarity_test.py
Nayef211: Just updated the tar-file to contain the contents of the dataset!

torchtext/datasets/amazonreviewpolarity.py
mthrok: > how to mock the value of `MD5` and the `hash_dict` being passed into `on_disk_cache`

IIUC, once the tar file is created, then you can compute the hash value of the tar, and patch the value of `torchtext.datasets.amazonreviewpolarity.MD5`. 
Mocking the `on_disk_cache` is an alternative way to do so, but IMO, it's less flexible, because it requires precise understanding of how and when the function is called, which can be changed anytime in the future. 

Something like the following should work.  the patch command in `setUpClass` then undo the patch in `tearDownClass`.
https://docs.python.org/3/library/unittest.mock.html#patch-methods-start-and-stop

```python
def setUpClass(cls):
    super().setUpClass()
    cls.root_dir = cls.get_base_temp_dir()
    cls.samples = get_mock_dataset(cls.root_dir)
    val = compute_hash(<the archive file>)
    cls.patcher = patch('torchtext.datasets.amazonreviewpolarity.MD5', val)
    cls.patcher.start()

def tearDownClass(cls):
    super().tearDownClass()
    cls.patcher.stop()
```

test/datasets/amazonreviewpolarity_test.py
mthrok: Should be doable but you want to be careful how much you want to monkey patch things. When test fixture gets complicated, it becomes error-prone, especially because money patching is not what we do in library code thus we are not as familiar with the technique. And test fixtures are not tested, it's much harder to find bugs in fixtures.

test/datasets/amazonreviewpolarity_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+    return mocked_data
+
+
+class TestAmazonReviewPolarity(TempDirMixin, TorchtextTestCase):
+    root_dir = None
+    samples = []
+
+    @classmethod
+    def setUpClass(cls):
```
mthrok, line 56: It's safer to call `super().setUpClass()`.

torchtext/datasets/amazonreviewpolarity.py
ejguan: With this argument, users can determine if they want to check the hash of cached files. You guys can decide if users would turn this off to skip hash check.
Note: If the concern comes from API forward compatibility, it can be switched to a keyword argument to move it to the the end.

test/datasets/amazonreviewpolarity_test.py
parmeet: Just to be clear, my query was to patch the URL (like the proposal to patch the MD5) and not the code (GDriveReader). Meaning that the patched URL would work like a normal URL but instead would fetch the mocked dataset. Is this the same understanding you had @Nayef211 , @mthrok ? I am not sure though if this is doable or if there are still concerns doing this?

torchtext/datasets/amazonreviewpolarity.py
Nayef211: @mthrok thanks for the example. Let me try it with your approach to see if I can get this working.

@ejguan thanks for the input. I think we were only adding the `validate_hash` arg to make it easier for testing since we don't expect the user to modify the downloaded files so they should have no need to skip the hash check. If we can get the patching to work, then I will get rid of the `validate_hash` arg.

erip: I think the patch should be

```python
with patch("torchdata.datapipes.iter.util.cacheholder._hash_check", return_value=True):
   ...
```

test/datasets/amazonreviewpolarity_test.py
Nayef211: Let me make sure I understand your suggestion correctly @parmeet. The reason we are patching the `MD5` hash is because the tests would actually fail without it. Patching the URL in this case doesn't make sense to me since the `on_disk_cache` datapipe just expects to verify whether the files that were downloaded by the `GDriveReader` datapipe exists locally. What would the benefit of patching the URL to return the `tar.gz` archive accomplish that wouldn't be done by the AmazonReviewPolarity dataset implementation?

erip: I think the logic we should be testing is:

1a. If the MD5 doesn't match, assert that we make an HTTP call
1a. If the MD5 does match, assert we read from cache
2a. If we've cached the contents of the tar, assert that we don't read the tar again
2b. If we haven't, assert we extract.
3. Test the parsing logic simply

torchtext/datasets/amazonreviewpolarity.py
Nayef211: > I think a slightly more to-the-spec patch would be
> 
> ```python
> # Who cares what the MD5 is, we care about the behavior if hashes match
> with patch("torchdata.datapipes.iter.util.cacheholder._hash_check", return_value=True):
>    ...
> ```

Actually that does seem like the better approach, thanks @erip!

test/datasets/amazonreviewpolarity_test.py
Nayef211: > I think the logic we should be testing is:
> 
> 1a. If the MD5 doesn't match, assert that we make an HTTP call 1a. If the MD5 does match, assert we read from cache 2a. If we've cached the contents of the tar, assert that we don't read the tar again 2b. If we haven't, assert we extract. 3. Test the parsing logic simply

I wonder if we have to test the logic for 1 and 2 since that's the expected behavior from the `OnDiskCacheHolderIterDataPipe` that we're using from torchdata. I assume their unit tests already do the check to verify that if the hash doesn't match, then retry the download and if the contents of an archive are cached, it shouldn't be extracted again. @ejguan would you be able to verify if this is indeed the case? Or do you think it's worthwhile adding tests for this as well?

I'm also following the same testing strategy used by torchaudio which assumes that any functions being used to implement the dataset has already been tested (in this case this would be datapipes). Instead we just test the API of the dataset to validate whether the expected input args result in the expected outputs. @mthrok could you weigh in with your thoughts as well?

parmeet: So as we are making the compressed file available in the folder, which is essentially the second step in the pipeline (the first being the download part). This means that any code/datapipe responsible for download part won't be executed/tested making it vulnerable for modifications.  The idea was simply to cover 100% of the code through testing :).

parmeet: > I'm also following the same testing strategy used by torchaudio which assumes that any functions being used to implement the dataset has already been tested (in this case this would be datapipes). 

I agree, we probably do not want to write tests that explicitly tests external library logic/workflow.

ejguan: `OnDiskCache` would check the hash, if it doesn't match, it will trigger the following ops (Downloader or Extract) in this case.

parmeet: > So as we are making the compressed file available in the folder, which is essentially the second step in the pipeline (the first being the download part). This means that any code/datapipe responsible for download part won't be executed/tested making it vulnerable for modifications. The idea was simply to cover 100% of the code through testing :).

BTW, as a very stupid example, if by mistake I end up changing GDriveReader to HttpReader, or change the url_dp to point to something else (instead of URL), I could completely bypass the current unit-testing.  That said, I don't want to block this PR for this. I don't mind merging it as such as it is already a huge improvement over what we had. Feel free to merge it without further due :)

parmeet (approved):
Overall LGTM! This is a huge improvement over the existing dataset testing strategy and will have overall improvement in dev-experience (much less time spent in CI testing as well as no flaky tests due to download issues). Thanks so much @Nayef211 for doing this!

test/datasets/amazonreviewpolarity_test.py
mthrok: > 1a. If the MD5 doesn't match, assert that we make an HTTP call
> 1a. If the MD5 does match, assert we read from cache
> 2a. If we've cached the contents of the tar, assert that we don't read the tar again
> 2b. If we haven't, assert we extract.
> 3. Test the parsing logic simply

Regarding the @erip's points I think they are valid. As I do not know which component implements which logic, if 1 and 2 are what torchdata is responsible for and tested there, I do not think we need to test it here.

> I'm also following the same testing strategy used by torchaudio which assumes that any functions being used to implement the dataset has already been tested (in this case this would be datapipes). Instead we just test the API of the dataset to validate whether the expected input args result in the expected outputs. @mthrok could you weigh in with your thoughts as well?

In the context of mocking and patching, I prefer to avoid patching the code as much as possible, because in my experience, patching requires more knowledge about the implementation detail, and as the library code changes, these implementation detail changes, thus it increases the maintenance cost. Which is why I said "be careful how much you want to monkey patch stuff". (Of course, the notable exception to this is network call and such, which otherwise would not be possible.)
The reason why I chose to mock dataset is because they won't change. The datasets by nature are not supposed to change. So even when the library code changes dramatically, the implementation should be centered around the data. Thus we would not need to change the test fixture that much. (and as a bonus, we found a unexpected bug when the mock data was introduced. The order of files returned by glob was not consistent across OS. This would not be caught by patching)
Note that in torchaudio repo, we do not test the download part and the dataset implementation does not check cache if not downloading a new archive. so the code coverage is not 100%. I personally consider this a reasonable trade-off.

Nayef211: Thanks for the detailed feedback and for describing how torchaudio approaches testing datasets @mthrok. Moving forward, I am going to stick to testing the core functionality of the dataset functions in this test suite. 

I still think @parmeet makes a valid point about getting better test coverage by patching `GDriveReader` datapipe to return the tar archive even if it adds a bit of maintenance overhead (like you said, its a tradeoff between maintenance and test coverage). I am going to merge this PR as is but let me add this patch in a followup PR so we can discuss further if it makes sense from a maintenance vs test coverage perspective.

torchtext/datasets/amazonreviewpolarity.py
Nayef211: @erip your suggestion for patching "torchdata.datapipes.iter.util.cacheholder._hash_check" to return `True` worked really well here! Would you mind reviewing this PR one more time before I merge it in?

erip:
LGTM! Very nice to see this coming together. Nice one, @Nayef211

torchtext/datasets/amazonreviewpolarity.py
erip: I think this looks good!

test/datasets/amazonreviewpolarity_test.py
```diff
@@ -0,0 +1,82 @@
...
+from ..common.torchtext_test_case import TorchtextTestCase
+
+
+def get_mock_dataset(root_dir):
+    """
+    root_dir: directory to the mocked dataset
+    """
+    base_dir = os.path.join(root_dir, "AmazonReviewPolarity")
+    temp_dataset_dir = os.path.join(base_dir, "temp_dataset_dir")
+    os.makedirs(temp_dataset_dir, exist_ok=True)
```
erip, line 21: Does this leak?

PR #1536, opened 2022-01-23
migrate Multi30k to datapipes.

erip:
Reference #1494


erip:
Unit tests look good, manual test looks good, confirmed a ~25% speedup on second call of `next(iter(ds))` due to caching. 🚀

parmeet (approved):
LGTM! Thanks :).

PR #1541, opened 2022-01-27
Implement ClipTokenizer that builds on top of GPT2BPETokenizer

abhinavarora:
This PR implements the CLIP tokenizer released by OpenAI:

- Implement a C++ CLIPEncoder that extends the the BPE encoder with the necessary modifications.
- Adds a Python interface for `CLIPTokenizer`
- Adds unit tests for the `CLIPTokenizer`
- Adds an entry transforms docs
- Fixes a bug in the `BPE_` method that is applicable only to CLIP and was identified with CLIP texting. 

## Testing

Test CLIP Tokenizer
```
pytest -k test_clip_ test/test_transforms.py
```

Test GPT2BPE Tokenizer
```
pytest -k test_gpt2_ test/test_transforms.py
```


abhinavarora:
Test failures seem to be coming from dataset testing failures!

torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -193,15 +192,15 @@ std::vector<std::string> GPT2BPEEncoder::BPE_(
...
   if (caching_enabled_ && cache_.contains(concatenated)) {
     return cache_.at(concatenated);
   }
 
   std::vector<std::string> tok_list = token_list;
-  auto pairs = _get_pairs(tok_list, seperator_);
+  auto pairs = get_pairs(tok_list, seperator_);
   if (pairs.empty()) {
-    return tok_list;
+    return {concatenated};
```
ebsmothers, line 203: Is this the bug you mentioned previously when trying to use GPT2 BPE for CLIP tokenization?

torchtext/csrc/clip_tokenizer.cpp
```diff
@@ -0,0 +1,137 @@
...
+    // ["a", "w", "some", "a", "w", "e"]
+    // Result: new_token_list = ["aw", "some", "aw", "e"]
+
+    auto parts = split_tokens(bigram, seperator_);
+    std::vector<std::string> new_token_list;
+    std::size_t i = 0;
+    while (i < tok_list.size()) {
+      auto j = list_str_index(tok_list, parts.first, i);
+      if (j != -1) {
+        for (int k = i; k < j; k++) new_token_list.push_back(tok_list[k]);
```
ebsmothers, line 63: Split into multiple lines?

ebsmothers (approved):
Thanks so much for implementing this! This looks great from my perspective.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
abhinavarora: Yes this is the bug. Although GPT-2 never really runs into this because pairs is not empty there. I came across this only when  testing with CLIP tokenizer.

torchtext/csrc/clip_tokenizer.cpp
abhinavarora: Thanks for sharing this. Currently this file is formatted with Google C++ Style . We are in the process of setting up clang-format for torchtext, so all formatting can be handled there.

torchtext/transforms.py
```diff
@@ -309,6 +309,105 @@ def __prepare_scriptable__(self):
...
+            return self._tokenize(input)
+        else:
+            raise TypeError("Input type not supported")
+
+    def __prepare_scriptable__(self):
+        r"""Return a JITable tokenizer.
+        """
+        if not self.is_jitable:
+            tokenizer_copy = deepcopy(self)
+            # Disable caching in script mode
```
parmeet, line 402: why do we disable caching in script mode? Also do we have some stats on how much we gain performance with caching?

abhinavarora: Disabling caching in script mode is a safe option. In future, we can design an API that can expose whether model developers want caching in script mode or not.

In script mode the tokenizer is used mostly for production and the cache can technically grow infinitely in such a case. This can lead to OOM errors which will be hard to debug.

test/test_transforms.py
```diff
@@ -291,3 +291,58 @@ def test_gpt2_bpe_tokenizer_save_load_torchscript(self):
         torch.save(tokenizer.__prepare_scriptable__(), tokenizer_path)
         loaded_tokenizer = torch.load(tokenizer_path)
         self._gpt2_bpe_tokenizer((loaded_tokenizer))
+
+
+class TestCLIPTokenizer(TorchtextTestCase):
+    def _load_tokenizer(self, test_scripting):
+        encoder_json = "clip_encoder.json"
+        bpe_vocab = "clip_vocab.bpe"
```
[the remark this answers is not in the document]
abhinavarora: Yes they are different. Since our unit test is more like an e2e test, it is better to rely on prod files which we don't expect to change at the moment.

torchtext/csrc/register_pybindings.cpp
```diff
@@ -178,6 +179,27 @@ PYBIND11_MODULE(_torchtext, m) {
...
           }));
 
+  py::class_<CLIPEncoder, c10::intrusive_ptr<CLIPEncoder>>(m, "CLIPEncoder")
+      .def(py::init<std::unordered_map<std::string, int64_t>,
+                    std::unordered_map<std::string, int64_t>, std::string,
+                    std::unordered_map<int64_t, std::string>, bool>())
+      .def_property_readonly("bpe_encoder_", &CLIPEncoder::GetBPEEncoder)
+      .def_property_readonly("bpe_merge_ranks_", &CLIPEncoder::GetBPEMergeRanks)
+      .def_readonly("seperator_", &CLIPEncoder::seperator_)
+      .def_property_readonly("byte_encoder_", &CLIPEncoder::GetByteEncoder)
```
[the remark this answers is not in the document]
parmeet: ahh right, my bad. Completely missed that part. Thanks!

torchtext/transforms.py
parmeet: Hmm, I see. Sounds good. Well, technically we can always limit how much cache we want to keep, but I agree that  that would require some API redesign for user-control etc. We can park this topic until then.

test/test_transforms.py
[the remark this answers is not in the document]
parmeet: As a follow-up, let's also think about how to expose the prod files for tokenizers to the users. If the user just want to use the pre-trained text processing files from established models (XLM-R, Roberta, CLIP), we might want to think how to best expose it. As of now we do not have a standard way of doing so.

parmeet (approved):
LGTM, thanks for adding this feature :)

ProGamerGov:
So, how soon until this makes it's way into the release versions?

And do I need to create a `.json` file to use CLIP's vocab file from here? https://github.com/openai/CLIP/blob/main/clip/bpe_simple_vocab_16e6.txt.gz

parmeet:
> So, how soon until this makes it's way into the release versions?

This should be part of our upcoming release in March.

> And do I need to create a `.json` file to use CLIP's vocab file from here? https://github.com/openai/CLIP/blob/main/clip/bpe_simple_vocab_16e6.txt.gz, and if so how do I create it?

Yes, you would need to pass the `encode json file` per the current interface exposed in transforms. You can create  it as per the code [here](https://github.com/openai/CLIP/blob/40f5484c1c74edd83cb9cf687c6ab92b28d8b656/clip/simple_tokenizer.py#L74) and save the dictionary as json file. cc: @abhinavarora

ProGamerGov:
@abhinavarora and @parmeet , seeing as the encoder file is just the output of the vocab file run through, I think that it might make sense to make the `encoder.json` file optional: https://github.com/pytorch/text/issues/1612

abhinavarora:
> @abhinavarora and @parmeet , seeing as the encoder file is just the output of the vocab file, I think that it might make sense to make the `encoder.json` file optional: #1612

@ProGamerGov I replied to you on #1612. Let's continue the discussion there :-) For now, you can use the json file and the txt file from HuggingFace which can be found [here](https://github.com/huggingface/transformers/blob/v4.16.2/src/transformers/models/clip/tokenization_clip.py#L36)

PR #1542, opened 2022-01-27
Add SST2 Mocked Unit Test

Nayef211:
Reference Issue: #1493

## Summary
- Added mocked unit test for SST dataset 
- Added `zip_equal` helper function to help test for unequal dataset lengths ([stackoverflow thread](https://stackoverflow.com/questions/32954486/zip-iterators-asserting-for-equal-length-in-python/32954700#32954700))

## Test
- ` pytest test/datasets/test_sst2.py`

## Followup Items
- [ ] Update AmazonReviewPolarity test to be consistent with changes in this PR

### DIFF
--- a/test/datasets/test_sst2.py
+++ b/test/datasets/test_sst2.py
@@ -46,7 +46,7 @@ def _get_mock_dataset(root_dir):
                 seed += 1
 
     compressed_dataset_path = os.path.join(base_dir, "SST-2.zip")
-    # create tar file from dataset folder
+    # create zip file from dataset folder
     with zipfile.ZipFile(compressed_dataset_path, "w") as zip_file:
         for file_name in ("train.tsv", "test.tsv", "dev.tsv"):
             txt_file = os.path.join(temp_dataset_dir, file_name)
--- a/test/datasets/test_sst2.py
+++ b/test/datasets/test_sst2.py
@@ -20,7 +20,6 @@ def _get_mock_dataset(root_dir):
     temp_dataset_dir = os.path.join(base_dir, "temp_dataset_dir")
     os.makedirs(temp_dataset_dir, exist_ok=True)
 
-    print(base_dir)
     seed = 1
     mocked_data = defaultdict(list)
     for file_name, (col1_name, col2_name) in zip(
--- a/test/datasets/test_sst2.py
+++ b/test/datasets/test_sst2.py
@@ -85,7 +85,7 @@ class TestSST2(TempDirMixin, TorchtextTestCase):
             n_iter += 1
         assert n_iter == len(self.samples[split])
 
-    @parameterized.expand(["train", "dev", "test"])
+    @parameterized.expand(["train", "test", "dev"])
     def test_sst2_split_argument(self, split):
         dataset1 = SST2(root=self.root_dir, split=split)
         (dataset2,) = SST2(root=self.root_dir, split=(split,))
--- a/test/datasets/test_sst2.py
+++ b/test/datasets/test_sst2.py
@@ -1,0 +1,101 @@
+import os
+import random
+import string
+import zipfile
+from collections import defaultdict
+from unittest.mock import patch
+
+from parameterized import parameterized
+from torchtext.datasets.sst2 import SST2
+
+from ..common.case_utils import TempDirMixin
+from ..common.torchtext_test_case import TorchtextTestCase
+
+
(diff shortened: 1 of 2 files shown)


test/datasets/test_sst2.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+    def test_sst2_split_argument(self, split1, split2):
+        with patch(
+            "torchdata.datapipes.iter.util.cacheholder._hash_check", return_value=True
+        ):
+            dataset1 = SST2(root=self.root_dir, split=split1)
+            (dataset2,) = SST2(root=self.root_dir, split=split2)
+
+            for d1, d2 in zip(dataset1, dataset2):
+                self.assertEqual(d1, d2)
```
mthrok, line 100: with regular `zip`, if one dataset is longer than the other, the reminder portions are ignored.

Probably it's worth adding a custom zip (until Python 3.10 becomes minimum supported version, then we can use `strict=True` in `zip`). https://stackoverflow.com/a/32954700

test/datasets/test_sst2.py
```diff
@@ -0,0 +1,100 @@
...
+                    expected_sample = self.samples[split][i]
+                    assert text == expected_sample[0]
+                    assert label == expected_sample[1]
+                    n_iter += 1
+            assert n_iter == len(self.samples[split])
+
+    @parameterized.expand(
+        [("train", ("train",)), ("dev", ("dev",)), ("test", ("test",))]
+    )
+    def test_sst2_split_argument(self, split1, split2):
```
mthrok, line 92: nit: I think you can also do 

```python
    @parameterized.expand(
         [("train", ), ("dev", ), ("test", )]
     )
     def test_sst2_split_argument(self, split):
        ...
             dataset1 = SST2(root=self.root_dir, split=split)
             (dataset2,) = SST2(root=self.root_dir, split=(split, ))
```

test/datasets/test_sst2.py
```diff
@@ -0,0 +1,100 @@
...
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.root_dir = cls.get_base_temp_dir()
+        cls.samples = _get_mock_dataset(cls.root_dir)
+
+    @parameterized.expand(["train", "test", "dev"])
+    def test_sst2(self, split):
+        with patch(
+            "torchdata.datapipes.iter.util.cacheholder._hash_check", return_value=True
+        ):
```
mthrok, line 72: Since this patch will be required on all the test cases, you can put it in `setUp` and `tearDown` methods.

test/datasets/test_sst2.py
```diff
@@ -0,0 +1,100 @@
...
+            dataset = SST2(root=self.root_dir, split=split)
+            n_iter = 0
+
+            if split == "test":
```
mthrok, line 76: nit: I think that the difference in looped logic among `"test"` and the others are small, and 

```python
for i, sample in enumerate(dataset):
    expected_sample = self.samples[split][i]
    assert sample[0] == expected_sample[0]
    if split != "test":
        assert sample[1] == expected_sample[1]
```

or maybe simply the following could work?

```python
for i, sample in enumerate(dataset):
    assert sample == self.samples[split][i]
```

test/datasets/test_sst2.py
Nayef211: I wasn't aware of this limitation of `zip`, thanks for bringing it up. I added a helper function `zip_equal` from the stackoverflow thread to deal with this!

test/datasets/test_sst2.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+        cls.patcher.stop()
+
+    @parameterized.expand(["train", "test", "dev"])
+    def test_sst2(self, split):
+        dataset = SST2(root=self.root_dir, split=split)
+        n_iter = 0
```
mthrok, line 80: The original reason I introduced `n_iter` is to make sure that the dataset returns the same number of items as in expected samples. Now this is resolved using `zip_equal`. Maybe this can be simplified.

```python
samples = list(dataset)
expected_samples = self.samples[split]
for sample, expected_sample in zip_equal(samples, expected_samples):
    assert sample == expected_sample
```

mthrok (approved):

parmeet (approved):
LGTM!

PR #1543, opened 2022-01-27
Convert get_mock_dataset fn in AmazonReviewPolarity to be private

Nayef211:
- [minor] Convert get_mock_dataset fn in AmazonReviewPolarity to be private since it's only used for dataset testing


parmeet (approved):
LGTM!

mthrok (approved):

PR #1545, opened 2022-01-28
migrate IWSLT2016 to datapipes.

erip:
Reference #1494 

This is a bit messy, but seems kosher. cc @ejguan for review

### DIFF
--- a/torchtext/data/datasets_utils.py
+++ b/torchtext/data/datasets_utils.py
@@ -67,7 +67,7 @@ def _clean_inner_tags_file(f_orig, base, stream):
         '<url', '<keywords', '<talkid', '<description', '<reviewer',
         '<translator', '<title', '<speaker', '<doc', '</doc'
     ]
-    f_txt = os.path.basename(f_orig.replace('.tags', ''))
+    f_txt = os.path.join(base, os.path.basename(f_orig.replace('.tags', '')))
     os.makedirs(base, exist_ok=True)
     with codecs.open(f_txt, mode='w', encoding='utf-8') as fd_txt:
         for line in stream.readlines():
--- a/torchtext/data/datasets_utils.py
+++ b/torchtext/data/datasets_utils.py
@@ -33,27 +33,23 @@ def _clean_xml_file(f_xml):
                 fd_txt.write(e.text.strip() + '\n')
 
 
-def _clean_inner_xml_file(f_xml, base, stream):
-    """Accepts an XML filename within a tarball and a stream of the byte contents
-    within that file and writes the cleaned contents to a new, untarred file
-    found in the provided base directory.
+def _clean_inner_xml_file(outfile, stream):
+    """Accepts an output filename and a stream of the byte contents of an XML file
+    within a tarball and writes the cleaned contents to a new, untarred file.
 
     Args:
-        f_orig: the full path of the XML file in the archive
-        base: the directory to which the new file should be written
-        stream: the byte datapipe of the contents of f_orig
+        outfile: the path to which the modified stream should be written
+        stream: the byte datapipe of the contents of the archived XML file
 
     Returns: the path to the newly-written file
     """
-    f_txt = os.path.basename(os.path.splitext(f_xml)[0])
-    os.makedirs(base, exist_ok=True)
-    out_file = os.path.join(base, f_txt)
-    with codecs.open(out_file, mode='w', encoding='utf-8') as fd_txt:
+    os.makedirs(os.path.dirname(outfile), exist_ok=True)
+    with codecs.open(outfile, mode='w', encoding='utf-8') as fd_txt:
         root = ET.fromstring(stream.read().decode("utf-8"))[0]
         for doc in root.findall('doc'):
             for e in doc.findall('seg'):
                 fd_txt.write(e.text.strip() + '\n')
-    return os.path.join(base, f_txt)
+    return outfile
 
 
 def _clean_tags_file(f_orig):
@@ -73,15 +69,13 @@ def _clean_tags_file(f_orig):
                 fd_txt.write(line.strip() + '\n')
 
 
-def _clean_inner_tags_file(f_orig, base, stream):
-    """Accepts a tags filename within a tarball and a stream of the byte contents
-    within that file and writes the cleaned contents to a new, untarred file
-    found in the provided base directory.
+def _clean_inner_tags_file(outfile, stream):
+    """Accepts an output filename and a stream of the byte contents of a tags file
+    within a tarball and writes the cleaned contents to a new, untarred file.
 
     Args:
-        f_orig: the full path of the tags file in the archive
-        base: the directory to which the new file should be written
-        stream: the byte datapipe of the contents of f_orig
+        outfile: the path to which the modified stream should be written
+        stream: the byte datapipe of the contents of the archived tags file
 
     Returns: the path to the newly-written file
     """
@@ -89,9 +83,8 @@ def _clean_inner_tags_file(f_orig, base, stream):
         '<url', '<keywords', '<talkid', '<description', '<reviewer',
         '<translator', '<title', '<speaker', '<doc', '</doc'
     ]
-    f_txt = os.path.join(base, os.path.basename(f_orig.replace('.tags', '')))
-    os.makedirs(base, exist_ok=True)
-    with codecs.open(f_txt, mode='w', encoding='utf-8') as fd_txt:
+    os.makedirs(os.path.dirname(outfile), exist_ok=True)
+    with codecs.open(outfile, mode='w', encoding='utf-8') as fd_txt:
         for line in stream.readlines():
             if not any(tag in line.decode("utf-8") for tag in xml_tags):
                 # TODO: Fix utf-8 next line mark
@@ -99,36 +92,32 @@ def _clean_inner_tags_file(f_orig, base, stream):
                 #                fd_txt.write(l.strip() + u"\u0085")
                 #                fd_txt.write(l.lstrip())
                 fd_txt.write(line.decode("utf-8").strip() + '\n')
-    return f_txt
+    return outfile
 
 
-def _rewrite_text_file(file, base, stream):
-    """Accepts a text filename within a tarball and a stream of the byte contents
-    within that file and writes the cleaned contents to a new, untarred file
-    found in the provided base directory.
+def _rewrite_text_file(outfile, stream):
+    """Accepts an output filename and a stream of the byte contents of a text file
+    within a tarball and writes the cleaned contents to a new, untarred file.
 
     Args:
-        f_orig: the full path of the text file in the archive
-        base: the directory to which the new file should be written
-        stream: the byte datapipe of the contents of f_orig
+        outfile: the path to which the modified stream should be written
+        stream: the byte datapipe of the contents of the archived text file
 
     Returns: the path to the newly-written file
     """
-    f_txt = os.path.basename(file)
-    os.makedirs(base, exist_ok=True)
-    out_file = os.path.join(base, f_txt)
-    with open(out_file, 'w', encoding='utf-8') as f:
+    os.makedirs(os.path.dirname(outfile), exist_ok=True)
+    with open(outfile, 'w', encoding='utf-8') as f:
         for line in stream.readlines():
-            f.write(line.decode("utf-8"))
-    return out_file
+            f.write(line.decode("utf-8") + "\n")
+    return outfile
 
 
-def _clean_files(fname, base, stream):
+def _clean_files(outfile, fname, stream):
     if 'xml' in fname:
-        return _clean_inner_xml_file(fname, base, stream)
+        return _clean_inner_xml_file(outfile, stream)
     elif "tags" in fname:
-        return _clean_inner_tags_file(fname, base, stream)
-    return _rewrite_text_file(fname, base, stream)
+        return _clean_inner_tags_file(outfile, stream)
+    return _rewrite_text_file(outfile, stream)
 
 
 def _create_data_from_json(data_path):
--- a/torchtext/data/datasets_utils.py
+++ b/torchtext/data/datasets_utils.py
@@ -89,6 +89,12 @@ def _rewrite_text_file(file, base, stream):
             f.write(line.decode("utf-8"))
     return out_file
 
+def _clean_files(fname, base, stream):
+    if 'xml' in fname:
+        return _clean_inner_xml_file(fname, base, stream)
+    elif "tags" in fname:
+        return _clean_inner_tags_file(fname, base, stream)
+    return _rewrite_text_file(fname, base, stream)
 
 def _create_data_from_json(data_path):
     with open(data_path) as json_file:
--- a/torchtext/data/datasets_utils.py
+++ b/torchtext/data/datasets_utils.py
@@ -34,6 +34,17 @@ def _clean_xml_file(f_xml):
 
 
 def _clean_inner_xml_file(f_xml, base, stream):
+    """Accepts an XML filename within a tarball and a stream of the byte contents
+    within that file and writes the cleaned contents to a new, untarred file
+    found in the provided base directory.
+
+    Args:
+        f_orig: the full path of the XML file in the archive
+        base: the directory to which the new file should be written
+        stream: the byte datapipe of the contents of f_orig
+
+    Returns: the path to the newly-written file
+    """
     f_txt = os.path.basename(os.path.splitext(f_xml)[0])
     os.makedirs(base, exist_ok=True)
     out_file = os.path.join(base, f_txt)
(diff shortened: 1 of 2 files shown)


erip:
I plan to simplify some of the logic here tomorrow.

erip:
OK, this is significantly cleaner now. I think this is ready for review!

torchtext/datasets/iwslt2016.py
```diff
@@ -225,50 +205,62 @@ def IWSLT2016(root='.data', split=('train', 'valid', 'test'), language_pair=('de
...
-        for item in zip(src_data_iter, tgt_data_iter):
-            yield item
-
-    return _RawTextIterableDataset(DATASET_NAME, NUM_LINES[split][num_lines_set_identifier[split]][tuple(sorted(language_pair))], _iter(src_data_iter, tgt_data_iter))
+            return fname.replace('.tags', '')
+        return fname
+
+    cache_decompressed_dp = cache_decompressed_dp.flatmap(FileLister)
+
+    def get_filepath(split, lang):
```
parmeet, line 241: Any specific reason we prefer function over a dict object? might be easy to read if we create object filepath instead

torchtext/datasets/iwslt2016.py
```diff
@@ -225,50 +205,62 @@ def IWSLT2016(root='.data', split=('train', 'valid', 'test'), language_pair=('de
...
-
-    def _iter(src_data_iter, tgt_data_iter):
-        for item in zip(src_data_iter, tgt_data_iter):
-            yield item
-
-    return _RawTextIterableDataset(DATASET_NAME, NUM_LINES[split][num_lines_set_identifier[split]][tuple(sorted(language_pair))], _iter(src_data_iter, tgt_data_iter))
+            return fname.replace('.tags', '')
+        return fname
+
+    cache_decompressed_dp = cache_decompressed_dp.flatmap(FileLister)
```
parmeet, line 239: So I guess this is doing the magic for flattening out all the inner files after extracting from inner tarballs right? I am trying to infer the role of FileLister based on your comment here  https://github.com/pytorch/text/issues/1494#issuecomment-1019716382 . It's not so intuitive that FileLister is doing the extraction. Could you please help understand a bit what this line is doing?

parmeet (commented):
Overall it LGTM! It's just not so easy to ensure the correctness with failing CI unit-tests on datasets due to cache failure (I am looking into it why the cache is failing). Could you please ensure it's working locally as well?

torchtext/datasets/iwslt2016.py
```diff
@@ -1,19 +1,23 @@
...
 import os
-from torchtext.utils import (download_from_url, extract_archive)
 from torchtext.data.datasets_utils import (
-    _RawTextIterableDataset,
     _wrap_split_argument,
     _clean_xml_file,
     _clean_tags_file,
-    _read_text_iterator,
 )
 from torchtext.data.datasets_utils import _create_dataset_directory
```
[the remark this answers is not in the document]
erip: I'm not sure I quite understand -- these imports have been removed, right?

torchtext/datasets/iwslt2016.py
erip: Yes, `flatmap(FileLister)` is not doing the extraction but is instead effectively linearizing the file tree of the directory structures which are the output of the `read_from_tar`.

So basically `read_from_tar()` will yield something like:

```
(/path/to/dir1/, /path/to/dir2/)`
```

And flatmapping the `FileLister` basically walks those trees in order.

Does this make sense?

torchtext/datasets/iwslt2016.py
erip: I don't think there was a particularly good reason for this decision from me. 😄  I can change this -- nice catch.

erip: Hmm, I guess if we remove the function we need to duplicate the code a bit... i.e.,

```
src_file_template = {...}[split][src_lang]
tgt_file_template = {...}[split][tgt_lang]
```

erip:
Hmm, tests are passing but playing in the REPL is breaking. I'll fix this and push...

torchtext/datasets/iwslt2016.py
parmeet: But we are calling `read_from_tar` only once here (https://github.com/pytorch/text/pull/1545/files#diff-6ea166aeffab6173bf513a4028091fcc22977b25d4b54fc98026ce0a0ad44476R227), which shall extract files from `2016-01.tgz`. I am trying to understand where do we extract from inner tarballs, for eg: for language pair (de-en)we would need to extract from inner tarball `de-en.tgz` . Sorry i may be missing something obvious here?

torchtext/datasets/iwslt2016.py
parmeet: hmm, not sure if i am missing something obvious here, but what I was proposing is to replace get_filepath function with the dictionary object it encapsulate (which is a nested dict). The call-site would then just need to index on the dictionary object instead of calling the function with corresponding arguments right?

torchtext/datasets/iwslt2016.py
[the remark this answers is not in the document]
erip: Oh, got it. 😄  yeah, I can do that!

torchtext/datasets/iwslt2016.py
erip: No, I'm the one missing the obvious things. Need much more coffee. Thanks for patience 😅

torchtext/datasets/iwslt2016.py
erip: No, you're not missing anything obvious. I think the code is not doing what I expect and I'm trying to figure it out. The issue seems to be about `filepath_fn` not doing what I think it should do since it's used to extract files, but not directories...

torchtext/datasets/iwslt2016.py
```diff
@@ -225,50 +231,60 @@ def IWSLT2016(root='.data', split=('train', 'valid', 'test'), language_pair=('de
...
-    # IWSLT dataset's url downloads a multilingual tgz.
-    # We need to take an extra step to pick out the specific language pair from it.
+    url_dp = IterableWrapper([URL])
+    cache_compressed_dp = url_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, _PATH),
+        hash_dict={os.path.join(root, _PATH): MD5},
+        hash_type="md5"
+    )
+    cache_compressed_dp = GDriveReader(cache_compressed_dp).end_caching(mode="wb", same_filepath_fn=True)
+    cache_compressed_dp = FileOpener(cache_compressed_dp, mode="b")
```
Nayef211, line 215: Can we move this line after line 249 (after calling `on_disk_cache`) to be in line with our other implemenations (i.e. [EnWik9](https://github.com/pytorch/text/pull/1512/files#diff-6f50d62a0a75985ae909dd7f4ba8d66f9ec7091d8bb374ddacb8937cc5c27a72R44-R50)). Also we should call this `cache_decompressed_dp`.

So this can look like
```
...
cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
        filepath_fn=lambda x: os.path.join(os.path.splitext(x[0])[0], iwslt_tar)
    )
cache_compressed_dp = FileOpener(cache_compressed_dp, mode="b").read_from_tar()
...
```

torchtext/datasets/iwslt2016.py
parmeet: haha, no worries :). I need to do the same, I am feeling quite dizzy looking at snowfall outside my window  😁

torchtext/datasets/iwslt2016.py
```diff
@@ -225,50 +231,60 @@ def IWSLT2016(root='.data', split=('train', 'valid', 'test'), language_pair=('de
...
+    cache_decompressed_dp = cache_decompressed_dp.flatmap(FileLister)
+
+    def get_filepath(f):
+        src_file, tgt_file = {
+            "train": _construct_filepath(f, src_train, tgt_train),
+            "valid": _construct_filepath(f, src_eval, tgt_eval),
+            "test": _construct_filepath(f, src_test, tgt_test)
+        }[split]
+
+        return src_file, tgt_file
```
Nayef211: Can we move this helper functions outside of the dataset function and make them private?

erip:
I'm getting a bit mixed up with the logic so I'm just going to brain dump here...

- We check the cache for the outermost tgz. If we don't have it, we fetch it.

- Next we read from the outermost tgz. In the original code we just extract the whole thing, so we want to wrap a `read_from_tar` call in caching. The `filepath_fn` here is not obvious to me... 

- Within the extraction, there's a tarfile created by `iwslt_tar = os.path.join(root, os.path.splitext(_PATH)[0], "texts", src_language, tgt_language, languages) + ".tgz"` which has contents we care about. In the original code this is extracted unconditionally. The `filepath_fn` is again not obvious here...

- Finally, we iterate through the files in this inner tgz and clean/read as needed. This logic after this is fine, I think...

Nayef211 (commented):
Thanks for taking on this migration @erip. This dataset is definitely more complex than the other's that we've worked on so far. Just left a couple of comments from my side but lmk if you agree with the suggestions 😄

torchtext/datasets/iwslt2016.py
Nayef211: Actually I noticed we're using `clean_files` and `get_filepath` fn in the IWSLT2017 dataset implementation as well. If they share similar logic, can we pull them out to `torchtext/data/datasets_utils.py`?

torchtext/datasets/iwslt2016.py
```diff
@@ -133,20 +135,27 @@ def _construct_filenames(filename, languages):
     return filenames
 
 
+def _construct_filepath(path, src_filename, tgt_filename):
+    src_path = None
+    tgt_path = None
+    src_path = path if src_filename in path else src_path
+    tgt_path = path if tgt_filename in path else tgt_path
```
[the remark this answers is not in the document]
erip: I completely removed this to simplify. Hope the new changes are a bit clearer.

parmeet:
> I'm getting a bit mixed up with the logic so I'm just going to brain dump here...

Thanks @erip for this exercise. This is certainly not as straightforward as other datasets.  My understanding about file_path fn in the context of caching is that it would look for all the files returned by that function and return True/False depending on whether they exist on filesystem or not (here is the code: https://github.com/pytorch/data/blob/160ce809f8c3ee686af0e3e701d7341827841eeb/torchdata/datapipes/iter/util/cacheholder.py#L182)

Based on this, I would like to think about the steps  as follows:

1) Download outer tar (with caching mechanism as usual)
2) Extract outer tar
   2.1) We would do this extraction only if the inner tar (let say en-de.tgz) we care about doesn't exist in the file system. So basically filepath_fn would return this path
3) Filter the files from step 2 such that we only get en-de.tgz. 
   3.1) Apply the same logic for caching
4) Filter the files we care about and apply the clean-up function
  4.1) Again we can check whether cleaned-up file exist on the filesystem or not using caching mechanism, and if not we apply the clean-up function

So in a way, there are 4 levels of caching, 1) Download, 2) Outer extraction, 3) inner extraction, 4) clean-up. I think caching is certainly quite involved after 1 , so i would suggest to break this PR into two. In this PR we only introduce caching for download as usual. Then we introduce the caching mechanism for other parts as follow-up PR. This would also make it bit easier to review. Let me know if this helps?


cc: @ejguan, to make sure I didn't misguided on the caching story :)

torchtext/datasets/iwslt2016.py
```diff
@@ -225,50 +205,62 @@ def IWSLT2016(root='.data', split=('train', 'valid', 'test'), language_pair=('de
...
+                "valid": tgt_eval,
+                "test": tgt_test,
+            }
+        }[lang][split]
+
+    cleaned_cache_decompressed_dp = cache_decompressed_dp.map(clean_files)
+
+    # Filters out irrelevant file given the filename templates filled with split and src/tgt codes
+    src_data_dp = cleaned_cache_decompressed_dp.filter(lambda x: get_filepath(split, src_language) in x)
+    tgt_data_dp = cleaned_cache_decompressed_dp.filter(lambda x: get_filepath(split, tgt_language) in x)
```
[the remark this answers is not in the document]
erip: I suppose `demux` could be used? Indeed, this was my first implementation but there were other factors that made the logic a bit messy.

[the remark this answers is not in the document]
ejguan: I would leave it as it is. Using `demux` is gonna make the code uglier here.

torchtext/datasets/iwslt2016.py
```diff
@@ -225,50 +205,62 @@ def IWSLT2016(root='.data', split=('train', 'valid', 'test'), language_pair=('de
...
-    for fname in extracted_files:
+    iwslt_tar = os.path.join(
+        "texts", src_language, tgt_language, languages
+    )
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
+        # Convert /path/to/downloaded/foo.tgz to /path/to/downloaded/foo/rest/of/path
+        filepath_fn=lambda x: os.path.join(os.path.splitext(x[0])[0], iwslt_tar)
+    )
+    cache_decompressed_dp = cache_decompressed_dp.read_from_tar()
+    cache_decompressed_dp = cache_decompressed_dp.end_caching(mode="wb")
```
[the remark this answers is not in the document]
ejguan: NVM, understood thanks to parmeet's comment

erip:
@parmeet that's helpful. I have things nearly good and am on the cleaning step. The issue currently is that when `clean_files` gets the filename, it's the name of the file within the tar (with its associated stream also available). This requires some ugly path-hacking in the map to emulate the in-place aspect of the update in the original code, but it should be doable.

erip:
OK, this looks better now. 

image

torchtext/datasets/iwslt2016.py
```diff
@@ -225,50 +204,67 @@ def IWSLT2016(root='.data', split=('train', 'valid', 'test'), language_pair=('de
...
+    src_filepath = file_path_by_lang_and_split[src_language][split]
+
+    cache_inner_src_decompressed_dp = cache_decompressed_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, "2016-01/texts/", src_language, tgt_language, languages, src_filepath)
+    )
+    cache_inner_src_decompressed_dp = FileOpener(cache_inner_src_decompressed_dp, mode="b").read_from_tar()
+    cache_inner_src_decompressed_dp = cache_inner_src_decompressed_dp.map(lambda x: _clean_files(x[0], os.path.splitext(os.path.dirname(os.path.dirname(x[0])))[0], x[1]))
+    cache_inner_src_decompressed_dp = cache_inner_src_decompressed_dp.filter(lambda x: src_filepath in x)
+    cache_inner_src_decompressed_dp = FileOpener(cache_inner_src_decompressed_dp, mode="b")
+    cache_inner_src_decompressed_dp = cache_inner_src_decompressed_dp.end_caching(mode="wb", same_filepath_fn=True)
```
parmeet, line 250: OK , I think we might need few changes here to optimize the whole thing. 
1) Cache file looks good (we are looking at the end state file after clean-up). But let's use variable  so that we can reuse it later in filtering part. 
2) We should reverse the order of filter and map so that we only clean the file we are interested in. 
3) The _clean_files function should return the stream (instead of saving it to file) so that that can be used to save the actual clean file through end_caching datapipe.

Also this mechanism is same for both src and trg, let's create a inner function to reuse across both the datapipes.

erip: One challenge with #2 is that the filenames get changed in the `_clean_inner*` functions, so we don't "know" them a priori. I haven't given much thought yet to whether we *could* know them...

torchtext/data/datasets_utils.py
```diff
@@ -33,6 +33,18 @@ def _clean_xml_file(f_xml):
                 fd_txt.write(e.text.strip() + '\n')
 
 
+def _clean_inner_xml_file(f_xml, base, stream):
```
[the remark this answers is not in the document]
erip: I have a TODO to rework these functions to return the stream instead, but the docstring has been pushed for the as-is functions.

torchtext/datasets/iwslt2016.py
erip: OK, I have applied similar logic to the "expected files" to create a mapping for "uncleaned files" and this seems like a nice solution. I am still returning the output file (i.e., the `filepath_fn` file) from `_clean_files` and it seems OK, but there doesn't seem to be any caching (subsequent calls to `next(iter(train))` take the same time).

erip: Scratch that -- I fixed it. 🚀  First read is ~2.6s and subsequent reads are ~600ms. It still seems like a lot of time, but maybe there are other things I've missed... This is starting to look a lot better now.

erip:
One lingering warning that I'm seeing is about child datapipes being reset. I suspect this is causing some latency issues but I'm not exactly sure why I'm seeing it. Maybe because I consume the datapipe when cleaning the files?

erip:
OK, I'm happier with this. 😸

torchtext/data/datasets_utils.py
```diff
@@ -33,6 +34,25 @@ def _clean_xml_file(f_xml):
                 fd_txt.write(e.text.strip() + '\n')
 
 
+def _clean_inner_xml_file(outfile, stream):
+    """Accepts an output filename and a stream of the byte contents of an XML file
+    within a tarball and writes the cleaned contents to a new, untarred file.
```
parmeet, line 39: I don't think it's necessary to mention `within a tarball and writes the cleaned contents to a new, untarred file.`. It could come from anywhere :)

parmeet:
> OK, I'm happier with this. 😸

woww, this looks great. Thanks so much @erip for reading well from my half-baked comments and working so hard on this PR :). This certainly is the most challenging so far. I think we should be good to go with what we have now!

parmeet (approved):
LGTM! Let's merge it once we address the minor comment here https://github.com/pytorch/text/pull/1545#discussion_r795105047

erip:
Done, @parmeet! :shipit:

PR #1549, opened 2022-01-28
[BUG-FIX] fix yelp dataset

parmeet:


parmeet:
Will do another PR, this also changed the format.

PR #1552, opened 2022-01-28
[BUG-FIX] fix penn dataset

parmeet:
This PR fix the bug in writing downloaded file to disk. It need be dumped as binary ('wb') and not as text/string ('w').


abhinavarora (approved):
LGTM!

PR #1554, opened 2022-01-30
Multi30k mocked testing

parmeet:
Reference issue #1493

### DIFF
--- a/test/datasets/test_multi30k.py
+++ b/test/datasets/test_multi30k.py
@@ -38,7 +38,7 @@ def _get_mock_dataset(root_dir):
     archive["train"] = os.path.join(base_dir, "training.tar.gz")
     archive["val"] = os.path.join(base_dir, "validation.tar.gz")
     archive["test"] = os.path.join(base_dir, "mmt16_task1_test.tar.gz")
-    # create tar file from dataset folder
+
     for split in ("train", "val", "test"):
         with tarfile.open(archive[split], "w:gz") as tar:
             tar.add(os.path.join(temp_dataset_dir, f"{split}.de"))
--- a/test/datasets/test_multi30k.py
+++ b/test/datasets/test_multi30k.py
@@ -1,0 +1,82 @@
+import os
+import random
+import string
+import tarfile
+from collections import defaultdict
+from unittest.mock import patch
+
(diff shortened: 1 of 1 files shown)


test/datasets/test_multi30k.py
```diff
@@ -0,0 +1,82 @@
...
+
+    seed = 1
+    mocked_data = defaultdict(list)
+    for file_name in ("train.de", "train.en", "val.de", "val.en", "test.de", "test.en"):
+        txt_file = os.path.join(temp_dataset_dir, file_name)
+        with open(txt_file, "w") as f:
+            for i in range(5):
+                rand_string = " ".join(
+                    random.choice(string.ascii_letters) for i in range(seed)
+                )
```
erip, line 31: One thought: since all of our datasets are utf-8 files, does it make sense to write unicode strings to make sure we don't have lingering bugs from default encodings when opening files? Maybe this is overkill, but it's been a big source of bugs when I did mostly windows development.

test/datasets/test_multi30k.py
```diff
@@ -0,0 +1,82 @@
...
+                content = f'{rand_string}\n'
```
Nayef211, line 32: nit: can we use double quotes here instead of single quotes

Nayef211 (approved):
Overall LGTM. Left some nit comments. Also think @erip has a valid point about writing unicode strings when we mock data. Should we do this for all our tests moving forward?

test/datasets/test_multi30k.py
parmeet: Ya, agreed. I think it is a good suggestion!

parmeet: I will keep it as a follow-up item as generating random UTF-8 string is not trivial.

parmeet:
> Overall LGTM. Left some nit comments. Also think @erip has a valid point about writing unicode strings when we mock data. Should we do this for all our tests moving forward?

Yes, let's add a follow-up item here or even better if we can also track this in the issue #1493 in general.

Nayef211:
> Yes, let's add a follow-up item here or even better if we can also track this in the issue #1493 in general.

Added this as a follow up item to the GH issue!

PR #1558, opened 2022-01-30
fix flake

erip:
somehow missed in most recent PR...


parmeet (approved):
LGTM! Thanks for fixing it.

PR #1555, opened 2022-01-30
Revert BC breaking change in PennTreebank dataset

parmeet:
Revert the BC breaking change in the output introduced in https://github.com/pytorch/text/pull/1511

PR #1559, opened 2022-01-31
revert the BC breaking change in enwik9 data

parmeet:


torchtext/datasets/enwik9.py
```diff
@@ -50,4 +50,4 @@ def EnWik9(root: str, split: Union[Tuple[str], str]):
     )
 
     data_dp = FileOpener(cache_decompressed_dp, mode="b")
-    return data_dp.readlines(decode=True, return_path=False)
+    return data_dp.readlines(decode=True, return_path=False, strip_newline=False)
```
abhinavarora, line 53: Usually, it is good to strip the new lines at the end of each line. Can we change this going forward?

abhinavarora: I know it breaks BC, but does it matter as we are revamping the library?

parmeet: > Usually, it is good to strip the new lines at the end of each line. Can we change this going forward?

It's a good point, I am not so sure why it is like that when they were first added. In-fact now when I think about it, it's not even consistent across our datasets. 

> I know it breaks BC, but does it matter as we are revamping the library?

Well, we wanted the migration to be not BC breaking. But if we do end up doing so, let's make sure to note all the BC breaking changes in release notes.

Nayef211: > Well, we wanted the migration to be not BC breaking. But if we do end up doing so, let's make sure to note all the BC breaking changes in release notes.
> 
> Edit: Also would be great not to mix the BC breaking changes with the migration.

I also updated the SST2 dataset to return a `int` for the labels even though it was returning strings before because that seemed like the expected behavior. Does it matter whether we do it in the same PR as the migration vs a different PR if we note all the BC breaking changes in the PR summary?

parmeet: Does it matter whether we do it in the same PR as the migration vs a different PR if we note all the BC breaking changes in the PR summary?

In general, it's a good practice to logically separate PRs that makes it easy for tracking purpose as well as to review :)

parmeet:
Before closing this and #1555, just wanted to check if we are all OK with removal of newlines from datasets? cc: @Nayef211 , @abhinavarora , @erip?

Follow-up if yes:
* Fix breaking tests for enwik9 and PennTreeBank by updating the hash code in  [raw_datasets.jsonl](https://github.com/pytorch/text/blob/main/test/asset/raw_datasets.jsonl) 
* Look for other datasets that return newlines and remove accordingly

Nayef211:
> Before closing this and #1555, just wanted to check if we are all OK with removal of newlines from datasets? cc: @Nayef211 , @abhinavarora , @erip?
> 
> Follow-up if yes:
> 
> * Fix breaking tests for enwik9 and PennTreeBank by updating the hash code in  [raw_datasets.jsonl](https://github.com/pytorch/text/blob/main/test/asset/raw_datasets.jsonl)
> * Look for other datasets that return newlines and remove accordingly

I think this makes sense to me. In addition to this, should we identify which datasets should return labels as `int` instead of `string` and update those as well?

erip:
These all seem reasonable to me and easy for users to "revert" independently in userland code if they need to.

parmeet:
Closing this PR as this is a reasonable BC breaking change.

parmeet:
In addition to this, should we identify which datasets should return labels as `int` instead of `string` and update those as well?

I think this is a reasonable change as well.

PR #1561, opened 2022-01-31
Add AmazonReviewFull Mocked Unit Test

VirgileHlav:
Reference Issue: #1493

## Summary
- Added mocked unit test for AmazonReviewFull dataset 

## Test
- ` pytest test/datasets/test_amazonreviewfull.py`

### DIFF
--- a/test/datasets/test_amazonreviewfull.py
+++ b/test/datasets/test_amazonreviewfull.py
@@ -26,7 +26,7 @@ def _get_mock_dataset(root_dir):
         txt_file = os.path.join(temp_dataset_dir, file_name)
         with open(txt_file, "w") as f:
             for i in range(5):
-                label = seed % 2 + 1
+                label = seed % 5 + 1
                 rand_string = " ".join(
                     random.choice(string.ascii_letters) for i in range(seed)
                 )
--- a/test/datasets/test_amazonreviewfull.py
+++ b/test/datasets/test_amazonreviewfull.py
@@ -1,0 +1,83 @@
+import os
+import random
+import string
+import tarfile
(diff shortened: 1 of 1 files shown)


erip:
LGTM! It may be the case that eventually we'll want to abstract some of these repeated concepts (getting the dataset, populating the mock, etc.) into some common classes, but that's maybe a good TODO for later.

Nayef211:
> LGTM! It may be the case that eventually we'll want to abstract some of these repeated concepts (getting the dataset, populating the mock, etc.) into some common classes, but that's maybe a good TODO for later.

LGTM as well! And totally agree with @erip. I will look into creating a base class that can handle some of the repeated logic across tests. For the time being, I think it's fine to get all of the tests implemented before the branch cut next week!

Nayef211 (approved):

test/datasets/test_amazonreviewfull.py
```diff
@@ -0,0 +1,83 @@
...
+    temp_dataset_dir = os.path.join(base_dir, "temp_dataset_dir")
+    os.makedirs(temp_dataset_dir, exist_ok=True)
+
+    seed = 1
+    mocked_data = defaultdict(list)
+    for file_name in ("train.csv", "test.csv"):
+        txt_file = os.path.join(temp_dataset_dir, file_name)
+        with open(txt_file, "w") as f:
+            for i in range(5):
+                label = seed % 2 + 1
```
Nayef211, line 29: If you look at the readme for the AmazonReviewFull dataset, it mentions that the label for the datasets range from 1 to 5. So this line should be `label = seed % 5 + 1`

PR #1563, opened 2022-02-01
mock up IWSLT2016 test for faster testing.

erip:
Reference #1493

### DIFF
--- a/test/datasets/test_iwslt2016.py
+++ b/test/datasets/test_iwslt2016.py
@@ -1,0 +1,83 @@
+import os
+import random
+import string
+from collections import defaultdict
+from unittest.mock import patch
+
+from parameterized import parameterized
+from torchtext.datasets.iwslt2016 import IWSLT2016
+from torchtext.data.datasets_utils import _generate_iwslt_files_for_lang_and_split
+
+from ..common.case_utils import TempDirMixin, zip_equal
+from ..common.torchtext_test_case import TorchtextTestCase
+
+
+def _get_mock_dataset(root_dir, split, src, tgt):
+    """
+    root_dir: directory to the mocked dataset
+    """
+    temp_dataset_dir = os.path.join(root_dir, f"IWSLT2016/2016-01/texts/{src}/{tgt}/{src}-{tgt}/")
+    os.makedirs(temp_dataset_dir, exist_ok=True)
+
+    seed = 1
+    mocked_data = defaultdict(lambda: defaultdict(list))
+    valid_set = "tst2013"
+    test_set = "tst2014"
+
+    files_for_split, _ = _generate_iwslt_files_for_lang_and_split(16, src, tgt, valid_set, test_set)
+    src_file = files_for_split[src][split]
+    tgt_file = files_for_split[tgt][split]
+    for file_name in (src_file, tgt_file):
+        txt_file = os.path.join(temp_dataset_dir, file_name)
+        with open(txt_file, "w") as f:
+            # Get file extension (i.e., the language) without the . prefix (.en -> en)
+            lang = os.path.splitext(file_name)[1][1:]
+            for i in range(5):
+                rand_line = " ".join(
+                    random.choice(string.ascii_letters) for i in range(seed)
+                )
+                dataset_line = f"{rand_line} {rand_line}\n"
+                # append line to correct dataset split
+                mocked_data[split][lang].append(dataset_line)
+                f.write(f'{rand_line} {rand_line}\n')
+                seed += 1
+
+    return list(zip(mocked_data[split][src], mocked_data[split][tgt]))
+
+
+class TestIWSLT2016(TempDirMixin, TorchtextTestCase):
+    root_dir = None
+    patcher = None
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.root_dir = cls.get_base_temp_dir()
+        cls.patcher = patch(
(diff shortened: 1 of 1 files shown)


test/datasets/test_iwslt2016.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @parameterized.expand([("train", "de", "en"), ("valid", "de", "en")])
+    def test_iwslt2016(self, split, src, tgt):
+        expected_samples = _get_mock_dataset(self.root_dir, split, src, tgt)
+
+        dataset = IWSLT2016(root=self.root_dir, split=split)
+
+        samples = list(dataset)
+
+        for sample, expected_sample in zip_equal(samples, expected_samples):
```
Nayef211, line 74: NIT: can we organize `sampes` and `expected_samples` similar to what we do in the [SST2 PR](https://github.com/pytorch/text/pull/1542/files#diff-59e6a3a1cb3a2dd765c0ea1110134a23c94307c3787674d5f00ee6027423b36eR81-R83) for consistency

test/datasets/test_iwslt2016.py
```diff
@@ -0,0 +1,83 @@
...
+    files_for_split, _ = _generate_iwslt_files_for_lang_and_split(16, src, tgt, valid_set, test_set)
+    src_file = files_for_split[src][split]
+    tgt_file = files_for_split[tgt][split]
+    for file_name in (src_file, tgt_file):
+        txt_file = os.path.join(temp_dataset_dir, file_name)
+        with open(txt_file, "w") as f:
+            # Get file extension (i.e., the language) without the . prefix (.en -> en)
+            lang = os.path.splitext(file_name)[1][1:]
+            for i in range(5):
+                rand_line = " ".join(
```
Nayef211, line 36: NIT: Can we rename this to `rand_string` for consistency with other tests?

test/datasets/test_iwslt2016.py
```diff
@@ -0,0 +1,83 @@
...
+        cls.patcher.start()
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.patcher.stop()
+        super().tearDownClass()
+
+    @parameterized.expand([("train", "de", "en"), ("valid", "de", "en")])
+    def test_iwslt2016(self, split, src, tgt):
+        expected_samples = _get_mock_dataset(self.root_dir, split, src, tgt)
```
Nayef211, line 68: Is there any specific reason why we don't want to generate all the mocked data within the `setUpClass` method and store it in `self.samples` like we do in the [other tests](https://github.com/pytorch/text/pull/1542/files#diff-59e6a3a1cb3a2dd765c0ea1110134a23c94307c3787674d5f00ee6027423b36eR66)?

test/datasets/test_iwslt2016.py
Nayef211: > Unfortunately it's not quite as straightforward unless we want to hardcode the language pairs in `setUpClass`. Otherwise there's no good way to parameterize them. This is required because the expected file name for caching is a function of the (src_lang, tgt_lang, split) which is somewhat unique.

Gotcha, I think it's okay to keep your current implementation. I didn't catch the fact that the ordering mattered here.

Nayef211 (approved):
Overall LGTM. I think we can merge this once we fix the nit comment I left about variable renaming!

erip:
Done, @Nayef211!

test/datasets/test_iwslt2016.py
```diff
@@ -0,0 +1,83 @@
...
+            "torchdata.datapipes.iter.util.cacheholder.OnDiskCacheHolderIterDataPipe._cache_check_fn", return_value=True
+        )
+        cls.patcher.start()
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.patcher.stop()
+        super().tearDownClass()
+
+    @parameterized.expand([("train", "de", "en"), ("valid", "de", "en")])
```
parmeet, line 66: IWSLT2016 also consist of `test` split, so ideally we should also include it in testing.

erip: Oops, yes. I had made a change and forgot to re-incorporate the test split.

test/datasets/test_iwslt2016.py
```diff
@@ -0,0 +1,83 @@
...
+                rand_string = " ".join(
+                    random.choice(string.ascii_letters) for i in range(seed)
+                )
+                dataset_line = f"{rand_string} {rand_string}\n"
+                # append line to correct dataset split
+                mocked_data[split][lang].append(dataset_line)
+                f.write(f'{rand_string} {rand_string}\n')
+                seed += 1
+
+    return list(zip(mocked_data[split][src], mocked_data[split][tgt]))
```
parmeet, line 45: Is there a reason we are creating a download archive `2016-01.tgz` like we are doing for other datasets?

parmeet: > Is there a reason we are not creating a download archive `2016-01.tgz` like we are doing for other datasets?
> 
> Edit: I think it quite important to start from the download archive, otherwise we can get into hard to find bugs specially when the compression pattern is complex like we have in here.

@erip just wanted to check if you do plan to follow-up on this as well?

erip: Yes, I can follow up on this. It will take a lot more thought since, as you mention, the clean up is quite involved. That said, I think it should be doable.

parmeet: > Yes, I can follow up on this. It will take a lot more thought since, as you mention, the clean up is quite involved. That said, I think it should be doable.

Sure, thanks @erip!

PR #1564, opened 2022-02-02
remove private functions whose responsibilities have been subsumed by datapipes. TODO: migrate WMT14 to remove remaining private dataset utils.

erip:
These functions are all private (and documented as such), so removing them should be fine. Once WMT14 is migrated we can also remove `_read_text_iterator` and `_download_extract_validate`. 

We will also likely want to formally deprecate `_RawTextIterableDataset` which is private, but I have the impression that people will be using it.

### DIFF
--- a/torchtext/data/datasets_utils.py
+++ b/torchtext/data/datasets_utils.py
@@ -25,15 +25,6 @@ for public consumption yet.
 """
 
 
-def _clean_xml_file(f_xml):
-    f_txt = os.path.splitext(f_xml)[0]
-    with codecs.open(f_txt, mode='w', encoding='utf-8') as fd_txt:
-        root = ET.parse(f_xml).getroot()[0]
-        for doc in root.findall('doc'):
-            for e in doc.findall('seg'):
-                fd_txt.write(e.text.strip() + '\n')
-
-
 def _clean_inner_xml_file(outfile, stream):
     """Accepts an output filename and a stream of the byte contents of an XML file
     and writes the cleaned contents to a new file on disk.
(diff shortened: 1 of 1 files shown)


Nayef211 (approved):
Thanks for helping clean up our library @erip! @parmeet I noticed WMT14 wasn't included in #1494. Do we plan on migrating the dataset out of experimental before our branch cut?

Nayef211:
> I don't see why not 😄 if we're already moving CC100 and SST2, we might as well cut over WMT14 as well. To be honest, I didn't even realize it hadn't been migrated yet. I can take that migration.
> 
> If it makes sense, I can do that first then we could rebase this and remove all private functions in one shot? I don't mind two PRs otherwise. Whatever's easiest.

Awesome that would be great @erip! I was actually going to suggest that we finish all the dataset migrations first before merging in this PR since we know other private functions that will need to be removed once all the migrations are complete. My only worry was that the CC-100 migration might take some more time since we still need to discuss how best to implement the dataset.

parmeet:
@erip , @Nayef211 I didn't included WMT14 because it was under experimental folder (meaning not yer part of our release). Sure, we can graduate it and make it part of the release. But as of now, I think the URL is not working and I disable the testing of WMT14 in this PR https://github.com/pytorch/text/pull/1404.

parmeet:
> We will also likely want to formally deprecate `_RawTextIterableDataset` which is private, but I have the impression that people will be using it.

hmm, this is a private class. Since the datasets would be migrated and we won't have any internal dependency, i wonder if there is any issue in removing this as well?

erip:
Some people are [apparently using it](https://github.com/search?q=%22_RawTextIterableDataset%22&type=code). I suppose if they really need it they can pin to a specific version.

parmeet:
> Some people are [apparently using it](https://github.com/search?q=%22_RawTextIterableDataset%22&type=code). I suppose if they really need it they can pin to a specific version.

hmm interesting. Yupp I guess that's always an option to depend on older version or copy it from there. I think from library perspective, it doesn't make sense to maintain a private function that no longer is used.

erip:
Related to the WMT14 migration -- I have some local changes, but it looks like the data in the GDrive has been removed because I can't actually download the dataset. 😬  Not sure how to proceed here...

parmeet:
> Related to the WMT14 migration -- I have some local changes, but it looks like the data in the GDrive has been removed because I can't actually download the dataset. 😬 Not sure how to proceed here...

Ya, I have noticed this earlier as well. Can we check on HF if they also use the same link? Or perhaps try to figure who hosted the dataset and check with them directly?

Nayef211:
@parmeet @erip do we want to land this PR before our branch cut today?

erip:
I think it's OK to merge and I can circle back to clean up the rest later.

Nayef211:
@erip can you actually merge in the latest changes from the master branch? There's a bunch of test failures since this branch is a little outdated

erip:
@Nayef211 rebased

PR #1566, opened 2022-02-03
Add DBpedia Mocked Unit Test

VirgileHlav:
Reference Issue: #1493

## Summary
- Added mocked unit test for DBpedia dataset 

## Test
- ` pytest test/datasets/test_dbpedia.py`


test/datasets/test_dbpedia.py
```diff
@@ -0,0 +1,85 @@
...
+    os.makedirs(temp_dataset_dir, exist_ok=True)
+
+    seed = 1
+    mocked_data = defaultdict(list)
+    for file_name in ["train.csv", "test.csv"]:
+        csv_file = os.path.join(temp_dataset_dir, file_name)
+        mocked_lines = mocked_data[os.path.splitext(file_name)[0]]
+        with open(csv_file, "w") as f:
+            for i in range(5):
+                label = seed % 11
```
Nayef211, line 30: DBpedia actually has labels ranging from 1-14. This should be `label = seed % 14 + 1`

test/datasets/test_dbpedia.py
```diff
@@ -0,0 +1,85 @@
...
+                rand_string = " ".join(
+                    random.choice(string.ascii_letters) for i in range(seed)
+                )
+                dataset_line = (label, f"{rand_string}")
+                f.write(f'{label},"{rand_string}"\n')
```
Nayef211, line 35: DBpedia has 3 items: the label, the title, and the contents. This should be changed to `f.write(f'{label},"{rand_string}","{rand_string}"\n')`. You can learn more about what's contained in each line of the dataset here https://huggingface.co/datasets/dbpedia_14#data-instances

test/datasets/test_dbpedia.py
```diff
@@ -0,0 +1,85 @@
...
+from ..common.case_utils import TempDirMixin, zip_equal
+from ..common.torchtext_test_case import TorchtextTestCase
+
+
+def _get_mock_dataset(root_dir):
+    """
+    root_dir: directory to the mocked dataset
+    """
+    base_dir = os.path.join(root_dir, "DBpedia")
+    temp_dataset_dir = os.path.join(base_dir, "dbpedia_csv")
```
Nayef211, line 20: Can you name this `temp_dataset_dir`. If the temp folder name is the same as the folder name within the tar, the tar extraction will be skipped. This is because the dataset implementation has some caching logic that checks whether the extracted files exists before going ahead with the tar archive extraction (this was actually missing for the DBpedia dataset but is being added here https://github.com/pytorch/text/pull/1571)

test/datasets/test_dbpedia.py
```diff
@@ -0,0 +1,85 @@
...
+                dataset_line = (label, f"{rand_string}")
+                f.write(f'{label},"{rand_string}"\n')
+
+                # append line to correct dataset split
+                mocked_lines.append(dataset_line)
+                seed += 1
+
+    compressed_dataset_path = os.path.join(base_dir, "dbpedia_csv.tar.gz")
+    # create gz file from dataset folder
+    with tarfile.open(compressed_dataset_path, "w:gz") as tar:
```
Nayef211, line 43: We can actually do something similar to what we're doing for the [AmazonReviewPolarity test ](https://github.com/Nayef211/text/blob/main/test/datasets/test_amazonreviewpolarity.py#L42-L44). This way we don't need the additional for-loop

test/datasets/test_dbpedia.py
VirgileHlav: Thanks for catching this

test/datasets/test_dbpedia.py
VirgileHlav: That's better, thanks

Nayef211 (approved):
LGTM!

PR #1569, opened 2022-02-03
Add UDPOS Mocked Unit Test

VirgileHlav:
Reference Issue: #1493

## Summary
- Added mocked unit test for UDPOS dataset 

## Test
- ` pytest test/datasets/test_udpos.py`


test/datasets/test_udpos.py
```diff
@@ -0,0 +1,87 @@
...
+        with open(txt_file, "w") as f:
+            for i in range(5):
+                label = seed % 2
+                rand_strings = [random.choice(string.ascii_letters) for i in range(seed)]
+                rand_label_1 = [random.choice(string.ascii_letters) for i in range(seed)]
+                rand_label_2 = [random.choice(string.ascii_letters) for i in range(seed)]
+                # one token per line (each sample ends with an extra \n)
+                for rand_string, label_1, label_2 in zip(rand_strings, rand_label_1, rand_label_2):
+                    f.write(f"{rand_string}\t{label_1}\t{label_2}\n")
+                f.write("\n")
```
erip, line 37: Maybe worth adding more examples than a single one to make sure the IOB reading for >1 example is tested too.

VirgileHlav: Having a look at the [udpos data](https://bitbucket.org/sivareddyg/public/downloads/en-ud-v2.zip), it is tokenized per line, so in my case each token has one character and is on one line. I assume this is what you meant by example, so I changed to randomly sample 1 to 10 characters instead for each token.

erip: By examples I mean "sentences" which are newline delimited:

```
a this this
b is is
c example example
d one one

e this this
f is is
g example example
h two two
```

erip: Ah, it looks OK. I somehow missed that that these were lists.

Nayef211 (approved):
LGTM. When you merge in this PR, can you also make sure to update https://github.com/pytorch/text/issues/1493 to check off that this dataset test is completed. Same goes for your other PRs 😄

PR #1582, opened 2022-02-04
[MINOR] Renamed AG_NEWS variable name to base_dir

Nayef211:
- Renamed ag_news dataset variable name from `temp_dataset_dir` to `base_dir` to be consistent with other tests


parmeet (approved):
LGTM!

PR #1583, opened 2022-02-04
add CC100 mocking test

parmeet:
Reference issue: #1493

### DIFF
--- a/test/datasets/test_cc100.py
+++ b/test/datasets/test_cc100.py
@@ -1,0 +1,66 @@
+import os
+import random
+import string
+import lzma
+from parameterized import parameterized
+from collections import defaultdict
+from unittest.mock import patch
+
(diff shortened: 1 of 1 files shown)


erip:
LGTM!

Nayef211 (approved):
Overall LGTM! I wonder if we should add the 2nd test for the split argument since this dataset also uses the `@_wrap_split_argument` decorator which means that it should be able to support both `"train"` and `("train",)` for the `split` argument?

parmeet:
> Overall LGTM! I wonder if we should add the 2nd test for the split argument since this dataset also uses the `@_wrap_split_argument` decorator which means that it should be able to support both `"train"` and `("train",)` for the `split` argument?

I think the split argument is bit superficial since the data does not really have splits. We might want to just remove this argument from the dataset.

PR #1585, opened 2022-02-05
Fix IWSLT2016 testing

erip:
This serves as a patch to the newly-added IWSLT2016 mock testing which addresses two issues:

1. Starting from the downloaded archive to test the extraction and cleaning pipeline more fully
2. Adds missed test split testing path

cc @parmeet

### DIFF
--- a/test/datasets/test_iwslt2016.py
+++ b/test/datasets/test_iwslt2016.py
@@ -14,6 +14,7 @@ from ..common.torchtext_test_case import TorchtextTestCase
 
 SUPPORTED_LANGPAIRS = [(k, e) for k, v in SUPPORTED_DATASETS["language_pair"].items() for e in v]
 
+
 def _generate_uncleaned_train():
     """Generate tags files"""
     file_contents = []
--- a/test/datasets/test_iwslt2016.py
+++ b/test/datasets/test_iwslt2016.py
@@ -17,6 +17,7 @@ SUPPORTED_LANGPAIRS = [(k, e) for k, v in SUPPORTED_DATASETS["language_pair"].it
 SUPPORTED_DEVTEST_SPLITS = SUPPORTED_DATASETS["valid_test"]
 DEV_TEST_SPLITS = [(dev, test) for dev, test in itertools.product(SUPPORTED_DEVTEST_SPLITS, repeat=2) if dev != test]
 
+
 def _generate_uncleaned_train():
     """Generate tags files"""
     file_contents = []
@@ -138,6 +139,7 @@ class TestIWSLT2016(TempDirMixin, TorchtextTestCase):
         if (dev_set not in SET_NOT_EXISTS[(src, tgt)] and test_set not in SET_NOT_EXISTS[(src, tgt)])
     ])
     def test_iwslt2016(self, split, src, tgt, dev_set, test_set):
+
         expected_samples = _get_mock_dataset(self.root_dir, split, src, tgt, dev_set, test_set)
 
         dataset = IWSLT2016(
--- a/test/datasets/test_iwslt2016.py
+++ b/test/datasets/test_iwslt2016.py
@@ -22,10 +22,11 @@ def _generate_uncleaned_train():
         '<translator', '<title', '<speaker', '<doc', '</doc'
     ]
     for i in range(100):
-        # Write one of the XML tags randomly to make sure we clean appropriately
         rand_string = " ".join(
             random.choice(string.ascii_letters) for i in range(10)
         )
+        # With a 10% change, add one of the XML tags which is cleaned
+        # to ensure cleaning happens appropriately
         if random.random() < 0.1:
             open_tag = random.choice(xml_tags) + ">"
             close_tag = "</" + open_tag[1:] + ">"
@@ -43,7 +44,6 @@ def _generate_uncleaned_valid():
     for doc_id in range(5):
         file_contents.append(f'<doc docid="{doc_id}" genre="lectures">')
         for seg_id in range(100):
-            # Write one of the XML tags randomly to make sure we clean appropriately
             rand_string = " ".join(
                 random.choice(string.ascii_letters) for i in range(10)
             )
--- a/test/datasets/test_iwslt2016.py
+++ b/test/datasets/test_iwslt2016.py
@@ -137,7 +137,7 @@ class TestIWSLT2016(TempDirMixin, TorchtextTestCase):
     def test_iwslt2016(self, split, src, tgt):
         expected_samples = _get_mock_dataset(self.root_dir, split, src, tgt)
 
-        dataset = IWSLT2016(root=self.root_dir, split=split)
+        dataset = IWSLT2016(root=self.root_dir, split=split, language_pair=(src, tgt))
 
         samples = list(dataset)
 
--- a/test/datasets/test_iwslt2016.py
+++ b/test/datasets/test_iwslt2016.py
@@ -1,6 +1,7 @@
 import os
 import random
 import string
+import tarfile
 from collections import defaultdict
 from unittest.mock import patch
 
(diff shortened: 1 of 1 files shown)


test/datasets/test_iwslt2016.py
```diff
@@ -12,35 +13,98 @@
...
+def _generate_uncleaned_train():
+    """Generate tags files"""
+    file_contents = []
+    examples = []
+    xml_tags = [
+        '<url', '<keywords', '<talkid', '<description', '<reviewer',
+        '<translator', '<title', '<speaker', '<doc', '</doc'
+    ]
+    for i in range(100):
+        # Write one of the XML tags randomly to make sure we clean appropriately
```
erip, line 25: This needs to be moved inside the `if` statement...

test/datasets/test_iwslt2016.py
```diff
@@ -12,35 +13,98 @@
...
+
+
+def _generate_uncleaned_valid():
+    file_contents = ["<root>"]
+    examples = []
+
+    for doc_id in range(5):
+        file_contents.append(f'<doc docid="{doc_id}" genre="lectures">')
+        for seg_id in range(100):
+            # Write one of the XML tags randomly to make sure we clean appropriately
```
erip, line 46: Same as above

test/datasets/test_iwslt2016.py
```diff
@@ -63,7 +127,11 @@ def tearDownClass(cls):
         cls.patcher.stop()
         super().tearDownClass()
 
-    @parameterized.expand([("train", "de", "en"), ("valid", "de", "en")])
+    @parameterized.expand([
```
erip, line 130: TODO: can we (should we?) parameterize by each supported lang pair?

test/datasets/test_iwslt2016.py
```diff
@@ -12,35 +13,98 @@
...
+    }[split]
+
+
 def _get_mock_dataset(root_dir, split, src, tgt):
     """
     root_dir: directory to the mocked dataset
     """
-    temp_dataset_dir = os.path.join(root_dir, f"IWSLT2016/2016-01/texts/{src}/{tgt}/{src}-{tgt}/")
-    os.makedirs(temp_dataset_dir, exist_ok=True)
+    outer_temp_dataset_dir = os.path.join(root_dir, f"IWSLT2016/2016-01/texts/{src}/{tgt}/")
```
Nayef211, line 73: I think we should make the first part of the path component `temp_dataset_dir` for the same reason as the comment here https://github.com/pytorch/text/pull/1566#discussion_r798932093

test/datasets/test_iwslt2016.py
Nayef211: For complete code coverage I think it would make sense to parameterize by each supported lang pair. I believe we could use the [`nested_params` fn ](https://github.com/pytorch/text/blob/main/test/common/parameterized_utils.py#L26)to generate the cartesian product of language pairs. The main issue I forsee with using that function is it would generate lang pairs that map to the same language (i.e `en ->en`) which I'm not sure would make sense for this dataset.

Nayef211 (commented):
I think overall this PR LGTM. @parmeet can I also get your review since you reviewed the initial IWSLT PR https://github.com/pytorch/text/pull/1563

test/datasets/test_iwslt2016.py
Nayef211: @parmeet also wanted to get your thoughts on this

parmeet: Yes, I would agree. Simulating the exact structure as the original dataset with all the possible use-cases would be the key for coverage.

erip:
It looks like my attempt at addressing the review feedback has broken the tests and unfortunately I can't test locally because the required nightly isn't available on OS X...  I'll try to address this ASAP.

Nayef211:
> It looks like my attempt at addressing the review feedback has broken the tests and unfortunately I can't test locally because the required nightly isn't available on OS X... I'll try to address this ASAP.

Thanks for all your help with the testing efforts @erip! I just want to see if you would have the bandwidth to add the `IWSLT2017` test as well once this PR is merged? It should be a simple matter of parameterizing this test to make it work with both `IWSLT2016` and `IWSLT2017`. 

The `torchtext` branch-cut is coming up this Thursday and we want to make sure all our tests are complete by this Wednesay!

erip:
Yes, absolutely!

test/datasets/test_iwslt2016.py
erip: Wait -- I don't think we _can_ make `temp_dataset_dir` the first path because we specifically have to pull these [inner files](https://github.com/pytorch/text/blob/main/torchtext/datasets/iwslt2016.py#L228) by name. This is slightly different than every other datasets, so some things might not work the same way in the mocked datasets...

erip:
> to generate the cartesian product of language pairs

IWSLT16 doesn't support [every langpair direction](https://github.com/pytorch/text/blob/main/torchtext/datasets/iwslt2016.py#L153-L165), so we would need to filter some pairs out. At that point, it might just make more sense to enumerate the pairs directly (with some logic for swapping the direction as needed). Thoughts?

parmeet:
> IWSLT16 doesn't support [every langpair direction](https://github.com/pytorch/text/blob/main/torchtext/datasets/iwslt2016.py#L153-L165), so we would need to filter some pairs out. At that point, it might just make more sense to enumerate the pairs directly (with some logic for swapping the direction as needed). Thoughts?

yupp, I agree. The total number of supported pairs are reasonable enough to enumerate explicitly.

erip:
@parmeet the middle ground seems to be leveraging the `SUPPORTED_DATASETS` attribute which gives the adjacency matrix for langpairs. I've incorporated that into testing but as I mention, the turnaround is a bit slow since I can't test locally ☹️

Nayef211:
> I can't test locally because the required nightly isn't available on OS X... I'll try to address this ASAP.

@erip  just wondering if you're talking about the torchtext nightly? Because we can build any of the pytorch/torchtext packages directly from source right? Also I'm happy to patch this PR to test your changes in a linux environment and give you feedback on any failures if you think that would be helpful!

erip:
No, it's pytorch nightly. I'm running into an error about datapipes_only not being a valid kwarg -- the same one that was causing CI errors in Windows. I have been trying to install fresh nightlies for a couple of days but no joy.

erip:
OK, a new nightly has been cut so I can test locally again 🥳  I have fixed one lingering issue and have a TODO to pass the right test sets... will follow a similar approach as langpair parameterization.

erip:
OK @parmeet @Nayef211 this is ready for review. The one thing I couldn't "figure out" was the `"temp_dataset_dir"` business. I think maybe that won't work here since we're looking for a very specific internal structure, but if you have thoughts about how to approach it I can try to add it.

test/datasets/test_iwslt2016.py
```diff
@@ -1,46 +1,128 @@
...
+
+    # create tar file from dataset folder
+    with tarfile.open(inner_compressed_dataset_path, "w:gz") as tar:
+        tar.add(inner_temp_dataset_dir, arcname=f"{src}-{tgt}")
+
+    outer_temp_dataset_path = os.path.join(
+        root_dir, "IWSLT2016", "2016-01.tgz"
+    )
+    with tarfile.open(outer_temp_dataset_path, "w:gz") as tar:
+        tar.add(outer_temp_dataset_dir, arcname="2016-01")
```
parmeet, line 125: Before adding `outer_temp_dataset_dir ` to the archive, we probably want to remove `inner_temp_dataset_dir` since the contents of this folder are already available in the `inner_compressed_dataset_path`, right?

erip: We could approach it that way -- my approach is to just re-read the cleaned files to generate the expected values. I don't know if there's a strong reason to prefer one vs. the other.

test/datasets/test_iwslt2016.py
```diff
@@ -1,46 +1,128 @@
...
+    }[split]
+
+
+def _get_mock_dataset(root_dir, split, src, tgt, valid_set, test_set):
     """
     root_dir: directory to the mocked dataset
     """
-    temp_dataset_dir = os.path.join(root_dir, f"IWSLT2016/2016-01/texts/{src}/{tgt}/{src}-{tgt}/")
-    os.makedirs(temp_dataset_dir, exist_ok=True)
+    outer_temp_dataset_dir = os.path.join(root_dir, f"IWSLT2016/2016-01/texts/{src}/{tgt}/")
```
Nayef211, line 78: If you change this path to be "...IWSLT2016/temp_dataset_dir/2016-01/..." you will notice that there are a lot of test failures. The purpose of naming it this is to ensure that the folder we are creating the mocked data is different from where the files/folders would be place when the archive is extracted (i.e. when calling the dataset). Without doing this, the archive will never be extracted since our dataset implementation checks to see whether a file/folder exists and if it does, the archive is never extracted.

When patching this PR locally and making the above change, I noticed we're getting `212 failed, 451 passed`. This is a sample error I'm getting from pytest:

```
FAILED test_iwslt2016.py::TestIWSLT2016::test_iwslt2016_219_train - AssertionError: 'H T G W w U w K b d\n' != 'b H K H c L t k I s\n'
```

test/datasets/test_iwslt2016.py
```diff
@@ -1,46 +1,128 @@
...
+
+    # create tar file from dataset folder
+    with tarfile.open(inner_compressed_dataset_path, "w:gz") as tar:
+        tar.add(inner_temp_dataset_dir, arcname=f"{src}-{tgt}")
+
+    outer_temp_dataset_path = os.path.join(
+        root_dir, "IWSLT2016", "2016-01.tgz"
+    )
+    with tarfile.open(outer_temp_dataset_path, "w:gz") as tar:
+        tar.add(outer_temp_dataset_dir, arcname="2016-01")
```
Nayef211, line 126: Another alternative to what I mentioned above is to add the following lines here to delete the outer temp dataset dir and in doing so ensure the archive will always be extracted:

```
import shutil
shutil.rmtree(outer_temp_dataset_dir)
```

Doing this gives me the following results: `110 failed, 553 passed`. This is a sample error I'm getting from pytest:
```
FAILED test_iwslt2016.py::TestIWSLT2016::test_iwslt2016_008_train - ValueError: Iterables have different lengths
```

Nayef211 (commented):
Just left some errors I found while patching your PR locally and removing the outer temp dir. Would you be able to investigate further @erip?

test/datasets/test_iwslt2016.py
erip: Oh, I understand the idea now. 👍  Let me give this a shot.

erip:
I'm having some trouble wrapping my mind around all the various nestings again and debugging all the issues accompanying it. I also have some personal things to attend to -- I understand there's some urgency here, so I'm not going to be offended if someone else wants to work on this. I can try to look again tonight.

parmeet:
> I'm having some trouble wrapping my mind around all the various nestings again and debugging all the issues accompanying it. I also have some personal things to attend to -- I understand there's some urgency here, so I'm not going to be offended if someone else wants to work on this. I can try to look again tonight.

Hey @erip, thanks so much for your efforts in this PR so far. Please do not bother. I know this is a complex test and since this is an improvement over what you have already contributed here https://github.com/pytorch/text/pull/1563, I don't necessarily call it a blocker for our branch-cut. I would be happy to take it over from here, otherwise we can keep this PR and do a cherry-picking into the Release branch later. cc: @Nayef211

Nayef211:
> Hey @erip, thanks so much for your efforts in this PR so far. Please do not bother. I know this is a complex test and since this is an improvement over what you have already contributed here #1563, I don't necessarily call it a blocker for our branch-cut. I would be happy to take it over from here, otherwise we can keep this PR and do a cherry-picking into the Release branch later. cc: @Nayef211

@parmeet if you have the bandwidth to take this on that would be great! We just have the test for the IWSLT2016 and IWSLT2017 datasets remaining before we can close #1493. I suspect the current caching issue should be an easy fix if you have a good understanding of the folder structure within the dataset archive.

parmeet:
> > Hey @erip, thanks so much for your efforts in this PR so far. Please do not bother. I know this is a complex test and since this is an improvement over what you have already contributed here #1563, I don't necessarily call it a blocker for our branch-cut. I would be happy to take it over from here, otherwise we can keep this PR and do a cherry-picking into the Release branch later. cc: @Nayef211
> 
> @parmeet if you have the bandwidth to take this on that would be great! We just have the test for the IWSLT2016 and IWSLT2017 datasets remaining before we can close #1493. I suspect the current caching issue should be an easy fix if you have a good understanding of the folder structure within the dataset archive.

sure, let me give it a try.

Nayef211:
Closing this now that https://github.com/pytorch/text/pull/1596 is merged

PR #1586, opened 2022-02-05
fix flake.

erip:


erip:
Oops, missed this with #1584. Closing!

PR #1587, opened 2022-02-05
[WIP] Remove real dataset caching and testing

parmeet:


erip:
This looks reasonable to me. The test failures seem to be related to unavailability of nightly on Windows, so I think this is OK to merge.

Issue #1589, opened 2022-02-07
Add Secondary Caching for Extracted Datasets Files


Nayef211:
## 🚀 Feature
We need to update our dataset implementations to add secondary caching for extracted files as a followup to https://github.com/pytorch/text/issues/1494. 

**Motivation**

Some of our datasets have a `cache_compressed_dp` and then a `cache_decompressed_dp` which is the behavior we want (i.e. [EnWik9](https://github.com/pytorch/text/blob/main/torchtext/datasets/enwik9.py#L35-L50)). On the other hand some of our other datasets only cache the downloaded archive file and not the files extracted from that archive (i.e. [SogouNews](https://github.com/pytorch/text/blob/main/torchtext/datasets/sogounews.py#L47-L51)).

## Backlog of Dataset Tests
The following datasets need to be updated to add the secondary caching mechanism:

- [x] SogouNews #1594
- [x] AmazonReviewFull #1594
- [x] CoNLL2000Chunking #1590 
- [x] DBpedia #1571 

cc @parmeet @abhinavarora @VirgileHlav @erip

erip:
Isn't it the case that some of these files (e.g., AG_NEWS) are just text files? So there's no `cache_decompressed_dp` that would exist?

Nayef211:
> Isn't it the case that some of these files (e.g., AG_NEWS) are just text files so there's no `cache_decompressed_dp` that would exist? Some of these definitely need to be fixed for sure.

Oops you're right thanks for catching this. I've removed AG_NEWS from this list. I may also have missed a dataset or two so it might be worth going through all the datasets again to make sure the list is accurate!

PR #1590, opened 2022-02-07
Updating Conll2000Chunking dataset to be consistent with other datasets

Nayef211:
Reference Issue: https://github.com/pytorch/text/issues/1589

## Summary
- Updated caching datapipe variable names to be consistent with other datasets
- Removed intermediate `conll2000chunking ` folder to download archived files to

### DIFF
--- a/torchtext/datasets/conll2000chunking.py
+++ b/torchtext/datasets/conll2000chunking.py
@@ -58,10 +58,8 @@ def CoNLL2000Chunking(root: str, split: Union[Tuple[str], str]):
     cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
         filepath_fn=lambda x: os.path.join(root, _EXTRACTED_FILES[split])
     )
-    cache_decompressed_dp = (
-        FileOpener(cache_decompressed_dp, mode="b")
-        .extract(file_type="gzip")
-        .filter(lambda x: _EXTRACTED_FILES[split] in x[0])
+    cache_decompressed_dp = FileOpener(cache_decompressed_dp, mode="b").extract(
+        file_type="gzip"
     )
     cache_decompressed_dp = cache_decompressed_dp.end_caching(
         mode="wb", same_filepath_fn=True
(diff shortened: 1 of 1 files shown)


torchtext/datasets/conll2000chunking.py
```diff
@@ -45,20 +45,27 @@ def CoNLL2000Chunking(root: str, split: Union[Tuple[str], str]):
     url_dp = IterableWrapper([URL[split]])
 
     # Cache and check HTTP response
-    cache_dp = url_dp.on_disk_cache(
-        filepath_fn=lambda x: os.path.join(root, "conll2000chunking", os.path.basename(URL[split])),
-        hash_dict={os.path.join(root, "conll2000chunking", os.path.basename(URL[split])): MD5[split]},
```
parmeet, line 50: Thanks for removing extra folder from the path. I think this would address the issue here https://github.com/pytorch/text/pull/1570#discussion_r800782606

torchtext/datasets/conll2000chunking.py
```diff
@@ -45,20 +45,27 @@ def CoNLL2000Chunking(root: str, split: Union[Tuple[str], str]):
...
     # Cache and check the gzip extraction for relevant split
-    cache_dp = cache_dp.on_disk_cache(
-        filepath_fn=lambda x: os.path.join(root, "conll2000chunking", _EXTRACTED_FILES[split])
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, _EXTRACTED_FILES[split])
+    )
+    cache_decompressed_dp = (
+        FileOpener(cache_decompressed_dp, mode="b")
+        .extract(file_type="gzip")
+        .filter(lambda x: _EXTRACTED_FILES[split] in x[0])
```
parmeet, line 64: I don't think it is necessary to do the filtering operation since we only have one de-compressed stream.

parmeet (approved):
Overall LGTM! Thanks for fixing the cache.

torchtext/datasets/conll2000chunking.py
Nayef211: You're right thanks for catching this. Just removed the filtering logic.

Issue #1588, opened 2022-02-06
Datasets Documentation


parmeet:
## 📚 Documentation
 
**Description**
Torchtext provides access to publicly hosted datasets for varying NLP tasks like  classification, machine translation, sequence tagging, Q&A, self-supervised learning etc. In terms of documentation available [here](https://pytorch.org/text/stable/datasets.html#), there is only minimal information provided to the user. Unfortunately, it does not necessarily cover dataset description, source, associated license or citation information, format etc that can help significantly improve user experience. For instance, popular [Datasets](https://huggingface.co/docs/datasets/) library from HuggingFace provides Data Cards for hosted datasets that covers many of these items (and more like leadership boards etc). [Here](https://huggingface.co/datasets/imdb) is an example for the IMDB dataset.
 
## Proposal
In this Issue we would like to improve and standardize the dataset documentation that would help provide better user experience. 
 
Items that we can potentially include (wherever available):
 
**summary**
A brief description of dataset
 
**source and legal**
Provide information on dataset source, licensing and citation information
 
**Statistics**
Provide information on number of examples in the dataset,  number of classes (for classification data), supported language-pairs (for MT data) etc.  
 
**Format**
Provide information on what to expect in the data. For eg for IMDB this may look like `Tuple( class label (type: int), text (type: str))` (this is just an example, other formats suggestions are welcome)

Nayef211:
Thanks for creating this issue @parmeet! Do you want to take on creating/standardizing the documentation for one of the datasets that we can use as an example moving forward when writing the docs for the rest of the datasets? Also would it be worthwhile keeping a checklist of which datasets still need to be worked on similar to what we've been doing for https://github.com/pytorch/text/issues/1493

parmeet:
Closing this issue, since the basic documentation has been added.

PR #1591, opened 2022-02-07
[BC-breaking] remove unnecessary split argument from datasets

parmeet:
This PR remove the split argument from CC100 and Enwik9 datasets. This is superficial just to comply with other datasets format. Specially datasets that serves unsupervised learning, it is often the case that it is just a collection of text. The dev/test split are not always necessarily available. 

BC-Breaking: This PR is BC-breaking for EnWiK9 dataset as we are removing the split argument.


torchtext/datasets/enwik9.py
```diff
@@ -22,10 +17,15 @@
 DATASET_NAME = "EnWik9"
 
 
-@_add_docstring_header(num_lines=NUM_LINES)
```
Nayef211, line 25: Any reason we're removing the `_add_docstring_header` decorator and manually adding the docstrings? Can we modify the `_add_docstring_header` to work with a dataset function that doesn't have a `split` arg?

Nayef211 (commented):
Overall the changes look good to me. A couple of follow-ups:
- Let's remove `test_enwik9_split_argument` from the corresponding test for the `EnWik9` dataset
- Should we make note of the BC breaking changes (removing the `split` arg from EnWik9) in our release notes?

torchtext/datasets/enwik9.py
parmeet: The decorator is doing a job of avoiding writing doc-string and make strong assumptions about datasets. But it's scope is fairly limited (for eg: it does not handle cases when there are no splits or when we have additional arguments other than root and split). Overall, I think we probably need to get rid of _add_doctring_header decorator in light of this issue #1588 anyway since we plan to expand the scope of documentation of what we currently have.

Nayef211 (commented):
I think we still need to remove the `split` argument when calling the EnWik9 dataset as well as the `@parameterized.expand(["train"])` decorator for the `test_enwik9` method to ensure all unit tests are passing

parmeet:
> I think we still need to remove the `split` argument when calling the EnWik9 dataset as well as the `@parameterized.expand(["train"])` decorator for the `test_enwik9` method to ensure all unit tests are passing.
> 
> On the bright side, it looks like our mocked tests are already providing useful signals when we make BC breaking changes to our datasets 😁

haha, true that! :)

Nayef211 (approved):
LGTM! Let's fix the stylecheck errors before merging

PR #1593, opened 2022-02-08
fixes platform-dependent expectation for Multi30k mocked test.

erip:
Tests currently fail on Windows because multi30k doesn't strip newlines and therefore the file is read with carriage returns: `"ABCD\r\n" != "ABC\n"`. This fixes the bug by stripping newlines from Multi30k to be consistent with other datasets.

Technically this is a BC and users can revert it in userland code by appending `os.linesep` to each line as needed.

### DIFF
--- a/test/datasets/test_multi30k.py
+++ b/test/datasets/test_multi30k.py
@@ -29,9 +29,8 @@ def _get_mock_dataset(root_dir):
                 rand_string = " ".join(
                     random.choice(string.ascii_letters) for i in range(seed)
                 )
-                content = f"{rand_string}\n"
-                f.write(content)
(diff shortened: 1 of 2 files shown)


parmeet:
Thanks @erip for the catch. I looked at other tests that are doing the same, but are not failing because the newline is stripped. Per our discussions in this PR https://github.com/pytorch/text/pull/1559, I wonder if we should instead strip newline from Multi30K dataset as well?

erip:
That would certainly be less kludgy! 🤣

parmeet (approved):
LGTM! Thanks for the fix.

PR #1594, opened 2022-02-08
Adding secondary caching to datasets

Nayef211:
Reference issue #1589

## Summary 
- Adding secondary caching to SogouNews and AmazonReviewFull datasets

## Test
- Delete cache and call `next(iter(dataset))` on the relevant datasets to verify data is being returned correctly

### DIFF
--- a/torchtext/datasets/amazonreviewfull.py
+++ b/torchtext/datasets/amazonreviewfull.py
@@ -44,16 +44,26 @@ def AmazonReviewFull(root: str, split: Union[Tuple[str], str]):
         raise ModuleNotFoundError("Package `torchdata` not found. Please install following instructions at `https://github.com/pytorch/data`")
 
     url_dp = IterableWrapper([URL])
-
-    cache_dp = url_dp.on_disk_cache(
+    cache_compressed_dp = url_dp.on_disk_cache(
         filepath_fn=lambda x: os.path.join(root, _PATH),
-        hash_dict={os.path.join(root, _PATH): MD5}, hash_type="md5"
+        hash_dict={os.path.join(root, _PATH): MD5},
+        hash_type="md5",
     )
-    cache_dp = GDriveReader(cache_dp).end_caching(mode="wb", same_filepath_fn=True)
-    cache_dp = FileOpener(cache_dp, mode="b")
+    cache_compressed_dp = GDriveReader(cache_compressed_dp).end_caching(
+        mode="wb", same_filepath_fn=True
+    )
 
-    extracted_files = cache_dp.read_from_tar()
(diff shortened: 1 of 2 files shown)


ejguan:
It seems `content-disposition` is needed to fetch `filename`.

Did you encounter such Error before? I think the implementation here is copied from TorchText https://github.com/pytorch/text/blob/4be2792101565ddf6dd79d1b7fffb7d55d63bf06/torchtext/_download_hooks.py#L21-L48

If you have never encountered it before, I think the problem comes from that we explicitly set streaming for the `request`. https://github.com/pytorch/data/blob/8942cf1d6cdb71013e7b93330a6b0d41445c5e6c/torchdata/datapipes/iter/load/online.py#L54-L57
We may need to switch back to normal request.

ejguan:
> I've been noticing this error for the last week or so for all of the datasets that are being downloaded from Google Drive.

I mean if you have encountered it before revamping Dataset into DataPipe. I believe when @parmeet implemented the GDriveDataPipe, we basically copied the code from TorchText as I linked in the previous comment.
And, what we have changed in TorchData are:
1. Add a context manager
2. Adding streaming to `response = session.get(url, stream=True)`

I am trying to figure out which part causes the issue. If TorchText doesn't have any error with previous implementation in TorchText. This Error has to be introduced by the changes we made (1 or 2).

And, I tried to reproduce it on my local machine but not succeed. Do you want to give a try by changing `stream` to False in your local env and test this PR?

Nayef211:
> And, I tried to reproduce it on my local machine but not succeed. Do you want to give a try by changing `stream` to False in your local env and test this PR?

I don't think the change is that simple since setting `stream=False` results in the following error:
```
RuntimeError: The computed hash d41d8cd98f00b204e9800998ecf8427e of /home/nayef211/.torchtext/cache/AmazonReviewFull/amazon_review_full_csv.tar.gz does not match the expectedhash 57d28bd5d930e772930baddf36641c7c. Delete the file manually and retry.
```
I wonder if it's because we need to modify additional logic if we start downloading the contents immediately instead of streaming. Should we create an issue on torchdata to keep track of this and move the discussions there?

parmeet:
@Nayef211, @ejguan This error existed in torchtext even before the migration, So I don't think any changes on torchdata would have triggered this. In my experience, I have seen this when the quota is exceeded. This is a transient error, though I never dig deeper into if there is a way to increase the quota limit or other alternatives to prevent this from happening. I would agree raising an issue on torchdata, so that we can discuss the problem and potential solutions in there.

parmeet (approved):
LGTM! Thanks for adding cache :)

PR #1596, opened 2022-02-09
IWSLT testing to start from compressed file

parmeet:
Original Credit: @erip 

This PR did some changes on top of https://github.com/pytorch/text/pull/1585 to fix failing tests. The main changes include:
* Removal of uncompressed directories to make sure the dataset is actually doing the extraction
* Removal of code block that conditionally write uncleaned files
* Creation of new base directory with every parameterized test call to make sure no artifacts are left from other tests

### DIFF
--- a/test/datasets/test_iwslt2016.py
+++ b/test/datasets/test_iwslt2016.py
@@ -14,6 +14,7 @@ from ..common.torchtext_test_case import TorchtextTestCase
 
 SUPPORTED_LANGPAIRS = [(k, e) for k, v in SUPPORTED_DATASETS["language_pair"].items() for e in v]
 
+
 def _generate_uncleaned_train():
     """Generate tags files"""
     file_contents = []
--- a/test/datasets/test_iwslt2016.py
+++ b/test/datasets/test_iwslt2016.py
@@ -17,6 +17,7 @@ SUPPORTED_LANGPAIRS = [(k, e) for k, v in SUPPORTED_DATASETS["language_pair"].it
 SUPPORTED_DEVTEST_SPLITS = SUPPORTED_DATASETS["valid_test"]
 DEV_TEST_SPLITS = [(dev, test) for dev, test in itertools.product(SUPPORTED_DEVTEST_SPLITS, repeat=2) if dev != test]
 
+
 def _generate_uncleaned_train():
     """Generate tags files"""
     file_contents = []
@@ -138,6 +139,7 @@ class TestIWSLT2016(TempDirMixin, TorchtextTestCase):
         if (dev_set not in SET_NOT_EXISTS[(src, tgt)] and test_set not in SET_NOT_EXISTS[(src, tgt)])
     ])
     def test_iwslt2016(self, split, src, tgt, dev_set, test_set):
+
         expected_samples = _get_mock_dataset(self.root_dir, split, src, tgt, dev_set, test_set)
 
         dataset = IWSLT2016(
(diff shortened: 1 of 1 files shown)


parmeet:
@erip hope the changes in the PR make sense. Let me know if you catch anything unusual or out of your expectation.

parmeet:
Oh, actually wait. I think I still haven't implemented it correctly. It's about the temporary directory and what we are compressing as part of it. Let's not review it yet @Nayef211, @erip

parmeet:
OK, I think we are good to go for review @Nayef211

test/datasets/test_iwslt2016.py
```diff
@@ -150,7 +152,7 @@ def tearDownClass(cls):
     ])
     def test_iwslt2016(self, split, src, tgt, dev_set, test_set):
 
-        root_dir = self.get_base_temp_dir()
+        root_dir = tempfile.TemporaryDirectory().name
```
erip, line 155: qq, why is this used in `test_iwslt2016`, but not in `test_iwslt2016_split_argument`?

parmeet: It's a good catch. I should use unique base directories in `test_iwslt2016_split_argument ` as well, let me push this change. 

To answer why we need unique base directories with every test: It seems keeping same base directory lead to tests failure that I didn't completely get grasp of. My suspicious is since we are writing and extracting partial (we only write files that are requested in test) dataset  in the same base directory with every parameterized test, this is leading to some weird conflicts in caching and extraction of dataset. So to avoid this situation, it's best to generate new base directory for every parameterized test. Although, I also do see need for better explanation of what's really going on under the hood.

Nayef211: Can we also use a context manager here to ensure that the temp directory get's cleaned up? Usually this would be handled by the `TempDirMixin` class but this test is a special case.

test/datasets/test_iwslt2016.py
```diff
@@ -165,8 +167,12 @@ def test_iwslt2016(self, split, src, tgt, dev_set, test_set):
     @parameterized.expand(["train", "valid", "test"])
     def test_iwslt2016_split_argument(self, split):
         root_dir = self.get_base_temp_dir()
```
Nayef211: If you are planning on removing the call to `get_base_temp_dir `, can we modify this test class to stop extending the `TempDirMixin` class?

test/datasets/test_iwslt2016.py
```diff
@@ -78,7 +79,8 @@ def _get_mock_dataset(root_dir, split, src, tgt, valid_set, test_set):
     """
 
     base_dir = os.path.join(root_dir, DATASET_NAME)
-    outer_temp_dataset_dir = os.path.join(base_dir, f"2016-01/texts/{src}/{tgt}/")
+    temp_dataset_dir = os.path.join(base_dir, 'temp_dataset_dir')
```
Nayef211, line 82: This addition of `temp_dataset_dir` is actually not necessary anymore now that you're deleting this directory at the end of this function. If the directory wasn't being deleted, this would be relevant since otherwise the caching logic would skip extracting the files (since the files exist within this dir). We can still keep this for consistency with the other tests.

test/datasets/test_iwslt2016.py
parmeet: good catch, let me do that!

test/datasets/test_iwslt2016.py
parmeet: that's a good point. let me see where to remove the redundancy.

parmeet: OK, so I think let's keep it consistent with other datasets test. I just removed the deletion of temp_datset_dir.

Nayef211 (approved):
LGTM! Thanks for helping fix the caching issue for the dataset!

erip:
This also looks good to me. Thanks for getting it over the finish line!

PR #1598, opened 2022-02-09
Add Mock test for IWSLT2017 dataset

parmeet:
Reference issue: #1493 

This PR do couple of things:
1) Add Mock test similar to #1596. Original credit: @erip due to #1585
2) Fix an issue on IWSLT2017 that created additional texts folder outside of extracted archive


parmeet:
@erip  would be great if you can also take a look and catch any obvious issues :)

Nayef211 (approved):
This PR looks good and seems consistent with the IWSLT2016 test. Since the two tests share so much similarity, do you think it would be feasible to parameterize them into one test class instead?

PR #1599, opened 2022-02-10
generate unicode strings to test utf-8 handling for all non-IWSLT dataset tests.

erip:
Partially address lingering TODO in #1493. 

TODO: add IWSLT which fails due to XML parsing errors -- need to investigate.


erip:
These test failures are actually good. 🔥

erip:
I have a Windows machine at home, so instead of using CI as a debugging mechanism I can just wrap this PR up once I'm home. I'm glad we caught these. 😄

Nayef211:
Thanks for taking this on @erip! Just a couple of questions from my side. 
- Do we know for certain that all of our datasets are UTF8 encoded?
- Do you have any initial guesses as to why these test failures are occurring? From a first glance it seems like it might be caused by the `LineReaderIterDataPipe` and the `CSVParserIterDataPipe` not handling UTF8 encoded strings correctly.
- Does this PR update all our dataset tests to use the UTF-8 encoded strings? If so it might be worthwhile mentioning that in the PR description as well.

erip:
1. I don't know for certain, but I strongly suspect it. I can confirm later by checking the legacy code to load these datasets -- if the encoding is set, that's evidence.
2. I suspect that the text readers aren't passing the appropriate encoding.
3. All but IWSLTs. Can update for clarity.

test/datasets/test_amazonreviewfull.py
```diff
@@ -0,0 +1,79 @@
...
+        base_dir, "amazon_review_full_csv.tar.gz"
+    )
+    # create tar file from dataset folder
+    with tarfile.open(compressed_dataset_path, "w:gz") as tar:
+        tar.add(temp_dataset_dir, arcname="amazon_review_full_csv")
+
+    return mocked_data
+
+
+class TestAmazonReviewFull(TempDirMixin, TorchtextTestCase):
```
parmeet, line 45: AmazonReviewFull and AmazonReviewPolarity are already merged into single Test case, so we no longer need this file :)

test/datasets/test_yelpreviewfull.py
```diff
@@ -0,0 +1,79 @@
...
+
+    compressed_dataset_path = os.path.join(base_dir, "yelp_review_full_csv.tar.gz")
+    # create gz file from dataset folder
+    with tarfile.open(compressed_dataset_path, "w:gz") as tar:
+        tar.add(temp_dataset_dir, arcname="yelp_review_full_csv")
+
+    return mocked_data
+
+
+class TestYelpReviewFull(TempDirMixin, TorchtextTestCase):
```
parmeet, line 45: same comment as for AmazonReview above. We can remove this file as well.

test/datasets/test_amazonreviewfull.py
erip: Oops -- looks like my merge conflict skills are a bit rusty. 😅

erip:
Tracked down the underlying issue with encoding here on Windows. See [here](https://github.com/pytorch/pytorch/issues/72713) for context.

parmeet:
> 2. I suspect that the text readers either aren't passing the appropriate encoding in 'r' mode or aren't decoding to utf-8 in 'b' mode

Seem like the culprit is not setting `decode` as `True` in our readers which is `False` by default https://github.com/pytorch/data/blob/9c6e5ddfcdf1061e3968ed5cd9d55754cc713965/torchdata/datapipes/iter/util/plain_text_reader.py#L90

erip:
> Seem like the culprit is not setting `decode` as `True` in our readers

Indeed, one option is to read the files in binary and decode appropriately. Ideally the `FileOpener` could handle opening the file with appropriate encoding/mode for us since extra downstream decoding is a bit of a pain. See the issue in upstream pytorch for the "better" option and alternatives.

parmeet:
> > Seem like the culprit is not setting `decode` as `True` in our readers
> 
> Indeed, one option is to read the files in binary and decode appropriately. Ideally the `FileOpener` could handle opening the file with appropriate encoding/mode for us since extra downstream decoding is a bit of a pain. See the issue in upstream pytorch for the "better" option and alternatives.

Agreed. This could (should) be handled at File Opening so that downstream readers do not have to worry about it. Thanks for picking this up @erip. Looking forward to the resolution at torchdata level. Meanwhile (since I am not sure if it will take longer to resolve APIs etc for FileOpener), can we try closing this PR by reading in binary mode and setting `decode` to True for downstream readers? We can then cherry-pick the changes once the FileOpener handles decoding scheme.

parmeet:
> Yes, that seems reasonable.

Great! Then let's try to land this before we make the branch-cut. cc: @Nayef211

erip:
OK, I think this should be gtg now @parmeet

parmeet (approved):
Great work @erip! This looks good to me. I will merge it once the CI is green for unit-testing.

parmeet:
Oh one more change @erip: I know IWSLT 16/17 test suit update is pending, but can we at-least update the reading/decoding part for actual datasets https://github.com/pytorch/text/blob/eb61b3fb1c586c45eda049590045303c91206515/torchtext/datasets/iwslt2016.py#L334-L338

erip:
Good catch, I'll fix that.

erip:
It looks like there's a lingering issue that I'm trying to debug with IMDB. The cache is written as [text](https://github.com/pytorch/text/blob/main/torchtext/datasets/imdb.py#L76-L78), but when I try to change it to being written as bytes and taking appropriate encoding compensation before the cache is written...

```patch
     cache_decompressed_dp = (
         cache_decompressed_dp.lines_to_paragraphs()
     )  # group by label in cache file
+    cache_decompressed_dp = cache_decompressed_dp.map(lambda x: (x[0], x[1].encode()))
     cache_decompressed_dp = cache_decompressed_dp.end_caching(
-        mode="wt",
-        filepath_fn=lambda x: os.path.join(root, decompressed_folder, split, x),
+        mode="wb",
+        filepath_fn=lambda x: os.path.join(root, decompressed_folder, split, x)
     )
```

I'm met with the following errors:

```log
test/datasets/test_imdb.py:84: in test_imdb_split_argument
    for d1, d2 in zip_equal(dataset1, dataset2):
test/common/case_utils.py:53: in zip_equal
    for combo in zip_longest(*iterables, fillvalue=sentinel):
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/_typing.py:366: in wrap_generator
    response = gen.send(None)
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/datapipes/iter/callable.py:95: in __iter__
    for data in self.datapipe:
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/_typing.py:366: in wrap_generator
    response = gen.send(None)
../data/torchdata/datapipes/iter/util/plain_text_reader.py:106: in __iter__
    for path, file in self.source_datapipe:
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/_typing.py:366: in wrap_generator
    response = gen.send(None)
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/datapipes/iter/fileopener.py:45: in __iter__
    yield from get_file_binaries_from_pathnames(self.datapipe, self.mode)
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/datapipes/utils/common.py:85: in get_file_binaries_from_pathnames
    for pathname in pathnames:
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/_typing.py:366: in wrap_generator
    response = gen.send(None)
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/datapipes/iter/combining.py:38: in __iter__
    for data in dp:
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/_typing.py:366: in wrap_generator
    response = gen.send(None)
../data/torchdata/datapipes/iter/util/saver.py:36: in __iter__
    for filepath, data in self.source_datapipe:
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/_typing.py:366: in wrap_generator
    response = gen.send(None)
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/datapipes/iter/callable.py:95: in __iter__
    for data in self.datapipe:
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/_typing.py:366: in wrap_generator
    response = gen.send(None)
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/datapipes/iter/callable.py:96: in __iter__
    yield self._apply_fn(data)
../../opt/miniconda3/envs/torchtext-dev/lib/python3.8/site-packages/torch/utils/data/datapipes/iter/callable.py:69: in _apply_fn
    res = self.fn(data[self.input_col])
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

fd = b'\xc3\xafp\xc4\xa9\xc8\xba\xc2\xbd\xc5\xb7\n\xc5\x9b\xe1\x9b\xac\xe2\xb1\xb4\xc7\xbf\xe2\xb1\xa3\xe2\xb1\xa4\xc2\xb6'

    def _read_bytes(fd):
>       return b"".join(fd)
E       TypeError: sequence item 0: expected a bytes-like object, int found

../data/torchdata/datapipes/iter/util/cacheholder.py:215: TypeError
```

erip:
The test seems OK if I pass `skip_read=True` to `end_caching`, but I'm not super comfortable with this decision because I don't quite understand what I'm losing from this decision...

parmeet:
> > The test seems OK if I pass `skip_read=True` to `end_caching`, but I'm not super comfortable with this decision because I don't quite understand what I'm losing from this...
> 
> It seems that `_read_bytes` is expecting a file-like object. But here since we are already mapping string to bytes (by doing encoding), this function doesn't work with bytes (although interesting `_read_str` works both with file-like object and str in which case it simply return the same string). I think it is save to use `skip_read=True` since we don't really need to read from stream. cc: @ejguan to make sure my understanding is right?

@erip could you please push this change, so that we can close this PR in prep for branch-cut? Thanks

erip:
I think this is good to merge @Nayef211 @parmeet

PR #1602, opened 2022-02-10
[FORMATTING] Updated formatting for all datasets

Nayef211:
- Updated formatting for all datasets using `black`


parmeet (approved):
LGTM! Can you please update summary on which formatter is used in here?

PR #1603, opened 2022-02-11
[DOCS] Added docs for text classification and language modeling datasets

Nayef211:
Reference issue: https://github.com/pytorch/text/issues/1588

## Description
Added docs for the following datasets
- Text Classification
  - AG_NEWS
  - SogouNews
  - DBpedia
  - YelpReviewPolarity
  - YelpReviewFull
  - YahooAnswers
  - AmazonReviewPolarity
  - AmazonReviewFull
  - IMDb
- Language Modeling
  - WikiText-2
  - WikiText103
  - PennTreebank

PR #1604, opened 2022-02-11
Added docs for CC100 and SST2

Nayef211:
## Description
- Added dataset docs for SST2 and CC100
- Converted splits to list in dosctring as a followup from https://github.com/pytorch/text/pull/1603#discussion_r804997079
- Reorganized datasets in documentation to be in alphabetical order


parmeet (approved):
LGTM! I think we miss AmazonReviewPolarity but we can do a follow-up. I will go ahead and merge this one.

PR #1616, opened 2022-02-16
prepare datasets for new encoding kwarg.

erip:
This may need to wait for appropriate nightlies to be published, but @ejguan suggested this feature will make it into pytorch 1.11 so that's a good sign.

### DIFF
--- a/torchtext/datasets/ag_news.py
+++ b/torchtext/datasets/ag_news.py
@@ -61,6 +61,5 @@ def AG_NEWS(root: str, split: Union[Tuple[str], str]):
     cache_dp = HttpReader(cache_dp)
     cache_dp = cache_dp.end_caching(mode="wb", same_filepath_fn=True)
 
-    # TODO: read in text mode with utf-8 encoding, see: https://github.com/pytorch/pytorch/issues/72713
-    data_dp = FileOpener(cache_dp, mode="b")
+    data_dp = FileOpener(cache_dp, encoding="utf-8")
     return data_dp.parse_csv().map(fn=lambda t: (int(t[0]), " ".join(t[1:])))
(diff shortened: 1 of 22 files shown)


ejguan:
Note: The PR with the change of encoding has already been merged into release/1.11 branch.

Nayef211:
> This may need to wait for appropriate nightlies to be published, but @ejguan suggested this feature will make it into pytorch 1.11 so that's a good sign.

I think we're using the latest `main` branch for torchdata for our unit tests so our tests should not be failing if the `encoding` kwarg is already present in the torchdata repo. 

Here are the code pointers to where we are installing torchdata
- [linux unittests](https://github.com/pytorch/text/blob/main/.circleci/unittest/linux/scripts/install.sh#L16-L17)
- [windows unittests](https://github.com/pytorch/text/blob/main/.circleci/unittest/windows/scripts/install.sh#L21-L22)

erip:
The encoding kwarg is in pytorch proper, not torchdata.

Nayef211:
> The encoding kwarg is in [pytorch proper](https://github.com/pytorch/pytorch/blob/78e481d07d3ec94d89badf3beff52ca75615a10e/torch/utils/data/datapipes/iter/fileopener.py), not torchdata.

Ahh understood, thanks for clarifying!

Nayef211:
Btw can we fix the merge conflicts with some of the datasets?

erip:
Oops -- yep!

parmeet:
Great to see this PR @erip! Hopefully we should soon be in position to land these changes :)

erip:
This is a good sign. Once nightlies with Windows (and OS X) are published, this should be gtg -- Linux nightlies have propagated and tests pass in those cases.

parmeet (approved):
Overall the changes LGTM! We can merge once after the windows tests are passing. Thanks.

erip:
New windows and OS X nightlies landed, so this should hopefully pass and be good to go. 🎆

Issue #1612, opened 2022-02-14
Make the `CLIPTokenizer`'s `encoder_json_path` variable optional, and use `dict(zip(vocab, range(len(vocab))))` instead


ProGamerGov:
## 🚀 Feature

https://github.com/pytorch/text/blob/main/torchtext/transforms.py#L312

In both CLIP and OpenCLIP, the encoder is simply just the vocab run through `dict(zip(vocab, range(len(vocab))))`, and it doesn't make a ton of sense to require a `encoder.json` file for this information. The `encoder.json` file requirement should be optional as the vocab file itself can be used to create the encoder, making the encoder file redundant.

https://github.com/openai/CLIP/blob/main/clip/simple_tokenizer.py#L74
https://github.com/mlfoundations/open_clip/blob/main/src/clip/tokenizer.py#L78

The current `clip_encoder.json` test asset is just the Python dict created by `dict(zip(vocab, range(len(vocab))))`, and thus will it makes a useful test for specifying an encoder, it's redundant: https://github.com/pytorch/text/blob/main/test/asset/clip_encoder.json

parmeet:
cc: @abhinavarora

abhinavarora:
Thank you @ProGamerGov for bringing this to my notice. I agree with you that this can get redundant. Currently we take both the encoder json and the vocab file to match the output of [HuggingFace's CLIP tokenizer](https://github.com/mlfoundations/open_clip/blob/main/src/clip/tokenizer.py#L67) which is a popular CLIP implementation. I am creating a PR to address this issue as well as another issue which deals with [hardcoding of vocab length in the CLIP implementation](https://github.com/mlfoundations/open_clip/blob/main/src/clip/tokenizer.py#L67). 

I will share a PR that will do the following:
1. Make encoder json optional 
2. In case encoder json is provided, infer the vocab length from it. This is to deal with the hardcoding in the CLIP implementation
3. Add an optional num_merges parameter to specify the number of merges when encoder json is not provided. This is again to deal with hardcoded parameters.

Let me know if this makes sense to you?

abhinavarora:
> @abhinavarora Yeah, that makes sense and sounds good to me!

@ProGamerGov Here is PR: #1622

ProGamerGov:
@abhinavarora Thank you for taking the time to implement the changes!

PR #1622, opened 2022-02-18
Modify CLIPTokenizer to either infer number of merges from encoder json or take it in constructor

abhinavarora:
# Problem 

@ebsmothers reported that current CLIPTokenizer runs into errors when original merges files is loaded from OpenAI. This happens because OpenAI codebase hardcodes the number of merges which torchtext did not do. 

Similarly @ProGamerGov reported #1612 . 

# Solution
This PR addresses these issues by doing the following:
1. Enable initializing CLIPTokenizer with just the bpe merges file, similar to OpenAI. 
2. Add a `num_merges` param that users can provide.
3. If encoder json is provided, use that to infer number of merges. 

# Testing
* Added new tests with an example that was failing earlier. 
* Added tests for different types of initialization

```
pytest test/test_transforms.py -k test_clip
```

### DIFF
--- a/torchtext/transforms.py
+++ b/torchtext/transforms.py
@@ -322,30 +322,44 @@ class CLIPTokenizer(Module):
     (a bit like sentencepiece) so a word will be encoded differently whether it
     is at the beginning of the sentence (without space) or not.
 
+    :param merges_path: Path to bpe merges file.
+    :type merges_path: str
     :param encoder_json_path: Path to BPE encoder json file.
     :type encoder_json_path: str
-    :param vocab_bpe_path: Path to bpe vocab file.
-    :type vocab_bpe_path: str
+    :param num_merges: Number of merges to read from the bpe merges file.
+    :type num_merges: int
     """
 
     _seperator: torch.jit.Final[str]
 
-    def __init__(
-        self,
-        encoder_json_path: str,
-        vocab_bpe_path: str,
-    ):
+    def __init__(self, merges_path: str, encoder_json_path: Optional[str] = None, num_merges: Optional[int] = None):
         super().__init__()
         self._seperator = "\u0001"
-        # load bpe encoder
-        with open(get_asset_local_path(encoder_json_path), "r", encoding="utf-8") as f:
-            bpe_encoder = json.load(f)
-        # load bpe vocab
-        with open(get_asset_local_path(vocab_bpe_path), "r", encoding="utf-8") as f:
-            bpe_vocab = f.read()
-        bpe_merge_ranks = {
-            self._seperator.join(merge_pair.split()): i for i, merge_pair in enumerate(bpe_vocab.split("\n")[1:-1])
-        }
+        # load bpe merges
+        with open(get_asset_local_path(merges_path), "r", encoding="utf-8") as f:
+            bpe_merges = f.read().split("\n")[1:]
+
+        if encoder_json_path:
+            # load bpe encoder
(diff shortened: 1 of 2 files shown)


ebsmothers (approved):
Looks great from my perspective. Thanks for the quick fix!

torchtext/transforms.py
```diff
@@ -322,30 +322,44 @@ class CLIPTokenizer(Module):
...
-        with open(get_asset_local_path(encoder_json_path), "r", encoding="utf-8") as f:
-            bpe_encoder = json.load(f)
-        # load bpe vocab
-        with open(get_asset_local_path(vocab_bpe_path), "r", encoding="utf-8") as f:
-            bpe_vocab = f.read()
-        bpe_merge_ranks = {
-            self._seperator.join(merge_pair.split()): i for i, merge_pair in enumerate(bpe_vocab.split("\n")[1:-1])
-        }
+        # load bpe merges
+        with open(get_asset_local_path(merges_path), "r", encoding="utf-8") as f:
```
erip, line 339: This may be a bit too much for this PR, but for sake of starting the conversation... is it possible to move the files from the constructor into a classmethod? The constructor is doing a lot of work and if people have their own merges, there's no direct way they can construct this without first writing them to a file.

torchtext/transforms.py
```diff
@@ -322,30 +322,44 @@ class CLIPTokenizer(Module):
...
     """
 
     _seperator: torch.jit.Final[str]
 
-    def __init__(
-        self,
-        encoder_json_path: str,
-        vocab_bpe_path: str,
-    ):
+    def __init__(self, merges_path: str, encoder_json_path: Optional[str] = None, num_merges: Optional[int] = None):
```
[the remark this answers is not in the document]
parmeet: I think it's OK. This is still to be released in upcoming cycle.

torchtext/transforms.py
parmeet: Sorry, I am not sure if I follow it completely. So as per the current interface user would have to provide the file paths. Are suggesting that the constructor should allow passing both file paths as well as direct `merges` object?

torchtext/transforms.py
```diff
@@ -322,30 +322,44 @@ class CLIPTokenizer(Module):
     (a bit like sentencepiece) so a word will be encoded differently whether it
     is at the beginning of the sentence (without space) or not.
```
parmeet, line 324: @abhinavarora As we discussed earlier for tokenizers doc-strings to provide information on standard out-of-the-box artifacts that we host, can we also update  to include example and provide artifacts path?

torchtext/transforms.py
erip: No, I'm suggesting that the constructor should not deal with files at all.

abhinavarora: I believe this is a broader discussion, not specific to this PR. I believe there has been a discussion around this before. I agree with you, that in future we should deal with file-like objects. I remember there were some concerns around this./ Maybe @parmeet may remember. 

@erip Let's track this in a separate issue, so that we can standardize this across all our methods that depend on external files.

erip: Sounds good -- opened #1624

torchtext/transforms.py
abhinavarora: Added in the latest commit!

abhinavarora:
The CircleCI error seems unrelated to this PR. Will merge this PR and debug later.

PR #1625, opened 2022-02-22
Build and test py3.10

atalman:
Build and test for Python 3.10

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -586,6 +586,9 @@ workflows:
       - binary_linux_wheel:
           name: binary_linux_wheel_py3.9
           python_version: '3.9'
+      - binary_linux_wheel:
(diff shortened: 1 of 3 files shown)


parmeet:
@atalman could please try rebasing on the main, the unit-test failures doesn't make sense. It is likely due to some conflicting changes in config.yml.

atalman:
@parmeet can you please take a look at this unit test failure ?  unittest_windows_py3.8

parmeet:
> @parmeet can you please take a look at this unit test failure ? unittest_windows_py3.8

Please ignore this test failure. It's some randomness that causing this test failure. In worst-case we will just disable it, but for now it shouldn't be a blocker.

parmeet (approved):
LGTM!

PR #1626, opened 2022-02-22
Modify CLIPTokenizer to either infer number of merges from encoder json or take it in constructor (#1622)

abhinavarora:


parmeet (approved):
LGTM!

PR #1635, opened 2022-03-01
[BC-breaking] rename Roberta Bundle

parmeet:
This PR changes the name of `RobertaModelBundle` -> `RobertaBundle` to keep consistency across domains. Unfortunately, we are already past deadline to cherry-pick this change for 0.12 release. We can incorporate this change in dot release.

### DIFF
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -8,7 +8,7 @@ torchtext.models
 .. currentmodule:: torchtext.models
 
 RobertaBundle
-------------------
+-------------
 
 .. autoclass:: RobertaBundle
    :members: transform
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -7,10 +7,10 @@ torchtext.models
 .. automodule:: torchtext.models
 .. currentmodule:: torchtext.models
 
-RobertaModelBundle
+RobertaBundle
 ------------------
 
-.. autoclass:: RobertaModelBundle
+.. autoclass:: RobertaBundle
    :members: transform
 
    .. automethod:: get_model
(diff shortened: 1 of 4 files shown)


Nayef211 (approved):
LGTM. Just a followup question. When we say dot release, do we mean something like `0.12.1`? And how do we create a dot release?

mthrok:
> This PR changes the name of `RobertaModelBundle` -> `RobertaBundle` to keep consistency across domains. Unfortunately, we are already past deadline to cherry-pick this change for 0.12 release. We can incorporate this change in dot release.

How do you gauge the impact? In audio, bundles are not expected to be instantiated by users directly, so the impact would be small. It is a public API for the sake of documenting instance method properly.

docs/source/models.rst
```diff
@@ -7,10 +7,10 @@ torchtext.models
 .. automodule:: torchtext.models
 .. currentmodule:: torchtext.models
 
-RobertaModelBundle
+RobertaBundle
 ------------------
```
mthrok, line 11:
```suggestion
-------------
```

mthrok (approved):

parmeet:
> When we say dot release, do we mean something like `0.12.1`? And how do we create a dot release?
Yes, that's right. We do not have to do explicitly anything if release eng. team already planning to do dot release for pytorch core (in which case they would do corresponding dot releases for domain libraries too). We can just cherry-pick in the same release branch for dot-releases.

parmeet:
> How do you gauge the impact? In audio, bundles are not expected to be instantiated by users directly, so the impact would be small.

It's a good question. Actually in text we are also exposing the [builder API](https://github.com/pytorch/text/blob/e6ba477a7bc769a075de35487d75d7f6439c1775/torchtext/models/roberta/bundler.py#L113). So I guess if users already make use of this API, they would be affected by backward incompatibility. That said, I would expect the impact to be small as well given that we are just introducing it in this release.

PR #1641, opened 2022-03-07
remove install instructions from tutorial

parmeet:
The installation instructions corresponds to nightly release. As we are getting close to release, we can remove them as users would be able to run this using the released binaries.


parmeet:
Actually i am thinking, we can wait to land this until after the release. We can then update the installation instructions accordingly.

Nayef211 (approved):

PR #1642, opened 2022-03-09
Add unicode generation to IWSLT tests (followup to #1608)

Nayef211:
- Followup to #1608 
- Addresses #1607 

### Testing
- `pytest test_iwslt2016.py`
- `pytest test_iwslt2017.py`


parmeet (approved):
LGTM! Thanks for fixing this. I think this is the reason for random test failure.

PR #1643, opened 2022-03-09
remove models from experimental

parmeet:
The utility function is simple enough to not be included as standard feature or utility function


Nayef211 (approved):
This removal sounds reasonable to me. Also looks like we weren't using this anywhere in our repo anyways.

PR #1645, opened 2022-03-10
remove experimental functional

parmeet:
summary:

* to_tensor is already added in main
* sequential_transform is already added in main


parmeet:
> Just wondering, but do we have an implementation of `ngrams` elsewhere? Or is this no longer something we want to maintain?

Yupp, we have an implementation [here](https://github.com/pytorch/text/blob/fcbe05745916d8a672f990a8aac40ca0f788f6da/torchtext/data/utils.py#L208) which is an iterator. The functor that is deleted here is just converting that iterator into list, so it is not as such a valuable feature to keep around.

PR #1646, opened 2022-03-10
remove experimental datasets

parmeet:
removing couple of datasets from experimental folder:
WMT14: (The download link is broken)
WMTNewsCrawl: This is one of the subsets of multiple datasets from [here](https://www.statmt.org/wmt11/translation-task.html) and there is no clear direction as to why this dataset is relevant

Removing them would also allow us to remove many utilities which are no longer needed to be maintained in core due to migration on datapipes.


Nayef211 (commented):
I do see fairseq has a couple of models that are trained on the `WMT14` dataset [here](https://github.com/pytorch/fairseq/blob/main/examples/language_model/README.md). I'm wondering, what the standard/bar that a dataset has to pass in order for us to maintain it in torchtext?

NicolasHug:
For torchvision the 2 main criteria for inclusion have been

- popular on papers with code. What defines "popular" is somewhat arbitrary and we don't have a definite threshold I believe
- someone / some team internally needs it

Nayef211:
> For torchvision the 2 main criteria for inclusion have been
> 
> * popular on papers with code. What defines "popular" is somewhat arbitrary and we don't have a definite threshold I believe
> * someone / some team internally needs it

In that case, let's delete this for now so we can also get rid of the additional utilities that are maintained in torchtext. If in the future, we deem that these datasets should be added, we can reimplement them using datapipes instead.

Nayef211 (approved):

PR #1647, opened 2022-03-10
clean-up examples folder

parmeet:
This PR cleans-up a number of stale examples:
* utils: We would like to deprecate download_from_url and extract_archive. These were maintained as they were needed for datasets before we moved them to DataPipes.
* ScriptVocab: It is just a wrapper around Vocab to show how the ScriptVocabulary in PyText can be constructed using torchtext Vocab. Since PyText is deprecated, this example is no longer relevant outside of PyText.
* pipeplines: The examples in pipelines are rudimentary and does not represent standard data pipelines. Many of them are non-functional due to changes in code-base. Going forward, we would like to create standard datapipes like RoBERTa and XLM-R pre-processing. 
* dataset: It is no longer relevant in context of DataPipes where batch can yielded directly using Batch API
* transforms: No longer relevant or depend on deprecated code


Nayef211 (approved):
These look fine to delete. One thought I had that may warrant further discussion is whether we want to create pipeline benchmarks that look similar to what we had in `pipelines.py`. I.e. supporting pipeline level benchmarks with components that are registered with pybind, torchbind, and are scripted.

parmeet:
> supporting pipeline level benchmarks with components that are registered with pybind, torchbind, and are scripted.

yupp I think this would make sense. Let's just re-write them using the new datasets and also potentially in context of standard data pre-processing text pipelines (XLM-R, RoBERTa etc).

PR #1649, opened 2022-03-10
update stable to point to 0.12

parmeet:


parmeet:
cc: @brianjo , @mattip

versions.html
```diff
@@ -21,20 +21,17 @@ <h1> PyTorch Documentation </h1>
           <a class="reference internal" href="main/">main (unstable)</a>
         </li>
         <li class="toctree-l1">
-          <a class="reference internal" href="0.12.0/">v0.12.0 (release candidate)</a>
+          <a class="reference internal" href="0.12.0/">v0.12.0 (stable candidate)</a>
```
brianjo, line 24: Should probably be (stable release) rather than candidate. Also if we rename the folder, then the pointer should go to 0.12 rather than 0.12.0.

brianjo (changes requested):

versions.html
nateanl: I saw PyTorch uses [`1.11`](https://pytorch.org/docs/1.11/) for this release folder instead of `1.11.0`, but considering we are publishing 1.11.1 minor release, it's better to keep `1.11.0` in this case?

mattip: Since typically for a point release the documentation should not change, I think it is better not to create too many duplicate documentation versions. There should be no visible changes between 1.11.0 and 1.11.1. By limiting the number of versions of documentation it keeps the version switcher more compact and also makes the site lighter. You could just expose the latest in the switcher, but then why bother keeping the older one around

nateanl: Sounds good. Indeed PyTorch 1.10 doesn't include dot release doc. We can rename the folder to `0.12`.

parmeet: As discussed, let's keep the folder name as such and we will follow up. Could I get stamp please, and I will land this. Thanks

Nayef211 (approved):

PR #1653, opened 2022-03-11
Add more test coverage

VirgileHlav:
Added more tests for raise/log cases and few untested functions in:
- `torchtext.nn.modules.multiheadattention.py`
- `torchtext.models.roberta.bundler.py`
- `torchtext.functional.py`

Following stats in [codecov.io/gh/pytorch/text](https://app.codecov.io/gh/pytorch/text).

### DIFF
--- a/test/data/test_modules.py
+++ b/test/data/test_modules.py
@@ -2,6 +2,7 @@ import torch
 from torch.nn import Linear
 from torch.nn.functional import multi_head_attention_forward as mha_forward
 from torchtext.nn import InProjContainer, MultiheadAttentionContainer, ScaledDotProduct
+from torchtext.nn.modules.multiheadattention import generate_square_subsequent_mask
 
 from ..common.torchtext_test_case import TorchtextTestCase
 
@@ -202,6 +203,24 @@ class TestModels(TorchtextTestCase):
                 value.expand(src_len, 2, embed_dim),
                 attn_mask=attn_mask_2D.expand(bsz * nhead, tgt_len, src_len),
             )
+        # attn_mask must be a 3D tensor.
+        with self.assertRaises(RuntimeError):
+            SDP(
+                query.expand(1, 2, 3, tgt_len, bsz * nhead, embed_dim),
+                key.expand(src_len, 2, embed_dim),
+                value.expand(src_len, 2, embed_dim),
+                attn_mask=attn_mask_2D,
+            )
+
+        # Only bool tensor is supported for attn_mask
+        attn_mask_2D_not_bool = torch.randint(0, 2, (tgt_len, src_len))
+        with self.assertRaises(RuntimeError):
+            SDP(
+                query.expand(1, 2, 3, tgt_len, bsz * nhead, embed_dim),
+                key.expand(src_len, 2, embed_dim),
+                value.expand(src_len, 2, embed_dim),
+                attn_mask=attn_mask_2D_not_bool,
+            )
 
         # attn_mask in a size of (1, tgt_len, src_len)
         # 2D tensor is not supported for attn_mask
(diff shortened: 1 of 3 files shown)


test/test_functional.py
```diff
@@ -68,6 +77,8 @@ def _add_token(self, test_scripting):
         func = functional.add_token
         if test_scripting:
             func = torch.jit.script(func)
+
+        # List[List[int]]
```
Nayef211, line 81: nit: can we prefix these comments with `case:`. At a first glance it almost looks like commented code.

test/test_functional.py
```diff
@@ -1,90 +1,78 @@
...
+        inputs, padding_value, expected_list = configs
         func = functional.to_tensor
         if test_scripting:
             func = torch.jit.script(func)
-        actual = func(input, padding_value=padding_value)
-        expected = torch.tensor([[1, 2, 0], [1, 2, 3]], dtype=torch.long)
-        torch.testing.assert_close(actual, expected)
+        else:
+            with self.assertRaises(TypeError):
+                func("test")
```
Nayef211, line 25: Just wondering if this additional assertion helps to increase test coverage by ensuring the exception handling implementation is working correctly?

Nayef211 (approved):
Overall LGTM. Thanks for helping improve the test coverage for our repo!

test/test_functional.py
VirgileHlav: Yes, this is to test that an input type that is not a `List[int]` or `List[List[int]]` raises a `TypeError` (see [here](https://github.com/pytorch/text/blob/main/torchtext/functional.py#L37)).

test/data/test_modules.py
```diff
@@ -213,11 +190,94 @@ def test_broadcast_scaled_dot_product(self):
...
-        # attn_mask's dim -3 is not equal to neither batch size or 1
-        with self.assertRaisesRegex(RuntimeError, "The size of the attn_mask is not correct."):
-            SDP(
-                query.expand(tgt_len, bsz * nhead, embed_dim),
-                key.expand(src_len, bsz * nhead, embed_dim),
-                value.expand(src_len, bsz * nhead, embed_dim),
-                attn_mask=attn_mask_2D.expand(2, tgt_len, src_len),
-            )
+
+    def test_assert_raises(self):
```
mthrok, line 194: This test name is generic and thus it is making it hard to understand what test is about.

Also can you add docstring so to tell the intended behavior of the test? Typically this should be "(Under some condition) this feature should produce this result"

test/data/test_modules.py
```diff
@@ -213,11 +190,94 @@ def test_broadcast_scaled_dot_product(self):
...
+        embed_dim, nhead, tgt_len, src_len, bsz = 10, 5, 6, 10, 64
+        SDP = ScaledDotProduct()
+        query = torch.rand((tgt_len, 1, embed_dim))
+        key = value = torch.rand((src_len, 1, embed_dim))
+        attn_mask_2D = torch.randint(0, 2, (tgt_len, src_len))
+
+        configs = [
```
mthrok, line 201: Looking at the following `for` loop, the configs here are independent each other. However in the current implementation, one of the configs fail to pass the test, then the rest of the configs are not tested.

Unit tests should be independent each other. So these configs are better extracted out of the test logic, using `parameterized.expand`.

test/test_functional.py
mthrok: Looking at the flow of the test logic, this assertion kind of showed us suddenly. 

Also that assertion does not require the parameterized input, so why not put it in a stand-alone test?

test/test_functional.py
```diff
@@ -1,90 +1,78 @@
...
+        [True, False],
+        [
+            [[[1, 2], [1, 2, 3]], [[1, 2], [1, 2]]],
+            [[1, 2, 3], [1, 2]],
+            [[["a", "b"], ["a", "b", "c"]], [["a", "b"], ["a", "b"]]],
+            [["a", "b", "c"], ["a", "b"]],
+        ]
+    )
+    def test_truncate(self, test_scripting, configs):
+        """test truncation on both sequence and batch of sequence with both str and int types"""
```
mthrok, line 41: This docstring does not explain what the actual testing is. Could you rephrase in a way that better explains what the passing criteria is? such as something should match with somethig

test/test_functional.py
```diff
@@ -1,90 +1,78 @@
...
+        inputs, expected = configs
         max_seq_len = 2
         func = functional.truncate
         if test_scripting:
             func = torch.jit.script(func)
+        else:
+            with self.assertRaises(TypeError):
+                func("test", max_seq_len=max_seq_len)
```
mthrok, line 49: Since this assertion has nothing to do with the parameterized part of the input values, it should be in a separate, stand-alone test.

test/test_functional.py
VirgileHlav: @mthrok thanks for the comments! I addressed all the issues in #1667.

test/data/test_modules.py
VirgileHlav: Done. See #1667.

test/data/test_modules.py
VirgileHlav: Done. See #1667.

test/models/test_models.py
```diff
@@ -135,3 +136,45 @@ def _train(model):
 
         self.assertEqual(model.encoder.state_dict(), encoder_current_state_dict)
         self.assertNotEqual(model.head.state_dict(), head_current_state_dict)
+
+    @pytest.fixture(autouse=True)
```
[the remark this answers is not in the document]
VirgileHlav: It does not, I changed the structure to use mock+patching. See #1667.

test/test_functional.py
VirgileHlav: Done. See #1667.

test/test_functional.py
VirgileHlav: Done. See #1667.

PR #1654, opened 2022-03-12
clean-up stale code

parmeet:
Summary:
* Removed stale code that is no longer used in the library due to changes in documentation decorators and Datasets migration 
* Removed unicode_csv_reader that is meant for  python 2


parmeet:
@VirgileHlav looks like removing stale code helps increase the coverage :)

Nayef211 (approved):

Issue #1662, opened 2022-03-25
How to install LTS (0.9.2)?


PolarisRisingWar:
## ❓ Questions and Help

**Description**

I've found that my PyTorch version is 1.8.2, so according to https://github.com/pytorch/text/#installation , the torchtext version is 0.9.2:
![image](https://user-images.githubusercontent.com/48322321/160079803-6149cea6-ccfa-4b8f-a392-894c1a018216.png)
But as I use `conda install -c pytorch torchtext` to install, the version I installed defaultly is 0.6.0. So I wander, is this version also OK for  me as the torchtext version 0.9.2 is the highest version I can install, or it's not OK as I can only install 0.9.2 version?

Nayef211:
I would recommend to update your torchtext version to 0.9.2 if you are using PyTorch version 1.8.2. Different versions of torchtext may work with different versions of PyTorch, but we don't provide any guarantees that they will be compatible and there won't be any breakages.

PolarisRisingWar:
I've uninstalled 0.6.0 version. But when I use `conda install -c pytorch torchtext==0.9.2`, it shows that this version is not in conda channel. When I use `pip install torchtext==0.9.2`, it shows that `ERROR: Could not find a version that satisfies the requirement torchtext==0.9.2 (from versions: 0.1.1, 0.2.0, 0.2.1, 0.2.3, 0.3.1, 0.4.0, 0.5.0, 0.6.0, 0.7.0, 0.8.0, 0.8.1, 0.9.0, 0.9.1, 0.10.0, 0.10.1, 0.11.0, 0.11.1, 0.11.2, 0.12.0)`
`ERROR: No matching distribution found for torchtext==0.9.2`  
How could I get this version? Is it correct?

PolarisRisingWar:
I tried `pip install torchtext==0.9.1`, and it forced my pytorch version changed to 1.8.1...

Nayef211:
Ahh I think in this case our version compatibility matrix might not be correct. I just took a look at all the torchtext versions available in conda [here](https://anaconda.org/pytorch/torchtext/files?page=2). If you click on the info for `torchtext 0.9.1` you will see  that the min version required for PyTorch is indeed `1.8.1`. 

Are you able to update to the latest versions of the package? Or is there a reason you want to use the older packages?

Nayef211:
[...]

PR #1665, opened 2022-03-28
Update README around installing LTS version

Nayef211:
## Descritption
- Adding more clarification around installing LTS binary distribution as a followup from #1662

## Changes
image

image


mthrok (approved):

abhinavarora (commented):
LGTM!

PR #1660, opened 2022-03-17
[CMake 3/3] Split source files with Python dependency to separate library

Nayef211:
**Reference Issue #1644**

## Description
- Updating torchtext extension to use CMake instead of PyTorch's `CppExtension` module
- This PR will be split up into 3 parts:
   1) Update C++ source files to use imports realtive to root directory (to fix CMake build errors) - #1666
   2) Create a single `_torchtext.so` library with all source files from `csrc/` folder. This is similar to the build artifacts created by the current build process - #1673
   3) Separate out source files that have a Python dependency to `_torchtext.so`. This will link against the `libtorchtext.so` library containing source files with pure C++ dependecies.

## Build Process
**Building the torctext project using `setup.py`**
```shell
python setup.py develop
```

**Building the extension manually using CMake**

```shell
# from root of repo
cd build
rm -rf * # clear all existing artifacts
cmake \
    -DCMAKE_PREFIX_PATH=`python -c 'import torch;print(torch.utils.cmake_prefix_path)'` \
    -DRE2_BUILD_TESTING:BOOL=OFF \
    -DBUILD_TESTING:BOOL=OFF \
    -DSPM_ENABLE_SHARED=OFF  \
    ..
cmake --build . --config Release
```

Issue #1644, opened 2022-03-09
Enable CMake Based Build System for Torchtext


Nayef211:
## 🚀 Feature
Create native C++ binaries using a CMake based build system to allow third-parties to integrate torchtext in their applications

### Motivation
Currently torchtext’s C++ extension and third party libraries are built using PyTorch’s [`CppExtension`](https://pytorch.org/docs/stable/cpp_extension.html#torch.utils.cpp_extension.CppExtension) module. Though this approach has some pros, the main downsides are that it makes integrating third party libraries challenging and the resulting library is linked against `libpython`, via `libtorch_python`. Due to the dependency on `libpython`, OSS torchtext does not have a standalone C++ binary that downstream users can depend on. This is a blocker for some customers who want to use torchtext operators in a non-Python environment (using TorchScript or as a C++ API). 

There are some issues created by users that further motivate this work:
- #1615
- #1255

### Pitch

We aim to resolve these problems by moving our build system to use CMake. Firstly, CMake should make it easier to integrate third party libraries simply by adding a reference to it’s own CMake file to build the library. Secondly, since CMake is language-agnostic, there is no need to link against libpython.

Before making any changes to torchtext’s build system, we need to ensure that there is adequate testing across all platforms in our OSS CI. This means enabling MacOS unit tests in CircleCI to catch platform specific build failures caused by the switch to CMake.

## Milestones 

- [x] Enable MacOS Unit tests #1672
- [x] Update C++ includes to use imports relative to root directory (to resolve CMake build errors) #1666
- [x] Switch the build mechanism from PyTorch’s `CppExtension` to CMake for torchtext and create a single `_torchtext.so` library with all source files from `csrc/` folder #1673
- [x] Separate out source files that have a Python dependency to `_torchtext.so` library which will link against a `libtorchtext.so` library containing source files with pure C++ dependencies #1660
- [x] Create a simple tutorial showcasing how to depend on `libtorchtext.so` within a simple C++ application #1817

PR #1673, opened 2022-04-01
[CMake 2/3] Add CMake Build to torchtext to create single `_torchtext` library

Nayef211:
**Reference Issue #1644**

## Description
- Updating torchtext extension to use CMake instead of PyTorch's `CppExtension` module
- This is the second part of #1660. We create a single `_torchtext.so` library with all source files from csrc/ folder. This is similar to the build artifacts created by the current build process.

## Build Process
**Building the torctext project using `setup.py`**
```shell
python setup.py develop
```

## Misc
closes https://github.com/pytorch/text/pull/1225


torchtext/__init__.py
```diff
@@ -1,11 +1,13 @@
 import os
+from ._extension import _init_extension
+
+_init_extension()
```
mthrok, line 4: Please check the compatibility of this initialization update on fbcode.

Also I recommend to call this function in `_extension.py` just for the sake of the simplification of `__init__.py`. (delegating the reponsibility to `_extension.py`, which yields a better locality.)

Still, the `_extension` module has to be imported first in `__init__.py`.

cc @parmeet

build_tools/setup_helpers/extension.py
```diff
@@ -1,189 +1,120 @@
...
+            f"-DCMAKE_BUILD_TYPE={cfg}",
+            f"-DCMAKE_PREFIX_PATH={torch.utils.cmake_prefix_path}",
+            f"-DCMAKE_INSTALL_PREFIX={extdir}",
+            "-DCMAKE_VERBOSE_MAKEFILE=ON",
+            f"-DPython_INCLUDE_DIR={distutils.sysconfig.get_python_inc()}",
+            "-DBUILD_TORCHTEXT_PYTHON_EXTENSION:BOOL=ON",
+            "-DRE2_BUILD_TESTING:BOOL=OFF",
+            "-DBUILD_TESTING:BOOL=OFF" "-DBUILD_SHARED_LIBS=OFF",
+            "-DCMAKE_POLICY_DEFAULT_CMP0063=NEW",
+            "-DCMAKE_CXX_FLAGS=" + _get_cxx11_abi(),
```
mthrok, line 82: This is supposed to be handled by TORCH_CXX_FLAGS.

However, https://github.com/pytorch/audio/issues/2287 suggests this does not work with manually built PyTorch, so I will follow up with this one.

build_tools/setup_helpers/extension.py
```diff
@@ -1,189 +1,120 @@
...
+
+        cmake_args = [
+            f"-DCMAKE_BUILD_TYPE={cfg}",
+            f"-DCMAKE_PREFIX_PATH={torch.utils.cmake_prefix_path}",
+            f"-DCMAKE_INSTALL_PREFIX={extdir}",
+            "-DCMAKE_VERBOSE_MAKEFILE=ON",
+            f"-DPython_INCLUDE_DIR={distutils.sysconfig.get_python_inc()}",
+            "-DBUILD_TORCHTEXT_PYTHON_EXTENSION:BOOL=ON",
+            "-DRE2_BUILD_TESTING:BOOL=OFF",
+            "-DBUILD_TESTING:BOOL=OFF" "-DBUILD_SHARED_LIBS=OFF",
```
mthrok, line 80:
```suggestion
            "-DBUILD_TESTING:BOOL=OFF", "-DBUILD_SHARED_LIBS=OFF",
```

Nayef211: Oops thanks for catching this!

build_tools/setup_helpers/extension.py
Nayef211: Gotcha, I will keep this in for now until the issue is resolved then.

torchtext/__init__.py
Nayef211: Updated the function call to reside within `_extension.py`. Will import this entire PR and test on fbcode before landing once we get all GH CI tests passing

torchtext/_extension.py
```diff
@@ -1,16 +1,66 @@
...
+    if not path.exists():
+        return False
+    torch.ops.load_library(path)
+    torch.classes.load_library(path)
+    return True
+
+
+def _init_extension():
+    if not _mod_utils.is_module_available("torchtext._torchtext"):
+        warnings.warn("torchtext C++ extension is not available.")
```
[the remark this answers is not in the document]
Nayef211: Changed this back to raise an `Exception` instead of a `warning`

.circleci/unittest/linux/scripts/setup_env.sh
```diff
@@ -36,7 +36,7 @@ conda activate "${env_dir}"
 
 
 # 3. Install minimal build tools
-pip --quiet install cmake ninja
```
mthrok, line 39: Maybe we need an equivalent change in windows setup script.

It seems like an old version of CMake can cause dynamic link issue

https://stackoverflow.com/a/37146557

build_tools/setup_helpers/extension.py
```diff
@@ -40,150 +57,97 @@ def _get_ela(debug):
...
-        extra_args = ["-GNinja"]
-        build_env.setdefault("CC", "cl")
-        build_env.setdefault("CXX", "cl")
-    else:
-        extra_args = []
-    subprocess.run(
-        args=[
-            "cmake",
+            "-DCMAKE_CXX_FLAGS=" + _get_cxx11_abi() + " " + _get_eca(cfg == "Debug"),
+            "-DCMAKE_EXE_LINKER_FLAGS=" + _get_ela(cfg == "Debug"),
```
mthrok, line 114: This affects flags of all the compilation.

In modern CMake it is recommended to approach this in more OOP-like fashion, using `target`s.

Once we verify this works, then we should move these dedicated location.

build_tools/setup_helpers/extension.py
```diff
@@ -1,189 +1,122 @@
...
-_ROOT_DIR = Path(__file__).parent.parent.parent.resolve()
-_CSRC_DIR = _ROOT_DIR / "torchtext" / "csrc"
-_TP_BASE_DIR = _ROOT_DIR / "third_party"
-_TP_INSTALL_DIR = _TP_BASE_DIR / "build"
-
-
-def _get_eca(debug):
-    eca = []
-    if platform.system() == "Windows":
-        eca += ["/MT"]
```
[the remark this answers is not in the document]
Nayef211: Re-added these flags to the compile commands

torchtext/csrc/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+  set(LIBTORCHTEXT_COMPILE_DEFINITIONS)
+
+  # See https://github.com/pytorch/pytorch/issues/38122
+  find_library(TORCH_PYTHON_LIBRARY torch_python PATHS "${TORCH_INSTALL_PREFIX}/lib")
+  if (WIN32)
+    find_package(Python3 ${PYTHON_VERSION} EXACT COMPONENTS Development)
+    set(ADDITIONAL_ITEMS Python3::Python)
+  endif()
+  function(define_extension name sources include_dirs link_libraries definitions)
+    add_library(${name} SHARED ${sources})
```
mthrok, line 14: Okay, so probably the correct way is to set this to `STATIC` when in windows.

```
if (MSVC)
    add_library(${name} STATIC ${sources})
else (MSVC)
    add_library(${name} SHARED ${sources})
endif()
```

.circleci/unittest/linux/scripts/install.sh
```diff
@@ -7,11 +7,27 @@ unset PYTORCH_VERSION
...
+(
+    if [ "${os}" == MacOSX ] ; then
+      # TODO: this can be removed as soon as linking issue could be resolved
+      #  see https://github.com/pytorch/pytorch/issues/62424 from details
+      MKL_CONSTRAINT='mkl==2021.2.0'
+    else
+      MKL_CONSTRAINT=''
+    fi
+    set -x
+    conda install -y -c "pytorch-${UPLOAD_CHANNEL}" "${CONDA_CHANNEL_FLAGS}" "$MKL_CONSTRAINT" pytorch cpuonly
```
mthrok, line 28: `CONDA_CHANNEL_FLAGS` and `MKL_CONSTANT` can be empty based on platforms, so the use of double-quotations here cause problem. The `command` will see it as option of empty string and it causes weird behavior.
```suggestion
    conda install -y -c "pytorch-${UPLOAD_CHANNEL}" ${CONDA_CHANNEL_FLAGS} ${MKL_CONSTRAINT} pytorch cpuonly
```

mthrok: BTW why does it have `cpuonly`? It does not hurt but it's not mandatory, is it? TorchText should build fine with non-CPU version of PyTorch (at least CUDA).

Nayef211: > CONDA_CHANNEL_FLAGS and MKL_CONSTANT can be empty based on platforms, so the use of double-quotations here cause problem. The command will see it as option of empty string and it causes weird behavior.

Gotcha will change this back to remove the double quotes

> BTW why does it have cpuonly? It does not hurt but it's not mandatory, is it? TorchText should build fine with non-CPU version of PyTorch (at least CUDA).

Not sure what the historical context was for this. But also torchtext doesn't need the CUDA version of PyTorch to run. So does it make sense to install the CUDA version when it's not necessary for our library? In addition the CUDA version of PyTorch is a larger than the CPU only version.

CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+  # RPATH stuff
+  set(CMAKE_MACOSX_RPATH ON)
+
+  set(CMAKE_SHARED_LIBRARY_SUFFIX ".so")
+endif()
+
+# Options
+option(BUILD_TORCHTEXT_PYTHON_EXTENSION "Build Python extension" OFF)
+
+set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH};${CMAKE_CURRENT_SOURCE_DIR}/cmake")
```
mthrok, line 52: note: Technically, this one is not necessary as there is no `cmake` directory which houses custom CMake scripts.
But it does not hurt to keep this as when such thing is added, it's customary to use `cmake` as directory name.

torchtext/__init__.py
```diff
@@ -1,21 +1,15 @@
 import os
 
+from torchtext import _extension  # noqa: F401
```
mthrok, line 3: Can you add a note that this is done for side-effect and it has to happen before anything?

torchtext/__init__.py
```diff
@@ -1,21 +1,15 @@
 import os
```
mthrok, line 1: off-topic, but it looks strange that the top-level module initialization script, which should be only doing initializations require os and do some stuff. These should be done in a dedicated submodule and only imported here.

torchtext/_extension.py
```diff
@@ -1,16 +1,66 @@
...
+            The expected case is `OSError` thrown by `ctype.DLL` when a dynamic dependency
+            is not found.
+            This behavior was chosen because the expected failure case is not recoverable.
+            If a dependency is missing, then users have to install it.
+    """
+    path = _get_lib_path(lib)
+    if not path.exists():
+        return False
+    torch.ops.load_library(path)
+    torch.classes.load_library(path)
```
mthrok, line 51: FYI: The implementation beneath these two different `load_library` are common, so only one of them are required.

torchtext/csrc/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+      TARGETS ${name}
+      LIBRARY DESTINATION .
+      RUNTIME DESTINATION .  # For Windows
+      )
+  endfunction()
+
+  set(
+    EXTENSION_SOURCES
+    clip_tokenizer.cpp
+    clip_tokenizer.h
```
mthrok, line 43: since all the headers are now looked up from the root directory of the repo, so I think it's less confusing not to have headers listed here.

mthrok (changes requested):
Overall looks good. Thanks for making this work!

build_tools/setup_helpers/extension.py
Nayef211: Updated to use `target`s. Thanks for the feedback here and I'm glad we figured out that switching to this approach resolved the Windows errors.

torchtext/__init__.py
Nayef211: Done

torchtext/__init__.py
Nayef211: Let me do this in a followup PR. Maybe we can define all global constants in a `_constants.py` file. Wdyt @parmeet?

torchtext/_extension.py
Nayef211: Hmm so I noticed audio actually makes a call to both. Any specific reason why it's done that way in torchaudio?

https://github.com/pytorch/audio/blob/main/torchaudio/_extension.py#L51-L52

torchtext/csrc/CMakeLists.txt
Nayef211: Done!

.circleci/unittest/linux/scripts/install.sh
```diff
@@ -7,11 +7,27 @@ unset PYTORCH_VERSION
...
+(
+    if [ "${os}" == MacOSX ] ; then
+      # TODO: this can be removed as soon as linking issue could be resolved
+      #  see https://github.com/pytorch/pytorch/issues/62424 from details
+      MKL_CONSTRAINT='mkl==2021.2.0'
+    else
+      MKL_CONSTRAINT=''
+    fi
+    set -x
+    conda install -y -c "pytorch-${UPLOAD_CHANNEL}" ${CONDA_CHANNEL_FLAGS} $MKL_CONSTRAINT pytorch cpuonly
```
mthrok, line 28:
```suggestion
    conda install -y -c "pytorch-${UPLOAD_CHANNEL}" ${CONDA_CHANNEL_FLAGS} ${MKL_CONSTRAINT} pytorch cpuonly
```

torchtext/_extension.py
```diff
@@ -1,16 +1,66 @@
-def _init_extension():
-    import importlib
-    import os
+import os
+from pathlib import Path
+
+import torch
+from torchtext._internal import module_utils as _mod_utils  # noqa: F401
```
mthrok, line 5: Is this `F401` necessary??

torchtext/_extension.py
mthrok: Yeah, I learned this fact myself only after I wrote this code. I have not updated audio yet.

CMakeLists.txt
Nayef211: Gotcha will keep that in mind. What sort of scripts would we put in a `cmake/` folder that wouldn't go in this CMake file?

mthrok (approved):
I thought I put "Approved" in the previous review, but somehow it was "Request changes".

The change looks good, this can be merged when all the builds / unit tests pass.

torchtext/_extension.py
Nayef211: Okay let me go with `torch.ops.load_library` since that seems to be the approach vision also uses https://github.com/pytorch/vision/blob/9f48c6274f66f71dd0cabe6a48bde84c8f239abe/torchvision/extension.py#L20

torchtext/_extension.py
Nayef211: You're right it's not since `_mod_utils` is used in the same file.

Nayef211:
@parmeet I'm seeing some test failures on this PR that have nothing to do with my changes. Do you know if it would be fine to land this?

PR #1684, opened 2022-04-18
Add benchmark for roberta prepoc pipelines

parmeet:
## Summary 
This PR adds benchmark for end-2-end preproc pipeline for RoBERTa based text pre-processing. All the pipelines are based on DataPipes.
* **Pipeline 1 (benchmark_roberta_datapipe):** Using operators/transforms implemented in torchtext and are directly applied on text
* **Pipeline 2 (benchmark_roberta_dataframe):** This pipeline first convert the input text (python tuples) into DataFrame of sizes batch_size on the fly. It then does text processing directly on dataframe.
  * **Pipeline 2.2 (Using Ops as UDFs):** This pipeline uses operators implemented in torchtext and are applied to TA dataframe as a UDFs using it's transform API  
  * **Pipeline 2.1 (Using Native Ops):** This pipeline used native operators implemented directly in TorchArrow 

## Results
Current end-2-end benchmarking results:
Settings: Macbook pro, SST2 dataset, batch_size 32, single run

Pipeline 1: ~ 5 seconds
Pipeline 2.1: ~28 seconds
Pipeline 2.2: ~73 seconds

### DIFF
--- a/benchmark/benchmark_roberta_pipeline.py
+++ b/benchmark/benchmark_roberta_pipeline.py
@@ -1,0 +1,145 @@
+from argparse import ArgumentParser
+from functools import partial
+from typing import Dict, Any
+
+import torcharrow as ta
+import torcharrow.dtypes as dt
+import torcharrow.pytorch as tap
+import torchtext.functional as F
+import torchtext.transforms as T
+from benchmark.utils import Timer
+from torch.hub import load_state_dict_from_url
+from torch.nn import Module
+from torch.utils.data import DataLoader
+from torchtext.datasets import DATASETS
+
+
+class RobertaTransformDataPipe(Module):
+    def __init__(self) -> None:
+        super().__init__()
+        # Instantiate various transforms
+
+        # Tokenizer to split input text into tokens
+        encoder_json_path = "https://download.pytorch.org/models/text/gpt2_bpe_encoder.json"
+        vocab_bpe_path = "https://download.pytorch.org/models/text/gpt2_bpe_vocab.bpe"
+        self.tokenizer = T.GPT2BPETokenizer(encoder_json_path, vocab_bpe_path)
+
+        # vocabulary converting tokens to IDs
+        vocab_path = "https://download.pytorch.org/models/text/roberta.vocab.pt"
+        self.vocab = T.VocabTransform(load_state_dict_from_url(vocab_path))
+
+        # Add BOS token to the beginning of sentence
+        self.add_bos = T.AddToken(token=0, begin=True)
+
+        # Add EOS token to the end of sentence
+        self.add_eos = T.AddToken(token=2, begin=False)
+
+    def forward(self, input: Dict[str, Any]) -> Dict[str, Any]:
+        tokens = self.tokenizer(input["text"])
+        tokens = F.truncate(tokens, max_seq_len=254)
+        tokens = self.vocab(tokens)
+        tokens = self.add_bos(tokens)
+        tokens = self.add_eos(tokens)
+        input["tokens"] = tokens
+        return input
+
+
+class RobertaTransformDataFrame(Module):
+    def __init__(self) -> None:
+        super().__init__()
+        # Instantiate various transforms
+
+        # Tokenizer to split input text into tokens
+        encoder_json_path = "https://download.pytorch.org/models/text/gpt2_bpe_encoder.json"
+        vocab_bpe_path = "https://download.pytorch.org/models/text/gpt2_bpe_vocab.bpe"
+        self.tokenizer = T.GPT2BPETokenizer(encoder_json_path, vocab_bpe_path)
+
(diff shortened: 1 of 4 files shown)


benchmark/benchmark_roberta_pipeline.py
```diff
@@ -0,0 +1,149 @@
...
+        tokens = self.add_eos(tokens)
+        input["tokens"] = tokens
+        return input
+
+
+def batch_tokenize(input: List[str], tokenizer: Callable) -> List[List[str]]:
+    return tokenizer(input)
+
+
+class RobertaTransformDataFrame(Module):
```
Nayef211, line 51: Same question as above for `RobertaTransformDataFrame`

benchmark/benchmark_roberta_pipeline.py
```diff
@@ -0,0 +1,149 @@
...
+
+        # Apply transformation on DataFrame
+        train_dp = train_dp.map(transform)
+
+        # Remove not required columns
+        train_dp = train_dp.map(lambda x: x.drop(["text"]))
+
+        # convert DataFrame to tensor (This will yeild named tuple)
+        train_dp = train_dp.map(lambda x: x.to_tensor({"tokens": tap.PadSequence(padding_value=1)}))
+
```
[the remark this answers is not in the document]
parmeet: hmm,  not really. I think the only common code is where we map the transform, right?

benchmark/benchmark_roberta_pipeline.py
```diff
@@ -0,0 +1,149 @@
...
+import torchtext.functional as F
+import torchtext.transforms as T
+from benchmark.utils import Timer
+from torch.hub import load_state_dict_from_url
+from torch.nn import Module
+from torch.utils.data import DataLoader
+from torchtext.datasets import DATASETS
+
+
+class RobertaTransformDataPipe(Module):
```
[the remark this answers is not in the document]
parmeet: I think it's a good point. Let me see if there is an elegant way to reuse the code.

benchmark/benchmark_roberta_pipeline.py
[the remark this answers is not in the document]
Nayef211: Actually taking another look the similarities are minor and I guess there's not really a point in overcomplicating the simple benchmarking pipeline. Leaving this as is sounds good to me!

benchmark/benchmark_roberta_pipeline.py
```diff
@@ -0,0 +1,170 @@
...
+        self.add_bos = T.AddToken(token=0, begin=True)
+
+        # Add EOS token to the end of sentence
+        self.add_eos = T.AddToken(token=2, begin=False)
+
+    def forward(self, input: ta.DataFrame) -> ta.DataFrame:
+        input["tokens"] = ta_F.bpe_tokenize(self.tokenizer, input["text"])
+        input["tokens"] = input["tokens"].list.slice(stop=254)
+        token_ids = ta_F.lookup_indices(self.vocab, input["tokens"])
+        token_ids._dtype = dt.List(dt.int64)
```
wenleix, line 100: this hack should be no longer needed once https://github.com/facebookresearch/torcharrow/pull/361 merged.

parmeet: Sounds good! Thanks @wenleix for the fix :)

erip: Note that torcharrow doesn't support even "relatively old" hardware (i.e., https://github.com/facebookresearch/torcharrow/issues/161) which may leak make these features unavailable to OS X users. If this becomes a hard dep on torchtext, some users may not even be able to use torchtext anymore (me included!)

wenleix: @erip : Thanks for raising the concern. Could you try it again? -- we should have recently fixed it issue and TorchArrow should be able to run on SSE-only machines (e.g. https://github.com/facebookresearch/torcharrow/pull/315)

benchmark/benchmark_roberta_pipeline.py
```diff
@@ -0,0 +1,87 @@
...
+        # TODO: Figure out how to create DataFrame of larger size and create smaller batches
+        train_dp = train_dp.dataframe(columns=columns, dataframe_size=batch_size)
+
+        # Apply transformation on DataFrame
+        train_dp = train_dp.map(transform)
+
+        # Remove not required columns
+        train_dp = train_dp.map(lambda x: x.drop(["text"]))
+
+        # convert DataFrame to tensor (This will yeild named tuple)
```
Nayef211, line 73: nit: "yield"

Nayef211 (approved):
LGTM! Thanks @parmeet for updating the benchmarks to reuse code from our previously implemented examples. These benchmarks will serve as a great starting point to understanding where the performance bottlenecks might be for our torcharrow integration work.

PR #1687, opened 2022-04-25
Fix smoke tests for linux

atalman:
This PR fixes the smoke tests for torchtext and python 3.9 and 3.10 We will need to rebuild docker image for this.
This is same as [PR audio](https://github.com/pytorch/audio/pull/2348)

### DIFF
--- a/.circleci/smoke_test/docker/Dockerfile
+++ b/.circleci/smoke_test/docker/Dockerfile
@@ -1,13 +1,7 @@
 # this Dockerfile is for torchtext smoke test, it will be created periodically via CI system
 # if you need to do it locally, follow below steps once you have Docker installed
-# assuming you're within the directory where this Dockerfile located
-#  $ docker build . -t torchtext/smoketest
-
-# if you want to push to aws ecr, make sure you have the rights to write to ECR, then run
-# $ eval $(aws ecr get-login --region us-east-1 --no-include-email)
-# $ export MYTAG=localbuild  ## you can choose whatever tag you like
-# $ docker tag torchtext/smoketest 308535385114.dkr.ecr.us-east-1.amazonaws.com/torchtext/smoke_test:${MYTAG}
-# $ docker push  308535385114.dkr.ecr.us-east-1.amazonaws.com/torchtext/smoke_test:${MYTAG}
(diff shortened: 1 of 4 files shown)


.circleci/smoke_test/docker/Dockerfile
```diff
@@ -25,12 +26,16 @@ RUN apt-get -qq update && apt-get -qq -y install curl bzip2 sox libsox-dev libso
...
+RUN conda create -y --name python3.9 python=3.9
+RUN conda create -y --name python3.10 python=3.10
+
 SHELL [ "/bin/bash", "-c" ]
 RUN echo "source /usr/local/etc/profile.d/conda.sh" >> ~/.bashrc
-RUN source /usr/local/etc/profile.d/conda.sh && conda activate python3.6 && conda install -y -c conda-forge sox && conda install -y numpy
 RUN source /usr/local/etc/profile.d/conda.sh && conda activate python3.7 && conda install -y -c conda-forge sox && conda install -y numpy
 RUN source /usr/local/etc/profile.d/conda.sh && conda activate python3.8 && conda install -y -c conda-forge sox && conda install -y numpy
+RUN source /usr/local/etc/profile.d/conda.sh && conda activate python3.9 && conda install -y -c conda-forge sox && conda install -y numpy
+RUN source /usr/local/etc/profile.d/conda.sh && conda activate python3.10 && conda install -y -c conda-forge sox && conda install -y numpy
```
mthrok, line 40: I think this is based off of torchaudio's Dockerfile and `sox` is not required in text. (technically it's no longer required in audio as well).

.circleci/smoke_test/docker/build_and_push.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
```
mthrok, line 2: Par our internal conversation we should probably move the comment in Dockerfile here and update it so that it does not mention AWS ECR but Docker hub.

mthrok (approved):

parmeet (approved):
LGTM! the test failures are unrelated, feel free to land after addressing the comments :)

PR #1688, opened 2022-04-26
fix pad transform test

parmeet:
This PR fixes the PAD transform test in scripting mode.

The [torch.nn.function.pad](https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html) function expect float value, and when providing with int,  it complains in script mode.


VirgileHlav (approved):
LGTM, thanks!

ebsmothers (approved):
Thanks for the fix!

Issue #532, opened 2019-05-03
Dataset.split with stratified=True is not reproducible


DianeBouchacourt:
When using TabularDataset.split with stratified = True, the sampling is not reproducible. After inspection, it looks like it is because of this line:
https://github.com/pytorch/text/blob/499e327ea53bdf67c648f5747ed26764283b968a/torchtext/data/dataset.py#L318

Indeed, the set of unique_strata returned is not always in the same order. Therefore, even though the samples drawn in rationed_split are the same, the different group in strata will be looked at in a different order :
https://github.com/pytorch/text/blob/499e327ea53bdf67c648f5747ed26764283b968a/torchtext/data/dataset.py#L329

A simple fix could be to add in stratify() the following line:
unique_strata = sorted(unique_strata)

mttk:
This sounds reasonable, could you send a PR?

Nayef211:
Closing the issue as we no longer maintain the `TabularDataset` in our library.

PR #1689, opened 2022-04-27
Fix Dataset.split to be always reproducible (pytorch#532)

sergioburdisso:
When `stratified=True` the `Dataset.split` was not reproducible due to
the inherently randomness of using the elements of a set in order (since
a set is unordered by definition). In our case the order of the output
elements (`list(strata_maps.values())`) depended on the order of the
items in `unique_strata` which was populated using a `set` object.

Resolves: https://github.com/pytorch/text/issues/532


Nayef211:
Hi @sergioburdisso. It looks like your repo is very outdates. Can you pull and merge in the latest changes from the main branch into your PR? Also I'm not sure if the feature you are trying to add is relevant to our current implementation of `Datasets`.

sergioburdisso:
Hi @Nayef211! firs of all thanks for your feedback on this. Yeah, I know that this fix is not relevant for the current implementation of `Datasets` since "legacy" has been removed in version `0.12.x`. However, for people that still have old projects or for whatever reason they're still using version `0.11.x` this fix is useful and in my case also it was necessary. That's why I applied this quick fix in the `11` branch, so that it could be hopefully included in the next `0.11.x` version. Was this the right branch? I'm not 100% sure though.

Nayef211:
> Hi @Nayef211! firs of all thanks for your feedback on this. Yeah, I know that this fix is not relevant for the current implementation of `Datasets` since "legacy" has been removed in version `0.12.x`. However, for people that still have old projects or for whatever reason they're still using version `0.11.x` this fix is useful and in my case also it was necessary. That's why I applied this quick fix in the `11` branch, so that it could be hopefully included in the next `0.11.x` version. Was this the right branch? I'm not 100% sure though.

Gotcha thanks for clarifying. In this case I would recommend keeping the changes on your fork. Since we are currently on torchtext `0.12.0` where `legacy` has been deleted, we will no longer be providing support for any fixes within the `legacy` folder. I don't think we plan to have any future `0.11.x` release to fix bugs from the initial `0.11` release.

@parmeet @mthrok could you confirm if what I'm saying makes sense?

parmeet:
> I don't think we plan to have any future `0.11.x` release to fix bugs from the initial `0.11` release.

That's right @Nayef211 . I don't think we plan to have dot releases for previous major release versions.

VirgileHlav:
Given the comments, closing this PR.

PR #1690, opened 2022-04-27
BetterTransformer support for torchtext

zrphercule:
Summary:
This diff created a fast path of using better transformer (torch.nn.TransformerEncoderLayer), with a converter from the existing torchtext transformer encoder layer to better transforomer.
The related tests are added in the following diff.

Differential Revision: D35948440


Nayef211:
Let's close this since https://github.com/pytorch/text/pull/1694 seems to fix the linting issues with this PR

PR #1693, opened 2022-04-29
Remove unneeded modules after using nn.Module for BetterTransformer

mikekgfb:
Summary: Remove unneeded modules after using nn.Module for BetterTransformer

Differential Revision: D36038830


Nayef211:
Closing this since the changes will be landed in https://github.com/pytorch/text/pull/1696

PR #1694, opened 2022-05-01
[fbsync] BetterTransformer support for torchtext (#1690)

parmeet:
Summary:
Pull Request resolved: https://github.com/pytorch/text/pull/1690

This diff created a fast path of using better transformer (torch.nn.TransformerEncoderLayer), with a converter from the existing torchtext transformer encoder layer to better transforomer.
The related tests are added in the following diff.

Reviewed By: parmeet

Differential Revision: D35948440

fbshipit-source-id: e69e12f2dd28edfea3176a10ee3d7d321d50c897


Nayef211 (approved):
LGTM. Test failures are not related

PR #1697, opened 2022-05-02
Model benchmark

parmeet:
Adding RoBERTa/XLM-R model benchmarks:
* Model forward
* Model forward/backward

As we made changes in the underlying model implementation, it is good to have the benchmark to track the performance gains/gaps. For instance, we recently migrated to BetterTransformer implementation and this benchmark can be used to assess the performance gains.

### DIFF
--- a/benchmark/benchmark_roberta_model.py
+++ b/benchmark/benchmark_roberta_model.py
@@ -1,9 +1,16 @@
 from argparse import ArgumentParser
 
 import torch
-import torchtext
 from benchmark.utils import Timer
 from torchtext.functional import to_tensor
+from torchtext.models import XLMR_BASE_ENCODER, XLMR_LARGE_ENCODER, ROBERTA_BASE_ENCODER, ROBERTA_LARGE_ENCODER
+
+ENCODERS = {
+    "xlmr_base": XLMR_BASE_ENCODER,
+    "xlmr_large": XLMR_LARGE_ENCODER,
+    "roberta_base": ROBERTA_BASE_ENCODER,
+    "roberta_large": ROBERTA_LARGE_ENCODER,
+}
 
 
 def basic_model_input(encoder):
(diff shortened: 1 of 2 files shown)


benchmark/benchmark_roberta_model.py
```diff
@@ -0,0 +1,64 @@
...
+    model_input = args.model_input
+
+    if encoder == "xlmr_base":
+        encoder = torchtext.models.XLMR_BASE_ENCODER
+    elif encoder == "xlmr_large":
+        encoder = torchtext.models.XLMR_LARGE_ENCODER
+    elif encoder == "roberta_base":
+        encoder = torchtext.models.ROBERTA_BASE_ENCODER
+    elif encoder == "roberta_large":
+        encoder = torchtext.models.ROBERTA_LARGE_ENCODER
```
VirgileHlav, line 34: Add a `raise` in case the string `encoder` is not recognized, otherwise line 36 will fail.

benchmark/benchmark_roberta_model.py
```diff
@@ -0,0 +1,64 @@
...
+    model = encoder.get_model()
+    if model_input == "basic":
+        model_input = basic_model_input(encoder)
+    else:
+        raise NotImplementedError("Given model input type is not implemented")
+
+    model.eval()
+
+    for _ in range(warmup_passes):
+        _ = model(model_input)
```
VirgileHlav, line 45: maybe just `model(model_input)`? same below

benchmark/utils.py
```diff
@@ -0,0 +1,28 @@
+import time
+
+
+class Timer:
```
VirgileHlav, line 4: Add top level summary of class for clarity?

benchmark/benchmark_roberta_model.py
```diff
@@ -0,0 +1,64 @@
...
+    model.zero_grad()
+
+
+def run(args):
+    encoder = args.encoder
+    num_passes = args.num_passes
+    warmup_passes = args.num_passes
+    model_input = args.model_input
+
+    if encoder == "xlmr_base":
```
VirgileHlav, line 27: Maybe use a dict and call `encoder = mapping_to_models[encoder]` to disentangle logic and correspondence between strings and models?

benchmark/utils.py
```diff
@@ -0,0 +1,28 @@
...
+
+        self._start = time.perf_counter()
+
+    def stop(self):
+        if self._start is None:
+            raise Exception("Timer is not running. Call .start() to start the timer.")
+
+        elapsed = time.perf_counter() - self._start
+
+        print(self._text + " ... Total running time: {}".format(elapsed))
```
VirgileHlav, line 21: Use a single string expression either `"{}".format(x)` or `f"{x}"`

benchmark/benchmark_roberta_model.py
```diff
@@ -0,0 +1,64 @@
...
+        for _ in range(num_passes):
+            _train(model, model_input)
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser()
+    parser.add_argument("--encoder", default="xlmr_base", type=str)
+    parser.add_argument("--num-passes", default=50, type=int)
+    parser.add_argument("--warmup-passes", default=10, type=int)
+    parser.add_argument("--model-input", default="basic", type=str)
```
VirgileHlav, line 63: Is there a reason `model-input` only has one accepted value? Otherwise we can remove it for now.

benchmark/utils.py
```diff
@@ -0,0 +1,28 @@
+import time
+
+
+class Timer:
+    def __init__(self, text="") -> None:
```
VirgileHlav, line 5: If not using `typing` in the file, ` -> None` is not required here.

VirgileHlav (commented):
Thanks @parmeet! I do not have the context behind this PR, but I added mostly python related suggestions.

benchmark/benchmark_roberta_model.py
parmeet: for now, I just created a simple input. But it is likely that someone want to implement and try to for complex input cases.

Nayef211 (approved):
Thanks for adding these benchmarks @parmeet. Overall LGTM. Do you think we can add some more context in the description to motivate the benchmarks? I assume this will be relevant to our TorchArrow integration efforts?

parmeet:
> Do you think we can add some more context in the description to motivate the benchmarks? 

sure will do!

> I assume this will be relevant to our TorchArrow integration efforts?

hmm, may be. This one specifically benchmarks the performance of model without taking pre-proc into account. May be it can be further extended to do training on real dataset with datapipes and TA based pre-processing...

PR #1698, opened 2022-05-02
Provide option to return splited tokens

parmeet:
# Issue:
The current GPT2BPE and CLIP Tokenizer transforms returns the encoded token IDs. Based on the user feedback, we learnt that for consistency, it would good to expose the option to return tokens directly instead of their encoded IDs. 

# Resolution:
This PR addresses the issue by adding  boolean option `return_tokens` in Tokenizer Transform initialization API that would let users to either return Tokens or their corresponding Token IDs (encoded as strings).

# Test:
pytest -v -s test/test_transforms.py::TestGPT2BPETokenizer
pytest -v -s test/test_transforms.py::TestCLIPTokenizer

### DIFF
--- a/torchtext/transforms.py
+++ b/torchtext/transforms.py
@@ -306,8 +306,8 @@ class GPT2BPETokenizer(Module):
         return isinstance(self.bpe, torch._C.ScriptObject)
 
     @torch.jit.export
-    def _tokenize(self, text: str) -> List[str]:
-        """Encode text into a list of tokens
+    def _encode(self, text: str) -> List[str]:
+        """Encode text into a list of tokens IDs
 
         Args:
             text: An input text string.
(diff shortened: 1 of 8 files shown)


torchtext/transforms.py
```diff
@@ -447,20 +470,44 @@ def _tokenize(self, text: str) -> List[str]:
...
+        Returns:
+            A list of bpe token ids represents each bpe tokens
+
+        For example: "awesome,awe"
+            --> bpe --> bpe tokens: ["aw", "esome"], [","], ["aw", "e"]
+        """
+        text = text.lower().strip()
+        return self.bpe.tokenize(text)
+
+    def forward(self, input: Any, return_tokens: bool = False) -> Any:
```
Nayef211, line 489: Same question as above

torchtext/transforms.py
```diff
@@ -327,20 +327,43 @@ def _tokenize(self, text: str) -> List[str]:
...
+
+        Returns:
+            A list of bpe token ids represents each bpe tokens
+
+        For example: "awesome,awe"
+            --> bpe --> bpe tokens: ["aw", "esome"], [","], ["aw", e]
+        """
+        return self.bpe.tokenize(text)
+
+    def forward(self, input: Any, return_tokens: bool = False) -> Any:
```
Nayef211, line 345: Just wondering if you think it makes more sense to pass this `return_tokens` flag into the `GPT2BPETokenizer` constructor instead? I would think that after creating a instance of this tokenizer, the user will consistently want to retrieve token token IDs or tokens instead of switching between the two options.

parmeet: hmm, it's a good point. That way we can keep the forward API cleaner as well :)

Nayef211 (approved):
LGTM. Before landing, can we update the PR description to state that the resolution is to pass a boolean flag in the constructor rather than the forward method. Can we also add the command used to test the PR changes?

Issue #1710, opened 2022-05-11
Add support for all datasets of the GLUE benchmark


VirgileHlav:
## 🚀 Feature

Add support for all 8 remaining datasets (SST-2 is already supported) of the [GLUE benchmark](https://gluebenchmark.com/): CoLA, MRPC, QQP, STS-B, MNLI, QNLI, RTE, WNLI.

**Motivation**

In itself adding support for all GLUE datasets has a lot of value: GLUE is one of the most widely used benchmark in the NLP community. Furthermore, our planned effort to develop a cohesive API for Multi-Task Learning requires us to enhance our suite our datasets, starting with GLUE.

**Additional context**

We have already created a streamlined dataset API, with a consistent use of DataPipes for dataset download and load operations, as well as a testing methodology relying on mock data (see #1493). This feature will only require to add support for these 8 datasets following that methodology.

## Datasets
 - [x] CoLA #1711
 - [x] MRPC #1712
 - [x] QQP #1713
 - [x] STS-B #1714
 - [x] MNLI #1715
 - [x] QNLI #1717
 - [x] RTE #1721
 - [x] WNLI #1724

parmeet:
Thanks @VirgileHlav  for creating this issue! Just a general comment regarding dataset implementation: let's avoid usage of lambda functions, please refer issue https://github.com/pytorch/text/issues/1716 for details.

Nayef211:
Thanks for creating this issue to keep track of the new datasets we are adding! One suggestion from my end would be to update our documentation for each new dataset we add [here](https://github.com/pytorch/text/blob/main/docs/source/datasets.rst).

parmeet:
Let's also make sure we add shuffle and sharding filters: Reference Issue: https://github.com/pytorch/text/issues/1727. Reference PR #1729

VirgileHlav:
Solved lint issues and added documentation + shuffle/sharding on all PRs.

PR #1714, opened 2022-05-11
Add support for STS-B dataset with unit tests

VirgileHlav:
## Summary
- Added support for STS-B dataset
- Added mocked unit tests for STS-B dataset 

## Test
- ` pytest test/datasets/test_stsb.py`

## Context
See #1710


torchtext/datasets/stsb.py
```diff
@@ -0,0 +1,94 @@
...
+        return os.path.join(root, os.path.basename(x))
+
+    def _extracted_filepath_fn(_=None):
+        return _filepath_fn(_EXTRACTED_FILES[split])
+
+    def _filter_fn(x):
+        return _EXTRACTED_FILES[split] in x[0]
+
+    def _modify_res(x):
+        return (int(x[3]), float(x[4]), x[5], x[6])
```
parmeet, line 74: Why do have float labels?

test/datasets/test_stsb.py
```diff
@@ -0,0 +1,79 @@
...
+    temp_dataset_dir = os.path.join(base_dir, "stsbenchmark")
+    os.makedirs(temp_dataset_dir, exist_ok=True)
+
+    seed = 1
+    mocked_data = defaultdict(list)
+    for file_name, name in zip(["sts-train.csv", "sts-dev.csv" "sts-test.csv"], ["train", "dev", "test"]):
+        txt_file = os.path.join(temp_dataset_dir, file_name)
+        with open(txt_file, "w", encoding="utf-8") as f:
+            for i in range(5):
+                label = seed % 5
```
parmeet, line 27: seem like the label is float, should we create float values here instead?

torchtext/datasets/stsb.py
```diff
@@ -0,0 +1,94 @@
...
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
+        filepath_fn=_extracted_filepath_fn
+    )
+    cache_decompressed_dp = (
+        FileOpener(cache_decompressed_dp, mode="b").read_from_tar().filter(_filter_fn)
+    )
+    cache_decompressed_dp = cache_decompressed_dp.end_caching(mode="wb", same_filepath_fn=True)
+
+    data_dp = FileOpener(cache_decompressed_dp, encoding="utf-8")
+    parsed_data = data_dp.parse_csv(delimiter='\t', quoting=csv.QUOTE_NONE).map(_modify_res)
```
parmeet, line 93: I remember `csv.QUPTE_NONE` was added to resolve some issue with parsing. Do you think we can introduce such strings in testing so that removing this flag would create parsing issue?

torchtext/datasets/stsb.py
VirgileHlav: The labels are a similarity score from 0 to 5, see https://arxiv.org/pdf/1708.00055.pdf.

test/datasets/test_stsb.py
VirgileHlav: Done

torchtext/datasets/stsb.py
VirgileHlav: Good idea, I adjusted the test and verified the new test fails in case we do not pass arg `quoting=csv.QUOTE_NONE`.

parmeet (approved):
LGTM! Thanks for addressing all the comments :)

PR #1717, opened 2022-05-11
Add support for QNLI dataset with unit tests

VirgileHlav:
## Summary
- Added support for QNLI dataset
- Added mocked unit tests for QNLI dataset 

## Test
- ` pytest test/datasets/test_qnli.py`

## Context
See #1710


test/datasets/test_qnli.py
```diff
@@ -0,0 +1,81 @@
...
+    for file_name in ("train.tsv", "dev.tsv", "test.tsv"):
+        txt_file = os.path.join(temp_dataset_dir, file_name)
+        with open(txt_file, "w", encoding="utf-8") as f:
+            f.write("index\tquestion\tsentence\tlabel\n")
+            for i in range(5):
+                label = seed % 2
+                rand_string_1 = get_random_unicode(seed)
+                rand_string_2 = get_random_unicode(seed+1)
+                dataset_line = (label, rand_string_1, rand_string_2)
+                label_str = 'entailment' if label == 1 else 'not_entailment'
```
parmeet, line 32: Per the convention we are returning integer IDs for the label. Wondering why are return labels in here? On related note, we were thinking of adding some helper info utility that shall provide the label mappings from IDs to original names.

VirgileHlav: We need the return label as a string here because this is the format of the input data file, so in order to write the mock data we cannot rely on integer labels. 

Having label mapping sounds like a good idea, we could have this part of the data loader maybe? But I think it should not be part of this PR, but be a wider change in one separate PR.

torchtext/datasets/qnli.py
```diff
@@ -0,0 +1,96 @@
...
+        hash_dict={_filepath_fn(URL): MD5},
+        hash_type="md5",
+    )
+    cache_compressed_dp = HttpReader(cache_compressed_dp).end_caching(mode="wb", same_filepath_fn=True)
+
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
+        filepath_fn=_extracted_filepath_fn
+    )
+    cache_decompressed_dp = (
+        FileOpener(cache_decompressed_dp, mode="b").read_from_zip().filter(_filter_fn)
```
Nayef211, line 90: Can we change this to `load_from_zip`? As per the [warning message in torchdata](https://github.com/pytorch/data/blob/c1d89fe9a1b06e610f32f823359771557b1ca12a/torchdata/datapipes/iter/util/ziparchiveloader.py#L89-L96), `read_from_zip` is deprecated

Nayef211 (approved):
LGTM

torchtext/datasets/qnli.py
```diff
@@ -0,0 +1,96 @@
...
+        return os.path.join(root, os.path.basename(x))
+
+    def _extracted_filepath_fn(_=None):
+        return os.path.join(root, _EXTRACTED_FILES[split])
+
+    def _filter_fn(x):
+        return _EXTRACTED_FILES[split] in x[0]
+
+    def _modify_res(x):
+        return (int(x[3]=="entailment"), x[1], x[2])
```
Nayef211, line 75: Same comment as https://github.com/pytorch/text/pull/1724#discussion_r877338959

torchtext/datasets/qnli.py
VirgileHlav: Thanks for catching this, done.

Issue #1716, opened 2022-05-11
Torchtext datasets failed during multi-processing due to lambda functions usage


parmeet:
Duplicate Issue https://github.com/pytorch/data/issues/397

# Issue
Currently, torchtext datasets make use of lambda functions that creates issue for MultiProcessing with DataLoader as pickle doesn't support serialization with lambda functions. 

Tracking the issue here as well for visibility 

# Suggested Resolution
Remove usage of lambda functions

cc: @NivekT  , @NicolasHug

NicolasHug:
Thanks @parmmet for keeping track of this!

I think there are still instances of the use of `lambdas` in the tutorials:

https://github.com/pytorch/text/blob/dfb53af611a6d6dc0a043a6eb26a1cec30501fc1/examples/tutorials/sst2_classification_non_distributed.py#L98

The DataLoader doesn't use multi-processing for now 

https://github.com/pytorch/text/blob/dfb53af611a6d6dc0a043a6eb26a1cec30501fc1/examples/tutorials/sst2_classification_non_distributed.py#L101

so it shouldn't be a problem in the tutorial script itself.

But in general, users might start tweaking the tutorial and set `num_workers=k`. This might cause problems because of the use of `lambdas`, but also with the RNG and shuffling issues that I raised in https://github.com/pytorch/data/issues/302.

Since these datasets are released already, I was wondering if torchtext had a plan to document these pitfalls before the next release?

parmeet:
Thanks @NicolasHug, sure will update the tutorial. 

Regarding some of the pitfalls, I think it's a good idea to document.  Will look into it and see to create a tracking issue that can be referenced in release notes or something :).

PR #1718, opened 2022-05-11
Replacing lambda functions with regular functions in all datasets

NivekT:
Replacing lambda functions with regular functions in all datasets because lambda functions are not serializable by `pickle` and serialization is necessary for DataLoader's multiprocessing feature.

If the changes are correct, it should not break any CI.

Fixes part of https://github.com/pytorch/data/issues/397 (TorchData team plans to improve DataLoader as well, see the comments within the issue)

Fixes https://github.com/pytorch/text/issues/1716

### DIFF
--- a/torchtext/datasets/ag_news.py
+++ b/torchtext/datasets/ag_news.py
@@ -52,14 +52,20 @@ def AG_NEWS(root: str, split: Union[Tuple[str], str]):
             "Package `torchdata` not found. Please install following instructions at `https://github.com/pytorch/data`"
         )
 
+    def _filepath_fn():
+        return os.path.join(root, split + ".csv")
+
+    def _modify_res(t):
+        return int(t[0]), " ".join(t[1:])
+
     url_dp = IterableWrapper([URL[split]])
     cache_dp = url_dp.on_disk_cache(
-        filepath_fn=lambda x: os.path.join(root, split + ".csv"),
-        hash_dict={os.path.join(root, split + ".csv"): MD5[split]},
+        filepath_fn=_filepath_fn,
+        hash_dict={_filepath_fn(): MD5[split]},
         hash_type="md5",
     )
     cache_dp = HttpReader(cache_dp)
     cache_dp = cache_dp.end_caching(mode="wb", same_filepath_fn=True)
 
     data_dp = FileOpener(cache_dp, encoding="utf-8")
-    return data_dp.parse_csv().map(fn=lambda t: (int(t[0]), " ".join(t[1:])))
+    return data_dp.parse_csv().map(fn=_modify_res)
--- a/torchtext/datasets/ag_news.py
+++ b/torchtext/datasets/ag_news.py
@@ -52,7 +52,7 @@ def AG_NEWS(root: str, split: Union[Tuple[str], str]):
             "Package `torchdata` not found. Please install following instructions at `https://github.com/pytorch/data`"
         )
 
-    def _filepath_fn():
+    def _filepath_fn(_=None):
         return os.path.join(root, split + ".csv")
 
     def _modify_res(t):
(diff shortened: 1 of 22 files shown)


parmeet (approved):
Thanks @NivekT . LGTM! Feel free to land once the CI passes w.r.t datasets.

torchtext/datasets/ag_news.py
```diff
@@ -52,14 +52,20 @@ def AG_NEWS(root: str, split: Union[Tuple[str], str]):
             "Package `torchdata` not found. Please install following instructions at `https://github.com/pytorch/data`"
         )
 
+    def _filepath_fn(_=None):
+        return os.path.join(root, split + ".csv")
+
+    def _modify_res(t):
+        return int(t[0]), " ".join(t[1:])
+
```
ejguan, line 60: It's better to use global function rather than local function here. Local functions are still non-picklable for pickle module.

torchtext/datasets/amazonreviewfull.py
```diff
@@ -58,21 +58,29 @@ def AmazonReviewFull(root: str, split: Union[Tuple[str], str]):
...
+
+    def _extracted_filepath_fn(_=None):
+        return os.path.join(root, _EXTRACTED_FILES[split])
+
+    def _filter_fn(x):
+        return _EXTRACTED_FILES[split] in x[0]
+
+    def _modify_res(t):
+        return int(t[0]), " ".join(t[1:])
+
```
ejguan, line 72: Ditto

ejguan (commented):
We need to replace local functions by global functions.

parmeet:
> I am sorry that I didn't have chance to look at this PR before landing. We need to replace local functions by global functions.

Oops Thanks for the catch @ejguan. I think we should add make amendment in our datatsets test suit ([here](https://github.com/pytorch/text/blob/ab76a046919e6358c7038512f3ee500dd17fb632/test/datasets/test_agnews.py#L36) ) about pickability. cc: @Nayef211

NivekT:
Let me fix this with another PR. Thanks for catching it.

Nayef211:
> Oops Thanks for the catch @ejguan. I think we should add make amendment in our datatsets test suit ([here](https://github.com/pytorch/text/blob/ab76a046919e6358c7038512f3ee500dd17fb632/test/datasets/test_agnews.py#L36) ) about pickability. cc: @Nayef211

@parmeet would you mind clarifying what you mean by "make amendment in our datatsets test suite"?

parmeet:
> @parmeet would you mind clarifying what you mean by "make amendment in our datatsets test suite"?

I meant to add tests to make sure we we can dump and load dataset objects using pickle. 

so for instance below code give error:
```python
from torchtext.datasets import AG_NEWS
import pickle
dp = AG_NEWS(split='train')
pickle.dump(dp, open("temp.pkl",'w'))
```
Error: AttributeError: Can't pickle local object 'AG_NEWS.<locals>._filepath_fn'


And in our tests we could just add another unit test that would dump and load the datapipe..

Nayef211:
> I meant to add tests to make sure we we can dump and load dataset objects using pickle.
> 
> so for instance below code give error:
> 
> ```python
> from torchtext.datasets import AG_NEWS
> import pickle
> dp = AG_NEWS(split='train')
> pickle.dump(dp, open("temp.pkl",'w'))
> ```
> 
> Error: AttributeError: Can't pickle local object 'AG_NEWS.._filepath_fn'
> 
> And in our tests we could just add another unit test that would dump and load the datapipe..

Gotcha thanks for clarifying. Let me create a PR to add these tests to our datasets!

parmeet:
> Let me create a PR to add these tests to our datasets!
Thanks @Nayef211 . One suggestion here would be to add a common test parameterized over all the datasets added to the repo instead of adding it to individual test suit :)

PR #1720, opened 2022-05-12
[fbsync] Enable model testing in FBCode

parmeet:
Summary:
This diff enables Model testing in FB code

Notes:
1. it only tests XLM-R models (base and large) in integration tests. We need to do a follow-up diff to enable RoBERTa testing since corresponding assets are missing in FBcode.

Edit: Addressed the Roberta model testing in this diff itself

2. parameterized was giving some weird long names to the test which was creating some unknown issue for running them in sandcastlle. Removed it for now to get the proper names for test.

Edit: refactored test suit since nested_params was creating long string names (400+ characters) for test methods due to RobertaBundle objects

Reviewed By: mikekgfb

Differential Revision: D35973306

fbshipit-source-id: 8a50d03466f60c8a4a0fbd5857611e68c92ebf08

### DIFF
--- a/test/integration_tests/test_models.py
+++ b/test/integration_tests/test_models.py
@@ -1,39 +1,25 @@
 import torch
-from torchtext.models import ROBERTA_BASE_ENCODER, ROBERTA_LARGE_ENCODER, XLMR_BASE_ENCODER, XLMR_LARGE_ENCODER
+from parameterized import parameterized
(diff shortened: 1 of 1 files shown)


Nayef211 (approved):
LGTM. Let's discuss offline about how we want to do testing in the future to reduce code duplication. I think @mthrok may be interested in this as well.

PR #1722, opened 2022-05-12
Updating sst2 tutorial to replace lambda usage

NivekT:
As it was pointed out in this comment:
https://github.com/pytorch/text/issues/1716#issuecomment-1124766084

### DIFF
--- a/examples/tutorials/sst2_classification_non_distributed.py
+++ b/examples/tutorials/sst2_classification_non_distributed.py
@@ -94,13 +94,18 @@ batch_size = 16
 train_datapipe = SST2(split="train")
 dev_datapipe = SST2(split="dev")
 
+
 # Transform the raw dataset using non-batched API (i.e apply transformation line by line)
-train_datapipe = train_datapipe.map(lambda x: (text_transform(x[0]), x[1]))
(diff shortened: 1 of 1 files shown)


examples/tutorials/sst2_classification_non_distributed.py
```diff
@@ -111,10 +116,14 @@
 #
 # ::
 #
+#   def batch_transform(x):
+#       return {"token_ids": text_transform(x["text"]), "target": label_transform(x["label"])}
```
NivekT, line 120: Note that `label_transform` was never defined in this file

Nayef211: Thanks for catching this. The `SST2` dataset returns an integer label in the range `0-1`. @parmeet do you think we can remove the `label_transform` all-together since we don't need to map a text label to an index?

parmeet: Yes, the target column need no transformation now. Thanks for the catch. We should remove it. 

`"target": label_transform(x["label"] -> "target": x["label"]`

parmeet (approved):
LGTM!

PR #1726, opened 2022-05-16
For Datasets, refactor local functions to be global so that they can be pickled

NivekT:
For Datasets, refactor local functions to be global so that they can be pickled

Fixes the issue pointed out here: https://github.com/pytorch/text/pull/1718#discussion_r873266766


ejguan (approved):
Thank you for unblocking TorchText!! This is awesome!

Issue #1727, opened 2022-05-17
Regarding adding shuffling and sharding datapipes to in-built datasets


parmeet:
## 🚀 Feature

**Motivation**

* To avoid pitfall with shuffling and sharding of datapipes in distributed training environments
* To ensure consistent experience of TorchData based datasets across domains.

**Pitch**

TorchText datasets return datapipes. In order to perform distributed computing, users would typically apply a sharding filter in order to shard the data across ranks. Furthermore, to make sure that we don’t shuffle data only within the corresponding shards, it is important to ensure that the sharding filter is applied after shuffling. As per the investigations from [TorchVision](https://github.com/pytorch/data/issues/302), this is not always obvious for users and could lead to suboptimal results if not being done in proper order.

We could do this by simply wrapping the datapipe at the very end

```python 
def MyDataSet(...):
    dp = ...
    dp = dp.shuffle().set_shuffle(False)
    dp = dp.sharding_filter()
    return dp
```

when users want to shuffle the dataset, they would simply set shuffle=True in DataLoader. Furthermore since the sharding filter is already applied, users do not have to explicitly call it when doing distributed training.

**Alternatives**

keep the datasets implementation as such and educate (tutorials, documentation) users to perform shuffling before sharding.

**Additional context**

* In addition to making sure that shuffling is always done before sharding, this also comes with the benefit that shuffling can be done before the datapipe contains heavy objects (like images) as shuffle datapipe creates a buffer internally to shuffle the corresponding data items. Hence for vision datasets, this is more than just convenient/helper utility.
 * If shuffle and sharding are done internally, it would mean that we must document the usage such that users do not apply shuffle and sharding again. 
 * We also want to ensure that users have similar experiences across domains and hence should have consistent solutions for common pitfalls.

cc: @NicolasHug , @ejguan , @kevinchn , @Nayef211 , @abhinavarora , @VirgileHlav

NicolasHug:
Thanks for agreeing to make these changes @parmeet , I think this will be positively impact users' experience with datapipes in the long run.

Considering the release and branch cut dates are approaching (fast!), please let me know if you'd like me to help by submitting PRs for the datapipes, or for the docs.

PR #1729, opened 2022-05-17
Add Shuffle and sharding datapipes to datasets

parmeet:
Reference Issue: #1727

### DIFF
--- a/test/datasets/common.py
+++ b/test/datasets/common.py
@@ -1,0 +1,25 @@
+from parameterized import parameterized
+from torch.utils.data.graph import traverse
+from torch.utils.data.graph_settings import get_all_graph_pipes
+from torchdata.datapipes.iter import Shuffler, ShardingFilter
+from torchtext.datasets import DATASETS
+
+from ..common.torchtext_test_case import TorchtextTestCase
+
+
+class TestShuffleShardDatasetWrapper(TorchtextTestCase):
+    # Note that for order i.e shuffle before sharding, TorchData will provide linter warning
+    # Modify this test when linter warning is available
+    @parameterized.expand(list(DATASETS.items()))
+    def test_shuffle_shard_wrapper(self, dataset_name, dataset_fn):
+        dp = dataset_fn()
+        if type(dp) == tuple:
+            dp = list(dp)
+        else:
+            dp = [dp]
+
+        for dp_split in dp:
+            dp_graph = get_all_graph_pipes(traverse(dp_split))
+            for annotation_dp_type in [Shuffler, ShardingFilter]:
+                if not any(isinstance(dp, annotation_dp_type) for dp in dp_graph):
+                    raise AssertionError(f"The dataset doesn't contain a {annotation_dp_type.__name__}() datapipe.")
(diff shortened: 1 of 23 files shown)


ejguan:
I enabled AWS extension last night. It seems the C-Extension is using glibc-2.29 during compilation. Will do a quick fix.

parmeet:
> Perhaps it might make sense to add a small test that makes sure all datapipes come with a sharding filter and a shuffler? 

Thanks @NicolasHug  for the suggestion. I think it is indeed a good idea. Let me add it as well.

VirgileHlav (commented):
Do we expect to have variations for shuffle and sharding between datasets, as for now all datasets use `.shuffle().set_shuffle(False).sharding_filter()`? If not, we could consider using instead a decorator to wrap this call on the returned object and clean up the code.

NicolasHug:
Good point @VirgileHlav .

In general you want to shard and shuffle on light objects (before decoding, before transforms) to avoid unnecessary computations, and to save memory. For now torchtext datasets yield light objects (simple text), but maybe in the future this will change?

In torchvision, we wrap in different places for each dataset, so using a decorator isn't an option.

VirgileHlav (approved):
@NicolasHug thanks for the clarification, in that case let's keep it as is for now. @parmeet I will also add this to the dataset effort in #1710.

Otherwise LGTM

parmeet:
Thanks @VirgileHlav SGTM! As @NicolasHug mentioned it could potentially vary from dataset to dataset. So if in some cases processing is needed at sample level, you could potentially shard the pipe before processing, but otherwise adding it at the end is fine.

test/datasets/common.py
```diff
@@ -0,0 +1,25 @@
...
+        if type(dp) == tuple:
+            dp = list(dp)
+        else:
+            dp = [dp]
+
+        for dp_split in dp:
+            dp_graph = get_all_graph_pipes(traverse(dp_split))
+            for annotation_dp_type in [Shuffler, ShardingFilter]:
+                if not any(isinstance(dp, annotation_dp_type) for dp in dp_graph):
+                    raise AssertionError(f"The dataset doesn't contain a {annotation_dp_type.__name__}() datapipe.")
```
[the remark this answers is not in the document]
ejguan: Thanks. I will do a fix for manylinux1 wheel first then add linter for you.

test/datasets/common.py
```diff
@@ -0,0 +1,25 @@
...
+from torchdata.datapipes.iter import Shuffler, ShardingFilter
+from torchtext.datasets import DATASETS
+
+from ..common.torchtext_test_case import TorchtextTestCase
+
+
+class TestShuffleShardDatasetWrapper(TorchtextTestCase):
+    # Note that for order i.e shuffle before sharding, TorchData will provide linter warning
+    # Modify this test when linter warning is available
+    @parameterized.expand(list(DATASETS.items()))
```
Nayef211, line 13: QQ: it looks like we're not using the `dataset_name` anywhere. Why don't we just pass in the `dataset_fn` to the test by doing something like

```python
@parameterized.expand(list(DATASETS.values()))
```

parmeet: Ahh my bad, thanks for the catch. Will fix it!

Nayef211: I actually realized that the `parameterized.expand` decorator complains when passing in the list of  `dataset_fn`. Lmk if you're able to figure out how to resolve the error

parmeet: Yupp, the problem is we need to pass tuples inside list. Just created PR to fix it https://github.com/pytorch/text/pull/1733

Nayef211: Awesome. Just incorporated this in my PR https://github.com/pytorch/text/pull/1732

PR #1731, opened 2022-05-17
[WIP] Fix broken torchdata install in CI

Nayef211:


Nayef211:
> Could you please re-run the workflow? TorchData has an updated nightly.

I reran the workflow and the errors with gclib seems to have gone away. Current test failures seem to relate to `test_download_glove_vectors`

PR #1736, opened 2022-05-21
Fix doc js initialization

mthrok:
1. Fix `collapsedSections is not defined` error.
PyTorch Sphinx theme seems to expect that `collapsedSections`
variable to be defined before its theme.js is executed.

2. Overwrite the link to GitHub project to torchtext
So that documentation link to torchtext's repository instead of
PyTorch core.


Nayef211 (approved):
LGTM!

PR #1737, opened 2022-05-23
Fix STSB and WikiTexts tests

### LINKED ISSUE #1735: Caching results of the `filter` will result to inconsistent cache state
Currently it blocks, but we just got lucky:

https://github.com/pytorch/text/blob/caaa8e3c08309fe2c40ae40efaf7c2adcf1a2c8a/torchtext/datasets/stsb.py#L85

Please change the order of `filter` and `end_caching`

parmeet:
This PR fixes issues with STSB and WikiTexts mocked testing. In both the cases mocked files were not written properly in their respective destinations. 

fixes #1735 

cc: @Nayef211 , @VirgileHlav , @VitalyFedyunin

### DIFF
--- a/test/datasets/test_wikitexts.py
+++ b/test/datasets/test_wikitexts.py
@@ -57,19 +57,19 @@ class TestWikiTexts(TempDirMixin, TorchtextTestCase):
     root_dir = None
     samples = []
 
-    @ classmethod
+    @classmethod
     def setUpClass(cls):
         super().setUpClass()
         cls.root_dir = cls.get_base_temp_dir()
         cls.patcher = patch("torchdata.datapipes.iter.util.cacheholder._hash_check", return_value=True)
         cls.patcher.start()
 
-    @ classmethod
+    @classmethod
     def tearDownClass(cls):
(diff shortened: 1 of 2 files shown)


test/datasets/test_wikitexts.py
```diff
@@ -55,19 +57,19 @@ class TestWikiTexts(TempDirMixin, TorchtextTestCase):
     root_dir = None
     samples = []
 
-    @classmethod
+    @ classmethod
```
Nayef211: nit: should we remove this space? same with the rest of the `@classmethod` and `@nested_params`  decorators

Nayef211 (approved):
LGTM thanks for fixing the tests!

test/datasets/test_wikitexts.py
parmeet: hmm,  must be by mistake..already fixed :)

parmeet:
> I wonder why the build is segfaulting when running `python setup.py clean`?

Ya, I am not so sure. I see same issues on audio CI (cc: @mthrok).

PR #1738, opened 2022-05-23
add test for shuffle before shard

parmeet:
Follow-up of #1729. We need to ensure that shuffle is applied before sharding.


VirgileHlav (approved):
Why is the `pywin32` dependency required?
LGTM

parmeet:
> Why is the `pywin32` dependency required?

Ya I was just trying to figure out the failures we have on windows machine due to portalocker and potential dependency on pywin32. Will remove it for this PR before landing.

PR #1740, opened 2022-05-25
Change root directory for datasets

parmeet:
Following conversations in this [post](https://fb.workplace.com/groups/473278361165956/permalink/517067430120382/) internally with @NicolasHug , this is the first step to change the root directory of datasets.

Follow-up:

- [ ] Enable setting root directory programmatically once the APIs are aligned.

### DIFF
--- a/torchtext/__init__.py
+++ b/torchtext/__init__.py
@@ -4,7 +4,7 @@ import os
 from torchtext import _extension  # noqa: F401
 
 _TEXT_BUCKET = "https://download.pytorch.org/models/text/"
-_CACHE_DIR = os.path.expanduser("~/.torchtext/cache")
+_CACHE_DIR = os.path.expanduser("~/.cache/torch/text")
 
 from . import data, datasets, experimental, functional, models, nn, transforms, utils, vocab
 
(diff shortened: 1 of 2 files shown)


VirgileHlav (approved):
Thanks @parmeet, that's a great direction to take! 
LGTM

parmeet:
> LGTM. Just to clarify, is `~/.cache/torch` the base directory used by the core PyTorch team to store downloaded assets? And now each domain will have its own subdirectory within that base directory?

That's right. For instance hub base dir is ~/.cache/torch/hub

torchtext/__init__.py
```diff
@@ -4,7 +4,7 @@
 from torchtext import _extension  # noqa: F401
 
 _TEXT_BUCKET = "https://download.pytorch.org/models/text/"
-_CACHE_DIR = os.path.expanduser("~/.torchtext/cache")
+_CACHE_DIR = os.path.expanduser("~/.cache/torch/text")
```
mthrok, line 7: `~/.cache/torch` is the default place of TORCH_HOME, which is governed by PyTorch core.
If users set TORCH_HOME to different place, this torchtext cache directory won't be affected. There is a pros and cons on either behavior. If If domain libraries are piggy-backing on it,  and the change is being made for the consistency, shouldn't that be accounted? Meaning that did you settle on how this should interact with TORCH_HOME?

mthrok: See https://pytorch.org/docs/stable/hub.html#where-are-my-downloaded-models-saved

parmeet: Thanks @mthrok for the catch. I missed to take TORCH_HOME into account in this PR :(. Let me make a quick fix.

PR #1741, opened 2022-05-25
Take TORCH_HOME env variable into account while setting the cache dir

parmeet:
Addresses comment https://github.com/pytorch/text/pull/1740#discussion_r882075619

### DIFF
--- a/torchtext/__init__.py
+++ b/torchtext/__init__.py
@@ -4,7 +4,11 @@ import os
 from torchtext import _extension  # noqa: F401
 
 _TEXT_BUCKET = "https://download.pytorch.org/models/text/"
-_CACHE_DIR = os.path.expanduser("~/.cache/torch/text")
+
+_TORCH_HOME = os.getenv("TORCH_HOME")
+if _TORCH_HOME is None:
+    _TORCH_HOME = "~/.cache/torch" #default
+_CACHE_DIR = os.path.expanduser(os.path.join(_TORCH_HOME, "text"))
 
 from . import data, datasets, experimental, functional, models, nn, transforms, utils, vocab
 


torchtext/__init__.py
```diff
@@ -4,7 +4,11 @@
 from torchtext import _extension  # noqa: F401
 
 _TEXT_BUCKET = "https://download.pytorch.org/models/text/"
-_CACHE_DIR = os.path.expanduser("~/.cache/torch/text")
+
+_TORCH_HOME = os.getenv("TORCH_HOME")
+if _TORCH_HOME is None:
+    _TORCH_HOME = "~/.cache/torch" #default
+_CACHE_DIR = os.path.expanduser(os.path.join(_TORCH_HOME, "text"))
```
mthrok, line 11: I think it can be as simple as

```
_CACHE_DIR = os.path.join(torch.hub.get_dir(), "text")
```

`torch.hub.get_dir` will do the check for you.

parmeet: hmm, I wonder if someone use the `set_dir` from hub, then wouldn't that path would deviate from `$TORCH_HOME`? In other words do we want to give preference to torch.hub root directly over `$TORCH_HOME` for domains? 

Another question for us now is if TORCH_HOME is not set do we want to default to torch.hub root :)

mthrok: At the beginning I thought the proposal was to piggy-back on PyTorch's default directory. Which has somewhat elaborated, but reasonable way of resolving the default. My view here is that I would avoid re-inventing the wheel so my suggestion was to reuse the implementation from PyTorch.

However, I get the sense that your view is that only `TORCH_HOME` should be considered and not necessarily how Torch Hub is resolving default path. I see that is another valid view and I don't have basis to say which is better.

But fundamentally, this kind of thinking is what happens when trying to incorporate yourself into something larger. I think simply using `~/.cache/torchtext` would avoid this kind of unnecessary design thinking, and allow users have complete and precise control over where it should be, without worrying how these configurations mix with PyTorch's configuration.

mthrok: Having said that, let me stamp on the PR to unblock you.

mthrok (approved):

PR #1745, opened 2022-05-26
Adding support for batch input in BERT Tokenizer with perf benchmark

parmeet:
This PR add support for batch tokenizer directly implemented at C++ kernel layer.

It also add benchmarking code for tokenizer. Following result is obtained by running tokenizer on 100000 samples.

```python
python benchmark/benchmark_bert_tokenizer.py --num-samples num_samples
```

**batch size 100**
Running TorchText BERT Tokenizer on non-batched input ... Total running time: 0.0014495829999998655
Running HF BERT Tokenizer (slow) on non-batched input ... Total running time: 0.01610425200000032
Running HF BERT Tokenizer (fast) on non-batched input ... Total running time: 0.0037178359999998634
Running TorchText BERT Tokenizer on batched input ... Total running time: 0.0007109650000001189
Running HF BERT Tokenizer (fast) on batched input ... Total running time: 0.0010531449999997555

**batch size 1000**
Running TorchText BERT Tokenizer on non-batched input ... Total running time: 0.021309904000000213
Running HF BERT Tokenizer (slow) on non-batched input ... Total running time: 0.383976283
Running HF BERT Tokenizer (fast) on non-batched input ... Total running time: 0.07770123400000006
Running TorchText BERT Tokenizer on batched input ... Total running time: 0.01940807400000022
Running HF BERT Tokenizer (fast) on batched input ... Total running time: 0.015243869999999937

**Observations**
*  ~15x on non-batched input compared to HF slow tokenizer (python based).
*  ~3x on non-batched input compared to HF fast tokenizer (Rust based)
*  torchtext is faster for batch processing up-to certain batch size but then as the batch size increases HF implementation becomes significantly faster (20-25%). This is likely because the kernel implementation of HF is faster compared to torchtext's but the overhead to call the binding for HF is higher compared to torchtext. 

Performance next steps:

- [ ] Perform analysis on the impact of batch size for batch processing.
- [ ] Identify and improve performance gaps in batch processing 
- [ ] Make use of timeit module in benchmark code

### DIFF
--- a/torchtext/csrc/register_pybindings.cpp
+++ b/torchtext/csrc/register_pybindings.cpp
@@ -224,27 +224,25 @@ PYBIND11_MODULE(_torchtext, m) {
           "batch_encode",
           [](const c10::intrusive_ptr<BERTEncoder>& self,
              const py::list& items) {
-            std::vector<std::vector<int64_t>> ids(items.size());
-            int64_t counter = 0;
+            std::vector<std::string> input;
             for (const auto& item : items) {
               Py_ssize_t length;
               const char* buffer = PyUnicode_AsUTF8AndSize(item.ptr(), &length);
-              ids[counter++] = self->Encode(std::string(buffer));
+              input.push_back(std::string(buffer));
             }
-            return ids;
+            return self->BatchEncode(input);
           })
       .def(
           "batch_tokenize",
           [](const c10::intrusive_ptr<BERTEncoder>& self,
              const py::list& items) {
-            std::vector<std::vector<std::string>> tokens(items.size());
-            int64_t counter = 0;
+            std::vector<std::string> input;
             for (const auto& item : items) {
               Py_ssize_t length;
               const char* buffer = PyUnicode_AsUTF8AndSize(item.ptr(), &length);
-              tokens[counter++] = self->Tokenize(std::string(buffer));
+              input.push_back(std::string(buffer));
             }
-            return tokens;
+            return self->BatchTokenize(input);
           })
       .def(py::pickle(
           // __getstate__
--- a/torchtext/csrc/register_pybindings.cpp
+++ b/torchtext/csrc/register_pybindings.cpp
@@ -220,6 +220,32 @@ PYBIND11_MODULE(_torchtext, m) {
       .def(py::init<const std::string, bool, c10::optional<bool>>())
       .def("encode", &BERTEncoder::Encode)
       .def("tokenize", &BERTEncoder::Tokenize)
+      .def(
+          "batch_encode",
+          [](const c10::intrusive_ptr<BERTEncoder>& self,
+             const py::list& items) {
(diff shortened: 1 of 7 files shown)


torchtext/transforms.py
```diff
@@ -597,6 +597,23 @@ def _encode(self, text: str) -> List[str]:
...
 
+    @torch.jit.export
+    def _batch_encode(self, text: List[str]) -> List[List[str]]:
+        """Encode text into a list of tokens IDs
+
+        Args:
+            text: An input text string.
+
+        Returns:
+            A list of token ids represents each sub-word
```
Nayef211, line 608: Can we update the docstring to reflect that we're operating on a list of strings and the output is a nested list of tokens

torchtext/transforms.py
```diff
@@ -597,6 +597,23 @@ def _encode(self, text: str) -> List[str]:
...
+
+        For example:
+            --> "Hello world!" --> token ids: [707, 5927, 11, 707, 68]
```
Nayef211, line 611: Also need to update the example

torchtext/transforms.py
```diff
@@ -612,6 +629,21 @@ def _tokenize(self, text: str) -> List[str]:
...
+
+        Args:
+            text: An input text string.
+
+        Returns:
+            A list of tokens (sub-words)
+
+        For example:
+            --> "Hello World!": ["Hello", "World", "!"]
+        """
```
Nayef211, line 644: Update docstring to reflect list input and nested list output

torchtext/csrc/register_pybindings.cpp
```diff
@@ -220,6 +220,32 @@ PYBIND11_MODULE(_torchtext, m) {
       .def(py::init<const std::string, bool, c10::optional<bool>>())
       .def("encode", &BERTEncoder::Encode)
       .def("tokenize", &BERTEncoder::Tokenize)
+      .def(
+          "batch_encode",
+          [](const c10::intrusive_ptr<BERTEncoder>& self,
```
Nayef211, line 225: Is there a reason we have the operator implementation in the registration file as opposed to the `bert_tokenizer.cpp` file? I assume we'd like to keep registration code separate from the actual implementation code?

Nayef211 (commented):
Left a couple of comments. In regards to the benchmark code, I wonder if we should try a really large value for `num_samples` so we can see how our batched implementation performs for very small and large batches. 

I also wonder if you've looked into using the `timeit` module which is meant for profiling and allows us to run a snippet of code several times to more accurately measure execution speed. I noticed that the `timeit` module has a `Timer` class built in. Could we potentially use this implementation instead of the one we have in `benchmark/utils.py`

torchtext/csrc/register_pybindings.cpp
parmeet: That's a good catch. I moved the core implementation back into the class.

PR #1749, opened 2022-05-27
Fix Mock tests due to change in datasets directory

### LINKED ISSUE #1747: All the mock tests are failing on CI
All the mock tests for datasets started failing recently on CI as well

So far only thing I could figure out is that all of them are just hanging when fetching the samples and then fail after a while. For eg: [here](https://github.com/pytorch/text/blob/d963bdb82b4c6c5170430fd7289b9dd0a2b9283a/test/datasets/test_agnews.py#L58) for AG_NEWS test.

Interestingly, there is no problem working with the actual datasets in IPython. 

cc: @Nayef211, @ejguan, @VitalyFedyunin

parmeet:
Fixes #1747


Nayef211 (approved):
LGTM

PR #1752, opened 2022-05-31
Use _get_torch_home standard utility from torch hub

parmeet:
Addresses https://github.com/pytorch/text/pull/1741#discussion_r884789886


parmeet:
The failures (test_download_extract_tar and test_download_extract_to_path) doesn't seem to be linked with this PR.

PR #1753, opened 2022-05-31
Removed ticks (``) from the url under is_module_available

suvigyajain0101:
Ticks can result in re-direction to 'https://github.com/pytorch/data%60' instead of 'https://github.com/pytorch/data', which is not a valid resource.
Tested in Colab.


suvigyajain0101:
> LGTM thanks for addressing this issue @suvigyajain0101!

Anytime. Cheers! :beers:

PR #1754, opened 2022-05-31
add header info for BERT tokenizer

parmeet:


Nayef211 (approved):
Stamping since the license info has already passed legal review

PR #1757, opened 2022-06-02
Fix test utils

parmeet:
The [validation.tar.gz](http://www.quest.dcs.shef.ac.uk/wmt16_files_mmt/validation.tar.gz)  file used in test_utils.py is broken. Ideally we should not rely on public assets for testing.

This PR creates a mock asset and upload it it _TEXT_BUCKET ("https://download.pytorch.org/models/text/") which is used for testing.


Nayef211 (approved):
LGTM

PR #1759, opened 2022-06-02
Fix bug in RTE and WNLI testing

parmeet:
This PR fixes issues RTE and WNLI testing:
* In both tests, "dataset" directory was not taken into account causing it to download the real data instead of working with mock data
* RTE dataset was not converting string labels into proper integer IDs
* RTE dataset was not parsed properly. Need to pass quoting=csv.QUOTE_NONE in parse_csv to fix the issue.

### DIFF
--- a/test/datasets/test_qnli.py
+++ b/test/datasets/test_qnli.py
@@ -54,7 +54,7 @@ class TestQNLI(TempDirMixin, TorchtextTestCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.root_dir = cls.get_base_temp_dir()
-        cls.samples = _get_mock_dataset(os.path.join(cls.root_dir, "datasets"))
+        cls.samples = _get_mock_dataset(cls.root_dir)
         cls.patcher = patch("torchdata.datapipes.iter.util.cacheholder._hash_check", return_value=True)
         cls.patcher.start()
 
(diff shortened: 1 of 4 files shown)


torchtext/datasets/rte.py
```diff
@@ -36,6 +36,8 @@
     "test": os.path.join("RTE", "test.tsv"),
 }
 
+MAP_LABELS = {"not_entailment": 1, "entailment": 0}
```
Nayef211, line 39: nit: can we make "entailment" entry come before "not_entailment"

test/datasets/test_qnli.py
```diff
@@ -54,7 +54,7 @@ class TestQNLI(TempDirMixin, TorchtextTestCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.root_dir = cls.get_base_temp_dir()
-        cls.samples = _get_mock_dataset(cls.root_dir)
+        cls.samples = _get_mock_dataset(os.path.join(cls.root_dir, "datasets"))
```
Nayef211, line 57: Can we remove this change from this PR. I will make all fixes corresponding to QNLI dataset in this PR https://github.com/pytorch/text/pull/1760

Nayef211 (approved):
LGTM. Let's merge after adding a PR summary

PR #1760, opened 2022-06-03
Fix bug in QNLI dataset and corresponding test

Nayef211:
# Description
- QNLI dataset had a bug where we weren't taking into account that the `test` split did not have a label entry
- Additionally the `root_dir` for the mock dataset was not set correctly so the entire dataset was downloaded
- Updated dataset implementation and corresponding test

# Testing
` pytest test/datasets/test_qnli.py`


parmeet (approved):
LGTM!

PR #1762, opened 2022-06-03
Set release version and base pytorch version

Nayef211:
- Setting release version and base pytorch version as part of the torchtext 0.13 branch cut process


parmeet (approved):
LGTM!

PR #1767, opened 2022-06-06
Updated doctree for 0.13.0 release candidate

Nayef211:


parmeet (commented):
Thanks @Nayef211 for this PR. As discussed, let's wait for the tag to be created. We should have docs auto-generated for us. If not, we can land this PR.

Nayef211:
> Thanks @Nayef211 for this PR. As discussed, let's wait for the tag to be created. We should have docs auto-generated for us. If not, we can land this PR.

Looks like the files have been autogenerated. Modified this PR to just update the doctree.

versions.html
```diff
@@ -20,20 +20,20 @@ <h1> PyTorch Documentation </h1>
         <li class="toctree-l1">
           <a class="reference internal" href="main/">main (unstable)</a>
         </li>
+        <li class="toctree-l1">
+          <a class="reference internal" href="0.13.0/">v0.9.0 (release candidate)</a>
```
parmeet, line 24: could you please update 0.9.0 to 0.13.0.

versions.html
```diff
@@ -20,20 +20,20 @@ <h1> PyTorch Documentation </h1>
...
+        </li>
         <li class="toctree-l1">
           <a class="reference internal" href="0.12.0/">v0.12.0 (stable release)</a>
         </li>
-       <li class="toctree-l1">
+        <li class="toctree-l1">
```
parmeet, line 29: not sure if it matter, but can we remove extra space in the front?

versions.html
```diff
@@ -20,20 +20,20 @@ <h1> PyTorch Documentation </h1>
...
         <li class="toctree-l1">
-          <a class="reference internal" href="0.9.0/">v0.9.0</a>
+          <a class="reference internal" href="0.10.0/">v0.10.0</a>
         </li>
       </ul>
     </div>
     </div></div>
   </body>
-</html>
+</html>
```
parmeet, line 39: looks like the newline is removed from the end?

versions.html
Nayef211: Oops thanks for the catch!

versions.html
Nayef211: So this is actually removing the additional space that was there before to make this line consistent with how we indent the other `<li` lines

versions.html
Nayef211: Added

parmeet (approved):
LGTM!

PR #1770, opened 2022-06-07
Add test to compare encoder inference on input with and without padding

erichan1:
Encoder inference should output identical embeddings if given a padded vs. non-padded input (assuming the non-padded data is the same). Check that this is the case. 

We make this check for four paths by setting the below two flags. 
- with_no_grad = [True, False] -> this is performing inference with and without torch.no_grad(). When torch.no_grad() is turned on, we hit the fast path of torch.nn.TransformerEncoder (BetterTransformer). 
- return_all_layers = [True, False]. When set to False, torch.nn.TransformerEncoder is directly called, and when set to True, torchtext encoder loops through torch.nn.TransformerEncoderLayer directly to save the intermediate outputs of each layer.

### DIFF
--- a/test/models/test_transformers.py
+++ b/test/models/test_transformers.py
@@ -14,7 +14,7 @@ class TestTransformers(TorchtextTestCase):
         from torchtext.models import RobertaEncoderConf, RobertaModel
 
         def encoder_inference(encoder, input_lst, with_no_grad):
-            if(with_no_grad):
+            if with_no_grad:
                 with torch.no_grad():
                     res = [encoder(eval_input) for eval_input in input_lst]
             else:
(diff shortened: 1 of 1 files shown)


test/models/test_transformers.py
```diff
@@ -0,0 +1,52 @@
...
+            max_seq_len = 514,
+            num_attention_heads=12, 
+            num_encoder_layers=2,
+            dropout=0.1,
+            scaling=None,
+            normalize_before=False,
+        )
+        model = RobertaModel(encoder_conf)
+        model = model.eval()
+        # TODO: make return_all_layers a property of RobertaEncoderConf so it can be passed as arg
```
erichan1, line 41: Currently, return_all_layers cannot be set from RobertaEncoderConf because it's not a property of the config. Thus we need to manually set it below. Not a big deal right now, but something to consider.

PR #1771, opened 2022-06-07
Pinning Utf8proc version

parmeet:
Fixing version to 2.5.0


mthrok:
Is this the same version as torcharrow/velox use?

parmeet:
> Is this the same version as torcharrow/velox use?

yupp, that's right.

mthrok (approved):

PR #1772, opened 2022-06-07
[cherry-pick] Add test to compare encoder inference on input with and without padding (#1770)

Nayef211:
* add test to compare encoder inference on input with and without padding

* remove bottom whitespace

* remove unnecessary import

* lint

### DIFF
--- a/test/models/test_transformers.py
+++ b/test/models/test_transformers.py
@@ -1,0 +1,51 @@
+import torch
+
+from ..common.parameterized_utils import nested_params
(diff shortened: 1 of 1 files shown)


parmeet:
> * add test to compare encoder inference on input with and without padding
> * remove bottom whitespace
> * remove unnecessary import
> * lint

not sure if we need this in the summary?

parmeet (approved):
LGTM!

PR #1775, opened 2022-06-10
Migrate MaskTransform from internal to experimental/transforms.py

pmabbo13:
# Description

Migrate MaskTransform from internal to open-source torchtext library and add unit tests for class

# Process

- Remove **_TorchTextVocabTransform_** dependency and add necessary variables from vocab object as input parameters
- Have class perform random masking only
- Add **__random_masking_** method to class, which was previously called from internal **_transform_utils.py_**
- add **_mask_mask_prob_** and **_rand_mask_prob_ clas**s variables to enable testing of random masking

# Testing

Can only test for values where the output is deterministic. This corresponds to testing

- masking probability at 0% and 100%
- proportion of masked tokens replaced with [MASK] at 0% and 100%
- proportion of masked tokens replaced with random token at 0% and 100%
- **_mask_bos_** parameter that indicates whether or not BOS tokens are eligible for masking

pytest test/experimental/test_transforms.py

### DIFF
--- a/torchtext/experimental/transforms.py
+++ b/torchtext/experimental/transforms.py
@@ -441,7 +441,7 @@ class MaskTransform(nn.Module):
         Function to mask tokens randomly.
 
         Inputs:
-            1) tokens: Tensor with token ids of shape (batch_size x seq_len)
+            1) tokens: Tensor with token ids of shape (batch_size x seq_len). Includes token ids for special tokens such as [BOS] and [PAD]
             2) mask_prob: Probability of masking a particular token
 
         Outputs:
--- a/torchtext/experimental/transforms.py
+++ b/torchtext/experimental/transforms.py
@@ -1,6 +1,7 @@
 import io
-from typing import List
+from typing import List, Tuple
 
+import numpy as np
 import torch
 import torch.nn as nn
 from torch import Tensor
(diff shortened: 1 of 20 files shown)


Nayef211:
Can you also remove the changes to the `utf8proc` submodule by following the suggestions here: https://stackoverflow.com/questions/10906554/how-do-i-revert-my-changes-to-a-git-submodule. In general you don't want to commit any changes from submodules. If ever you see changes showing up from the submodule folder, you can run `git submodule update` to get the latest changes from that submodule into your repo (basically to keep the submodule up to date with the changes committed to their repo).

torchtext/experimental/transforms.py
```diff
@@ -340,3 +341,147 @@ def forward(self, tokens: List[str]) -> Tensor:
...
+
+        Inputs:
+            1) tokens: Tensor with token ids of shape (batch_size x seq_len)
+            2) mask_prob: Probability of masking a particular token
+
+        Outputs:
+            mask: Tensor with same shape as input tokens (batch_size x seq_len)
+                with masked tokens represented by a 1 and everything else as 0.
+        """
+        batch_size, seq_len = tokens.size()
```
parmeet, line 454: nit: Looks like tokens are used only to get the sizes. Should we pass the size directly instead?

torchtext/experimental/transforms.py
```diff
@@ -340,3 +341,147 @@ def forward(self, tokens: List[str]) -> Tensor:
...
+    """
+    The transform chooses mask_prob% (example 15%) of the token positions at random for
+    prediction.
+
+    If the i-th token is chosen, we replace the i-th token with
+    (1) the [MASK] token 80% of the time
+    (2) a random token 10% of the time
+    (3) the unchanged i-th token 10% of the time.
+
+    NOTE: this transform is not torchscriptible.
```
parmeet, line 356: nit: we can probably remove this statement.

torchtext/experimental/transforms.py
```diff
@@ -340,3 +341,147 @@ def forward(self, tokens: List[str]) -> Tensor:
...
+        self.pad_idx: int = pad_idx
+        self.mask_prob: float = mask_prob
+        self.mask_bos: bool = mask_bos
+
+    def forward(self, tokens: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        """
+        Applies mask to input tokens.
+
+        Inputs:
+            tokens: Tensor with token ids of shape (batch_size x seq_len)
```
parmeet, line 415: Let's also mention that that tokens are expected to contain any special tokens like [BOS] and [PAD] tokens ids.

parmeet (commented):
Thanks @pmabbo13 for the PR. Overall Looks great! I have left minor comments. Please feel free to publish the PR out for draft state.

Nayef211:
> Can you also remove the changes to the `utf8proc` submodule by following the suggestions here: https://stackoverflow.com/questions/10906554/how-do-i-revert-my-changes-to-a-git-submodule. In general you don't want to commit any changes from submodules. If ever you see changes showing up from the submodule folder, you can run `git submodule update` to get the latest changes from that submodule into your repo (basically to keep the submodule up to date with the changes committed to their repo).

Just following up with the exact instructions on how to revert the changes in the `utf8proc` submodule. In this PR, we accidentally updated the submodule to a more recent commit `f0b3707` compared to the commit hash we have specified in the main branch of the torchtext repo `0890a53 ` (change set can be viewed [here](https://github.com/JuliaStrings/utf8proc/compare/0890a538bf8238cded9be0c81171f57e43f2c755...f0b370716b9175125e24d0f16ba43de9322370e3)). To revert the changes in the submodule, we can simply go into the submodule folder and reset the HEAD to the commit that's specified in the torchtext repo. 

Steps:
1. `cd third_party/utf8proc`
2. `git reset --hard 0890a538bf8238cded9be0c81171f57e43f2c755`
3. `cd ../..`
4. `git add .`
5. `git commit -am "Resetting utf8proc submodule head to 0890a538bf8238cded9be0c81171f57e43f2c755"`
6. `git push`
7. Verify changes to submodule are removed from this PR

Nayef211 (commented):
This PR looks great! Left a couple of minor comments and suggestions. Thanks for thoroughly testing all of the edge cases for the `MaskTransform` implementation! 

In the future, as @parmeet suggested, feel free to move your PR out of draft state before requesting a review (as the action of requesting a review should indicate that your PR is no longer in draft form).

torchtext/experimental/transforms.py
pmabbo13: @parmeet I think just a few lines below it is invoked to convert the mask from a numpy array to a tensor (as shown below). I suppose we could change it by explicitly stating whether we want the tensor to be allocated on a cpu or cuda device, though I'm not sure if it must match the device on which `tokens` is allocated (which as far as I've checked isn't explicitly defined in the script)

`mask = torch.from_numpy(mask).to(tokens.device)`

torchtext/experimental/transforms.py
```diff
@@ -340,3 +341,144 @@ def forward(self, tokens: List[str]) -> Tensor:
...
+            mask: Tensor with same shape as input tokens (batch_size x seq_len)
+                with masked tokens represented by a 1 and everything else as 0.
+        """
+        batch_size, seq_len = tokens.size()
+        num_masked_per_seq = int(seq_len * mask_prob)
+
+        mask = np.zeros((batch_size, seq_len), dtype=np.int_)
+        mask[:, :num_masked_per_seq] = 1
+        for row in mask:
+            np.random.shuffle(row)
```
mthrok, line 458: Can we replace NumPy with PyTorch?

This brings in NumPy dependency and it uses NumPy's global RNG, which makes it harder to control the randomness of the code.

torchtext/experimental/transforms.py
pmabbo13: To follow up, this issue may be resolved if we refactor to remove the numpy dependency entirely (as suggested by @mthrok in later comments)

torchtext/experimental/transforms.py
```diff
@@ -451,7 +451,7 @@ def _random_masking(self, tokens: torch.tensor, mask_prob: float) -> torch.Tenso
         batch_size, seq_len = tokens.size()
         num_masked_per_seq = int(seq_len * mask_prob)
 
-        mask = torch.zeros((batch_size, seq_len), dtype=torch.int)
+        mask = torch.zeros((batch_size, seq_len), dtype=torch.int).to(tokens.device)
```
Nayef211, line 454: @parmeet do we know if the `to(tokens.device)` is required here? AFAIK all of the text preprocessing happens on CPU right now so there would never be a need to move the tensor to GPU (since `tokens.device` should always be `cpu`). Does this make sense?

PR #1782, opened 2022-06-13
Add missing type hints

ProGamerGov:
I added the `-> None` type hint to all the functions in `torchtext/` and `test/` that were missing it, using some [simple regex code](https://stackoverflow.com/questions/64948233/how-to-add-none-to-the-end-of-init-functions-with-notepad).

Adding `-> None` to `__init__` functions is specified by PEP-0484: https://www.python.org/dev/peps/pep-0484/


Nayef211 (approved):
This looks great @ProGamerGov! Thanks for improving typing in torchtext!

PR #1789, opened 2022-06-14
Add CNN-DM dataset to torchtext

pmabbo13:
# Description
Add CNNDM dataset to TorchText using TorchData datapipes

# Process

1. Download CNN and DailyMail data from [here](https://cs.nyu.edu/~kcho/DMQA/). These downloads are cached, however the extractions from the tar files are not yet cached. See follow-up items. 
2. Download URL lists from [here](https://github.com/becxer/cnn-dailymail/), which designate which stories belong to the train/dev/test splits. These download are not yet cached. See follow-up items.
3. Use URL list to filter for stories that belong to target split.
4. Parse these stories to separate the article text from the abstract text and return a datapipe that yields (article, abstract) for each story.

# Testing
Create mock dataset that mimics the format of raw CNNDM files and ensure that the datapipe yields the correct output on this mock dataset

`pytest test/datasets/test_cnndm.py`

# Follow-Up Items

- [ ] Cache the extracted tar files for cnn_stories.tgz and dailymail_stories.tgz
- [ ] Cache URL lists (either on disk or using LRUCache) pending benchmark tests on how long it takes to instantiate them on the fly vs loading from cache


torchtext/data/datasets_utils.py
```diff
@@ -310,3 +311,97 @@ def __iter__(self):
...
+    def _hashhex(self, s):
+        """Returns a heximal formated SHA1 hash of the input string."""
+        h = hashlib.sha1()
+        h.update(s)
+        return h.hexdigest()
+
+    def __iter__(self):
+
+        for _, url in self.source_datapipe:
+
```
Nayef211, line 403: nit: remove new line

torchtext/data/datasets_utils.py
```diff
@@ -310,3 +311,97 @@ def __iter__(self):
...
+            return line
+        if line == "":
+            return line
+        if line[-1] in self.END_TOKENS:
+            return line
+        # print line[-1]
+        return line + " ."
+
+    def __iter__(self):
+
```
Nayef211, line 353: nit: remove new line

torchtext/datasets/cnndm.py
```diff
@@ -0,0 +1,100 @@
...
+    "train": "https://raw.githubusercontent.com/abisee/cnn-dailymail/master/url_lists/all_train.txt",
+    "val": "https://raw.githubusercontent.com/abisee/cnn-dailymail/master/url_lists/all_val.txt",
+    "test": "https://raw.githubusercontent.com/abisee/cnn-dailymail/master/url_lists/all_test.txt",
+}
+
+URL_LIST_MD5 = {
+    "train": "c8ca98cfcb6cf3f99a404552568490bc",
+    "val": "83a3c483b3ed38b1392285bed668bfee",
+    "test": "4f3ac04669934dbc746b7061e68a0258",
+}
```
Nayef211, line 31: Can we delete this if we're not planning to use it anywhere?

torchtext/data/datasets_utils.py
```diff
@@ -310,3 +311,97 @@ def __iter__(self):
                     columns[i].append(column)
         if len(columns) > 0:
             yield columns
+
+
+@functional_datapipe("parse_cnndm_data")
```
Nayef211, line 316: For both of the `parse_cnndm` datapipes, let's mention in docstrings that the code is inspired from https://github.com/abisee/cnn-dailymail/blob/master/make_datafiles.py

torchtext/datasets/cnndm.py
```diff
@@ -0,0 +1,100 @@
...
+def _extracted_filepath_fn(root: str, source: str, t):
+    return os.path.join(root, _EXTRACTED_FOLDERS[source])
+
+
+def _filter_fn(story_fnames, x):
+    return os.path.basename(x[0]) in story_fnames
+
+
+def _get_split_list(split: str):
+
```
Nayef211, line 64: nit: remove new line

torchtext/datasets/cnndm.py
```diff
@@ -0,0 +1,100 @@
...
+    "cnn": os.path.join("cnn", "stories"),
+    "daily_mail": os.path.join("dailymail", "stories"),
+}
+
+
+def _filepath_fn(root: str, source: str, _=None):
+    return os.path.join(root, PATH_LIST[source])
+
+
+def _extracted_filepath_fn(root: str, source: str, t):
```
Nayef211, line 55: leave a comment here saying that this function will be used to cache the contents of the tar file

Nayef211 (commented):
Overall the implementation looks great! Some followup items:
- Update the PR description
- Add a section in the description for followup items with a checklist of features that have yet to be implemented. We can then add a link to followup PRs that implement those features and check them off (similar to https://github.com/pytorch/text/pull/1542)
- Add unittests for the dataset

Nayef211 (approved):
LGTM! Thanks for addressing all the PR comments. Let's wait for a review from @parmeet before we merge this in. 😄

torchtext/datasets/cnndm.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def _filter_fn(story_fnames, x):
+    return os.path.basename(x[0]) in story_fnames
+
+
+def _get_split_list(split: str):
+    url_dp = IterableWrapper([URL_LIST[split]])
+    online_dp = OnlineReader(url_dp)
+    return online_dp.readlines().parse_cnndm_split()
```
parmeet, line 61: `parse_cnndm_split` is essentially mapping URLs to corresponding hash+'.story' right? If so, instead of creating DP, could we just create a private function and then use `map`.

test/datasets/test_cnndm.py
```diff
@@ -0,0 +1,97 @@
...
+            filename = h.hexdigest() + ".story"
+            txt_file = os.path.join(source_dir, filename)
+
+            with open(txt_file, "w", encoding=("utf-8")) as f:
+                article = get_random_unicode(seed) + "."
+                abstract = get_random_unicode(seed + 1) + "."
+                dataset_line = (article, abstract)
+                f.writelines([article, "\n@highlight\n", abstract])
+                # append line to correct dataset split
+                mocked_data[split].append(dataset_line)
```
parmeet, line 42: nit: I wonder if we should create multiple URLs per story type (CNN, dailymail) like the real datasets? In general I think it's fine what we did here.

parmeet (approved):

torchtext/datasets/cnndm.py
Nayef211: Do we have any guidance on when to create a DP to do some processing vs using map + private function? Are we only creating a DP for parsing through the dataset?

Nayef211: I spoke with @parmeet offline and I think the sentiment was that if the processing is simple and is mapping some input text into some output text as is done by `parse_cnndm_split` then we should use the `map` approach. If on the otherhand the processing complex as is the case for `parse_cnndm_data` then we should convert the helper function into a datapipe. Sorry for the extra work to change this back to a private fn!

PR #1792, opened 2022-06-15
Fix push on release reference name

atalman:
Fix push on release reference name
We want to compare it against refs/heads/release rather then release
Tests: https://github.com/atalman/vision/commit/af17cd95d2d43ca13354fb700e2da42108dd5a87
Sets correctly release chanell (wheels): https://github.com/atalman/vision/runs/6901327010?check_suite_focus=true


parmeet (approved):
LGTM!

PR #1801, opened 2022-06-22
Adding Benchmark for TA ops

parmeet:
This PR add benchmark comparison code for TA ops: Vocab and GPT2BPE tokenizer

Results:

Running times for SST2  train dataset on Macbook pro:

torchtext's GPT2BPE tokenizer: 3.13
torchtext's vocab: 0.32
torcharrow's GPT2BPE tokenizer: 59.13
torcharrow's vocab: 0.03

Interestingly GPT2BPE op is much slower compare to torchtext's implementation whereas vocab op is much faster compared to torchtext's implementation. Need further investigations.

### DIFF
--- a/benchmark/benchmark_torcharrow_ops.py
+++ b/benchmark/benchmark_torcharrow_ops.py
@@ -1,0 +1,56 @@
+import sys, os
+
+import torcharrow as ta
+import torchtext.transforms as T
+from benchmark.utils import Timer
+from torcharrow import functional as ta_F
+from torchtext._download_hooks import load_state_dict_from_url
+from torchtext.datasets import SST2
+
+sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), "../examples"))
+from data_pipeline.roberta_dataframe import init_ta_gpt2bpe_encoder, init_ta_gpt2bpe_vocab
+
+
+def run_torchtext_ops():
+    # tokenizer converting text into tokens
+    encoder_json_path = "https://download.pytorch.org/models/text/gpt2_bpe_encoder.json"
+    vocab_bpe_path = "https://download.pytorch.org/models/text/gpt2_bpe_vocab.bpe"
+    tokenizer = T.GPT2BPETokenizer(encoder_json_path, vocab_bpe_path)
+
+    # vocabulary converting tokens to IDs
+    vocab_path = "https://download.pytorch.org/models/text/roberta.vocab.pt"
+    vocab = T.VocabTransform(load_state_dict_from_url(vocab_path))
+
+    # dataset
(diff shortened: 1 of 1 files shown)


wenleix (approved):
Thanks for putting the benchmark! The vocab indeed looks quite fast (wow , can we check the result is correct )

> Interestingly GPT2BPE op is much slower compare to torchtext's implementation

Yeah this seems surprisingly slow, especially since the code seems to only have 1 line: https://github.com/pytorch/torcharrow/blob/d5026fa2276fbf7eabf909f7ecba9bbb7eef4cd6/csrc/velox/functions/text/bpe_tokenize.h#L25

I can imagine there are two string copies we can optimize (one for `text.str()` and one for `result.copy_from`). But this magnitude of difference usually means there is some unexpected super slow path... maybe we should quickly check with Velox engineers whether some APIs are not optimized (e.g. `copy_from` may not work well yet over `std::vector<std::string>`?)

parmeet:
> can we check the result is correct 

yupp I verified it..

> Yeah this seems surprisingly slow, especially since the code seems to only have 1 line: https://github.com/pytorch/torcharrow/blob/d5026fa2276fbf7eabf909f7ecba9bbb7eef4cd6/csrc/velox/functions/text/bpe_tokenize.h#L25

Yupp, we are doing something similar for vocab too, but instead of copying over we are using push_back. I don't know if it should give such huge performance regression. 

> But this magnitude of difference usually means there is some unexpected super slow path

Ya, we could use push_back instead of copy to verify if the regression is coming from here..But in general, I agree there is likely something else causing this regression.

cc: @Nayef211

Nayef211:
> > But this magnitude of difference usually means there is some unexpected super slow path
> 
> Ya, we could use push_back instead of copy to verify if the regression is coming from here..But in general, I agree there is likely something else causing this regression.

Looping in @laithsakka who had some suggestions on efficient string copying in velox. Would you mind taking a look to see if there are any obvious things we can improve on.

wenleix:
@Nayef211 Thanks. Would it be possible to also set up the benchmark internally? (say, we can run the benchmark with one `buck` command; given we have internally TA built with Text ops -- thanks for setting it up!, I would assume it's quite close? ~) -- this helps internal Velox engineers (and also me) to look into that as well? ^_^

Nayef211:
> @Nayef211 Thanks. Would it be possible to also set up the benchmark internally? (say, we can run the benchmark with one `buck` command; given we have internally TA built with Text ops -- thanks for setting it up!, I would assume it's quite close? ~) -- this helps internal Velox engineers (and also me) to look into that as well? ^_^

Sure I can get set that up. Will tag you in the diff once it's ready!

PR #1806, opened 2022-06-23
Define TORCHTEXT_API macro for visibility control

mthrok:
## Context:

TorchText uses dual-binding (PyBind11 and TorchBind) to make custom operations available in Python.
The both binding eventually calls the same implementation contained in `libtorchtext.so`.
The ones bound via PyBind11 (the ones in `torchtext._torchtext`) calls into `libtorchtext.so`.

Untitled drawing

This means that `libtorchtext.so` has to make the symbols (APIs) used by `torchtext._torchtext` visible.

However, the default visibility of symbols in shared libraries are different in Windows.
On Windows all the symbols are by default hidden. To work around this, we use `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` to expose all the symbols.
There is an upper limit of visible symbols one library fine can contain, and this can be problematic in the future.
(Although it is unlikely that torchtext will hit the limit, unless it introduces custom CUDA kernels.)

A better approach is to selectively mark the symbols that should be visible as visible.

## Summary of the change set

This commit introduces `TORCHTEXT_API` macro which annotates functions with proper visibility.
The core logic was taken from https://github.com/pytorch/pytorch/blob/bcc02769bef1d7b89bec724223284958b7c5b564/c10/macros/Export.h

The behavior is as follow;

For non-Windows: It is always `__attribute__((__visibility__("default")))`
For Windows:
If the header is included from the compilation unit of `libtorchtext`, then it resolves to `__declspec(dllexport)`. otherwise it resolves to `__declspec(dllimport)`.

This allows to remove `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS`.

### DIFF
--- a/torchtext/csrc/CMakeLists.txt
+++ b/torchtext/csrc/CMakeLists.txt
@@ -41,7 +41,8 @@ set(
   )
 
 set(
-  LIBTORCHTEXT_COMPILE_DEFINITIONS)
+  LIBTORCHTEXT_COMPILE_DEFINITIONS
+  TORCHTEXT_BUILD_MAIN_LIB)
 
 function (define_library name source include_dirs link_libraries compile_defs)
   add_library(${name} SHARED ${source})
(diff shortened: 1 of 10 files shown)


Nayef211 (approved):
LGTM! Thanks @mthrok for introducing the change to add correct visibility for Windows systems.

PR #1807, opened 2022-06-24
Adding benchmarks for add_tokens operator

Nayef211:
## Description
- Add benchmark comparison code for TA ops: `add_tokens`
- Followup to https://github.com/pytorch/text/pull/1801

## Results
All times are for SST2 train dataset on MacBook Pro
```
torchtext's add tokens operation (string): 0.382
torchtext's add tokens operation (int): 0.431

torcharrow's add tokens operation (string): 3.652
torcharrow's add tokens operation (int): 0.075
```

## Takeaways
- add_tokens operation on strings are ~10x slower on torcharrow
- add_tokens operation on ints are ~5x faster on torcharrow

### DIFF
--- a/benchmark/benchmark_torcharrow_ops.py
+++ b/benchmark/benchmark_torcharrow_ops.py
@@ -21,6 +21,12 @@ def run_torchtext_ops():
     vocab_path = "https://download.pytorch.org/models/text/roberta.vocab.pt"
     vocab = T.VocabTransform(load_state_dict_from_url(vocab_path))
 
+    # add token to beginning or end of sentence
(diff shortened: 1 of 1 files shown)


parmeet (approved):
Thanks @Nayef211, for adding this benchmark. Results looks quite interesting. i wonder why the string operations are too slow? Could this be the reason we see similar behavior when comparing vocab and GPT2BPE results..For vocab we are working with ints and GPT2BPE we are dealing with strings. One potential fix would be to revert the changes in this PR https://github.com/pytorch/torcharrow/pull/353 and  check if working with strings is the cause of regression?

ccL @wenleix

PR #1812, opened 2022-06-29
[WIP] Add T5 Model to TorchText

pmabbo13:
# Description

Add T5 architecture to torchtext

# Process

The T5 architecture is very similar to the architecture of a traditional transformer. The main differences are that rather than using positional embeddings, it computes a relative attention bias that encodes the relative position of a token within a sequence. This position bias is then passed into each layer and used to compute the attention scores. T5 also uses a simplified layer normalization (root mean square normalization) which occurs at the start of every attention and feed-forward block.

Incorporating relative attention bias requires under the hood changes to the MultiHeadAttention module. We can use HF's implementation for computing relative attention bias and modify the source code for torch.nn.MultiHeadAttention to incorporate relative attention bias. We can also create our own layer normalization, similarly to HF.

Given the above components, we can then define our own T5EncoderLayer, T5DecoderLayer, T5Encoder, T5Decoder, and T5 modules, all of which can inherit from torch.nn.TransformerEncoderLayer, torch.nn.TransformerDecoderLayer, etc.

# Testing
not yet implemented

### DIFF
--- a/torchtext/prototype/t5/modules.py
+++ b/torchtext/prototype/t5/modules.py
@@ -114,7 +114,8 @@ def _t5_scaled_dot_product_attention(
             have shape :math:`(B, Nt, Ns)`
     """
     B, Nt, E = q.shape
-    q = q / math.sqrt(E)
+    # NOTE: HF implementation does not perform this normalization. For the sake of matching test results, we have commented it out
+    # q = q / math.sqrt(E)
 
     n_heads, tgt_len, src_len = position_bias.size()[1:]
     assert B % n_heads == 0
@@ -491,6 +492,11 @@ class T5MultiheadAttention(nn.MultiheadAttention):
         super(T5MultiheadAttention, self).__init__(
             embed_dim, num_heads, dropout, bias, add_bias_kv, add_zero_attn, kdim, vdim, batch_first, device, dtype
         )
+        factory_kwargs = {"device": device, "dtype": dtype}
+        self.q_proj_weight = nn.Parameter(torch.empty((embed_dim, embed_dim), **factory_kwargs))
+        self.k_proj_weight = nn.Parameter(torch.empty((embed_dim, self.kdim), **factory_kwargs))
+        self.v_proj_weight = nn.Parameter(torch.empty((embed_dim, self.vdim), **factory_kwargs))
+        self.register_parameter("in_proj_weight", None)
 
     def forward(
         self,
@@ -513,62 +519,35 @@ class T5MultiheadAttention(nn.MultiheadAttention):
         if self.batch_first and is_batched:
             query, key, value = [x.transpose(1, 0) for x in (query, key, value)]
 
-        if not self._qkv_same_embed_dim:
-            attn_output, attn_output_weights, position_bias = t5_multi_head_attention_forward(
-                query,
-                key,
-                value,
-                self.embed_dim,
-                self.num_heads,
-                self.in_proj_weight,
-                self.in_proj_bias,
-                self.bias_k,
-                self.bias_v,
-                self.add_zero_attn,
-                self.dropout,
-                self.out_proj.weight,
-                self.out_proj.bias,
-                compute_relative_attention_bias=compute_relative_attention_bias,
-                relative_attention_num_buckets=relative_attention_num_buckets,
-                relative_attention_max_distance=relative_attention_max_distance,
-                relative_attention_bias=relative_attention_bias,
-                position_bias=position_bias,
-                training=self.training,
-                key_padding_mask=key_padding_mask,
-                need_weights=need_weights,
-                attn_mask=attn_mask,
-                use_separate_proj_weight=True,
-                q_proj_weight=self.q_proj_weight,
-                k_proj_weight=self.k_proj_weight,
-                v_proj_weight=self.v_proj_weight,
-                average_attn_weights=average_attn_weights,
-            )
-        else:
-            attn_output, attn_output_weights, position_bias = t5_multi_head_attention_forward(
-                query,
-                key,
-                value,
-                self.embed_dim,
-                self.num_heads,
-                self.in_proj_weight,
-                self.in_proj_bias,
-                self.bias_k,
-                self.bias_v,
-                self.add_zero_attn,
-                self.dropout,
-                self.out_proj.weight,
-                self.out_proj.bias,
-                compute_relative_attention_bias=compute_relative_attention_bias,
-                relative_attention_num_buckets=relative_attention_num_buckets,
-                relative_attention_max_distance=relative_attention_max_distance,
-                relative_attention_bias=relative_attention_bias,
-                position_bias=position_bias,
-                training=self.training,
-                key_padding_mask=key_padding_mask,
-                need_weights=need_weights,
-                attn_mask=attn_mask,
-                average_attn_weights=average_attn_weights,
-            )
+        attn_output, attn_output_weights, position_bias = t5_multi_head_attention_forward(
+            query,
+            key,
+            value,
+            self.embed_dim,
+            self.num_heads,
+            self.in_proj_weight,
+            self.in_proj_bias,
+            self.bias_k,
+            self.bias_v,
+            self.add_zero_attn,
+            self.dropout,
+            self.out_proj.weight,
+            self.out_proj.bias,
+            compute_relative_attention_bias=compute_relative_attention_bias,
+            relative_attention_num_buckets=relative_attention_num_buckets,
+            relative_attention_max_distance=relative_attention_max_distance,
+            relative_attention_bias=relative_attention_bias,
+            position_bias=position_bias,
+            training=self.training,
+            key_padding_mask=key_padding_mask,
+            need_weights=need_weights,
+            attn_mask=attn_mask,
+            use_separate_proj_weight=True,
+            q_proj_weight=self.q_proj_weight,
+            k_proj_weight=self.k_proj_weight,
+            v_proj_weight=self.v_proj_weight,
+            average_attn_weights=average_attn_weights,
+        )
         if self.batch_first and is_batched:
             return attn_output.transpose(1, 0), attn_output_weights, position_bias
         else:
@@ -747,15 +726,15 @@ class T5Encoder(nn.TransformerEncoder):
         Shape:
             see the docs in Transformer class.
         """
-
         output = src
+        all_outputs = ()
         position_bias = None
         for mod in self.layers:
+            all_outputs = all_outputs + (output,)
             output, position_bias = mod(
                 output, src_mask=mask, src_key_padding_mask=src_key_padding_mask, position_bias=position_bias
             )
-
-        return output
+        return output, all_outputs, position_bias
 
 
 class T5EncoderModel(nn.Module):
@@ -806,14 +785,16 @@ class T5EncoderModel(nn.Module):
         )
         self.norm = T5LayerNorm(d_model)
         self.encoder = T5Encoder(self.encoder_layer, relative_attention_num_buckets, num_heads, num_layers)
-        self.dropout = nn.Dropout(dropout)
+        self.dropout1 = nn.Dropout(dropout)
+        self.dropout2 = nn.Dropout(dropout)
 
     def forward(self, tokens: torch.Tensor):
 
         padding_mask = tokens.eq(self.padding_idx)
-        embeddings = self.dropout(self.token_embeddings(tokens))
-        encoder_output = self.encoder(embeddings, src_key_padding_mask=padding_mask)
+        embeddings = self.dropout1(self.token_embeddings(tokens))
+        encoder_output, all_hidden_states, position_bias = self.encoder(embeddings, src_key_padding_mask=padding_mask)
         encoder_output = self.norm(encoder_output)
-        encoder_output = self.dropout(encoder_output)
+        last_hidden_state = self.dropout2(encoder_output)
+        all_hidden_states = all_hidden_states + (last_hidden_state,)
 
-        return encoder_output
+        return last_hidden_state, all_hidden_states, position_bias


torchtext/prototype/t5/modules.py
```diff
@@ -74,3 +77,60 @@ def _compute_bias(
     values = relative_attention_bias(relative_position_bucket)  # shape (query_length, key_length, num_heads)
     values = values.permute([2, 0, 1]).unsqueeze(0)  # shape (1, num_heads, query_length, key_length)
     return values
+
+
```
pmabbo13, line 84: `_t5_scaled_dot_product_attention()` is modified from [pytorch implementation](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/functional.py#L4814) to incorporate relative attention bias when computing the attention scores. `position_bias` is passed in as a parameter (`line 87`) and is added to the attention score computation in lines `116-121, 126, 130`

torchtext/prototype/t5/modules.py
```diff
@@ -582,3 +584,30 @@ def forward(
             return attn_output.transpose(1, 0), attn_output_weights, position_bias
         else:
             return attn_output, attn_output_weights, position_bias
+
+
```
pmabbo13, line 577: T5 model uses a root-mean-square layer normalization. `T5LayerNorm` implements this and is taken directly [from HF](https://github.com/huggingface/transformers/blob/360719a6a4e86c99cb852e844d30712eabb04baf/src/transformers/models/t5/modeling_t5.py#L239)

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,819 @@
...
+    attn += position_bias
+    attn = F.softmax(attn, dim=-1)
+    if dropout_p > 0.0:
+        attn = F.dropout(attn, p=dropout_p)
+    # (B, Nt, Ns) x (B, Ns, E) -> (B, Nt, E)
+    output = torch.bmm(attn, v)
+    return output, attn
+
+
```
pmabbo13, line 141: `t5_multi_head_attention_forward()` is modified from its [pytorch implementation](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/functional.py#L4909) to incorporate relative attention bias. We've added  parameters `compute_relative_attention_bias`, `relative_attention_bias`, `relative_attention_num_buckets`, `relative_attention_max_distance`, and `position_bias` which are used in lines `432-444, 449`. This implementation was inspired by a similar [HF implementation](https://github.com/huggingface/transformers/blob/360719a6a4e86c99cb852e844d30712eabb04baf/src/transformers/models/t5/modeling_t5.py#L513)

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,819 @@
...
+        return attn_output, attn_output_weights, position_bias
+
+    else:
+        if not is_batched:
+            # squeeze the output if input was unbatched
+            attn_output = attn_output.squeeze(1)
+
+        return attn_output, None, position_bias
+
+
```
pmabbo13, line 474: `T5MultiheadAttention` inherits from [pytorch's MultiheadAttention](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/modules/activation.py#L882). The forward method is a trimmed down version of theirs that includes necessary parameters to incorporate relative attention bias. Pytorch's implementation includes some torchscript optimizations ([lines 1069-1126](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/modules/activation.py#L1059)) that I wasn't too sure about so I omitted for now.

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,819 @@
...
+        variance = hidden_states.to(torch.float32).pow(2).mean(-1, keepdim=True)
+        hidden_states = hidden_states * torch.rsqrt(variance + self.variance_epsilon)
+
+        # convert into half-precision if necessary
+        if self.weight.dtype in [torch.float16, torch.bfloat16]:
+            hidden_states = hidden_states.to(self.weight.dtype)
+
+        return self.weight * hidden_states
+
+
```
pmabbo13, line 604: `T5EncoderLayer` inherits from [pytorch's TransformerEncoderLayer](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/modules/transformer.py#L301). It is initialized to incorporate relative attention bias parameters, use `T5MultiheadAttention` and `T5LayerNorm`, and remove the bias from linear layers (lines `627-636`). The forward method is also taken from the pytorch implementation and modified to incorporate position bias (lines `656-694`). The pytorch method included some torchscript optimizations ([lines 410-457](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/modules/transformer.py#L410)) that I omitted for now.

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,819 @@
...
+            position_bias = attn[2]
+
+        return self.dropout1(x), position_bias
+
+    # feed forward block
+    def _ff_block(self, x: Tensor) -> Tensor:
+        x = self.linear2(self.dropout(self.activation(self.linear1(x))))
+        return self.dropout2(x)
+
+
```
pmabbo13, line 701: `T5Encoder` inherits from [pytorch's TransformerEncoder](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/modules/transformer.py#L165). The initialization is mostly the same, except that we only compute relative attention bias in the first layer, and the resulting tensor is passed up to the higher layers to avoid re-computing. Lines `732-735` incorporate this distinction. The forward method is a trimmed down version of the pytorch implementation's, which included torchscript optimizations ([lines 206-232](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/modules/transformer.py#L206)) that I've omitted for now. They've also implemented something called [convert_to_nested](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/modules/transformer.py#L235) which I'm not too sure of its purpose.

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,819 @@
...
+        output = src
+        position_bias = None
+        for mod in self.layers:
+            output, position_bias = mod(
+                output, src_mask=mask, src_key_padding_mask=src_key_padding_mask, position_bias=position_bias
+            )
+
+        return output
+
+
```
pmabbo13, line 760: `T5EncoderModel` is the complete encoder model that includes the word embeddings, and the final layer normalizations and dropouts

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,819 @@
...
+        num_heads,
+        dropout=0.0,
+        bias=False,
+        add_bias_kv=False,
+        add_zero_attn=False,
+        kdim=None,
+        vdim=None,
+        batch_first=False,
+        device=None,
+        dtype=None,
```
Nayef211, line 488: When inheriting from a Parent class, I would make use of `*args` and `**kwargs` to pass in multiple arguments and keyword arguments to your parent class. In the `T5MultiheadAttention` class, you don't make use of any of the input args during initialization so using `*args` and `**kwargs` and passing them to the Parent class for initialization would be less verbose.

@mthrok wanted to double check if you guys follow this practice in torchaudio?

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,819 @@
...
+        activation: Union[str, Callable[[Tensor], Tensor]] = F.relu,
+        layer_norm_eps: float = 1e-6,
+        batch_first: bool = False,
+        norm_first: bool = True,
+        compute_relative_attention_bias: bool = False,
+        relative_attention_num_buckets: int = 32,
+        relative_attention_max_distance: int = 128,
+        relative_attention_bias: Optional[Tensor] = None,
+        device=None,
+        dtype=None,
```
Nayef211, line 621: Same comment as above about using `*args` and `**kwargs`

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,819 @@
...
+        num_heads: int,
+        num_layers: int,
+        batch_first: bool,
+        relative_attention_num_buckets: int,
+        relative_attention_max_distance: int,
+        padding_idx: int,
+        max_seq_len: int,
+        vocab_size: int,
+    ) -> None:
+        super().__init__()
```
Nayef211, line 778: nit: Let's add docstrings explaining the purpose of the class and all of the input arguments before we actually merge in the implementation

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,819 @@
...
+            norm_first=True,
+            relative_attention_num_buckets=relative_attention_num_buckets,
+            relative_attention_max_distance=relative_attention_max_distance,
+        )
+        self.norm = T5LayerNorm(d_model)
+        self.encoder = T5Encoder(self.encoder_layer, relative_attention_num_buckets, num_heads, num_layers)
+        self.dropout = nn.Dropout(dropout)
+
+    def forward(self, tokens: torch.Tensor):
+
```
Nayef211, line 812: nit: remove new line

Nayef211 (commented):
Overall the implementation looks fine to me in terms of class separation and design. To verify correctness, lets compare the outputs of the HF implementation with ours as we discussed.

@abhinavarora it would be useful to get your feedback on whether all of the changes from the T5 paper are represented correctly in the code.

torchtext/prototype/t5/modules.py
mthrok: I think it depends on the context. There are pros and cons for both approach. If the signature of the constructor is a public API, then explicitly listing could be helpful when documenting. (at the same time, it could appear cluttered) Even if it is not public, if it is expected that the signature can change many time then, explicit is better.

Assuming that this class is just a component of the T5 model class, won't change, and not intended for direct use, the use of `args` and `kwargs` will make the code shorter, and I think that is reasonable.

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,819 @@
...
+        add_bias_kv=False,
+        add_zero_attn=False,
+        kdim=None,
+        vdim=None,
+        batch_first=False,
+        device=None,
+        dtype=None,
+    ) -> None:
+
+        super(T5MultiheadAttention, self).__init__(
```
mthrok, line 491: `super(<CLASS_NAME>, self)` is the syntax from Python 2 era. It can be as simple as

```suggestion
        super().__init__(
```

torchtext/prototype/t5/modules.py
mthrok: Is it absolutely necessary that this implementation is a plain function, instead of a method in T5 class?

I see that many attributes from the model instance is being passed, which makes me wonder what is the value-add in T5 model class implementation.

PyTorch's implementation (which was originated from torchtext) is trying to cover many use cases, so the signature is complex, where as this implementation is only required to serve the needs of T5 model, can't it be simplified?

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,819 @@
...
+        encoder_layer,
+        relative_attention_num_buckets,
+        num_heads,
+        num_layers=12,
+        norm=None,
+        enable_nested_tensor=True,
+    ):
+        super(T5Encoder, self).__init__(encoder_layer, num_layers, norm, enable_nested_tensor)
+
+        first_layer = copy.deepcopy(encoder_layer)
```
mthrok, line 732: I am not sure if the use of deep copy is guaranteed to keep working with PyTorch modules.

Instead of expecting an instance, it is more elegant to accept parameters required to build the component and instantiate encoders as much as needed.

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,76 @@
+import math
+
+import torch
+
+
```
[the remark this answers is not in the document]
abhinavarora: I saw [this](https://github.com/pytorch/pytorch/blob/dc169d53aa266560750ea25ee0cf31c7e614550d/aten/src/ATen/native/Distributions.h#L120) as one way of citing in Pytorch core. Maybe we could do something similar.

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,800 @@
...
+from typing import Optional, Tuple, Union, Callable
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch import Tensor
+
+
+# NOTE: taken from HF; used to compute relative attention bias
+def _relative_position_bucket(relative_position, bidirectional=True, num_buckets=32, max_distance=128):
```
abhinavarora, line 13: Let's add type annotations for consistency?

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,800 @@
...
+    relative_position_if_large = torch.min(
+        relative_position_if_large, torch.full_like(relative_position_if_large, num_buckets - 1)
+    )
+
+    relative_buckets += torch.where(is_small, relative_position, relative_position_if_large)
+    return relative_buckets
+
+
+# NOTE: modified from HF; used to compute relative attention bias
+def _compute_bias(
```
abhinavarora, line 59: Let's add type annotations for consistency?

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,800 @@
...
+        num_buckets=relative_attention_num_buckets,
+        max_distance=relative_attention_max_distance,
+    )
+    values = relative_attention_bias(relative_position_bucket)  # shape (query_length, key_length, num_heads)
+    values = values.permute([2, 0, 1]).unsqueeze(0)  # shape (1, num_heads, query_length, key_length)
+    return values
+
+
+# NOTE: modified from torch.nn.functional._scaled_dot_product_attention to incorporate relative attention bias
+def _t5_scaled_dot_product_attention(
```
parmeet, line 86: @pmabbo13  for easier follow-up, could you please add comment directly in the source code that is modified to incorporate relative attention bias?

torchtext/prototype/t5/modules.py
parmeet: I agree with @mthrok with regard to usage. I think in general, it wouldn't hurt to keep the arguments explicit :) (Imaging some enthusiastic user who quickly want to try out this new T5 style MHA in their work, it would be bit easier, of-course it is at their risk since we haven't yet made the API public).

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,800 @@
...
+            v_proj_weight=self.v_proj_weight,
+            average_attn_weights=average_attn_weights,
+        )
+        if self.batch_first and is_batched:
+            return attn_output.transpose(1, 0), attn_output_weights, position_bias
+        else:
+            return attn_output, attn_output_weights, position_bias
+
+
+# NOTE: Taken from HF
```
parmeet, line 557: Let's also provide the link of the source code.

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,800 @@
...
+    ).to(torch.long)
+    relative_position_if_large = torch.min(
+        relative_position_if_large, torch.full_like(relative_position_if_large, num_buckets - 1)
+    )
+
+    relative_buckets += torch.where(is_small, relative_position, relative_position_if_large)
+    return relative_buckets
+
+
+# NOTE: modified from HF; used to compute relative attention bias
```
parmeet, line 58: Let's also provide the link to original source code in comment.

torchtext/prototype/t5/modules.py
```diff
@@ -0,0 +1,800 @@
...
+import warnings
+from typing import Optional, Tuple, Union, Callable
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch import Tensor
+
+
+# NOTE: taken from HF; used to compute relative attention bias
```
parmeet, line 12: Let's also provide the link to source code as comment

Issue #1797, opened 2022-06-20
Unable to compile torchtext v0.13 with CUDA


cdeepali:
## 🐛 Bug

Seeing the following error while building torchtext with FORCE_CUDA=1:

```
-- Found CUDA: /usr/local/cuda (found version "11.4")
-- The CUDA compiler identification is unknown
-- Detecting CUDA compiler ABI info
```

```
CMake Error in <myhome>/work/build/temp.linux-ppc64le-3.9/CMakeFiles/CMakeTmp/CMakeLists.txt:
  CUDA_ARCHITECTURES is empty for target "cmTC_94ffb".


CMake Error in <myhome>/work/build/temp.linux-ppc64le-3.9/CMakeFiles/CMakeTmp/CMakeLists.txt:
  CUDA_ARCHITECTURES is empty for target "cmTC_94ffb".


CMake Error at <myhome>/_build_env/share/cmake-3.19/Modules/CMakeDetermineCompilerABI.cmake:48 (try_compile):
  Failed to generate test project build system.
```
**To Reproduce** :

Build torchtext tag https://github.com/pytorch/text/releases/tag/v0.13.0-rc2 with FORCE_CUDA=1

**Expected behavior** .
It should build successfully.

**Environment**

OS: Red Hat Enterprise Linux 8.5 (Ootpa) (ppc64le)
GCC version: Anaconda GCC 11.2
Python version: 3.9.12 (main, Apr  5 2022, 07:09:29)
CUDA runtime version: 11.4.152
```

**Comment by parmeet** [comment]
Hi @cdeepali, torchtext does not depend on CUDA. Is there a reason you are building torchtext with FORCE_CUDA=1?

**Comment by cdeepali** [comment]
Yes thats correct. But my build is failing without it as well, so I had just tried passing FORCE_CUDA=1. But my build fails even if I don't pass it.

**Comment by atalman** [comment]
@cdeepali could you please post complete compile log with errors ?

**Comment by cdeepali** [comment]
I am actually trying to build torchtext when CUDA enabled pytorch is installed in the build environment. As a result cmake detects CUDA on its own and eventually fails:

full log:

```
++ python -V
++ cut -d ' ' -f 2
++ cut -d . -f 1,2
+ PYTHON_VERSION=3.9
[...]

PR #1814, opened 2022-06-29
Fixing build when CUDA enabled torch is installed

Nayef211:
- Addresses #1797


CMakeLists.txt
```diff
@@ -51,13 +51,15 @@ option(BUILD_TORCHTEXT_PYTHON_EXTENSION "Build Python extension" OFF)
 
 set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH};${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
-find_package(Torch REQUIRED)
+set(TORCH_INSTALL_PREFIX $ENV{TORCH_INSTALL_PREFIX})
```
mthrok, line 54: I think this should be an CMake option and being passed from [`setup.py`](https://github.com/pytorch/text/blob/main/tools/setup_helpers/extension.py#L62). So that the logic is hidden from user.

Nayef211: Done. Lmk if the changes look okay

CMakeLists.txt
```diff
@@ -50,14 +50,16 @@ endif()
 option(BUILD_TORCHTEXT_PYTHON_EXTENSION "Build Python extension" OFF)
 
 set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH};${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+set(TORCH_INSTALL_PREFIX "${CMAKE_PREFIX_PATH}/../.." CACHE STRING "Install path for torch")
```
mthrok, line 53: This seems to be overwriting the value provided by Python process. Is this correct? I think ‘option’ is more suited here.

Nayef211: According to the [docs](https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry), "Since cache entries are meant to provide user-settable values this does not overwrite existing cache entries by default. Use the FORCE option to overwrite existing entries.".

IIUC, `option` is only used for booleans ([docs](https://cmake.org/cmake/help/latest/command/option.html)).

mthrok: Ah I see. I mistook them.

mthrok (approved):
LGTM in condition that you tested this manually and confirmed it works.

PR #1815, opened 2022-07-05
[Docs] Updating usage docs for Regex Tokenizer

Nayef211:
## Description
- Updating docs for regex tokenizer to include 
  - caveats with RE2 regex library
  - tokenization example with a single regex pattern
- Adding the tokenizer to generated docs

### DIFF
--- a/docs/source/transforms.rst
+++ b/docs/source/transforms.rst
@@ -44,7 +44,6 @@ BERTTokenizer
 
    .. automethod:: forward
 
-
 VocabTransform
 --------------
 
--- a/docs/source/transforms.rst
+++ b/docs/source/transforms.rst
@@ -17,28 +17,28 @@ SentencePieceTokenizer
    .. automethod:: forward
 
 GPT2BPETokenizer
-----------------------
+----------------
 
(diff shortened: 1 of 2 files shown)


docs/source/transforms.rst
```diff
@@ -30,14 +30,20 @@ CLIPTokenizer
 
    .. automethod:: forward
 
+RegexTokenizer
+----------------------
```
parmeet, line 34:
```suggestion
RegexTokenizer
--------------
```

docs/source/transforms.rst
```diff
@@ -30,14 +30,20 @@ CLIPTokenizer
...
+
+.. autoclass:: RegexTokenizer
+
+   .. automethod:: forward
+
 BERTTokenizer
 ----------------------
```
parmeet, line 41:
```suggestion
BERTTokenizer
-------------
```

docs/source/transforms.rst
parmeet: nit: underline equal length as text

parmeet (approved):
LGTM!

Issue #1756, opened 2022-06-01
Multi30K dataset link is broken


xuzhao9:
The link to Multi30K dataset at `http://www.quest.dcs.shef.ac.uk/wmt16_files_mmt/training.tar.gz` is broken: https://github.com/pytorch/text/blob/73bf4fa8cedc12d910ab76190e446bd2e47a8325/torchtext/datasets/multi30k.py#L16

parmeet:
Yupp, looks like the server is down :(.

muskbing:
> Yupp, looks like the server is down :(.

Any solutions?

chenghan1995:
I meet the same problem, is there any solution?

chenghan1995:
![image](https://user-images.githubusercontent.com/37642597/172782290-c90ae43f-4f58-434a-bdb3-3c6df9b423bd.png)

chenghan1995:
![image](https://user-images.githubusercontent.com/37642597/172782370-26dc20dc-3d44-46b3-b22d-9126cb43acfa.png)

Nayef211:
@chenghan1995 @muskbing unfortunately, we're not responsible for hosting the datasets. I'd recommend waiting for their server to come back up or reaching out directly to the organization that hosts the dataset. In this case this would be the University of Sheffield.

cc @parmeet I wonder if you know of a way to get in contact with the team that hosts this dataset?

muskbing:
> @chenghan1995 @muskbing unfortunately, we're not responsible for hosting the datasets. I'd recommend waiting for their server to come back up or reaching out directly to the organization that hosts the dataset. In this case this would be the University of Sheffield.
> 
> cc @parmeet I wonder if you know of a way to get in contact with the team that hosts this dataset?

I've send email to the owner of the dataset
https://www.statmt.org/wmt16/multimodal-task.html#task1
![image](https://user-images.githubusercontent.com/53995462/172785451-f55927ba-59b4-44b9-8d67-d95a2aa99405.png)

email address is:lspecia@gmail.com

But there is no response. 

I wander is there anyone who has the data file

neychev:
Found a local copy of the dataset and uploaded it to github (it's rather small). For now it is available via this link: https://github.com/neychev/small_DL_repo/tree/master/datasets/Multi30k

Just in case, all rights belong to the original authors of the dataset, this is only a temporal copy for convenience.

neychev:
Example code to make it work (tested on Colab):

```
!pip install torchdata

[...]

PR #1816, opened 2022-07-06
Fix Multi30k dataset urls

Nayef211:
# Description
- Closes https://github.com/pytorch/text/issues/1756
- Will update the URLs to the original once the server is back up

### DIFF
--- a/torchtext/datasets/multi30k.py
+++ b/torchtext/datasets/multi30k.py
@@ -12,6 +12,7 @@ if is_module_available("torchdata"):
     from torchdata.datapipes.iter import FileOpener, IterableWrapper
     from torchtext._download_hooks import HttpReader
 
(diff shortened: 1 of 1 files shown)


parmeet (approved):
LGTM, thanks for providing the temporary fix. Although i am bit concerned changing the hash for test dataset (not sure what leads to this change), let's go ahead with this change to make the dataset usable.

Nayef211:
> LGTM, thanks for providing the temporary fix. Although i am bit concerned changing the hash for test dataset (not sure what leads to this change), let's go ahead with this change to make the dataset usable.

@parmeet I was wondering the same thing but unfortunately I am not able to compare this test split against the test split of the original dataset as I don't have it readily available 😞

PR #1820, opened 2022-07-11
Use TA functional for adding tokens to the beginning and end of input

parmeet:
This PR updates example to use TA functional for adding tokens to the beginning and end of input.


Nayef211 (approved):
LGTM

PR #1826, opened 2022-07-12
add layer norm module for t5 model

pmabbo13:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #1829
* #1828
* #1827
* __->__ #1826
* #1833
* #1832
* #1831
* #1830
* #1825

PR #1827, opened 2022-07-12
add t5 layer module that can be used for both encoder or decoder stack

pmabbo13:
* #1829
* #1828
* __->__ #1827
* #1826
* #1833
* #1832
* #1831
* #1830
* #1825


torchtext/prototype/t5/modules.py
```diff
@@ -557,3 +557,170 @@ def forward(self, hidden_states):
             hidden_states = hidden_states.to(self.weight.dtype)
 
         return self.weight * hidden_states
+
+
+class T5Layer(nn.Module):
```
Nayef211, line 562: Add comment with link to where this is taken from. Also let's make sure to use permalinks (i.e. something like this https://github.com/huggingface/transformers/blob/8581a798c0a48fca07b29ce2ca2ef55adcae8c7e/src/transformers/models/t5/modeling_t5.py#L239).

Read https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files for more info

Nayef211 (commented):
Let's highlight what the modifications are with the use of a PR comment

pmabbo13:
> Let's highlight what the modifications are with the use of a PR comment

updated!

pmabbo13:
# Description

The T5Layer implementation is very similar to the [nn.TransformerEncoderLayer](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/modules/transformer.py#L301) and [nn.TransformerDecoderLayer](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/modules/transformer.py#L483) implementations. The main differences are that:

1. T5Layer generalizes to be used as either an encoder or decoder layer. For the encoder we perform self-attention + feed-forward. For the decoder we perform self-attention + cross-attention + feed-forward. If the boolean parameter `is_decoder = True`, the forward method will perform cross-attention after performing self-attention. Otherwise, it will skip the cross-attention step and go directly to the feed-forward step.
2. T5Layer takes in parameters `relative_attention_num_buckets`, `relative_attention_max_distance`, `compute_relative_attention` and `relative_attention_bias`. This is information passed into the self-attention block so that it can include a relative attention bias term when computing attention scores.

torchtext/prototype/t5/modules.py
```diff
@@ -444,3 +444,171 @@ def forward(self, hidden_states):
...
+                d_model, nhead, is_decoder=is_decoder, dropout=dropout, device=device, dtype=dtype
+            )
+            self.norm3 = T5LayerNorm(d_model, eps=layer_norm_eps)
+            self.dropout4 = nn.Dropout(dropout)
+
+        if isinstance(activation, str):
+            if activation == "relu":
+                self.activation = F.relu
+            elif activation == "gelu":
+                self.activation = F.gelu
```
parmeet, line 528: raise exception if the activation str is neither `relu` or `gelu`

parmeet (approved):
Overall LGTM!

PR #1828, opened 2022-07-12
add t5 stack that can function as either the encoder or decoder of a t5 model

pmabbo13:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #1829
* __->__ #1828
* #1827
* #1826
* #1833
* #1832
* #1831
* #1830
* #1825

PR #1829, opened 2022-07-12
add t5 model that can function as both encodery-only or encoder-decoder model

pmabbo13:
* __->__ #1829
* #1828
* #1827
* #1826
* #1833
* #1832
* #1831
* #1830
* #1825


torchtext/prototype/t5/model.py
```diff
@@ -0,0 +1,178 @@
+from typing import Optional, Union, Callable
+
+import torch
+import torch.nn as nn
+from torch import Tensor
+
+from .modules import T5Stack, T5LayerNorm
+
+
+class T5Model(nn.Module):
```
Nayef211, line 10: permalink to original implementation from HF

Nayef211 (commented):
@parmeet I wonder if you think it makes sense to update our docs to include the `T5Model` class as well as the yet to be created `T5Bundle`. Similarly is there a reason why we don't include the `RobertaModel` class in the docstring if it's a public facing component?

pmabbo13:
# Description

The T5Model implementation is very similar to the [nn.Transformer](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/modules/transformer.py#L16) implementation, with some additional functionality

1. Takes a tokenized encoder input sequence and decoder input sequence and transforms them into word embeddings.
2. Computes the padding masks for the input sequences based on the `padding_idx` input argument when initializing the model.
3. Generates a causal mask for decoder self-attention unless one has already been provided via the `decoder_mask` input argument to the forward method.
4. Returns the output of the final layers to the encoder and decoder, the output at each layer of the encoder and decoder, the self-attentions scores of each layer of the encoder and decoder, the cross-attention scores of each layer of the decoder
5. Based on input parameter `encoder_only` when initializing the model, will just run the encoder portion and output its corresponding results.

torchtext/prototype/t5/model.py
```diff
@@ -0,0 +1,189 @@
...
+            self.dropout3 = nn.Dropout(dropout)
+            self.dropout4 = nn.Dropout(dropout)
+
+    def forward(
+        self,
+        encoder_tokens: Tensor,
+        decoder_tokens: Tensor = None,
+        encoder_mask: Optional[Tensor] = None,
+        decoder_mask: Optional[Tensor] = None,
+    ) -> Tensor:
```
parmeet, line 123: Let's change the annotation to match output type (Dict[str, Tensor])?

parmeet (approved):
Overall LGTM!

parmeet:
> I wonder if you think it makes sense to update our docs to include the `T5Model` class as well as the yet to be created `T5Bundle`. 

This is just a prototype feature. So it may not be necessary to include it in the docs. I don't see other domains doing this as well?

> Similarly is there a reason why we don't include the `RobertaModel` class in the docstring if it's a public facing component?

hmm, that's a good catch. Not really. Looks like we missed including the docs for it, perhaps we were only focusing on the Roberta Bundler API that expose this model to users..

Nayef211 (approved):

PR #1832, opened 2022-07-13
computing attention scores using relative attention bias

pmabbo13:
* (to be filled)

### DIFF
--- a/torchtext/prototype/t5/modules.py
+++ b/torchtext/prototype/t5/modules.py
@@ -14,7 +14,7 @@
 # */
 
 import math
-from typing import Optional, Union, Callable
+from typing import Optional, Union, Tuple, Callable
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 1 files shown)


torchtext/prototype/t5/modules.py
```diff
@@ -64,6 +65,65 @@ def __init__(
     def forward():
         pass
 
+    def _t5_scaled_dot_product_attention(
```
Nayef211, line 68: Can we leave a comment with the link to the original implementation?

Nayef211 (approved):

pmabbo13:
# Description

Having computed the relative attention bias term, this method computes the attention scores.  The implementation is very similar to the [nn.Functional._scaled_dot_product_attention](https://github.com/pytorch/pytorch/blob/5953fd9133c0bdcc0158acf1472fac403bc5f636/torch/nn/functional.py#L4814), expect that we pass in `position_bias` as an input argument so that relative attention bias can be incorporated in the computation of the attention scores.

Since the input tensors to this function are 4-dimensional, we replace the `torch.baddbmm`  and `torch.bmm` with `torch.matmul`. Since we are no longer using `torch.baddbmm` where `attn_mask` was passed as an input argument, we instead add the `attn_mask` directly to `position_bias` to ensure the mask is still applied before the softmax is taken to get the final attention scores.

PR #1836, opened 2022-07-13
Add TA  Tensor creation  operation to the benchmark

parmeet:
This PR adds the tensor creation operation to the TA ops benchmark code.

### Current Results

Running on MacBook pro for SST2 dataset with 67349 rows with average number of tokens per row is 11.35.

Torchtext  tensor creation process: ~1.19 sec
TorchArrow tensor creation process: ~21.91 sec


wenleix (approved):
LGTM

PR #1841, opened 2022-07-18
fix OBO error while reading vocab files with empty lines in BERT tokenizer

### LINKED ISSUE #1840: Off by one (OBO) error in torchtext implementation of BERTTokenizer
The output token IDs are off by one for certain vocab files. Per the initial investigation by @geoffreyangus, apparently torchtext is not taking newlie \n into account in the vocab file, whereas HF does count it hence the offset by 1.

Reproducible gist by @geoffreyangus https://gist.github.com/geoffreyangus/ae8a68900a46e4b1b3fec4926b7fa51b

parmeet:
fixes #1840

### DIFF
--- a/torchtext/csrc/bert_tokenizer.cpp
+++ b/torchtext/csrc/bert_tokenizer.cpp
@@ -29,6 +29,31 @@ std::string BERTEncoder::kUnkToken = "[UNK]";
 
 int kMaxCharsPerWords = 100;
 
+static std::vector<std::string> _read_vocab(std::string file_path) {
+  std::ifstream fin(file_path, std::ios::in);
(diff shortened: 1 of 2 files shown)


Nayef211 (approved):
Overall LGTM. I was wondering if we considered modifying the existing [`_load_vocab_from_file`](https://github.com/pytorch/text/blob/e9640517ee92389a9a14ee6d29e6aea1f841d0d7/torchtext/csrc/vocab.cpp#L265) implementation to add a flag to include newline characters? I know that the existing implementation has the options for multithreading as well as other perf optimizations (which I guess may not be too relevant to load in a smaller vocab file as might be the case for the BERT vocab).

PR #1846, opened 2022-07-18
Bundler API for TorchText T5 Model

pmabbo13:


Nayef211 (approved):
LGTM! Do we plan to add the end-to-end test in a followup PR? For reference, the integration tests can be found [here](https://github.com/pytorch/text/blob/main/test/integration_tests/test_models.py) and the unit tests for the existing RobertaBundle can be found [here](https://github.com/pytorch/text/blob/main/test/models/test_models.py)

torchtext/prototype/models/t5/bundler.py
```diff
@@ -0,0 +1,152 @@
...
+        >>> encoder_conf = T5Conf(encoder_only=True)
+        >>> model = T5Bundle.build_model(config=encoder_conf, checkpoint=model_weights_path)
+    """
+
+    _config: T5Conf
+    _path: Optional[str] = None
+
+    def get_model(
+        self,
+        *,
```
pmabbo13, line 50: I saw for the Roberta bundler that `get_model` and `build_model` accepted `*` as an input argument, however I'm not too sure why that is since I don't see it being used in the methods. I kept them here for now, but want to verify their purpose before removing.

pmabbo13:
> LGTM! Do we plan to add the end-to-end test in a followup PR? For reference, the integration tests can be found [here](https://github.com/pytorch/text/blob/main/test/integration_tests/test_models.py) and the unit tests for the existing RobertaBundle can be found here (https://github.com/pytorch/text/blob/main/test/models/test_models.py)

I intended to add the tests in a separate PR, but I have no opposition to adding it to this PR.

torchtext/prototype/models/t5/bundler.py
parmeet: This will enforce user to  pass all arguments after `*` using keywords only in function call. They can no longer be passed as positional arguments. This may be useful when changing API in the future since the arguments were passed using their names, changing their positions would not cause breakage in existing code base.

torchtext/prototype/models/t5/bundler.py
```diff
@@ -0,0 +1,152 @@
...
+        >>> output = model(model_input)['encoder_output']
+        >>> output.shape
+        torch.Size([2, 6, 768])
+
+    Example - Pretrained base t5 model
+        >>> import torch, torchtext
+        >>> t5_base = torchtext.prototype.models.T5_BASE
+        >>> model = t5_base.get_model()
+        >>> encoder_input = torch.tensor([[1,2,3,4,5,6],[7,8,9,0,0,0]])
+        >>> decoder_input = torch.tensor([[0],[0]])
```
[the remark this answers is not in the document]
parmeet: I wonder if can avoid passing this at all and construct the decoder input internally since we already know the padding IDX from the model config?

parmeet:
> Overall LGTM! Do we plan to add integration tests in follow-up PR?

Oops just saw the answer here https://github.com/pytorch/text/pull/1846#issuecomment-1189144246. Feel free to ignore :)

Nayef211:
> I intended to add the tests in a separate PR, but I have no opposition to adding it to this PR.

Gotcha that should be fine. I do agree with @parmeet's comment about not expecting the user to pass in the `decoder_input` if we already know the padding indices from the input config to the model.

PR #1848, opened 2022-07-19
Testing T5Model

pmabbo13:
# Description
Add integration tests for both the T5 model (encoder-only and encoder-decoder)
# Process
* Upload expected output for example input sequence, sourced from HuggingFace after passing in same input to corresponding HF model -> `test/asset/t5.base.encoder.output.pt`, `test/asset/t5.base.output`
* Add integration tests to verify that T5_BASE_ENCODER and T5_BASE outputs match the references from HF
* Test that bundler methods ( `build_model()`, `get_model()`, `config()` ) are working as expected, and returning the correct error messages when invalid input args are provided.

# Testing
`pytest test/prototype/models/test_models.py`

# Follow-Up

- [ ] Add testing for model training after LM head has been added (need LM head otherwise cannot compute cross-entropy loss)
- [ ] Investigate if model is torch-scriptable. This is unclear at the moment because the pytorch transformer implementation indicates that transformers are not torch-scriptable if normalization happens first (which it does for T5), and when the query, key, and values are not equal (which is the case for transformer and t5 decoders).


Nayef211 (approved):
LGTM! When we decide to move the model and the tests to the main folder, we should look into parameterizing the model and integration tests since all of the tests are almost exactly the same as the ones we have [here](https://github.com/pytorch/text/blob/main/test/integration_tests/test_models.py) and [here](https://github.com/pytorch/text/blob/main/test/models/test_models.py).

test/prototype/models/test_models.py
```diff
@@ -0,0 +1,130 @@
...
+                num_attention_heads=2,
+                num_encoder_layers=2,
+            )
+            T5Bundle.build_model(
+                config=dummy_t5_conf,
+                freeze_model=True,
+                checkpoint=1,
+            )
+
+    def test_t5_bundler_conf_property(self):
```
parmeet, line 105: nit: Perhaps we can test only with encoder-decoder conf, seems like testing both with encoder only and encoder-decoder conf could be redundant?

pmabbo13: sure! do you mean for all the tests (since the only difference between testing the encoder and encoder-decoder models is the value passed for `encoder-only` when initializing the configs), or just remove the redundancy for `test_t5_bundler_conf_property()`?

parmeet: right, so we could either test with encode-only or encoder-decoder conf, but doing with both might be redundant.

parmeet: And I  meant for tests within `test_t5_bundler_conf_property()` only.

pmabbo13: okay, i'll remove the one for encoder-only. thanks!

PR #1852, opened 2022-07-21
T5Transform text pre-processing for t5 model

pmabbo13:
# Description
Add a transformation class that takes string inputs and prepares them to be passed into a T5 encoder. The transformation class should also have a decode method that translates token ids back into strings. This will be used to translate the sequences generated by the T5 decoder.

# Process
T5Transform is instantiated by providing a path to a pre-trained SentencePiece model, the maximum sequence length (used for truncation), the padding index, and the end-of-sequence index.

Its forward method accepts a single string, or a batch of strings, and uses the pre-trained SentencePiece model to tokenize the string(s) and translate the tokens to their corresponding ids. Then the resulting sequences are truncated and an end-of-sequence token is added to each. Finally, the sequences are padded to the length of the longest sequence in the batch.

Its decode method accepts a single list of token ids, or a batch of these lists to represent multiple sequences. The pre-trained SentencePiece model is then used to translate them back into tokens and merge the tokens into a single string per sequence.

# Test

We test that the forward method correctly translates an input string (batched and un-batched) into the appropriate token ids, with the special tokens added where necessary. We also test that the decode method correctly translates token ids (batched and un-batched) into the correct strings. The torch-scripted versions of these transforms are also tested.

`pytest test/prototype/models/test_transforms.py`


test/prototype/test_transforms.py
```diff
@@ -256,3 +257,26 @@ def test_mask_transform_mask_bos(self):
             masked_tokens, _, _ = mask_transform(self.sample_token_ids)
             exp_tokens = torch.tensor([[5, 5, 5, 5, 5], [5, 5, 5, 4, 4]])
             self.assertEqual(exp_tokens, masked_tokens)
+
+    def _t5tokenizer(self, test_scripting):
+        asset_name = "https://huggingface.co/t5-base/resolve/main/spiece.model"
```
Nayef211, line 262: We don't try not to rely on any network downloads for our unit tests. For other tests related to the SPM tokenizer, you will see we rely on an `spm_example.model` file contained in the assets directory. Can we use the same asset when testing our T5Transform?

pmabbo13: sorry, left this from my own local tests. will update

test/prototype/test_transforms.py
```diff
@@ -256,3 +257,26 @@ def test_mask_transform_mask_bos(self):
...
+        if test_scripting:
+            transform = torch.jit.script(transform)
+
+        actual = transform(["Hello World!, how are you?"])
+        expected = torch.tensor([[8774, 1150, 55, 6, 149, 33, 25, 58, 1]])
+        self.assertEqual(actual, expected)
+
+        actual = transform("Hello World!, how are you?")
+        expected = torch.tensor([8774, 1150, 55, 6, 149, 33, 25, 58, 1])
+        self.assertEqual(actual, expected)
```
Nayef211, line 274: Are there any differences between these 2 test cases?

pmabbo13: yeah, in the first case the input is passed in as a list (format for handling batched sequences) vs the input for the second case is just a single string. I can make the first test more obviously batched by adding a second sequence if that's easier to follow.

torchtext/prototype/models/t5/t5_transform.py
```diff
@@ -0,0 +1,47 @@
...
+
+class T5Transform(nn.Module):
+    def __init__(self, sp_model_path: str, max_seq_len: int, eos_idx: int, padding_idx: int):
+        super().__init__()
+
+        self.sp_model = load_sp_model(get_asset_local_path(sp_model_path))
+        self.max_seq_len = max_seq_len
+        self.eos_idx = eos_idx
+        self.padding_idx = padding_idx
+        self.pipeline = T.Sequential(T.Truncate(self.max_seq_len), T.AddToken(token=self.eos_idx, begin=False))
```
parmeet, line 19: Curious if you have tried to add self.encode to T.Sequential and see if it works?

pmabbo13: I have but it does not work. The error is that Sequential only accepts Module subclasses

test/prototype/models/test_transforms.py
```diff
@@ -0,0 +1,58 @@
+import torch
+from test.common.assets import get_asset_path
+from test.common.torchtext_test_case import TorchtextTestCase
+from torchtext.prototype.models import T5Transform
+
+
+class TestTransforms(TorchtextTestCase):
+    def _t5tokenizer(self, test_scripting):
+        asset_name = "t5_tokenizer_base.model"
```
parmeet, line 9: Instead of adding new asset file, we should probably work with existing assets if available. In this case, shall we try working with `spm_example.model` ?

test/prototype/models/test_transforms.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def test_t5tokenizer(self):
+        """test tokenization on string input (batched and un-batched)"""
+        self._t5tokenizer(test_scripting=False)
+
+    def test_t5tokenizer_jit(self):
+        """test tokenization with scripting on string input (batched and un-batched)"""
+        self._t5tokenizer(test_scripting=True)
+
+    def _t5tokenizer_decode(self, test_scripting):
```
parmeet, line 35: nit: I wonder if we just have both `decode` and `encode` methods tested in the same function i.e `_t5tokeniser`?

test/prototype/models/test_transforms.py
pmabbo13: @Nayef211 and I were actually debating how best to approach this, because if we used `spm_example.model` then we'd essentially be testing for functional correctness. But since T5Transform is so similar to SentencePieceTokenizer except that it includes additional transformations specific to T5, we thought it made more sense to tailor the test towards t5 specifically as opposed to a general spm model.

Nayef211: @parmeet if we use the existing `spm_example.model`, these tests do not add as much value as we already have specific tests for the SentencePiece tokenizer. As @pmabbo13 mentioned, if we want to test that the output of the T5Transform is equal to that of the T5Transform in HF, then it would make sense to make use of the spm model specific to T5. Also the asset is around 700 KB which is less than some of the existing assets we've checked in. Lmk what you think!

torchtext/prototype/models/t5/t5_transform.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def __init__(self, sp_model_path: str, max_seq_len: int, eos_idx: int, padding_idx: int):
+        super().__init__()
+        self.sp_model = load_sp_model(get_asset_local_path(sp_model_path))
+        self.max_seq_len = max_seq_len
+        self.eos_idx = eos_idx
+        self.padding_idx = padding_idx
+        self.pipeline = T.Sequential(T.Truncate(self.max_seq_len), T.AddToken(token=self.eos_idx, begin=False))
+
+    def forward(self, input: Any) -> Any:
```
Nayef211, line 40: Is there a reason we specify the input as `Any` instead of `Union[str, List[str]]`?

torchtext/prototype/models/t5/t5_transform.py
```diff
@@ -0,0 +1,85 @@
...
+        :type input: Union[str, List[str]]
+        :return: Tokenized text that has been truncated, appended with end-of-sequence token, and padded
+        :rtype: Union[List[int], List[List[int]]]
+        """
+        tokens = self.encode(input)
+        out = to_tensor(self.pipeline(tokens), padding_value=self.padding_idx)
+        return out
+
+    @torch.jit.export
+    def encode(self, input: Any) -> Any:
```
Nayef211, line 52: same question about typing

torchtext/prototype/models/t5/t5_transform.py
```diff
@@ -0,0 +1,85 @@
...
+            for text in input:
+                tokens.append(self.sp_model.EncodeAsIds(text))
+            return tokens
+        elif torch.jit.isinstance(input, str):
+            return self.sp_model.EncodeAsIds(input)
+        else:
+            raise TypeError("Input type not supported")
+
+    @torch.jit.export
+    def decode(self, input: Any) -> Any:
```
Nayef211, line 70: same question about typing

test/prototype/models/test_transforms.py
parmeet: I understand the overall sentiment here and it's a good argument for adding the actual asset file. But then this make me wonder if we are really unit-testing the functional correctness of the transform implementation or actually testing the asset file :). 

That said, I think we would also be needing this for integration testing, since we need a real output in there instead of dummy output from any SPM model file. So I think I agree with you both, adding the actually asset file would make sense!

torchtext/prototype/models/t5/t5_transform.py
pmabbo13: I initially had them typed, but noticed that [SentencePieceTokenizer](https://github.com/pytorch/text/blob/af03d7101c3c86aee9ecb0dd724bb0bf5b2b09da/torchtext/transforms.py#L56) had them as `Any` so deferred to that. I will revert it back.

parmeet: > Is there a reason we specify the input as `Any` instead of `Union[str, List[str]]`?

I guess @pmabbo13 might have followed what we did in our transform implementation where we always use `Any` as type for input. The reason for this is to ensure when transforms are combined in `SequentialTransform`, the overall transform is still scriptable. More details about this issue can be found  [here](https://github.com/pytorch/text/pull/1453#issue-1071329083). As for T5Transform, if we do not expect it to be used in `SequentialTransform` and treat it as a standalone one, I agree that we could just use the right annotation types as suggested above.

parmeet (approved):
LGTM! Thanks @pmabbo13 for adding the transform class.

PR #1856, opened 2022-07-22
Add text transform to T5 bundler

pmabbo13:
# Description
Expose `T5Transform` in `T5Bundle`, which allows for text pre-processing of input strings before feeding them into a T5 model.

# Process
Create a callable attribute in `T5Bundle` that will be used to pre-process the input text to a T5 model and also decode the model's output back into text. The transform objects for the encoder-only and encoder-decoder base models are instantiated using `T5Transform`, which calls on a pre-trained SentencePiece model.

# Testing

Integration tests were updated to accept input strings, which are then passed to the transform method. The resulting token ids are fed into the model, and the output tensors are checked against results from the reference HuggingFace models.

# Follow-Up

Currently, the decoder outputs hidden states of the final layer. Once the generator is added, we can test the transform's `decode` method.


parmeet (approved):
LGTM!

PR #1857, opened 2022-07-26
[cherry-pick] Fixing build when CUDA enabled torch is installed (#1814)

atalman:
* Fixing build when cuda torch is installed

* Passing in TORCH_INSTALL_PREFIX from extension.py


Nayef211 (approved):
LGTM

PR #1835, opened 2022-07-13
Set MACOSX_DEPLOYMENT_TARGET=10.9 for binary job

mthrok:
Recent CircleCI migration https://github.com/pytorch/text/pull/1818
silently bumped the minimum supported macOS version to 11.

PyTorch still supports 10.9 and the ecosystem still uses 10.9.
Issue: https://github.com/pytorch/text/issues/1834

This commit sets MACOSX_DEPLOYMENT_TARGET=10.9, so that binary
distribution are compatible on macOS=10.9.

Issue #1834, opened 2022-07-13
Add Nightly releases for macos < 11 with x86_64


ejguan:
## 🚀 Feature

After 2022/07/07, the nightly releases for TorchText on macos are updated to 11_0 with `amd64`. However, there are still use cases with `x86_64`. Apparently, PyTorch still provides releases for both platforms.

**Motivation**

This causes CI failure from TorchData side as GHA is still using 10.9 version of mac machines to execute tests. When the workflow tries to install torchtext, it will fall back to the nightly releases from 2022/07/07 with the corresponding PyTorch nightly of 2022/07/07 . As we landed a private BC-breaking change from PyTorch Core, the CI becomes red with the domain tests.

Edit: ~And, for my local environment, we still have a bunch of people using mac without M-series chips. Without such support, we have to compile from source of TorchText~

**Alternatives**

Compiled from source

mthrok:
Maybe it is related to https://github.com/pytorch/text/pull/1818
The migration to Xcode 12.5 might change the default CPU type. 

However, the latest nightly build like [this one](https://app.circleci.com/pipelines/github/pytorch/text/6077/workflows/b642c26d-5c7d-40ad-802d-1af32ef33794/jobs/207400) suggests it is built for `x86_64`.

What Python version on which platform are you seeing the issue?

mthrok:
That makes sense. I think this issue from audio has the same cause.

https://github.com/pytorch/audio/issues/2536

mthrok:
Made a PR #1835 to fix it.

While looking at the job log from the recent nightly, it seems that the compilation is done for 10.9

```
copying build/lib.macosx-10.9-x86_64-3.8/torchtext/vocab/vectors.py -> build/bdist.macosx-10.9-x86_64/wheel/torchtext/vocab
copying build/lib.macosx-10.9-x86_64-3.8/torchtext/vocab/vocab.py -> build/bdist.macosx-10.9-x86_64/wheel/torchtext/vocab
copying build/lib.macosx-10.9-x86_64-3.8/torchtext/vocab/__init__.py -> build/bdist.macosx-10.9-x86_64/wheel/torchtext/vocab
copying build/lib.macosx-10.9-x86_64-3.8/torchtext/vocab/vocab_factory.py -> build/bdist.macosx-10.9-x86_64/wheel/torchtext/vocab
[...]

PR #1818, opened 2022-07-07
Update CircleCI Xcode image

mthrok:
CircleCI is removing Xcode 12.4.0 image on August, and there was a planned
burnout on July 6th. [[detail](https://discuss.circleci.com/t/xcode-image-deprecation/44294?mkt_tok=NDg1LVpNSC02MjYAAAGFbbxbX7nSPCzN0MCKN078pw0VLJ-TMdICr8_gouRNYBM8C55RL8NDKLXA_9CQGPqnhJE5lsSFdetLRF-nH7iBLzoPGBfYpf2vuJ-XkW_C4__4)]

https://app.circleci.com/pipelines/github/pytorch/text/5996/workflows/0bfa021d-b532-4385-b406-c56373c8e9ce/jobs/204435

This commit updates Xcode image to 12.5

PR #1858, opened 2022-07-26
Set MACOSX_DEPLOYMENT_TARGET=10.9 for binary job (#1835)

atalman:
Recent CircleCI migration https://github.com/pytorch/text/pull/1818
silently bumped the minimum supported macOS version to 11.

PyTorch still supports 10.9 and the ecosystem still uses 10.9.
Issue: https://github.com/pytorch/text/issues/1834

This commit sets MACOSX_DEPLOYMENT_TARGET=10.9, so that binary
distribution are compatible with macOS=10.9.


Nayef211 (approved):

parmeet (approved):
LGTM!

PR #1862, opened 2022-07-26
Prepare T5 Model for Language Generation

pmabbo13:
# Description

Add a T5 model that can be used for language generation tasks.

# Process

For language generation, the outputs of the decoder pass through a linear layer which projects the hidden states to the vocab. The boolean parameter `linear_head` was added to `T5Conf`, which allows the user to specify if they want the model to include this final linear layer and return its output. A bundler object, `T5_BASE_GENERATION` was also created to instantiate such a model using pre-trained weights under the base configuration. These weights have been uploaded to the AWS and Manifold buckets as `t5.base.generation.pt`.

# Testing
Tests that were used to check the outputs of the `T5_BASE` and `T5_BASE_ENCODER` models are also used to test the outputs of `T5_BASE_GENERATION`. Additionally, the linear head in the generation models allows us to compute the cross entropy loss. Therefore, we can now also test the train functionality.

`pytest test/prototype/models/test_models.py`

# Follow-Up

- [ ] `test/prototype/models/test_models.py::TestModels::test_t5_bundler_train` appears to be failing in select environments (ex. for the most recent commit it failed in `windows_py3.8` and `macos_py3.10`, though the exact environment(s) it fails in varies with each new, yet unrelated, commit. It's unclear what the source of this failure is, and why it doesn't consistantly fail in the same environments.


parmeet (approved):
Thanks for adding Language generation capability to T5 model, overall LGTM!

torchtext/prototype/models/t5/model.py
```diff
@@ -191,6 +202,12 @@ def forward(
             decoder_output = self.dropout4(decoder_output)
             decoder_hidden_states = decoder_hidden_states + (decoder_output,)
 
+            if self.linear_head:
+                if self.tie_word_embeddings:
+                    decoder_output = decoder_output * (self.config.embedding_dim ** -0.5)
```
Nayef211, line 207: Can we mention what the purpose of `tie_word_embeddings` is and leave a link to Google's/HF's implementation? At a cursory glance, it's not obvious why we multiply the decoder output with `(self.config.embedding_dim ** -0.5)`

Nayef211 (approved):
LGTM

torchtext/prototype/models/t5/model.py
pmabbo13: So it looks like `tie_word_embeddings` is just a flag to indicate whether or not the encoder and decoder use the same embedding weights. It looks like this configuration exists in HF across a wider set of models, but for T5 it is defaulted to True. Likewise, our implementation has it so that the encoder and decoder always share the same embeddings. I'm guessing that the reason why the rescaling is dependent on this flag being True is because it rescales the decoder output by the embedding dimension, which I guess only works if the encoder and decoder share the same embedding size?

Given that our T5 model will always share these embeddings, I'm wondering if it makes sense to keep the rescaling, but get rid of the flag, and just include a code pointer to where HF is doing a similar rescaling.

pmabbo13: As per our live conversation, I will remove `tie_word_embeddings` but keep the re-scaling.

PR #1864, opened 2022-07-28
Tutorial on using T5 model for text summarization

pmabbo13:


parmeet (approved):
Overall, LGTM! Thank you @pmabbo13 for adding this end-2-end tutorial, this would be quite helpful for users to on-board T5 model. Let's make sure the tutorial can be build locally before merge :).

examples/tutorials/cnndm_summarization.py
```diff
@@ -0,0 +1,309 @@
...
+# batched and non-batched text input (i.e. one can either pass a single sentence or a list of sentences), however
+# the T5 model expects the input to be batched.
+#
+
+from torchtext.prototype.models import T5Transform
+
+padding_idx = 0
+eos_idx = 1
+max_seq_len = 512
+t5_sp_model_path = r"https://download.pytorch.org/models/text/t5_tokenizer_base.model"
```
Nayef211, line 54: nit: I don't think we need to treat this URL as a raw string since there's no backslashes present

examples/tutorials/cnndm_summarization.py
```diff
@@ -0,0 +1,309 @@
...
+#
+#
+from torchtext.prototype.models import T5_BASE_GENERATION
+
+
+t5_base = T5_BASE_GENERATION
+transform = t5_base.transform()
+model = t5_base.get_model()
+model.to(DEVICE)
+
```
Nayef211, line 143: Is it worthwhile explicitly calling `model.eval()` since we're using the model for inference. I know you mentioned that unless the `training` parameter is set to true, dropout is set to 0, but this may not be obvious to the user in this tutorial.

Nayef211 (approved):
LGTM! Great work on adding a descriptive tutorial that is very easy to follow!

PR #1866, opened 2022-07-29
Changing version - required for m1 builds

atalman:
Changing version - required for m1 builds
Ref: https://github.com/pytorch/vision/pull/6334


mthrok:
Since this is release branch, isn't it more appropriate to remove `a0`?

PR #1869, opened 2022-08-02
Updating T5 demo to use beam search for generator

pmabbo13:
# Description

Update T5 tutorial to use a beam search for decoding, as opposed to a greedy search.

# Process

A beam search was implemented, which keeps track of the log probability of multiple sequences generated for a single input sequence, and prunes them at each iteration to only keep the top `k` most likely. This allows the generator to create sequences that are more probable than those generated by a greedy search. 

We find that the sequences generated via beam search tend to be longer, but still mostly capture the main points expressed in the target summaries. We expect an improvement in the summaries with the addition of constraints such as a length penalty, ngram limit, min length, max length, etc. 

# Testing

We tested the logic be ascertaining that the sequences generated when `beam_size=1` were the same as those generated under a greedy decoder --> see Generate Summaries section of [this notebook](https://gist.github.com/pmabbo13/2698f755448ef5de936a7c5c28dc8f1f)

Run `BUILD_GALLERY=1 make 'SPHINXOPTS=-W' html` in `docs` and review rendered document in `docs/build/html/tutorials/cnndm_summarization.html`

### DIFF
--- a/examples/tutorials/cnndm_summarization.py
+++ b/examples/tutorials/cnndm_summarization.py
@@ -149,21 +149,76 @@ model.to(DEVICE)
 #
 # We can define a sequence generator to produce an output sequence based on the input sequence provided. This calls on the
 # model's encoder and decoder, and iteratively expands the decoded sequences until the end-of-sequence token is generated
-# for all sequences in the batch. The `greedy_generator` method shown below uses a greedy search (i.e. expands the sequence
-# based on the most probable next word).
+# for all sequences in the batch. The `generate` method shown below uses a beam search to generate the sequences.
 #
 
 from torch import Tensor
 from torchtext.prototype.models import T5Model
 
 
-def greedy_generator(
-    encoder_tokens: Tensor,
-    eos_idx: int,
-    model: T5Model,
-) -> Tensor:
+def beam_search(
+    beam_size: int,
+    step: int,
+    bsz: int,
+    decoder_output: Tensor,
+    decoder_tokens: Tensor,
+    scores: Tensor,
+    incomplete_sentences: Tensor,
+):
+    probs = F.log_softmax(decoder_output[:, -1], dim=-1)
+    top = torch.topk(probs, beam_size)
+
+    # N is number of sequences in decoder_tokens, L is length of sequences, B is beam_size
+    # decoder_tokens has shape (N,L) -> (N,B,L)
+    # top.indices has shape (N,B) - > (N,B,1)
+    # x has shape (N,B,L+1)
+    # note that when step == 1, N = batch_size, and when step > 1, N = batch_size * beam_size
+    x = torch.cat([decoder_tokens.unsqueeze(1).repeat(1, beam_size, 1), top.indices.unsqueeze(-1)], dim=-1)
+
+    # beams are first created for a given sequence
+    if step == 1:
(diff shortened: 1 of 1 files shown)


examples/tutorials/cnndm_summarization.py
```diff
@@ -149,21 +149,76 @@ def apply_prefix(task, x):
 #
 # We can define a sequence generator to produce an output sequence based on the input sequence provided. This calls on the
 # model's encoder and decoder, and iteratively expands the decoded sequences until the end-of-sequence token is generated
-# for all sequences in the batch. The `greedy_generator` method shown below uses a greedy search (i.e. expands the sequence
-# based on the most probable next word).
+# for all sequences in the batch. The `generate` method shown below uses a beam search to generate the sequences.
 #
 
 from torch import Tensor
 from torchtext.prototype.models import T5Model
```
pmabbo13, line 158: One thing to note is that when generating the first tokens of the sequences, `decoder_tokens` has shape `(batch_size, 1)`. Since we are using a beam search, at that first iteration each sequence has `k` many tokens with which to start the sequence. Since the decoder expects `decoder_tokens` to be 2D, with one sequence per row, we treat each beam as its own sequence such that `decoder_tokens` now has shape `(batch_size * beam_size, 2)`, where the first `k` rows are the beams belonging to the original sequence 1, the next `k` rows are the beams belonging to the original sequence 2, etc. Since the decoder requires the encoder outputs, we must also reshape modify the encoder output since it has the output for a `batch_size` number of sequence. Lines `239-244` define a new order where we repeat each encoder output related to each given original sequence `k` times. This is so that that as we pass in each beam as its own sequence, the decoder still has the corresponding encoder output for the sequence.

examples/tutorials/cnndm_summarization.py
```diff
@@ -149,21 +149,76 @@ def apply_prefix(task, x):
...
 
 
-def greedy_generator(
-    encoder_tokens: Tensor,
-    eos_idx: int,
-    model: T5Model,
-) -> Tensor:
+def beam_search(
+    beam_size: int,
+    step: int,
```
pmabbo13, line 161: `step` here is equivalent to the current length of sequences in `decoder_tokens`. The first time `beam_search` is called, `step=1` because `decoder_tokens` is initialized to have the padding token as the starter token to each sequence.

examples/tutorials/cnndm_summarization.py
```diff
@@ -149,21 +149,76 @@ def apply_prefix(task, x):
 #
 # We can define a sequence generator to produce an output sequence based on the input sequence provided. This calls on the
 # model's encoder and decoder, and iteratively expands the decoded sequences until the end-of-sequence token is generated
-# for all sequences in the batch. The `greedy_generator` method shown below uses a greedy search (i.e. expands the sequence
-# based on the most probable next word).
+# for all sequences in the batch. The `generate` method shown below uses a beam search to generate the sequences.
```
Nayef211, line 152: Should we mention somewhere that we use a beam size of 3 to generate summaries in this tutorial. Also might be worthwhile mentioning that larger beam sizes can result in better generation at the cost of computational complexity and that `beam_size == 1` is equivalent to a greedy decoder.

Nayef211 (approved):

parmeet (approved):
LGTM!

PR #1870, opened 2022-08-03
Update CircleCI Xcode image (#1818)

atalman:
CircleCI is removing Xcode 12.4.0 image on August, and there was a planned
burnout on July 6th.

https://app.circleci.com/pipelines/github/pytorch/text/5996/workflows/0bfa021d-b532-4385-b406-c56373c8e9ce/jobs/204435

This commit updates Xcode image to 12.5

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -203,7 +203,7 @@ jobs:
   binary_macos_wheel:
     <<: *binary_common
     macos:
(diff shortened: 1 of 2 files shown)


datumbox (commented):
Looked good to me! I assume just to be safe, we keep the version on the same major as previously instead of moving to the latest one.

PR #1871, opened 2022-08-03
Make BERT benchmark code more robust

parmeet:
Updated benchmark code to run on pre-defined number of samples and batch size. By running on higher number of samples gives more robust statistics because 1) we show more variable length samples to tokenizer 2) we are running for larger number of batches instead of just 1 as currently the case.

## Benchmark results
**Number or samples:** 100000

### non-batched input

TorchText BERT Tokenizer: 1.7653241670000002
HF BERT Tokenizer (slow): 27.455106365
HF BERT Tokenizer (fast): 5.351107693000003

### Batched input

**Batch-size: 50**
TorchText BERT Tokenizer: 1.376252063
HF BERT Tokenizer (fast): 1.5889374279999995

**Batch-size: 100**
TorchText BERT Tokenizer: 1.3049638119999996
HF BERT Tokenizer (fast): 1.4069846630000002

**Batch-size: 200**
TorchText BERT Tokenizer: 1.275028583
HF BERT Tokenizer (fast): 1.2769447180000002

**Batch-size: 400**
TorchText BERT Tokenizer: 1.3523340929999996
HF BERT Tokenizer (fast): 1.2558808729999997

Apparently, for HF tokenizer the operator call is slower compared to torchtext but the backend implementation is faster. This is why after increase batch size to certain point, HF tokenizer (fast) is more performant compared to torchtext's implementation.

### DIFF
--- a/benchmark/benchmark_bert_tokenizer.py
+++ b/benchmark/benchmark_bert_tokenizer.py
@@ -14,29 +14,36 @@ def benchmark_bert_tokenizer(args):
     tt_tokenizer = tt_bert_tokenizer(VOCAB_FILE, return_tokens=True)
     hf_tokenizer_slow = hf_bert_tokenizer_slow.from_pretrained("bert-base-uncased")
     hf_tokenizer_fast = hf_tokenizer_lib.from_pretrained("bert-base-uncased")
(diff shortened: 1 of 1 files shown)


Nayef211 (approved):
LGTM

PR #1874, opened 2022-08-05
Add torchdata to testing requirements in requirements.txt

ronakmal:
Torchdata is a requirement to run the test suite, but it doesn't look like this is documented anywhere. This change proposes adding it to the testing section of the requirements.txt.


parmeet (approved):
LGTM!

PR #1880, opened 2022-08-12
[WIP] wrapper class for end-to-end t5 model

pmabbo13:
# Description
Add a wrapper class for the T5 model that takes text as input and outputs the generated text from the decoder

# Process
The wrapper class performs the following end-to-end process:
1. transforms input text using `T5Transform` to get truncated and padded token IDs
2. passes resulting model input through t5 model, using a simple beam search generator to produce the output sequence
3. uses transform pipeline to translate the generator output (sequence of token IDs) to text.

Wrapper class accepts either: `configuration` parameter which is a string indicating which bundler object to use (currently only supports `'base'`, but will update once other configurations are added), or parameters `checkpoint`, `t5_conf`, and `transform` if custom model architecture and weights are to be used.

# Testing
Feed wrapper class input text and ensure that it returns the expected output. Tests are performed for wrapper instantiation using both `configuration` and custom settings. Scripted and non-scripted versions are also tested.

### DIFF
--- a/test/prototype/integration_tests/test_models.py
+++ b/test/prototype/integration_tests/test_models.py
@@ -53,3 +53,49 @@ class TestT5(TorchtextTestCase):
         self._t5_model(
             is_jit=is_jit, t5_model=T5_BASE_GENERATION, expected_asset_name=expected_asset_name, test_text=test_text
         )
+
+    @parameterized.expand([("jit", True), ("not_jit", False)])
+    def test_t5_wrapper(self, name, is_jit) -> None:
+
+        from torchtext.prototype.models.t5.wrapper import T5Wrapper
+
+        test_text = ["translate English to French: I want to eat pizza for dinner."]
+        expected_text = ["Je veux manger de la pizza pour le dîner."]
+        beam_size = 3
+        max_seq_len = 512
+        model = T5Wrapper(configuration="base")
+        if is_jit:
+            model = torch.jit.script(model)
+
+        output_text = model(test_text, beam_size, max_seq_len)
+        self.assertEqual(output_text, expected_text)
+
+    @parameterized.expand([("jit", True), ("not_jit", False)])
+    def test_t5_wrapper_checkpoing(self, name, is_jit) -> None:
+
+        from torchtext.prototype.models import T5Conf, T5Transform
+        from torchtext.prototype.models.t5.wrapper import T5Wrapper
+
+        test_text = ["translate English to French: I want to eat pizza for dinner."]
+        expected_text = ["Je veux manger de la pizza pour le dîner."]
+        beam_size = 3
+        max_seq_len = 512
+        config = T5Conf(encoder_only=False, linear_head=True)
+        transform = T5Transform(
+            "https://download.pytorch.org/models/text/t5_tokenizer_base.model",
+            max_seq_len=512,
+            eos_idx=1,
+            padding_idx=0,
+        )
+        model = T5Wrapper(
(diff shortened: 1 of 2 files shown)


test/prototype/integration_tests/test_models.py
```diff
@@ -53,3 +53,49 @@ def test_t5_base_generation_model(self, name, is_jit) -> None:
         self._t5_model(
             is_jit=is_jit, t5_model=T5_BASE_GENERATION, expected_asset_name=expected_asset_name, test_text=test_text
         )
+
+    @parameterized.expand([("jit", True), ("not_jit", False)])
+    def test_t5_wrapper(self, name, is_jit) -> None:
+
+        from torchtext.prototype.models.t5.wrapper import T5Wrapper
```
Nayef211, line 60: Why do we do this import here instead of at the very top of the file?

test/prototype/integration_tests/test_models.py
```diff
@@ -53,3 +53,49 @@ def test_t5_base_generation_model(self, name, is_jit) -> None:
...
+            model = torch.jit.script(model)
+
+        output_text = model(test_text, beam_size, max_seq_len)
+        self.assertEqual(output_text, expected_text)
+
+    @parameterized.expand([("jit", True), ("not_jit", False)])
+    def test_t5_wrapper_checkpoing(self, name, is_jit) -> None:
+
+        from torchtext.prototype.models import T5Conf, T5Transform
+        from torchtext.prototype.models.t5.wrapper import T5Wrapper
```
Nayef211, line 77: Same question as above

torchtext/prototype/models/t5/wrapper.py
```diff
@@ -0,0 +1,191 @@
...
+            self.bundler = T5Bundle(_path=checkpoint, _config=t5_config, transform=lambda: transform)
+            self.model = self.bundler.build_model(
+                config=t5_config, freeze_model=freeze_model, checkpoint=checkpoint, strict=strict, dl_kwargs=dl_kwargs
+            )
+        else:
+            self.bundler = T5_BASE_GENERATION
+            self.model = self.bundler.get_model()
+
+        self.transform = self.bundler.transform()
+        self.config = self.bundler._config
```
Nayef211, line 55: Why do we store `self.config`? Is this used anywhere?

torchtext/prototype/models/t5/wrapper.py
```diff
@@ -0,0 +1,191 @@
...
+        if step == 1:
+            # x has shape (batch_size, B, L+1) -> (batch_size * B,L+1)
+            # new_scores has shape (batch_size,B)
+            # incomplete_sentences has shape (batch_size * B) = (N)
+            new_decoder_tokens = x.view(-1, step + 1)
+            new_scores = top.values
+            new_incomplete_sentences = incomplete_sentences
+
+        # beams already exist, want to expand each beam into possible new tokens to add
+        # and for all expanded beams beloning to the same sequences, choose the top k
```
Nayef211, line 87: nit: belonging

Nayef211 (approved):
Overall LGTM. How difficult would it be to enable the wrapper class to work with other model configurations?

pmabbo13:
> Overall LGTM. How difficult would it be to enable the wrapper class to work with other model configurations?

It would be straight forward. We could update `line 43` in `wrapper.py` to assert that the value passed in for `configuration` is in `('base', 'small', 'large', '3b', '11b')`, and then use if statements to load the bundler object that corresponds to the input configuration.

PR #1881, opened 2022-08-15
Update xcode version to 14.0 in CI

Nayef211:
## Description
- CI jobs on MacOS have been failing with the following [error message](https://app.circleci.com/pipelines/github/pytorch/text/6558/workflows/f3d78aac-bd05-4c72-aad9-d469a869b4fb/jobs/225889)
```
failed to create host: Image xcode:12.5 is not supported
```
- According to the [CircleCI website](https://circleci.com/docs/xcode-policy?utm_source=google&utm_medium=sem&utm_campaign=sem-google-dg--uscan-en-dsa-maxConv-auth-brand&utm_term=g_-_c__dsa_&utm_content=&gclid=Cj0KCQjw3eeXBhD7ARIsAHjssr9nFtoGHLNQhM3Ef0HE9-XQbAmvuCFfEuboIbNneZJl_Syu3IqUyOAaAjTSEALw_wcB), Xcode 12 will be deprecated once Xcode 15 reaches RC which is likely the cause for this failure
- In this PR we bump the Xcode version to be `14.0` in line with [torchvision](https://github.com/pytorch/vision/blob/main/.circleci/config.yml#L905)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -207,7 +207,7 @@ jobs:
   binary_macos_wheel:
     <<: *binary_common
     macos:
-      xcode: "12.5"
+      xcode: "14.0"
     steps:
       - checkout
       - designate_upload_channel
@@ -233,7 +233,7 @@ jobs:
   binary_macos_conda:
     <<: *binary_common
     macos:
-      xcode: "12.5"
+      xcode: "14.0"
     steps:
       - checkout
       - designate_upload_channel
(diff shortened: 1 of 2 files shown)


datumbox (approved):
LGTM, thanks!

BTW TorchVision has switched directly to v14 back when it was beta and we didn't face any issues. I believe Andrey switched text to 12.5 just to be safe, but I think the latest shouldn't cause you any problems. Worth testing.

PR #1882, opened 2022-08-16
Graduate `MaskTransform` from prototype

joecummings:
**What**? Graduating the `MaskTransform` class created by @pmabbo13 from the `prototype` dir to the main dir. This involves the following changes:

- Move `MaskTransform` class
- Move `MaskTransformTest` class
- Clean up unnecessary imports

**Why?** Part of [T113713985](https://www.internalfb.com/intern/tasks/?t=113713985)

**Test Plan**?
- `pre-commit run`
- `pytest test/test_transforms.py`
- `pytest test/prototype/test_transforms.py`


Nayef211:
> Only failing the Windows 10, Python 3.10 unit tests, but this doesn't seem to be a valid error.

Yup you're right, this is a long standing error with dependency resolution that we need to spend some more time to look into

Nayef211 (approved):
LGTM thanks for working on this! 

Just as an FYI for next time, we should probably add a line describing what [T113713985](https://www.internalfb.com/intern/tasks/?t=113713985) is about. Tasks are internal and cannot be viewed by anyone from the OSS community that is not a Meta employee.

PR #1884, opened 2022-08-18
Allow CNNDM to be imported from torchtext.datasets

pmabbo13:
# Description

Currently can only import CNNDM from `torchtext.datasets.cnndm`. Update so that can import from `torchtext.datasetes`

# Process

Add CNNDM to `torchtext.datasets.__init__.py` and rename certain class objects to align with expected names in `torchtext.datasets.__init__.py`. Update CNNDM import statements.

# Test

`pytest test/datasets/test_cnndm.py` (this imports CNNDM, so import statement is tested)


Nayef211 (approved):
LGTM thanks for fixing this

PR #1885, opened 2022-08-26
Remove dependency on the torch::jit::script::Module for mobile builds

joecummings:
Summary: In order to resolve linkage errors. Specifically when vocab getting build for "mobile" version it can't resolve symbols for torch::jit::script::Module

Reviewed By: Nayef211

Differential Revision: D38771271

fbshipit-source-id: 693b656f2a17af9fa5a7a1904742557f902edb55


mthrok:
Needs to update the corresponding file lists in https://github.com/pytorch/text/blob/main/torchtext/csrc/CMakeLists.txt.

torchtext/csrc/vocab_factory.h
```diff
@@ -1,56 +1,24 @@
 #include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
```
Nayef211, line 2: @joecummings I would actually move this header to the corresponding `.cpp` file since we don't do any type casting to STL containers in the header. Sorry for the extra work here.

Nayef211 (approved):
LGTM! Rest of the test failures related to running out of disk storage should be fixed in follow up PRs (i.e. https://github.com/pytorch/text/pull/1886 and https://github.com/pytorch/text/pull/1889)

joecummings:
`unittest_windows_py3.10` failure noted and being investigated in #1890

PR #1886, opened 2022-08-29
Delete model checkpoints within integration test

Nayef211:
## Description
- Currently tests in CI are failing due to running out of disk space (i.e. see [example failure](https://app.circleci.com/pipelines/github/pytorch/text/6623/workflows/25d5e336-a916-4855-a9ad-b99b684b979a/jobs/228196))
- This happens because some of the checkpoints downloaded in the integration test are quite large (around 3GB) and several checkpoints are downloaded for the TF and the RoBERTa models
- We mitigate this by removing the checkpoints at the end of every test. The tradeoff is that this will cause more network usage since a single checkpoint may need to be downloaded several times (as several tests may rely on the same checkpoint)
- We also convert `load_state_dict_from_url` within `_download_hooks.py` into a partial function with the default value for `model_dir` set to `_CACHE_DIR`. This ensures that all text assets are downloaded to `$TORCH_HOME/text` folder which is in line with where we download all of our datasets (see the following [internal post](https://fb.workplace.com/groups/473278361165956/permalink/517067430120382/) for relevant discussion) .

## Testing
- Ensure all CI tests are passing
- `pytest test/prototype/integration_tests/test_models.py`
- `pytest test/integration_tests/test_models.py`

### DIFF
--- a/test/integration_tests/test_models.py
+++ b/test/integration_tests/test_models.py
@@ -6,8 +6,9 @@ from torchtext.models import (
     XLMR_BASE_ENCODER,
     XLMR_LARGE_ENCODER,
 )
+from torchtext.utils import get_asset_local_path
 
-from ..common.assets import get_asset_path
+from ..common.assets import conditional_remove, get_asset_path
 from ..common.torchtext_test_case import TorchtextTestCase
 
 
@@ -30,6 +31,10 @@ class TestRobertaEncoders(TorchtextTestCase):
         actual = model(model_input)
         expected = torch.load(expected_asset_path)
         torch.testing.assert_close(actual, expected)
+
+        # delete checkpoint from cache
+        model_checkpoint_path = get_asset_local_path(encoder._path)
+        conditional_remove(model_checkpoint_path)
 
     @parameterized.expand([("jit", True), ("not_jit", False)])
     def test_xlmr_base_model(self, name, is_jit):
--- a/torchtext/_download_hooks.py
+++ b/torchtext/_download_hooks.py
@@ -1,14 +1,19 @@
 import re
+from functools import partial
 
 import requests
 
 # This is to allow monkey-patching in fbcode
-from torch.hub import load_state_dict_from_url  # noqa
+from torch.hub import load_state_dict_from_url as torchhub_load_state_dict_from_url  # noqa
+from torchtext import _CACHE_DIR
 from torchtext._internal.module_utils import is_module_available
 from tqdm import tqdm
 
 if is_module_available("torchdata"):
     from torchdata.datapipes.iter import HttpReader, GDriveReader  # noqa F401
+
+# we set the model_dir to be equal to _CACHE_DIR so that all checkpoint downloads go into the text specific cache
+load_state_dict_from_url = partial(torchhub_load_state_dict_from_url, model_dir=_CACHE_DIR)
 
 
 def _stream_response(r, chunk_size=16 * 1024):
--- a/test/prototype/integration_tests/test_models.py
+++ b/test/prototype/integration_tests/test_models.py
@@ -1,6 +1,6 @@
 import torch
 from parameterized import parameterized
-from test.common.assets import get_asset_path
+from test.common.assets import conditional_remove, get_asset_path
 from test.common.parameterized_utils import nested_params
 from test.common.torchtext_test_case import TorchtextTestCase
 from torchtext.prototype.models import (
@@ -17,6 +17,7 @@ from torchtext.prototype.models import (
     T5Transform,
 )
 from torchtext.prototype.models.t5.wrapper import T5Wrapper
+from torchtext.utils import get_asset_local_path
 
 
 BUNDLERS = {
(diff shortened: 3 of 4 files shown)


torchtext/_download_hooks.py
```diff
@@ -1,15 +1,20 @@
...
+from torch.hub import load_state_dict_from_url as torchhub_load_state_dict_from_url  # noqa
+from torchtext import _CACHE_DIR
 from torchtext._internal.module_utils import is_module_available
 from tqdm import tqdm
 
 if is_module_available("torchdata"):
     from torchdata.datapipes.iter import HttpReader, GDriveReader  # noqa F401
 
+# we set the model_dir to be equal to _CACHE_DIR so that all checkpoint downloads go into the text specific cache
+load_state_dict_from_url = partial(torchhub_load_state_dict_from_url, model_dir=_CACHE_DIR)
```
mthrok, line 16: The use of partial makes it looks like a hack to me. 
But IIUC, this approach is inlined with decisions on other so it seems better to be a regular function.

mthrok: On a separate note, `torchtext._CACHE_DIR` is a constant string, so this won't allow client code to change the cache directory at runtime. That seems to be inflexible to me. Perhaps get rid of `torchtext._CACHE_DIR` and change it to a function.

test/integration_tests/test_models.py
```diff
@@ -6,8 +6,9 @@
     XLMR_BASE_ENCODER,
     XLMR_LARGE_ENCODER,
 )
+from torchtext.utils import get_asset_local_path
 
-from ..common.assets import get_asset_path
+from ..common.assets import conditional_remove, get_asset_path
```
mthrok, line 11: Mixing fixtures/assets from unit tests with integration tests is a code smell for me.

torchtext/_download_hooks.py
mthrok: Consolidating directories for caches is good, however I'd keep pre-trained models and dataset caches in separate sub directory.

mthrok (commented):
I think, instead of manually calling `conditional_remove`, the deletion logic should be automated and the pre-condition should be configured in a way it's independent from rest of the test/library logic. (i.e. it should be turned on/off independent of what ever library/test logics are)

That will avoid the case where future contributors forget to add `conditional_remove`.

torchtext/_download_hooks.py
Nayef211: I ended up deciding to scrap this change alltogether and just follow the approach torchaudio takes. This means that all model checkpoints will still be downloaded to the `torch.hub.home` directory.

I think it should be fine for `torchtext._CACHE_DIR` to be a constant string as it's only used to download dataset and other asset files from a given URL (not model checkpoints). If clients wanted to modify the root part of `torchtext._CACHE_DIR`, they can just update the `$TORCH_HOME` env variable as that is respected by the `_get_torch_home()` method.

test/integration_tests/test_models.py
Nayef211: Removed and replaced with pytest fixtures as per your recommendation

mthrok (approved):
stamp

test/integration_tests/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+        help=(
+            "When provided, tests will use temporary directory as Torch Hub directory. "
+            "Downloaded models will be deleted after each test."
+        ),
+    )
+
+
+@pytest.fixture(scope="class")
+def temp_hub_dir(tmp_path_factory, pytestconfig):
+    if not pytestconfig.getoption("--use-tmp-hub-dir"):
```
h-vetinari, line 20: Follow-up in #1889 aside, I don't understand how this can work at all.

Both `tmp_path_factory` and `pytestconfig` appear _nowhere else_ in the code base (as far as GH search and `git grep` go, at least), and based on the pytest [docs](https://docs.pytest.org/en/7.2.x/example/simple.html#pass-different-values-to-a-test-function-depending-on-command-line-options) this should probably be
```suggestion
@pytest.fixture(scope="class")
def temp_hub_dir(tmp_path_factory, request):
    if not request.config.getoption("--use-tmp-hub-dir"):
```
but that still leaves me mystified as to where `tmp_path_factory` is coming from.

h-vetinari (commented):
Trying to get torchtext [into](https://github.com/conda-forge/staged-recipes/pull/17129) conda-forge (xref #2023), and currently stumbling over this fixture when trying to run the tests.

test/integration_tests/conftest.py
h-vetinari: > but that still leaves me mystified as to where `tmp_path_factory` is coming from.

Ah, just found that it's a fixture that comes with pytest by [default](https://docs.pytest.org/en/7.2.x/how-to/tmp_path.html#tmp-path-factory-example).

h-vetinari: OK, and so's `pytestconfig`. In any case, my suspicion now is that this only works when called from _within_ the `integration_tests` folder, because (I guess) recursive `conftest.py`s are not evaluated when the CLI is called from a folder that's further out.

Nayef211: @h-vetinari yup I think you're understanding is correct. Does only being able to run this test from within the `integration_tests` folder pose an issue for adding torchtext to `conda-forge`?

h-vetinari: > Does only being able to run this test from within the `integration_tests` folder pose an issue for adding torchtext to `conda-forge`?

No, I just had to realise the reason, after that it's simple. ;)

PR #1888, opened 2022-08-30
Export symbols in `common.h` file

Nayef211:
## Description
- https://github.com/pytorch/text/pull/1806 added a `TORCHTEXT_API` to export relevant symbols on Windows platform
- We add the macro to the methods in `common.h` as these are used by functions in `vocab_factory.h`

### DIFF
--- a/torchtext/csrc/common.cpp
+++ b/torchtext/csrc/common.cpp
@@ -2,10 +2,7 @@
 
 #include <fstream>
 #include <ios>
-#include <iostream>
 #include <limits>
-#include <string>
-#include <vector>
 
 namespace torchtext {
 namespace impl {
--- a/torchtext/csrc/common.cpp
+++ b/torchtext/csrc/common.cpp
@@ -1,9 +1,12 @@
+#include <torchtext/csrc/common.h>
+
 #include <fstream>
 #include <ios>
 #include <iostream>
 #include <limits>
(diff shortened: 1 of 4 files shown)


Nayef211:
All of the MacOS failures should be fixed by https://github.com/pytorch/text/pull/1886. The Windows unittest failure looks new and will require further investigation but is not caused by this PR.

joecummings (approved):
This looks good, thanks for doing this!

One possible nit: I usually squash (fold in the mercurial world) unimportant commits like "Fixing imports" so there's less clutter in the commit history. If this isn't common practice w/ torchtext, no worries, but thought I'd mention it.

Nayef211:
> This looks good, thanks for doing this!
> 
> One possible nit: I usually squash (fold in the mercurial world) unimportant commits like "Fixing imports" so there's less clutter in the commit history. If this isn't common practice w/ torchtext, no worries, but thought I'd mention it.

That's a good point. In torchtext, we generally keep all commits (even unimportant ones related to linting), but when we merge we create a PR against the `main` branch, we always use the "Squash and Merge" option. This way all commits are preserved in the PR/feature branches but commits in the `main` branch always map 1 to 1 with the corresponding PR. Lmk if this practice makes sense to you, if not we can discuss what the best practice should be moving forward.

PR #1894, opened 2022-09-01
Make comment paths dynamic

joecummings:
### Description

Change hardcoded module path in comments to be dynamic.


Nayef211 (approved):
LGTM, thanks for the fix.

PR #1896, opened 2022-09-12
Turn off mask checking for torchtext which is known to have a legal mask

mikekgfb:
Summary: Turn off mask checking for torchtext which is known to have a legal mask

Reviewed By: zrphercule

Differential Revision: D39445703


Nayef211:
Closing as this was autosynced to fbsync once internal diff landed

PR #1899, opened 2022-09-13
Update doc theme to the latest

Nayef211:
- Deleting outdated docs on experimental datasets
- Update `pytorch_sphinx_theme` to latest commit
- This also updates torchtext docs with new Terms / Privacy links pointing to Linux foundation

Note we still pin the theme version so that our customization does not break randomly similar to https://github.com/pytorch/audio/pull/2679


mthrok (approved):
The intended change looks good but there are some other render issues observed.

Screen Shot 2022-09-16 at 8 08 40 AM


Follow up:
- pin versions of Sphinx-related tools ([example](https://github.com/pytorch/audio/blob/main/docs/requirements.txt#L11-L14))
- [Override CSS](https://github.com/pytorch/audio/blob/c2870b6e8b157ce34b5c9e09a1a4507726c73ef6/docs/source/_static/css/custom.css#L18-L25) so that example gallery is rendered properly.

Nayef211:
> Follow up:
> 
> * pin versions of Sphinx-related tools ([example](https://github.com/pytorch/audio/blob/main/docs/requirements.txt#L11-L14))
> * [Override CSS](https://github.com/pytorch/audio/blob/c2870b6e8b157ce34b5c9e09a1a4507726c73ef6/docs/source/_static/css/custom.css#L18-L25) so that example gallery is rendered properly.

Thanks for the feedback. Let me create a followup PR to do this.

> nit: I'd have split the PR to deletion and theme update.

I think the reason I did this in one is because @svekars mentioned the following
> To update torchtext docs to pick up this change, you need to rebuild your docs. It can be a super simple PR that just changes a word or fixes a typo under the docs/source dir. 

In hindsight, maybe the update to sphinx-theme was enough to rebuild the docs.

Issue #1380, opened 2021-08-23
torchtext nightly conda package is currently uninstallable


seemethere:
## 🐛 Bug
**Describe the bug**
Conda package installs of `torchtext` are currently borked due to a dependency issue

**To Reproduce**
1. Run the following:
```
$ conda install -c "pytorch-nightly" -c "nvidia" -y python=3.9 torchtext cpuonly
```

<details>
<summary> Long and mostly useless output from conda install command </summary>

```
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: \
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
failed

UnsatisfiableError: The following specifications were found to be incompatible with each other:

Output in format: Requested package -> Available versions

Package zlib conflicts for:
conda[version='>=4.9.2'] -> python[version='>=3.6,<3.7.0a0'] -> zlib[version='>=1.2.11,<1.3.0a0']
python=3.9 -> zlib[version='>=1.2.11,<1.3.0a0']
sqlite -> zlib[version='>=1.2.11,<1.3.0a0']
pyopenssl -> python -> zlib[version='>=1.2.11,<1.3.0a0']
brotlipy -> python[version='>=3.7,<3.8.0a0'] -> zlib[version='>=1.2.11,<1.3.0a0']
zlib
idna -> python[version='>=3.5'] -> zlib[version='>=1.2.11,<1.3.0a0']
certifi -> python[version='>=3.6,<3.7.0a0'] -> zlib[version='>=1.2.11,<1.3.0a0']
chardet -> python[version='>=3.9,<3.10.0a0'] -> zlib[version='>=1.2.11,<1.3.0a0']
setuptools -> python[version='>=3.8,<3.9.0a0'] -> zlib[version='>=1.2.11,<1.3.0a0']
six -> python -> zlib[version='>=1.2.11,<1.3.0a0']
pysocks -> python[version='>=3.7,<3.8.0a0'] -> zlib[version='>=1.2.11,<1.3.0a0']
conda-package-handling -> python[version='>=3.6,<3.7.0a0'] -> zlib[version='>=1.2.11,<1.3.0a0']
[...]

PR #1905, opened 2022-09-15
[Cherry Pick] move PATH_MANAGER to OSS

Nayef211:
Summary:
## Problem:
pytext got "No module named 'pytorch'" in issue https://github.com/facebookresearch/pytext/issues/1706

It's due to `from pytorch.text.fb.utils import PATH_MANAGER` is internal only but imported in pytext. Actually, `pytorch/text/fb/utils/__init__.py` should be open sourced.

## Solution:
This diff moved it to OSS as `from torchtext.utils import PATH_MANAGER` and updated all the references

Reviewed By: Nayef211

Differential Revision: D39292896

fbshipit-source-id: c0046d62e64145b60ad9a5298b366f0f1a348369

### DIFF
--- a/.circleci/unittest/linux/scripts/environment.yml
+++ b/.circleci/unittest/linux/scripts/environment.yml
@@ -5,6 +5,7 @@ dependencies:
   - pip
   - pip:
       - dataclasses
+      - iopath
       - nltk
       - requests
       - revtok
--- a/.circleci/unittest/windows/scripts/environment.yml
+++ b/.circleci/unittest/windows/scripts/environment.yml
@@ -7,6 +7,7 @@ dependencies:
   - spacy
   - pip:
       - dataclasses
+      - iopath
       - nltk
       - requests
       - revtok
--- a/packaging/torchtext/meta.yaml
+++ b/packaging/torchtext/meta.yaml
@@ -23,6 +23,7 @@ requirements:
     - python
     - requests
     - tqdm
+    - iopath
     {{ environ.get('CONDA_PYTORCH_CONSTRAINT') }}
 
 build:
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,7 @@ tqdm
 
 # Downloading data and other files
 requests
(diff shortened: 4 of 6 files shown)


joecummings (commented):
Looks like this needs linting.

hudeven:
Looks like we need to include iopath in requirements.txt

mthrok:
Is `iopath` fully available in OSS space?

There was an attempt to do this in the past, but reverted due to https://github.com/pytorch/text/issues/1380

Nayef211:
> Is `iopath` fully available in OSS space?
> 
> There was an attempt to do this in the past, but reverted due to #1380

It seems like `iopath` is still only available from `conda-forge` and according to #1380, we don't want to have a dependency on that channel. @hudeven should we revert this diff altogether or are there any other alternatives we can pursue?

cc @seemethere

hudeven:
I see. I can revert it then. Let's ask iopath's maintainer to release it to pypi

hudeven:
@Nayef211  I noticed that iopath is available in conda. The latest version is uploaded 1 year ago and likely to be outdated though: https://anaconda.org/iopath/iopath

Nayef211:
> @Nayef211 I noticed that iopath is available in conda. The latest version is uploaded 1 year ago and likely to be outdated though: https://anaconda.org/iopath/iopath

Yup in https://github.com/pytorch/text/issues/1380#issuecomment-904121322 @seemethere mentions that we don't want a dependency on the `conda-forge` channel which is the only channel that `iopath` is available in. He might have more context as to why this is the case.

mthrok:
> we don't want a dependency on the `conda-forge` channel 

Dependency resolution is very complex and adding (requiring) `conda-forge` channel often makes it impossible to install PyTorch due to constrains posed by other libraries.

joecummings:
@Nayef211 Can this be closed?

Nayef211:
> @Nayef211 Can this be closed?

Yup the original diff was reverted

PR #1909, opened 2022-09-20
Add repository double-conversion support for loongarch

nmanhong:
Add warehouse double-conversion to support loongarch, which has been tested on loongarch and can be passed

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,6 +9,7 @@
 [submodule "third_party/double-conversion"]
 	path = third_party/double-conversion
 	url = https://github.com/google/double-conversion
+	branch = main
 	ignore = dirty
 [submodule "third_party/utf8proc"]
 	path = third_party/utf8proc
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,7 +9,7 @@
 [submodule "third_party/double-conversion"]
 	path = third_party/double-conversion
 	url = https://github.com/google/double-conversion
-	branch = main
+	branch = master
 	ignore = dirty
 [submodule "third_party/utf8proc"]
 	path = third_party/utf8proc


.gitmodules
```diff
@@ -9,6 +9,7 @@
 [submodule "third_party/double-conversion"]
 	path = third_party/double-conversion
 	url = https://github.com/google/double-conversion
+	branch = master
```
mthrok, line 12: Does this mean each `git submodule` will fetch the latest commit from `master` branch?
I think it is preferable to pin the commit to the current latest in this case.

mthrok (commented):
@Nayef211 @malfet Are you familiar with loongarch?
I guess this PR is safe as long as the basic double-conversion functionalities are tested on upstream.

.gitmodules
nmanhong: According to my understanding, it should not be pulled from the latest every time, but when adding submodules, a branch is specified.  it should be from the commit 256ac80956 of this branch at that time. Manually update this submodule. Maybe I don't know much about github.

joecummings: I think this should be `main` branch

Nayef211: @nmanhong I think if you want to update the submodule, you can go into the directory and run `git fetch` and `git merge` the upstream branch to update the local code. In this case the upstream branch would be the `master` branch. AFAIK this doesn't need to be specified in the `.gitmodules` file.

Nayef211:
> @Nayef211 @malfet Are you familiar with loongarch? I guess this PR is safe as long as the basic double-conversion functionalities are tested on upstream.

I'm not, but the changes aren't breaking any new CI tests. I'm okay with merging it unless you think we should get eyes from someone that's more familiar with `loongarch`

.gitmodules
Nayef211: @nmanhong I'm happy to merge this once we've removed the following line from this PR.

PR #1911, opened 2022-09-20
update documentation to reflect IMDB output

acxz:
When attempting to use the IMDB api, I got results that were different from what the docs suggested. This PR attempts to update the docs with the correct output of the IMDB api.

### DIFF
--- a/torchtext/datasets/imdb.py
+++ b/torchtext/datasets/imdb.py
@@ -82,8 +82,8 @@ def IMDB(root: str, split: Union[Tuple[str], str]):
         root: Directory where the datasets are saved. Default: os.path.expanduser('~/.torchtext/cache')
         split: split or splits to be returned. Can be a string or tuple of strings. Default: (`train`, `test`)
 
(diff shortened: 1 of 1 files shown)


Nayef211:
@acxz I think this might have been an oversight when the dataset was implemented. To be consistent with our other text classification dataset, we do want to return labels as integers. If you want to take this on, it would be as simple as adding a  `MAP_LABELS` similar to how we do it in the QNLI dataset.
https://github.com/pytorch/text/blob/fd49d42c6c90fc4b556b63227f3509292bbdba11/torchtext/datasets/qnli.py#L40

acxz:
Ah I see, I'll take a stab at this in the upcoming weekend. Thanks for the implementation pointers.

Nayef211:
@acxz lmk if you still plan on taking this on, otherwise I'm happy to submit a PR for it this week 😄

acxz:
Yeah, actually if you could take it on that would be great. Ended up traveling this weekend and this week is looking a bit busy. Thanks for that!

joecummings:
Closed in favor of #1914

PR #1914, opened 2022-09-27
Resolve inconsistency in IMDB label output

Nayef211:
## Description
- As per https://github.com/pytorch/text/pull/1911#issuecomment-1253111364, we need to update the label output to be integers instead of strings

## Testing
- `pytest test/torchtext_unittest/datasets/test_imdb.py`

Closes #1911


joecummings (approved):
lgtm

PR #1916, opened 2022-09-29
[Feature] Added capability to add special tokens in GPT2BPEEncoder and avoid splitting on them

reachsumit:
## Description

Add add_special_tokens feature to GPT2BPETokenizer

This change is targeted towards a requirement posted internally. It adds a new function `add_special_tokens` to `GPT2BPETokenizer` in order to enable the user to specify a dictionary of token that should not be changed during tokenization. Any newly specified token shall also be added to the vocabulary.

The functionality is same as HuggingFace's `add_special_tokens` feature.

## Types of changes

## Changes made
* Made changes to C++ code (`GPT2BPEEncoder` class) to support the `add_special_tokens` functionality. Also added the logic to ensure those special token are not split during tokenization.
* Updated torch bindings and py bindings as required.
* Added `add_special_tokens` function to Python interface (`GPT2BPETokenizer` class).
* Added unit test to extensively test the tokenization process.

## Testing
* Verified tokenization output to be consistent with HuggingFace's tokenizer.
* No issue identified with any of the unit tests.


torchtext/csrc/gpt2_bpe_tokenizer.h
```diff
@@ -55,8 +58,13 @@ struct GPT2BPEEncoder : torch::CustomClassHolder {
  private:
   const int64_t inf_;
   // Encode byte into an unicode character.
-  std::vector<std::string> ByteEncode_(std::string token);
+  std::vector<std::string> ByteEncode_(
+      std::string token,
+      bool is_never_split_token);
   int64_t GetBPEMergeRank_(std::string pair);
+  int64_t added_to_vocab_tokens_count;
+  //   std::set<std::string> bpe_never_split_set_;
```
joecummings, line 66: Remove uncommented code

torchtext/transforms.py
```diff
@@ -1,7 +1,7 @@
 import json
 from copy import deepcopy
 from functools import lru_cache
-from typing import Any, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
```
joecummings, line 4: Use more general `Mapping`

torchtext/transforms.py
```diff
@@ -349,6 +359,26 @@ def _tokenize(self, text: str) -> List[str]:
         """
         return self.bpe.tokenize(text)
 
+    def add_special_tokens(self, special_tokens_dict: Dict[str, Union[str, List[str]]]) -> int:
```
joecummings, line 362: How did you decide on this implementation? As I remember from the post, the inspiration was primarily from HuggingFace's implementation: https://huggingface.co/docs/transformers/internal/tokenization_utils#transformers.SpecialTokensMixin.add_tokens

reachsumit: This implementation is very similar to the HuggingFace's. 

For HuggingFace:
* Mixin defines the special tokens, and the structure enforced on the input: https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils_base.py#L896
* `PreTrainedTokenizer` keeps a separate encoder map for special tokens that are added later: https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils.py#L430 (also creates a trie to save all those tokens that are not to be split: https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils.py#L445)
* Added tokens encoder map is checked first when doing encoding: https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils.py#L586
* Regex based approach is used to distinguish between non-special and special tokens: https://github.com/huggingface/transformers/blob/3f936df66287f557c6528912a9a68d7850913b9b/src/transformers/tokenization_utils.py#L511

Of course, HuggingFace's implementation is more generic and is extensible to all of their tokenizers through inheritance. `GPT2Tokenizer` inherits `PreTrainedTokenizer` which inherits `PreTrainedTokenizerBase`, which implements two mixins: SpecialTokensMixin, PushToHubMixin. Comparatively torchtext's tokenizers all inherit `nn.Module` with their respective (and fairly separated imo) C++ implementation.

This pull request is the minimal implementation of their general Python approach. That's why I kept the Python API to be the same. A more general implementation is possible as C++ supports multiple inheritance. But it might take a bit more time and require a broader discussion with your team.

torchtext/transforms.py
reachsumit: done. Thank you!

torchtext/csrc/gpt2_bpe_tokenizer.h
reachsumit: Thanks, Joe! I removed the comment.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -282,14 +364,47 @@ std::vector<int64_t> GPT2BPEEncoder::Encode(const std::string& text) {
...
+      continue;
+    bpe_never_split_set_.insert(token);
+    if (!bpe_encoder_.contains(token)) {
+      added_tokens_encoder.insert(
+          token, bpe_encoder_.size() + added_tokens_encoder.size());
+      newly_added++;
+    }
+  }
+
+  added_to_vocab_tokens_count += newly_added;
```
Nayef211, line 404: I might be missing something here but where exactly is this variable used?

torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -61,28 +63,94 @@ std::vector<std::string> gpt2_bpe_pre_tokenizer(std::string input) {
...
+          } else { // push last whitespace char as a token if it is not a space
+            tokens.push_back(token.substr(token.length() - 1));
+          }
+        }
+      } else if (prepend_space) {
+        tokens.push_back(" " + token);
+        prepend_space = false;
+      } else {
+        tokens.push_back(token);
+      }
```
Nayef211, line 153: I'm having a bit of a hard time following all of the additional logic to this method. Having a more detailed explanation via code comments + adding a few lines to the commented pseudocode would really help with readability for someone that is new to our codebase.

Nayef211 (commented):
Thanks @reachsumit for taking the time to implement a highly requested feature to GPT2BPEEncoder. Let me know if my comments make sense. Happy to also have some offline discussions as needed before approving this PR! 😄

torchtext/csrc/register_pybindings.cpp
```diff
@@ -179,6 +179,16 @@ PYBIND11_MODULE(_torchtext, m) {
...
+      .def(
+          "add_special_tokens",
+          [](const c10::intrusive_ptr<GPT2BPEEncoder>& self,
+             const std::unordered_map<std::string, std::string>& items,
+             const std::vector<std::string>& additional) {
+            c10::Dict<std::string, std::string> d;
+            for (const auto& item : items)
+              d.insert(item.first, item.second);
+            return (self->AddSpecialTokens(d, additional));
+          })
```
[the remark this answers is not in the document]
reachsumit: I originally tried using the `std::unordered_map`, but I get the following compilation error (`unsupported input type: std::unordered_map<Key, Value>. Please use Dict<Key, Value> instead.`) on doing so.

```
/Users/sumitkumar/mambaforge/envs/bootcamp-rl/lib/python3.10/site-packages/torch/include/ATen/core/boxing/impl/make_boxed_from_unboxed_functor.h:164:5: error: static_assert failed "You tried to register a kernel with an unsupported input type: std::unordered_map<Key, Value>. Please use Dict<Key, Value> instead."
    static_assert(AllowDeprecatedTypes,
    ^             ~~~~~~~~~~~~~~~~~~~~
```

So, I switched to `c10::Dict` instead. But it turned out that the Python dictionary doesn't directly convert to our custom `c10:Dict` map, and it gives me following error if I try to do so.

```
TypeError: add_special_tokens(): incompatible function arguments. The following argument types are supported:
    1. (self: torchtext._torchtext.GPT2BPEEncoder, arg0: c10::Dict<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, long long>) -> int

Invoked with: <torchtext._torchtext.GPT2BPEEncoder object at 0x12428de70>, {}

Did you forget to `#include <pybind11/stl.h>`? Or <pybind11/complex.h>,
<pybind11/functional.h>, <pybind11/chrono.h>, etc. Some automatic
conversions are optional and require extra headers to be included
when compiling your pybind11 module.
```

So, I chose to pass Python `dict` to C++ `unordered_map`, but convert it bindings to avoid the first error. This is similar to the logic for constructor.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -61,28 +63,94 @@ std::vector<std::string> gpt2_bpe_pre_tokenizer(std::string input) {
   // - ELSE, add token to return list
```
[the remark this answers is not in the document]
reachsumit: I added more comments in `AddSpecialTokens` to indicate the usage of this set.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
reachsumit: I originally meant to use this variable to return the vocab size (original encoder size + added encoder size), similar to how HuggingFace API provisions it. But it appears that we don't really have any existing requirement for fetching size, so I just removed it in the latest commit.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
reachsumit: Sorry, for the confusion. I agree that the logic is complex to understand just by looking at the code. I added some detailed explanation and example on the steps in the latest commit. Let me know if that is sufficient or any further details would help.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -61,28 +63,122 @@ std::vector<std::string> gpt2_bpe_pre_tokenizer(std::string input) {
...
-    if (is_whitespace(token)) {
-      prepend_space = false;
-      if (inp.empty()) { // token is last token
-        tokens.push_back(token);
-      } else {
-        if (token.length() > 1) {
-          tokens.push_back(token.substr(0, token.length() - 1));
+  std::vector<std::string> index_matches;
+
+  /* Notes on handling Special Tokens:
```
reachsumit, line 69: This thread is also very helpful for understanding spacing in BPE tokenization: https://discuss.huggingface.co/t/bpe-tokenizers-and-spaces-before-words/475

torchtext/transforms.py
```diff
@@ -349,6 +359,26 @@ def _tokenize(self, text: str) -> List[str]:
...
+        :rtype: int
+        """
+        for key in special_tokens_dict.keys():
+            assert (
+                key in self.SPECIAL_TOKENS_ATTRIBUTES
+            ), f"Key '{key}' is not in the special token list: {self.SPECIAL_TOKENS_ATTRIBUTES}"
+
+        return self.bpe.add_special_tokens(
+            {k: v for k, v in special_tokens_dict.items() if k != "additional_special_tokens"},
+            special_tokens_dict.get("additional_special_tokens", []),
```
reachsumit, line 379: The reason for keeping standard special tokens separate from "additional" special tokens is to be able to easily provision access to those standard tokens later. (like this: https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils_base.py#L979).

torchtext/transforms.py
```diff
@@ -294,6 +294,16 @@ class GPT2BPETokenizer(Module):
     def __init__(self, encoder_json_path: str, vocab_bpe_path: str, return_tokens: bool = False) -> None:
         super().__init__()
         self._seperator = "\u0001"
+        self.SPECIAL_TOKENS_ATTRIBUTES = [
```
joecummings, line 297: These special tokens are fairly common across all tokenizers. Would it make sense to pull these out into `commons` or `utils` so that we can reuse them across multiple transforms? cc @Nayef211

reachsumit: I moved `SPECIAL_TOKENS_ATTRIBUTES` to `utils` in latest commit.

Nayef211: @mthrok taught me about the [YAGNI principle](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) which I believe is quite applicable here. Let's not move these out into `commons` or `utils` until we see a concrete case where they can be reused. I also think it would make more sense for `SPECIAL_TOKENS_ATTRIBUTES` to be a class variable of `GPT2BPETokenizer` rather than an instance variable as these attributes would not be changing across instances. Lmk if this makes sense.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
Nayef211: I think the explanation you added really helps with the code readability, thanks!

Nayef211 (approved):
Thanks for addressing all the comments @reachsumit! The changes LGTM other than a few nits. I will merge this PR once those are resolved!

reachsumit:
Thanks for your comments, @Nayef211 ! I addressed them in the latest commits and also rebased the change on latest main branch. Let me know if in case of any concerns.

PR #1918, opened 2022-09-30
[Feature] Add ability to load HF checkpoints into T5 model

joecummings:
## Changes

* Adds a `build_model_from_huggingface_ckpt` static method to the `T5Bundler`. 

## Testing

* Adds integration tests to check that an HF model loaded with HF weights and a TorchText T5 model loaded with HF weights both return the same thing when applied to data

Future considerations: This does not support loading from external URLs - planning on adding this in a follow-up Diff from the internal FB side so I can test with Manifold (+ possible GDrive, static public links, Github).

### DIFF
--- a/test/integration_tests/prototype/test_models.py
+++ b/test/integration_tests/prototype/test_models.py
@@ -151,13 +151,15 @@ class TestLoadFromHFCheckpoints(TorchtextTestCase):
         # check that encoder layers match
         for i in range(config.num_encoder_layers + 1):
             if i < config.num_encoder_layers:
+                hf_output_sa = hf_output.attentions[i] if encoder_only else hf_output.encoder_attentions[i]
                 # self-attention scores
                 assert torch.equal(
-                    our_output["encoder_sa_scores"][i], hf_output.encoder_attentions[i]
+                    our_output["encoder_sa_scores"][i], hf_output_sa
                 ), f"Mismatched self-attention scores for encoder layer {i}"
+            hf_output_hs = hf_output.hidden_states[i] if encoder_only else hf_output.encoder_hidden_states[i]
             # encoder hidden states
             assert torch.equal(
-                our_output["encoder_hidden_states"][i], hf_output.encoder_hidden_states[i]
+                our_output["encoder_hidden_states"][i], hf_output_hs
             ), f"Mismatched hidden states for encoder layer {i}"
 
         if not encoder_only:
--- a/test/integration_tests/prototype/test_models.py
+++ b/test/integration_tests/prototype/test_models.py
@@ -141,33 +141,26 @@ class TestT5WrapperCheckpoint(TorchtextTestCase):
 
 
 class TestLoadFromHFCheckpoints(TorchtextTestCase):
+    def setUp(self) -> None:
+        self.encoder_input_ids = torch.tensor([[1, 2, 3, 4, 5, 6], [7, 8, 9, 0, 0, 0]])
+        self.encoder_padding_mask = torch.tensor([[1, 1, 1, 1, 1, 1], [1, 1, 1, 0, 0, 0]])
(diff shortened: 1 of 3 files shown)


torchtext/prototype/models/t5/bundler.py
```diff
@@ -133,6 +135,129 @@ def build_model(
...
+        *,
+        freeze_model: bool = False,
+        strict: bool = True,
+    ) -> T5Model:
+        config_path = f"{ckpt_path}/config.json"
+        model_path = f"{ckpt_path}/pytorch_model.bin"
+
+        assert os.path.exists(model_path), f"No PyTorch model found at {model_path}"
+
+        with open(config_path, "r") as handle:
```
joecummings, line 150: This is OSS compatible, but not internally compatible. Coming in a follow up commit.

test/integration_tests/prototype/test_models.py
```diff
@@ -135,3 +138,127 @@ def test_t5_wrapper_checkpoint(self, name) -> None:
...
+            model_path = f"{tmp_dir}/hf_t5_small_gen"
+
+            t5_small_gen = T5ForConditionalGeneration.from_pretrained("t5-small")
+            t5_small_gen.save_pretrained(model_path)
+
+            encoder_input_ids = torch.tensor([[1, 2, 3, 4, 5, 6], [7, 8, 9, 0, 0, 0]])
+            encoder_padding_mask = torch.tensor([[1, 1, 1, 1, 1, 1], [1, 1, 1, 0, 0, 0]])
+
+            decoder_input_ids = torch.tensor([[7, 8, 9, 0, 0, 0], [10, 11, 12, 0, 0, 0]])
+            decoder_padding_mask = torch.tensor([[1, 1, 1, 0, 0, 0], [1, 1, 1, 0, 0, 0]])
```
Nayef211, line 249: Since these seem to be reused across tests, can we store them in the test class within the `setUp` method?

test/integration_tests/prototype/test_models.py
```diff
@@ -135,3 +138,127 @@ def test_t5_wrapper_checkpoint(self, name) -> None:
...
+            for i in range(config.num_encoder_layers + 1):
+                if i < config.num_encoder_layers:
+                    # self-attention scores
+                    assert torch.equal(
+                        our_output["encoder_sa_scores"][i], hf_output.encoder_attentions[i]
+                    ), f"Mismatched self-attention scores for encoder layer {i}"
+                # encoder hidden states
+                assert torch.equal(
+                    our_output["encoder_hidden_states"][i], hf_output.encoder_hidden_states[i]
+                ), f"Mismatched hidden states for encoder layer {i}"
```
Nayef211, line 169: This logic seems like an exact copy of that above. Can't we just do the check for the decoder layers if it's not an encoder only model?

torchtext/prototype/models/t5/bundler.py
```diff
@@ -142,14 +142,25 @@ def build_model_from_huggingface_ckpt(
         freeze_model: bool = False,
         strict: bool = True,
     ) -> T5Model:
+        """Build T5Model model from a HuggingFace checkpoint.
+
+        Note: Only works with Huggingface models saved in the PyTorch format. Will not work \
```
Nayef211, line 147: nit: do we need the "\" in the docstrings?

Nayef211 (approved):
LGTM! Thanks for adding this capability @joecummings! I would recommend pulling in these changes and ensuring it works with internal Meta infra before merging in the PR. 😄

torchtext/prototype/models/t5/bundler.py
joecummings: Sorry - not entirely sure what you're referring to here?

Nayef211: Sorry updated the comment. Looks like backslash characters don't show up when you put quotes around them which was a noob mistake 😅

PR #1919, opened 2022-10-01
Add decoding capability to GPT2BPE tokenizer

reachsumit:
## Description

Add add_special_tokens feature to GPT2BPETokenizer

This change is targeted towards a requirement posted internally. It adds a new function `add_special_tokens` to `GPT2BPETokenizer` in order to enable the user to specify a dictionary of token that should not be changed during tokenization. Any newly specified token shall also be added to the vocabulary.

The change adds the capability to decode tokens ids back to text. This change is required following the internal discussion.

## Types of changes

## Changes made
* Made changes to C++ code (`GPT2BPEEncoder` class) to support the `decode` functionality.
* Updated torch bindings and py bindings as required.
* Added `decode` function to Python interface (`GPT2BPETokenizer` class).
* Added unit test to test the decoding process.

## Testing
* No issue identified with any of the unit tests.

### DIFF
--- a/test/torchtext_unittest/test_transforms.py
+++ b/test/torchtext_unittest/test_transforms.py
@@ -364,10 +364,32 @@ class TestGPT2BPETokenizer(TorchtextTestCase):
             else:
                 self.assertEqual(tokenizer(txt), expected_token_ids[idx])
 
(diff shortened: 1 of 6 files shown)


Nayef211 (approved):
LGTM!

PR #1924, opened 2022-10-04
Update pkg_helpers.bash for PyTorch 1.13 release

abhinavarora:


mthrok (commented):
Why is this necessary?

The current proposed change is touching the code path for nightly build. So I don't think this is correct.

Has something about the release process changed that require change in the bash script?

abhinavarora:
> Why is this necessary?
> 
> The current proposed change is touching the code path for nightly build. So I don't think this is correct.
> 
> Has something about the release process changed that require change in the bash script?

Hey @mthrok I was just following the step 5 from [this wiki](https://www.internalfb.com/intern/wiki/TorchAudio/Release_Process/Runbook/). Please let me know if this is stale?

mthrok:
Can you rebase on https://github.com/pytorch/text/pull/1923 and see if it makes difference? I believe it doesn’t make a difference but let’s see.

mthrok:
In the previous release, this change was made as a patch. But I believe that the correct fix is to make Arm Mac build job aware of PYTORCH_VERSION environment variable.

PR #1925, opened 2022-10-04
Update decoding logic to handle special tokens

reachsumit:
## Description

Update decoding logic to handle special tokens

This change is targeted is a follow up to recently added `add_special_tokens` and `decode` features to `GPT2BPETokenizer`. 

The change adds the capability to decode special tokens ids back to text.

## Types of changes

## Changes made
* Made changes to C++ code (`GPT2BPEEncoder` class) to update the `decode` functionality for special tokens.
* Added unit test to test the new decoding process.

## Testing
* No issue identified with unit tests.

### DIFF
--- a/torchtext/csrc/gpt2_bpe_tokenizer.cpp
+++ b/torchtext/csrc/gpt2_bpe_tokenizer.cpp
@@ -400,7 +400,8 @@ std::vector<int64_t> GPT2BPEEncoder::Encode(const std::string& text) {
 
 std::string GPT2BPEEncoder::Decode(const std::vector<int64_t>& tokens) {
   std::string text;
-  std::vector<bool> special_token_flags(tokens.size());
+  bool is_prev_special = false;
+  bool is_current_special = false;
   // setup converter for converting wide chars to/from chars
   using convert_type = std::codecvt_utf8<wchar_t>;
   std::wstring_convert<convert_type, wchar_t> converter;
@@ -412,13 +413,13 @@ std::string GPT2BPEEncoder::Decode(const std::vector<int64_t>& tokens) {
     if (added_tokens_decoder_.contains(token)) {
       // string is a special token from extended vocab
       decoded_token = added_tokens_decoder_.at(token);
-      special_token_flags[tok_idx] = true;
+      is_current_special = true;
     } else {
       const std::string str = bpe_decoder_.at(token);
       if (bpe_never_split_set_.find(str) != bpe_never_split_set_.end()) {
         // string is a special token from known vocab
         decoded_token = str;
-        special_token_flags[tok_idx] = true;
+        is_current_special = true;
       } else {
         // string is a regular token from known vocab
         const std::wstring ws = converter.from_bytes(str);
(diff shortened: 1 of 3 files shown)


torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -397,18 +397,45 @@ std::vector<int64_t> GPT2BPEEncoder::Encode(const std::string& text) {
...
+          // get output character from byte decoder for each wide character
+          unsigned char uchr = byte_decoder_.at(converter.to_bytes(wchr));
+          decoded_token.push_back(uchr);
+        }
+      }
+    }
+
+    // fix left space(s) for special tokens
+    if (special_token_flags[tok_idx] == true &&
+        (tok_idx > 0 && special_token_flags[tok_idx - 1] == false)) {
```
Nayef211, line 432: Just want to double check, does `&& special_token_flags[tok_idx - 1] == false` only ensure that we append a space if the token before the special token is a regular token? If so, why do we do this?

Nayef211 (approved):
Overall changes LGTM. I would also wait for a stamp from @abhinavarora before merging this.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
reachsumit: It ensures that we insert space only if the last token wasn't special, because if the last token was special, it would have added space after it (through code at line 438). So we avoid having one extra space in case when two special tokens are next to each other.

Nayef211:
> @Nayef211 @reachsumit What is the status of this? Can we rebase, merge and let Jin's team know they have this capability now?

Changes looked fine to me but wanted a review from @abhinavarora as well to make sure I didn't miss anything!

reachsumit:
@joecummings I believe there is no additional PR required to support Jin's team's usecase. After this PR is approved and merged, we should be good to go.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -397,18 +397,45 @@ std::vector<int64_t> GPT2BPEEncoder::Encode(const std::string& text) {
...
+        const std::wstring ws = converter.from_bytes(str);
+        for (wchar_t wchr : ws) {
+          // get output character from byte decoder for each wide character
+          unsigned char uchr = byte_decoder_.at(converter.to_bytes(wchr));
+          decoded_token.push_back(uchr);
+        }
+      }
+    }
+
+    // fix left space(s) for special tokens
```
abhinavarora, line 430: Let's add an example in comments to make this clearer?

torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -397,18 +397,45 @@ std::vector<int64_t> GPT2BPEEncoder::Encode(const std::string& text) {
 
 std::string GPT2BPEEncoder::Decode(const std::vector<int64_t>& tokens) {
   std::string text;
+  std::vector<bool> special_token_flags(tokens.size());
```
abhinavarora, line 400: Creating a vector seems to be an overkill. From the logic it seems we should be good with 2 bools. Is this correct or am I missing something?

abhinavarora (commented):
Left few comments. @reachsumit could you take a look? Logic looks fine to me.

reachsumit:
Rebased on latest `main` to fix conflicts, and addressed recent comments.
* Added examples for adding back spaces while decoding.
* Replaced using boolean array with two boolean variables.

torchtext/csrc/gpt2_bpe_tokenizer.cpp
```diff
@@ -400,19 +400,67 @@ std::vector<int64_t> GPT2BPEEncoder::Encode(const std::string& text) {
...
+        decoded_token = str;
+        is_current_special = true;
+      } else {
+        // string is a regular token from known vocab
+        const std::wstring ws = converter.from_bytes(str);
+        for (wchar_t wchr : ws) {
+          // get output character from byte decoder for each wide character
+          unsigned char uchr = byte_decoder_.at(converter.to_bytes(wchr));
+          decoded_token.push_back(uchr);
+          is_current_special = false;
```
abhinavarora, line 430: How about setting this as `false` once at the beginning of the loop?

abhinavarora (approved):
Overall LGTM, left a minor comment.

PR #1934, opened 2022-10-11
update dataset RTE information

Geaming-CHN:
NUM_LINES in torchtext/datasets/rte.py is wrong or maybe out of date (line 24). Update the information in the part

### DIFF
--- a/torchtext/datasets/rte.py
+++ b/torchtext/datasets/rte.py
@@ -22,9 +22,9 @@ URL = "https://dl.fbaipublicfiles.com/glue/data/RTE.zip"
 MD5 = "bef554d0cafd4ab6743488101c638539"
 
 NUM_LINES = {
-    "train": 67349,
-    "dev": 872,
-    "test": 1821,
+    "train": 2490,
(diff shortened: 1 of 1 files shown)


joecummings:
Hi @Geaming-CHN, thanks for the contribution! Can you please link in the PR where you found these updated numbers?

Geaming-CHN:
> Hi @Geaming-CHN, thanks for the contribution! Can you please link in the PR where you found these updated numbers?

this link might help. https://www.tensorflow.org/datasets/catalog/glue#gluerte
or you just use the following codes in torchtext also can get the correct numbers
```python
from torchtext.datasets import RTE
d1, d2, d3 = RTE()
print(len(list(d1)))
print(len(list(d2)))
print(len(list(d3)))
```
```
2490
277
3000
```
so u can see it, the code runs correctly but the numbers are wrong

Nayef211 (approved):
@Geaming-CHN thanks for catching this, I think this must have been an oversight when the dataset was added.

PR #1936, opened 2022-10-11
Add Character Level BPE Tokenizer

rshraga:
Summary: This change adds a character level BPE tokenizer to the set of available transforms. It takes a pre-trained encoder dict (i.e vocab dict) and merge list as input. It is not using C++ for encoding / decoding at this time.

Differential Revision: D40186470


joecummings:
According to a discussion with @Nayef211, this should be closed and can be cherry-picked into `main` after it passes all tests in Phabricator.

Issue #1949, opened 2022-10-18
Ensure `main` and `fbsync` are both in sync


Nayef211:
## Description
- Currently the `fbsync` branch is [172 commits ahead](https://github.com/pytorch/text/compare/main...fbsync), [749 commits behind](https://github.com/pytorch/text/compare/fbsync...main) main.
- We want to ensure that `main` and `fbsync` branches are both up to date which is a 2 stage process
   1. Merge all commits from `main` to `fbsync` and resolve any merge conflicts
   2. Merge all commits from `fbsync` to `main` and resolve any merge conflicts 

<img width="485" alt="image" src="https://user-images.githubusercontent.com/22487263/196543598-656368a3-409e-4df1-8af7-40acc5e9c23a.png">

PR #1950, opened 2022-10-18
[MERGE 2/2] merge `fbsync` branch to `main`

Nayef211:
## Description
- Refer to #1949


rshraga:
Retried failing build and unittest to see if resolved. Is fbsync healthy overall? Where in the UI can we see that fbsync / all these commits are in a good state?

Nayef211:
> Retried failing build and unittest to see if resolved. Is fbsync healthy overall? Where in the UI can we see that fbsync / all these commits are in a good state?

Looks like all tests are passing now, thanks for doing the retry. You can see the status of the commits in `fbsync` (i.e. if all tests are passing) here https://github.com/pytorch/text/commits/fbsync. However this is not a relevant signal since we don't generally create PRs against the `fbsync` branch. Rather once a diff gets commited on fbcode, a corresponding commit is created in the `fbsync` branch bypassing all CI tests.

joecummings (approved):

Issue #1951, opened 2022-10-19
Error when importing torchtext: "No module named 'regex'"


seemethere:
We're observing errors on torchtext when attempting to import wheels due to a missing dependency.

Error looks like:
```
ERROR conda.cli.main_run:execute(41): `conda run python -c import torchtext; print('package version is ', torchtext.__version__)` failed. (See above for error)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/__w/_temp/conda_environment_3282699477/lib/python3.7/site-packages/torchtext/__init__.py", line 12, in <module>
    from . import data, datasets, prototype, functional, models, nn, transforms, utils, vocab, experimental
  File "/__w/_temp/conda_environment_3282699477/lib/python3.7/site-packages/torchtext/models/__init__.py", line 1, in <module>
    from .roberta import *  # noqa: F401, F403
  File "/__w/_temp/conda_environment_3282699477/lib/python3.7/site-packages/torchtext/models/roberta/__init__.py", line 1, in <module>
    from .bundler import (
  File "/__w/_temp/conda_environment_3282699477/lib/python3.7/site-packages/torchtext/models/roberta/bundler.py", line 13, in <module>
    import torchtext.transforms as T
  File "/__w/_temp/conda_environment_3282699477/lib/python3.7/site-packages/torchtext/transforms.py", line 6, in <module>
    import regex as re
ModuleNotFoundError: No module named 'regex'
```

Issue was introduced in #1946 and was fixed for conda but not fixed for wheels

Issue #1702, opened 2022-05-05
Add TorchData as a required dependency in TorchText for 0.13.0 release


Nayef211:
## 🚀 Feature

TorchData is currently an optional dependency which means users have to `pip install torchdata` if they want to be able to use our datasets. Since datasets are a core feature of TorchText, we want to improve the user experience by making TorchData a required dependency.

After the next `torch` and `torchdata` branch cuts for version `1.12`, we want to take a hard dependency on the `torchdata` nightlies.

Nayef211:
cc @atalman @malfet @parmeet @ejguan @seemethere

Nayef211:
@atalman @malfet @ejguan do you guys know if we have the go ahead to make TorchData as a required dep moving forward? I know we were aiming to do this with the last release but it got pushed back to the 0.13 release instead.

ejguan:
Yeah, it would be great if we TorchData are aware of the requirements/criteria to make this happen. Then, we can investigate if we have any step missing.

ejguan:
After #1979 is fixed, could we start to make torchdata as a hard dependency in the main branch? Then, we will be able to spend time on figuring out what's wrong with the installation script.

Nayef211:
Closed by #1985

PR #1952, opened 2022-10-19
Add regex to "install_requires"

seemethere:
Fixes https://github.com/pytorch/text/issues/1951

### DIFF
--- a/.circleci/unittest/linux/scripts/environment.yml
+++ b/.circleci/unittest/linux/scripts/environment.yml
@@ -15,5 +15,6 @@ dependencies:
       - spacy
       - tqdm
       - expecttest
+      - regex
       - https://github.com/explosion/spacy-models/releases/download/de_core_news_sm-3.0.0/de_core_news_sm-3.0.0.tar.gz#egg=de_core_news_sm==3.0.0
       - https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.0.0/en_core_web_sm-3.0.0.tar.gz#egg=en_core_web_sm==3.0.0
--- a/.circleci/unittest/windows/scripts/environment.yml
+++ b/.circleci/unittest/windows/scripts/environment.yml
@@ -17,5 +17,6 @@ dependencies:
       - tqdm
       - certifi
       - expecttest
+      - regex
       - https://github.com/explosion/spacy-models/releases/download/de_core_news_sm-3.0.0/de_core_news_sm-3.0.0.tar.gz#egg=de_core_news_sm==3.0.0
       - https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.0.0/en_core_web_sm-3.0.0.tar.gz#egg=en_core_web_sm==3.0.0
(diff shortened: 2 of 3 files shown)


osalpekar (approved):
Thanks! I'm wondering how this escaped smoke tests?

malfet (approved):
LGTM, but can we please get an ack from torchtext team that it is indeed the case. Also, why isn't there a torchdata dependency?

joecummings:
The dependency on regex is relatively recent, but yes I'm also confused as to how this escaped smoke tests. Looking into this now. 

Also, it looks like you might need to add this here, as well: https://github.com/pytorch/text/blob/5eb33ce1f7447df5069b6cfb55b8177c9bbaff08/.circleci/unittest/windows/scripts/environment.yml#L8

CC @Nayef211 for question about Torchdata.

seemethere:
> The dependency on regex is relatively recent, but yes I'm also confused as to how this escaped smoke tests. Looking into this now.
> 
> Also, it looks like you might need to add `regex` as a dependency here, as well:
> 
> https://github.com/pytorch/text/blob/5eb33ce1f7447df5069b6cfb55b8177c9bbaff08/.circleci/unittest/windows/scripts/environment.yml#L8
> 
> CC @Nayef211 for question about Torchdata.

I'm pretty sure regex gets installed as a transitive dependency for the unittest workflows which might explain the discrepancy. Also the binary builds don't run smoke tests (like importing torchtext) so it's understandable how this wasn't caught.

seemethere:
As well, why do we even need this dependency in the first place? Is the built-in `re` module not good enough for the usecase of `findall`?

Nayef211:
> As well, why do we even need this dependency in the first place? Is the built-in `re` module not good enough for the usecase of `findall`?
> 
> cc @rshraga

I'm not 100% sure why we didn't use `re` but you can find some relevant discussions around adding using regex in the [internal diff thread ](https://www.internalfb.com/diff/D40186470?dst_version_fbid=5257229001071858&transaction_fbid=476429154443011). IIUC this is a temporary addition and will be removed as soon as @rshraga migrates the regex pattern matching logic to a C++ operator.

> Also, why isn't there a torchdata dependency?

@malfet see https://github.com/pytorch/text/issues/1702 where we tagged you guys in asking what the process would be to make torchdata a required dep. I believe there was also an [internal chat thread](https://fb.workplace.com/chat/t/5361854030576540) about this but I'm not sure there was a resolution.

seemethere:
> > As well, why do we even need this dependency in the first place? Is the built-in `re` module not good enough for the usecase of `findall`?
> > cc @rshraga
> 
> I'm not 100% sure why we didn't use `re` but you can find some relevant discussions around adding using regex in the [internal diff thread ](https://www.internalfb.com/diff/D40186470?dst_version_fbid=5257229001071858&transaction_fbid=476429154443011). IIUC this is a temporary addition and will be removed as soon as @rshraga migrates the regex pattern matching logic to a C++ operator.
> 
> > Also, why isn't there a torchdata dependency?
> 
> @malfet see #1702 where we tagged you guys in asking what the process would be to make torchdata a required dep. I believe there was also an [internal chat thread](https://fb.workplace.com/chat/t/5361854030576540) about this but I'm not sure there was a resolution.

Actually switching it to `re` provides no negative results so I'm just going to submit a new PR to do that instead

seemethere:
Closing in favor of https://github.com/pytorch/text/pull/1953

PR #1958, opened 2022-10-21
Add Torchdata as a requirement and remove conditional imports of Torchdata

abhinavarora:


Nayef211:
For conda builds I think you need to add `torchdata` as a runtime dependency to the [packaging/torchtext/meta.yaml file](https://github.com/pytorch/text/blob/main/packaging/torchtext/meta.yaml#L23).

malfet:
Imo this PR needs to land into trunk first and then cherry-picked into release branch

PR #1969, opened 2022-10-27
[Nova] Mac x86 Unittests

osalpekar:
This should be unblocked after https://github.com/pytorch/test-infra/pull/968 is merged.

### DIFF
--- a/.github/workflows/test-macos-cpu.yml
+++ b/.github/workflows/test-macos-cpu.yml
@@ -43,8 +43,8 @@ jobs:
         export MKL_CONSTRAINT='mkl==2021.2.0'
 
         # Create Conda Env
-        conda create -yp ci_env python="${PYTHON_VERSION}"
-        conda activate /work/ci_env
+        conda create -yp ./ci_env python="${PYTHON_VERSION}"
+        conda activate ./ci_env
         python3 -m pip --quiet install cmake>=3.18.0 ninja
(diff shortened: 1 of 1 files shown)


.github/workflows/test-macos-cpu.yml
```diff
@@ -0,0 +1,72 @@
...
+
+jobs:
+  tests:
+    strategy:
+      matrix:
+        python_version: ["3.7", "3.8", "3.9", "3.10"]
+      fail-fast: false
+    uses: pytorch/test-infra/.github/workflows/macos_job.yml@main
+    with:
+      runner: macos-12
```
osalpekar, line 23: I'm going to make the runner arg a matrix entry to support both mac x86 and M1 from a single workflow. (will need to see if this causes any dependency issues while running the tests)

osalpekar: Will do this in a follow-up since there currently aren't M1 tests so this is slightly lower-pri.

osalpekar:
Unittests for Mac on py3.10 take ~40 mins even on CircleCI (see https://app.circleci.com/pipelines/github/pytorch/text/7086/workflows/c9b4abf6-705a-43fd-b76f-a1378af16ed7/jobs/245546), so I'll bump up the timeout - thinking of doing this in a follow-up PR.

seemethere (approved):

PR #1974, opened 2022-10-28
[Nova] Pass the smoke test scripts to Reusable Workflow

osalpekar:
If the smoke-test script is not passed to the reusable workflow, it will just use the naive smoke test implemented in the reusable workflow. Passing it (as we do here) will ensure the smoke test script that exists in this repo gets run after the build is complete.

### DIFF
--- a/.github/workflows/build-conda-linux.yml
+++ b/.github/workflows/build-conda-linux.yml
@@ -39,6 +39,7 @@ jobs:
       pre-script: ${{ matrix.pre-script }}
       post-script: ${{ matrix.post-script }}
       package-name: ${{ matrix.package-name }}
+      smoke-test-script: ${{ matrix.smoke-test-script }}
       # Using "development" as trigger event so these binaries are not uploaded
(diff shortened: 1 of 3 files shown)


weiwangmeta (commented):
Do we need build-conda-macos.yml? or why do we need to split conda/wheels for linux?

osalpekar:
> Do we need build-conda-macos.yml? or why do we need to split conda/wheels for linux?

build-conda-macos.yml is introduced in https://github.com/pytorch/text/pull/1973 which is not merged yet. That PR already passes the smoke test script.

weiwangmeta:
> > Do we need build-conda-macos.yml? or why do we need to split conda/wheels for linux?
> 
> build-conda-macos.yml is introduced in #1973 which is not merged yet. That PR already passes the smoke test script.

Ah I missed that. Let me approve this PR.

weiwangmeta (approved):
LGTM

PR #1975, opened 2022-10-28
[Nova] Add M1 Unittests

osalpekar:


osalpekar:
torchtext doesn't have M1 tests at the moment, so no need to increase support right off the bat. We can revisit this next half.

PR #1977, opened 2022-10-31
Ensure decoder mask is on same device as other tensors

joecummings:
In testing on a Bento notebook, located another device mismatch bug that - if user was using a GPU and decoder - the decoder mask would be on the CPU. 

Test plan: N2615118

Should unblock AI for CS workflow and others.


Nayef211 (approved):
LGTM

PR #1994, opened 2022-11-29
[Nova] Remove CUDA binary builds

osalpekar:
TorchText only has CPU binary builds, but our build matrix was including cuda builds as well. This should remove these unnecessary builds.


joecummings (approved):
lgtm 👍

PR #1997, opened 2022-11-30
Move read_from_tar to load_from_tar

ejguan:
Forward fix caused by deprecating PRs: https://github.com/pytorch/pytorch/pull/89794 and https://github.com/pytorch/data/pull/909


osalpekar (approved):
Thanks for the quick fix!

Issue #2001, opened 2022-12-07
One of the three datasets returned by `Multi30k` seems to be bugged.


raaaaaymond:
## 🐛 Bug

**Describe the bug** A clear and concise description of what the bug is.

The testing data returned by `Multi30k`  doesn't match the expected SHA256 hash. The precise error is:

```
RuntimeError: The computed hash 0681be16a532912288a91ddd573594fbdd57c0fbb81486eff7c55247e35326c2 of C:\Users\raaaa/.cache\torch\text\datasets\Multi30k\mmt16_task1_test.tar.gz does not match the expectedhash 6d1ca1dba99e2c5dd54cae1226ff11c2551e6ce63527ebb072a1f70f72a5cd36. Delete the file manually and retry.
This exception is thrown by __iter__ of HashCheckerIterDataPipe(hash_dict={'C:\\Users\\raaaa/.cache\\torch\\text\\datasets\\Multi30k\\mmt16_task1_test.tar.gz': '6d1ca1dba99e2c5dd54cae1226ff11c2551e6ce63527ebb072a1f70f72a5cd36'}, hash_type='sha256', rewind=True, source_datapipe=MapperIterDataPipe)
```

I've done what the message suggested; I deleted the files manually and did it again, but the same error occurs.

**To Reproduce** Steps to reproduce the behavior:

Paste the following into a new Python file and run it.

```
import torchtext

def _main():
    train, val, test = torchtext.datasets.Multi30k(language_pair=("de", "en"))
    # The following works fine because `val` and `train` datasets are fine.
    # for thing in val:
    #     print(thing)
    #     break
    # Invoking the generator (which is `test`) in the following way triggers the error.
    for thing in test:
        print(thing)
        break


if __name__ == "__main__":
    _main()
```

You should see the error I pasted above.

**Expected behavior** A clear and concise description of what you expected to happen.

I expect no error.

**Environment**

PyTorch version: 1.13.0+cu117
Is debug build: False
CUDA used to build PyTorch: 11.7
ROCM used to build PyTorch: N/A

OS: Microsoft Windows 11 Pro
GCC version: (x86_64-posix-seh, Built by strawberryperl.com project) 8.3.0
Clang version: Could not collect
CMake version: version 3.20.2
Libc version: N/A

Python version: 3.10.1 (tags/v3.10.1:2cd268a, Dec  6 2021, 19:10:37) [MSC v.1929 64 bit (AMD64)] (64-bit runtime)
[...]

PR #2003, opened 2022-12-07
update multi30k test dataset hash

rshraga:
**Description**
- This diff changes the expected hash of the test data set back to what it was before https://github.com/pytorch/text/pull/1816 as that hash seems to be expected now. 
- fixes the bug reported here https://github.com/pytorch/text/issues/2001

**Testing**
```
import torchtext

def _main():
    train, val, test = torchtext.datasets.Multi30k(language_pair=("de", "en"))
    for thing in test:
        print(thing)
        break

if __name__ == "__main__":
    _main()
```

Above script now works

### DIFF
--- a/torchtext/datasets/multi30k.py
+++ b/torchtext/datasets/multi30k.py
@@ -20,7 +20,7 @@ URL = {
 MD5 = {
     "train": "20140d013d05dd9a72dfde46478663ba05737ce983f478f960c1123c6671be5e",
     "valid": "a7aa20e9ebd5ba5adce7909498b94410996040857154dab029851af3a866da8c",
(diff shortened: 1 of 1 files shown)


joecummings:
Why was this changed in the first place here? Commit hash: c7e5a07890bf0a2eb1e8baa3e768f2e2a3091eba

@Nayef211

Nayef211 (approved):
LGTM thanks for the fix

mikelcq:
I installed the latest torch 1.13.1,  torchtext 0.14.1.
Why I still have the exact error?   even the printed hash log is the same as #2001 described.

PR #2011, opened 2022-12-19
Beginning of generation utils and necessary refactors of T5 Model 

joecummings:
## Context

We aim to add generation utils that support a number of encoder/decoder and decoder-based models. To do so, we also have to rework our current encoder/decoder model, T5.

## Changes

1. Separated logic for encoder and decoder into self-contained `nn.Modules`. 
    1a. Move dropout layers and norms to `T5Encoder` and `T5Decoder`
    1b. Pass `token_embeddings` to the encoder if constructed through the `T5Model`. Now the encoder can take in tokenized text *or* embedded text.
    1c. Add `get_encoder` and `get_decoder` getter functions (not torchscriptable ATM)
    1d. Update type annotations to allow for padding_masks and encoder_outputs
    1e. Change `T5Encoder` and `T5Decoder` return types to dictionaries

2. Added `GenerationUtils` class and `greedy_search` generation technique
    2a. Added deprecation warning to `T5Wrapper` until `beam_search` is added.

3. Froze configs to avoid mutating the model unnecessarily

## Testing

- One notebook showing parity with HuggingFace's T5 model
- One notebook showing that HuggingFace models can be used with the `GenerationUtil`
- Integration tests for the new endpoints and additional functions

## Notes

- `T5Wrapper` is no longer Torchscriptable
- Was not able to guarantee parity between our greedy search and HuggingFace's despite having nearly identical implementations
- As part of a discussion with @forresti wrt onboarding T5 V1.1 and T5-FLAN, might need to refactor to include Gated-GeLU activation. This will come in a follow-up PR but documenting here.


notebooks/hf_vs_tt_t5.ipynb
```diff
@@ -0,0 +1,83 @@
+{
```
joecummings, line 1: I know we verify completeness often w/ internal notebooks - I thought for those that show parity with HuggingFace or external libraries, we could put those notebooks in the actual repo. Seems like a better way to keep track rather than some Bento notebooks w/ scattered ownership.

torchtext/prototype/models/t5/model.py
```diff
@@ -1,14 +1,14 @@
...
 import torch
 import torch.nn as nn
 from torch import Tensor
 
-from .modules import T5Encoder, T5Decoder, T5LayerNorm
+from .modules import T5Encoder, T5Decoder
 
 
-@dataclass
+@dataclass(frozen=True)
```
joecummings, line 11: Freezing this as we probably don't want people to be able to overwrite configs and still try to use the model - much more likely to run into bugs that way.

torchtext/prototype/models/t5/model.py
```diff
@@ -136,12 +131,24 @@ def __init__(
             for p in self.parameters():
                 p.requires_grad = False
 
+    def prepare_inputs_for_generation(self, input_ids, encoder_outputs, **kwargs):
```
joecummings, line 134: @Nayef211 Any advice on what to do here? I know `**kwargs` are disallowed by TorchScript but I don't know of an elegant way to generalize the generation utils without allowing kwargs.

torchtext/prototype/models/t5/model.py
```diff
@@ -136,12 +131,24 @@ def __init__(
...
+    def get_encoder(self) -> T5Encoder:
+        return self.encoder
+
+    def get_decoder(self) -> T5Decoder:
+        return self.decoder
+
     def forward(
         self,
-        encoder_tokens: Tensor,
+        encoder_tokens: Optional[Tensor] = None,
```
joecummings, line 145: Don't have to include `encoder_tokens` if `encoder_outputs` are already provided.

torchtext/prototype/models/t5/modules.py
```diff
@@ -805,11 +805,13 @@ def __init__(
         layer_norm_eps: float = 1e-6,
         relative_attention_num_buckets: int = 32,
         relative_attention_max_distance: int = 128,
+        token_embeddings: Optional[nn.Module] = None,
```
joecummings, line 808: I don't think this should mess up the backprop for the entire model, but can you confirm @Nayef211 ?

torchtext/prototype/models/t5/modules.py
```diff
@@ -830,24 +832,44 @@ def __init__(
...
         )
         self.num_layers = num_layers
+        self.norm = T5LayerNorm(d_model)
+        self.dropout1 = nn.Dropout(dropout)
+        self.dropout2 = nn.Dropout(dropout)
 
     def forward(
         self,
-        tgt: Tensor,
+        tgt: Optional[Tensor] = None,
```
joecummings, line 841: Similar to the forward of the entire model, if the target is already embedded, no need to inlcude the raw tokenized tgt.

torchtext/prototype/generate.py
```diff
@@ -0,0 +1,110 @@
...
+        Returns:
+            Tensor of Tensors containing output sequences as ids.
+
+        `Note`: If one beam is provided or no beams are specified, the generation method will default to greedy search.
+        """
+        model_kwargs = {}
+
+        if self.is_encoder_decoder:
+            encoder = self.model.get_encoder()
+            model_kwargs["encoder_outputs"] = encoder(inputs)
```
joecummings, line 100: This should be the necessary args for the `forward` method of whatever model is being used in decoding.

torchtext/prototype/generate.py
```diff
@@ -0,0 +1,110 @@
+from typing import Optional
+
+import torch
+import torch.nn.functional as F
+from torch import nn
+
+
+class GenerationUtil:
```
joecummings, line 8: Does this whole class have to be torchscriptable, as well??

joecummings: That would make it extremely difficult to incorporate other models.

torchtext/prototype/generate.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    def __init__(self, model: nn.Module, is_encoder_decoder: bool = True) -> None:
+        self.model = model
+        self.is_encoder_decoder = is_encoder_decoder
+
+    def _prepare_decoder_ids_for_generation(
+        self, batch_size: int, device: Optional[torch.device] = None, **model_kwargs
+    ):
+        if model_kwargs is not None and "decoder_input_ids" in model_kwargs:
+            return model_kwargs.pop("decoder_input_ids")
```
rshraga, line 24: why pass around model_kwargs dict instead of just having an optional decoder_input_ids param?

torchtext/prototype/models/t5/model.py
joecummings: Freezing the config won't make it impossible to try a smaller model or modified architecture. It just means that once they instantiate the config and pass the config to the model, they won't be able to modify it. 

Example:
```
config = T5Config(encoder_only=True)
t5_model = T5Model(config=config)
t5_model.config.encoder_only = False  # Currently allowed; with freezing config, this would throw an error
```

Nayef211: Just to follow up here, in the example you just showed, would it affect the model behavior if users did end up changing the config after instantiating the model? IIUC the config is only used during model instantiation anyways. That being said I don't see any issues with freezing the config.

torchtext/prototype/models/t5/model.py
Nayef211: +1

torchtext/prototype/models/t5/model.py
```diff
@@ -136,12 +131,24 @@ def __init__(
...
         decoder_tokens: Optional[Tensor] = None,
         encoder_mask: Optional[Tensor] = None,
         decoder_mask: Optional[Tensor] = None,
+        encoder_padding_mask: Optional[Tensor] = None,
+        decoder_padding_mask: Optional[Tensor] = None,
+        encoder_outputs: Optional[Mapping[str, Tensor]] = None,
```
Nayef211, line 151: Please add these args to the docstring

torchtext/prototype/models/t5/modules.py
```diff
@@ -830,24 +832,44 @@ def __init__(
             ]
         )
         self.num_layers = num_layers
+        self.norm = T5LayerNorm(d_model)
+        self.dropout1 = nn.Dropout(dropout)
+        self.dropout2 = nn.Dropout(dropout)
```
Nayef211, line 837: Why did we decide to move this from the model to the encoder/decoder?

notebooks/hf_vs_tt_t5.ipynb
joecummings: Sure, but for a quick fix you can right click on expand dots on the top right of this file and select "View file" and it'll give you a notebook view.

joecummings:
@Nayef211 Can I get some 👀  on this again when you have a chance?

torchtext/prototype/models/t5/modules.py
joecummings: Keeps the entire encoder forward method self-contained.

torchtext/prototype/models/t5/model.py
joecummings: It wouldn't affect the model behavior, but it would throw an error saying "Config cannot be modified", which I think is what we want. It would be considered undefined behavior if someone e.g. instantiated a model without a decoder and then went back and changed the config to say that it did have a decoder.

joecummings:
@atalman @osalpekar Is this failing integration test related to Nova migration? The process seems to be killed with no helpful error and the integration tests pass on my local machine.

atalman:
> @atalman @osalpekar Is this failing integration test related to Nova migration? The process seems to be killed with no helpful error and the integration tests pass on my local machine.

@joecummings looks like integration tests are running out of memory, code 137: [3796110184](https://github.com/pytorch/text/actions/runs/3796110184/jobs/6455887023)

joecummings:
> > @atalman @osalpekar Is this failing integration test related to Nova migration? The process seems to be killed with no helpful error and the integration tests pass on my local machine.
> 
> @joecummings looks like integration tests are running out of memory, code 137: [3796110184](https://github.com/pytorch/text/actions/runs/3796110184/jobs/6455887023)

Silly follow-up question, but how would I go about allocating more memory for these integration tests?

torchtext/prototype/models/t5/wrapper.py
```diff
@@ -47,6 +48,8 @@ def __init__(
             strict (bool): Passed to :func: `torch.nn.Module.load_state_dict` method. (Default: `False`)
             dl_kwargs (dictionary of keyword arguments): Passed to :func:`torch.hub.load_state_dict_from_url`. (Default: `None`)
         """
+        warnings.warn("`T5Wrapper` is being deprecated. Please use new `GenerationUtils`.", category=DeprecationWarning)
```
pbontrager, line 51: GenerationUtils, if made in an nn.Module, could be treated as a generic wrapper for any LLM. This might be easier for the user but would break from the Huggingface design. It would allow for generation parameters to be saved with the model.

pbontrager (commented):
Great to have sampling as part of the library!

torchtext/prototype/generate.py
```diff
@@ -0,0 +1,141 @@
...
+            if eos_idx is not None:
+                unfinished_sequences = unfinished_sequences.mul((next_tokens != eos_idx).long())
+
+            # Stop iterating once all sequences are finished or exceed the max_len
+            if unfinished_sequences.max() == 0 or len(input_ids[0]) >= max_len:
+                break
+
+        return input_ids
+
+    def beam_search(self, input_ids: torch.Tensor, num_beams: int, max_len: Optional[int]) -> torch.Tensor:
```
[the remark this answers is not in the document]
joecummings: Idk about inheriting from `GenerationUtil`, but as a standalone class, this makes sense.

PR #2015, opened 2022-12-21
Rename variable to avoid conflict with PIP system variable PIP_PREFIX

weiwangmeta:
Fix failures in  https://github.com/pytorch/text/actions/runs/3742919848/jobs/6354468925
The failure was because the script defined "PIP_PREFIX = --pre " then pip install --pre torch torchtext command would install into the PIP_PREFIX directory, in this case '--pre ' directory. 

Solution: rename the PIP_PREFIX to TEXT_PIP_PREFIX to avoid this conflict.

### DIFF
--- a/.github/scripts/validate_binaries.sh
+++ b/.github/scripts/validate_binaries.sh
@@ -14,7 +14,7 @@ export PIP_DOWNLOAD_URL="https://download.pytorch.org/whl/cpu"
 export PIP_PREFIX=""
 
 if [[ ${CHANNEL} = 'nightly' ]]; then
-    export PIP_PREFIX="--pre "
+    export TEXT_PIP_PREFIX="--pre "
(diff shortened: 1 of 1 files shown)


seemethere (approved):

atalman (approved):
LGTM

.github/scripts/validate_binaries.sh
```diff
@@ -14,7 +14,7 @@ export PIP_DOWNLOAD_URL="https://download.pytorch.org/whl/cpu"
 export PIP_PREFIX=""
```
malfet, line 14: Didn't you need to modify this one as well?

weiwangmeta: Good catch!

weiwangmeta: Happen to be a no-op, but yeah, I should have modified it.

PR #2017, opened 2022-12-27
Beam search w/ Flashlight Text

joecummings:


torchtext/prototype/generate.py
```diff
@@ -0,0 +1,324 @@
...
+        
+        Returns:
+            Tensor of the generated sequences.
+        """
+        if self.is_encoder_decoder:
+            encoder_output_key = "last_hidden_state" if self.is_huggingface_model else "encoder_output"
+            encoder_output = model_kwargs["encoder_outputs"][encoder_output_key]
+
+        def update_func(emissions, N, T, prev_step_token_idxs, prev_step_model_states, timestep):
+            # `emissions` and `N` are unused in this current implementation
```
joecummings, line 140: Could change this around to take in emissions, but AFAIK there is no easy way to get the actual tensor from the data_ptr.

torchtext/prototype/generate.py
```diff
@@ -0,0 +1,324 @@
...
+            # Makeshift padding so that we can stack the tensors
+            while len(final_tokens) < max_len:
+                final_tokens += [0]
+            
+            # Convert from list to tensors
+            final_tokens_as_tensors = torch.Tensor(final_tokens).to(torch.long)
+
+            return final_tokens_as_tensors
+
+        if num_python_workers > 1:
```
joecummings, line 255: PyTorch has a MP module (essentially a clone of Python's MP); however, this relies on `pickle`, which means all of these functions would have to be at a global level. Very open to suggestions here.

Issue #2018, opened 2022-12-29
Increase memory for integration tests


joecummings:
PR: https://github.com/pytorch/text/pull/2011

See failing integration test

Nayef211:
@joecummings the above URL is broken for me. Can we update?

joecummings:
> @joecummings the above URL is broken for me. Can we update?

updated

PR #2024, opened 2023-01-18
[BUG] Fix memory leak with Regex C++ operator

### LINKED ISSUE #2020: [GPT2BPETokenizer] Out-of-memory issue during training when special token was added through "add_special_tokens()"
## 🐛 Bug

**Describe the bug** The training would OOM when using GPT2BPETokenizer with special tokens added through add_special_tokens()

**To Reproduce**

Can reproduce the issue with a minimal notebook example

With only one special tokens added, the memory footprint would increase quickly as the tokenization proceeds (see screenshot):
```
tokenizer = GPT2BPETokenizer(
    encoder_json_path="https://download.pytorch.org/models/text/gpt2_bpe_encoder.json",
    vocab_bpe_path="https://download.pytorch.org/models/text/gpt2_bpe_vocab.bpe",
)
additional_special_tokens = ["TOKEN_0"]	
tokenizer.add_special_tokens(
    special_tokens_dict={"additional_special_tokens": additional_special_tokens}
)]	
tokenizer.add_special_tokens(
    special_tokens_dict={"additional_special_tokens": additional_special_tokens}
)

i = 0
while True:
    tokenizer.forward(["hello", "world", "random", "stuff"])
    if i % 10000 == 0:
        process = psutil.Process(os.getpid())
        memory_in_megabytes = process.memory_info().rss / (1024 * 1024)
        print(f"with special token: {i}, {memory_in_megabytes} MB")  # in MB
    i += 1
```

image

Tried another run without adding any special token, the memory footprint is pretty stable
```
tokenizer = GPT2BPETokenizer(
    encoder_json_path="https://download.pytorch.org/models/text/gpt2_bpe_encoder.json",
    vocab_bpe_path="https://download.pytorch.org/models/text/gpt2_bpe_vocab.bpe",
)

i = 0
while True:
    tokenizer.forward(["hello", "world", "random", "stuff"])
    if i % 10000 == 0:
        process = psutil.Process(os.getpid())
        memory_in_megabytes = process.memory_info().rss / (1024 * 1024)
        print(f"with special token: {i}, {memory_in_megabytes} MB")  # in MB
    i += 1
```

image

**Expected behavior** The memory footprint should not increase too much when special tokens were added

Nayef211:
## Description
- We noticed that the memory footprint of `GPT2BPETokenizer` increases as tokenization occurs only when special tokens are added. After further investigation, the reason for this is due to a memory leak from the additional `Regex` object that is created whenever special tokens have been added to the tokenizer class.
https://github.com/pytorch/text/blob/8c01462995d6c34ff1b123ed6b81a7497e14b96b/torchtext/csrc/gpt2_bpe_tokenizer.cpp#L123
- Looking into the Regex implementation we dynamically create a heap allocated object (a pointer to a RE2 object) without ever freeing it. The fix is simply to create a destructor to free the dynamically allocated memory.
https://github.com/pytorch/text/blob/8c01462995d6c34ff1b123ed6b81a7497e14b96b/torchtext/csrc/regex.cpp#L6

## Test Plan
After the fix, the memory footprint doesn't change with additional forward calls to the tokenizer.
```
tokenizer = GPT2BPETokenizer(
    encoder_json_path="https://download.pytorch.org/models/text/gpt2_bpe_encoder.json",
    vocab_bpe_path="https://download.pytorch.org/models/text/gpt2_bpe_vocab.bpe",
)
additional_special_tokens = ["TOKEN_0"]	
tokenizer.add_special_tokens(
    special_tokens_dict={"additional_special_tokens": additional_special_tokens}
)]	
tokenizer.add_special_tokens(
    special_tokens_dict={"additional_special_tokens": additional_special_tokens}
)

i = 0
while True:
    tokenizer.forward(["hello", "world", "random", "stuff"])
    if i % 10000 == 0:
        process = psutil.Process(os.getpid())
        memory_in_megabytes = process.memory_info().rss / (1024 * 1024)
        print(f"with special token: {i}, {memory_in_megabytes} MB")  # in MB
    i += 1
```
image

Resolves #2020


rshraga:
looks great! Are the test failures expected / unrelated?

joecummings:
> looks great! Are the test failures expected / unrelated?

Yep, we need to increase our memory usage for the Integration Tests, see #2018

Nayef211:
> Looks like the commit history may be a little messy -> can you rebase @Nayef211 ?

Yeah I have a couple of commits that were made to the `main` branch of my forked repo rather than the original repo. Since we squash and merge changes, it shouldn't affect the commit history of the `main` branch.

joecummings (approved):
lgtm

torchtext/csrc/regex.cpp
```diff
@@ -6,6 +6,10 @@ Regex::Regex(const std::string& re_str) : re_str_(re_str) {
   compiled_pattern_ = new RE2(re_str_);
 }
 
+Regex::~Regex() {
+  delete compiled_pattern_;
+}
+
```
mthrok, line 12: Would this `Regex` object created a lot?
Is `RE* compiled_pattern_` attribute changed during the lifetime of `Regex`?

If the answer is No to both, it seems more semantic to use a smart pointer like `unique_ptr` and get rid of `new ~ delete` idiom entirely.
But if this `Regex` is expected to be instantiated so many times, 
then the overhead might not be negligible.

Nayef211: https://github.com/pytorch/text/blob/8c01462995d6c34ff1b123ed6b81a7497e14b96b/torchtext/csrc/gpt2_bpe_tokenizer.cpp#L123 
Yeah if you look at the use case here the `Regex` object is created every time `gpt2_bpe_pre_tokenizer()` function is called. And this function is called for every token in the list of tokens passed into forward. I don't believe it's possible to update the `RE* compiled_pattern_` attribute (or at least the RE2 library doesn't provide any getters or setters to update this attrbiute).

An alternative solution to prevent instantiation of `Regex` object so many times would be for us to create a C++ method that allow us to use the same `Regex` object but change the `compiled_pattern_` by creating a new `RE2` object (which would still require us to do `new` and `delete`).

Let me know your thoughts here and I can do this as a followup.

mthrok: then I think the current implementation should be fine.

mthrok (approved):

PR #2025, opened 2023-01-18
added gpu test

rshraga:
Added linux GPU unit testing

### DIFF
--- a/.github/workflows/test-linux-gpu.yml
+++ b/.github/workflows/test-linux-gpu.yml
@@ -1,0 +1,72 @@
+name: Unit-tests on Linux GPU
+
+on:
+  pull_request:
+  push:
+    branches:
+      - nightly
+      - main
+      - release/*
+  workflow_dispatch:
+
+env:
+  CHANNEL: "nightly"
+
+jobs:
+  tests:
+    strategy:
+      matrix:
+        python_version: ["3.8"]
+        cuda_arch_version: ["11.6"]
+      fail-fast: false
+    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
+    with:
+      runner: linux.g5.4xlarge.nvidia.gpu
+      repository: pytorch/text
+      gpu-arch-type: cuda
+      gpu-arch-version: ${{ matrix.cuda_arch_version }}
+      timeout: 120
+      script: |
+        # Mark Build Directory Safe
+        git config --global --add safe.directory /__w/text/text
+
+        # Set up Environment Variables
+        export PYTHON_VERSION="${{ matrix.python_version }}"
+        export VERSION="${{ matrix.cuda_arch_version }}"
+        export CUDATOOLKIT="pytorch-cuda=${VERSION}"
+
+        # Set CHANNEL
+        if [[ (${GITHUB_EVENT_NAME} = 'pull_request' && (${GITHUB_BASE_REF} = 'release'*)) || (${GITHUB_REF} = 'refs/heads/release'*) ]]; then
+          export CHANNEL=test
+        else
+          export CHANNEL=nightly
+        fi
+
+        # Create Conda Env
+        conda create -yp ci_env python="${PYTHON_VERSION}"
+        conda activate /work/ci_env
+        python3 -m pip --quiet install cmake>=3.18.0 ninja
+        conda env update --file ".circleci/unittest/linux/scripts/environment.yml" --prune
+
+        # TorchText-specific Setup
+        printf "* Downloading SpaCy English models\n"
+        python -m spacy download en_core_web_sm
+        printf "* Downloading SpaCy German models\n"
+        python -m spacy download de_core_news_sm
+
+        # Install PyTorch, Torchvision, and TorchData
+        set -ex
+        conda install \
+          --yes \
+          -c "pytorch-${CHANNEL}" \
+          -c nvidia "pytorch-${CHANNEL}"::pytorch[build="*${VERSION}*"] \
+          "${CUDATOOLKIT}"
+        printf "Installing torchdata nightly\n"
+        python3 -m pip install --pre torchdata --extra-index-url https://download.pytorch.org/whl/nightly/cpu
+        python3 setup.py develop
+        python3 -m pip install parameterized
+
+        # Run Tests
+        python3 -m torch.utils.collect_env
+        cd test
+        python3 -m pytest --cov=torchtext --junitxml=test-results/junit.xml -v --durations 20 torchtext_unittest


joecummings:
This is awesome, @rshraga, thank you! As part of this PR, can you add a very simple GPU test for T5? That will also make sure that we have the architecture to ensure GPU tests only run on the GPU and CPU tests only run on the CPU. I think you can borrow some logic from torchaudio on how to do this. And let's get @mthrok 's eyes on this, too.

.github/workflows/test-linux-gpu.yml
```diff
@@ -0,0 +1,72 @@
...
+        set -ex
+        conda install \
+          --yes \
+          -c "pytorch-${CHANNEL}" \
+          -c nvidia "pytorch-${CHANNEL}"::pytorch[build="*${VERSION}*"] \
+          "${CUDATOOLKIT}"
+        printf "Installing torchdata nightly\n"
+        python3 -m pip install --pre torchdata --extra-index-url https://download.pytorch.org/whl/nightly/cpu
+        python3 setup.py develop
+        python3 -m pip install parameterized
```
mthrok, line 67: Feel like `parameterized` can be incorporated into `.circleci/unittest/linux/scripts/environment.yml`

.github/workflows/test-linux-gpu.yml
```diff
@@ -0,0 +1,72 @@
...
+  pull_request:
+  push:
+    branches:
+      - nightly
+      - main
+      - release/*
+  workflow_dispatch:
+
+env:
+  CHANNEL: "nightly"
```
mthrok, line 13: Is there a advantage of defining the env var here?

mthrok (commented):
Overall it looks good.

rshraga:
@mthrok thank you for the comments! Will address these in the next version. When I check the result of unit tests in "Unit-tests on Linux GPU" workflow, I see that the new GPU ones did not even run (only the cpu ones are there) 
image

I checked and saw that the instance used for the job is g5.4xlarge which has a GPU so I am not sure why the GPU tests get skipped. Do you have any ideas how I could debug and get these tests to run?

mthrok:
> @mthrok thank you for the comments! Will address these in the next version. When I check the result of unit tests in "Unit-tests on Linux GPU" workflow, I see that the new GPU ones did not even run (only the cpu ones are there) image
> 
> I checked and saw that the instance used for the job is g5.4xlarge which has a GPU so I am not sure why the GPU tests get skipped. Do you have any ideas how I could debug and get these tests to run?

This is very peculiar. GPU tests are not even shown skipped. They are not recognized.

Couple of suggestions for debugging
1. Install torchtext with `python setup.py install` instead of `python setup.py develop`.
2. Run `python3 -m torch.utils.collect_env` after `cd test`.

Nayef211 (commented):
Since we're no longer just enabling GPU tests, can we update the PR description to go into details on the changes that were made and the reasoning behind them?

.github/workflows/test-linux-gpu.yml
rshraga: Just following the pattern from other workflows. If you suggest fixing / cleanup can we do so to all of them?

.github/workflows/test-linux-gpu.yml
```diff
@@ -0,0 +1,72 @@
...
+        fi
+
+        # Create Conda Env
+        conda create -yp ci_env python="${PYTHON_VERSION}"
+        conda activate /work/ci_env
+        python3 -m pip --quiet install cmake>=3.18.0 ninja
+        conda env update --file ".circleci/unittest/linux/scripts/environment.yml" --prune
+
+        # TorchText-specific Setup
+        printf "* Downloading SpaCy English models\n"
```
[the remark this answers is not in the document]
rshraga: I just copied most of this over from the other github workflow files. If improvements are needed, can we do that in separate PR?

test/torchtext_unittest/models/models_test_impl.py
```diff
@@ -2,30 +2,42 @@
 from unittest.mock import patch
 
 import torch
-import torchtext
 from torch.nn import functional as torch_F
 
-from ..common.torchtext_test_case import TorchtextTestCase
```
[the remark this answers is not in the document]
rshraga: Yes because pytest will run all tests for all super classes of `TorchtextTestCase` So `BaseTestModels` must not extend it directly

.github/workflows/test-linux-gpu.yml
```diff
@@ -0,0 +1,72 @@
...
+        python -m spacy download en_core_web_sm
+        printf "* Downloading SpaCy German models\n"
+        python -m spacy download de_core_news_sm
+
+        # Install PyTorch, Torchvision, and TorchData
```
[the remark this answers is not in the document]
rshraga: removed

test/torchtext_unittest/common/case_utils.py
```diff
@@ -37,11 +38,34 @@ def get_temp_path(self, *paths):
...
+
+    dtype = None
+    device = None
+
+    def setUp(self):
+        super().setUp()
+        torch.random.manual_seed(2434)
+
+    @property
+    def complex_dtype(self):
```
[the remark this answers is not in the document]
rshraga: removed, thanks!

test/torchtext_unittest/models/models_gpu_test.py
```diff
@@ -1,11 +1,11 @@
+import pytest
 import torch
 
-from ..common.case_utils import skipIfNoCuda
 from ..common.torchtext_test_case import TorchtextTestCase
 from .models_test_impl import BaseTestModels
 
 
-@skipIfNoCuda
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is not available")
```
Nayef211, line 8: Any reason we're no longer using a custom `skipIfNoCuda` decorator to do this check? Wouldn't it be cleaner for the future when we have many gpu specific unit tests?

Nayef211 (approved):

Nayef211:
Another thing to consider is whether we gain any benefits from running all tests on GPU enabled hosts. Since these tests are already being run on CPU hosts, do we want to save time and resources by only selecting GPU specific tests to be run on GPU hosts?

test/torchtext_unittest/models/models_gpu_test.py
mthrok: Is the issue with GPU test not executed resolved?
If not then this appears to be no1 suspect for me. 

The unit test suite in torchaudio (which torchtext’s suite is based off of), was originally (and still) written with Python unittest module. No pytest code is used. (pytest command is used to run the code though) The reason for that is 1. to reuse the helper functions from PyTorch and 2. fbcode was not supporting pytest. 

One of the biggest downsides is that ‘unittest’ does not natively support parameterization. So we used ‘parameterized’. ‘parameterized’ (and ‘pytest’ as well) employs many tricks to manipulate meta data and class/type objects, which we wouldn’t (and shouldn’t ) do in normal circumstances writing a library, so they often show strange behavior. 

So far unittest + parameterized for writing tests then using pytest for running them worked fine. 
But this one seems to be combining three to write tests, which we avoided in audio repo and could have some unexpected effects under the hood.

test/torchtext_unittest/models/models_test_impl.py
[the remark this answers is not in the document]
mthrok: That makes sense. Thanks for clarifying

.github/workflows/test-linux-gpu.yml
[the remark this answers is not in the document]
mthrok: Yeah it’s just a random thought. Not necessary to follow ip

test/torchtext_unittest/models/models_gpu_test.py
rshraga: Yea using skipIfNoCuda was not working properly, but using pytest.mark.skipif (as vision uses) works fine.

rshraga:
> Another thing to consider is whether we gain any benefits from running all tests on GPU enabled hosts. Since these tests are already being run on CPU hosts, do we want to save time and resources by only selecting GPU specific tests to be run on GPU hosts?

The pattern for the other libraries is like it is here: i.e cpu machines skip gpu tests, while gpu machines run everything. I think to do what you are describing would require keeping all cpu tests for all code and all gpu tests for all code in separate folders. This would make things look a bit messy imo

test/torchtext_unittest/models/models_gpu_test.py
mthrok: torchvision's test suites are based on pytest. So that's a bit different.

.github/workflows/test-linux-gpu.yml
```diff
@@ -0,0 +1,72 @@
...
+
+        # Set CHANNEL
+        if [[ (${GITHUB_EVENT_NAME} = 'pull_request' && (${GITHUB_BASE_REF} = 'release'*)) || (${GITHUB_REF} = 'refs/heads/release'*) ]]; then
+          export CHANNEL=test
+        else
+          export CHANNEL=nightly
+        fi
+
+        # Create Conda Env
+        conda create -yp ci_env python="${PYTHON_VERSION}"
```
mthrok, line 46: Can you make all the call to `conda` and `pip` quiet? The CI log is super long but not providing value much.

joecummings:
> Another thing to consider is whether we gain any benefits from running all tests on GPU enabled hosts. Since these tests are already being run on CPU hosts, do we want to save time and resources by only selecting GPU specific tests to be run on GPU hosts?

I would be in favor of this approach. Currently the only thing that needs to be run on GPU is modeling and our tests already take a long time.

mthrok:
> > Another thing to consider is whether we gain any benefits from running all tests on GPU enabled hosts. Since these tests are already being run on CPU hosts, do we want to save time and resources by only selecting GPU specific tests to be run on GPU hosts?
> 
> I would be in favor of this approach. Currently the only thing that needs to be run on GPU is modeling and our tests already take a long time.

I gave some thoughts and it's not much relevant here but if you have custom kernel implementations for both CPU/GPU, then it's worth running the tests for both implementations on CPU-only build and GPU-only build.

If the GPU code is written in Python only (which I believe is the case for torchtext), then running tests on GPU machine to check that devices are properly propagated should be enough, because it's PyTorch's premise that they implement the basic ops in CPU and GPU properly.

Nayef211:
> I gave some thoughts and it's not much relevant here but if you have custom kernel implementations for both CPU/GPU, then it's worth running the tests for both implementations on CPU-only build and GPU-only build.
> 
> If the GPU code is written in Python only (which I believe is the case for torchtext), then running tests on GPU machine to check that devices are properly propagated should be enough, because it's PyTorch's premise that they implement the basic ops in CPU and GPU properly.

@mthrok, just to make sure I understand, are you suggesting we do go forward with only running GPU specific tests on GPU hosts or to go with the status quo of running all tests on GPUs?

mthrok:
> > I gave some thoughts and it's not much relevant here but if you have custom kernel implementations for both CPU/GPU, then it's worth running the tests for both implementations on CPU-only build and GPU-only build.
> > If the GPU code is written in Python only (which I believe is the case for torchtext), then running tests on GPU machine to check that devices are properly propagated should be enough, because it's PyTorch's premise that they implement the basic ops in CPU and GPU properly.
> 
> @mthrok, just to make sure I understand, are you suggesting we do go forward with only running GPU specific tests on GPU hosts or to go with the status quo of running all tests on GPUs?

I think it's okay to run just GPU tests for Python-based implementations. But when you have custom ops with CPU and GPU implementations, I think it's good to test both CPU and GPU of those ops.

Nayef211:
LGTM! Feel free to merge :)

.github/workflows/test-linux-gpu.yml
```diff
@@ -0,0 +1,73 @@
...
+
+env:
+  CHANNEL: "nightly"
+
+jobs:
+  tests:
+    strategy:
+      matrix:
+        python_version: ["3.8"]
+        cuda_arch_version: ["11.6"]
```
mthrok, line 20: They are dropping the support for 11.6, so might be better to move to 11.7.

https://pytorch.slack.com/archives/C2077MFDL/p1675256463971369

Nayef211: Thanks for taking care of this in https://github.com/pytorch/text/pull/2040

PR #2027, opened 2023-01-18
flan t5 torchtext

forresti:
Summary:
[d2go] flan t5 torchtext

This enables support for FLAN-T5 in Torchtext.

So far, we have only enabled FLAN-T5 encoder-only models. If we need to have an encoder-decoder model, it would be straightforward to add support for that.

Differential Revision: D42159825


Nayef211:
Fyi @forresti, you don't need to export your diffs to Github. Once you're diff lands, it will get autosynced to the `fbsync` branch on Github and then do a manual cherry pick to bring the changes over to the `main` branch.

PR #2029, opened 2023-01-20
[CHERRY PICK] flan t5 torchtext (#2027)

Nayef211:
Summary:
Pull Request resolved: https://github.com/pytorch/text/pull/2027

[d2go] flan t5 torchtext

This enables support for FLAN-T5 in Torchtext.

So far, we have only enabled FLAN-T5 encoder-only models. If we need to have an encoder-decoder model, it would be straightforward to add support for that.

Reviewed By: joecummings

Differential Revision: D42159825

fbshipit-source-id: 6c2a4430df890131e18d3ebe40bba35ecc6b25b8


joecummings (approved):
lgtm

PR #2034, opened 2023-01-26
[TEST] Add torchdata version to smoke tests

Nayef211:
## Description
- Add torchdata import and version checks (ensuring the nightly is within 3 days of the current date) to smoke tests

### DIFF
--- a/test/smoke_tests/smoke_tests.py
+++ b/test/smoke_tests/smoke_tests.py
@@ -15,12 +15,9 @@ def validateTorchdataVersion():
         year=int(torchdata_date_str[:4]), month=int(torchdata_date_str[4:6]), day=int(torchdata_date_str[6:])
     )
 
-    print("[DEBUG1]", torchdata.__version__, torchdata_date_str)
-    print("[DEBUG2]", torchdata_datetime, datetime.now() - timedelta(days=3))
-    raise Exception(f"[DEBUG3] {torchdata_datetime}, {datetime.now() - timedelta(days=3)}")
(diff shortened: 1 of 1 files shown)


test/smoke_tests/smoke_tests.py
```diff
@@ -1,6 +1,24 @@
...
+    from datetime import datetime, timedelta
+
+    # validate torchdata version is within 3 days of current day
+    torchdata_date_str = torchdata.__version__[-8:]
+    torchdata_datetime = datetime(
+        year=int(torchdata_date_str[:4]), month=int(torchdata_date_str[4:6]), day=int(torchdata_date_str[6:])
+    )
+
+    assert torchdata_datetime >= datetime.now() - timedelta(
+        days=3
```
atalman, line 19: Can we make 3 a const ? something like for both torchdata and torchtext
```
NIGHTLY_ALLOWED_DELTA = 3

  
    date_t_str = re.findall("dev\d+", torchdata.__version__)
    date_t_delta = datetime.now() - datetime.strptime(date_t_str[0][3:], format_dt)
    if date_t_delta.days >= NIGHTLY_ALLOWED_DELTA:
        raise RuntimeError(
            f"the binaries are from {date_t_str} and are more than {NIGHTLY_ALLOWED_DELTA} days old!"
        )
```

Nayef211: Updated the PR with your suggestions

atalman (approved):
LGTM

Issue #2036, opened 2023-01-27
Drop Python 3.7 support


Nayef211:
Following PyTorch Core's recommendation to drop Python 3.7 support

See https://github.com/pytorch/pytorch/issues/80513

PR #2037, opened 2023-01-27
Drop support for Python 3.7

Nayef211:
According to https://github.com/pytorch/pytorch/issues/80513, PyTorch Core will drop support for 3.7 with the upcoming `torch==2.0.0` release. We should do the same.

Closes https://github.com/pytorch/text/issues/2036


atalman (approved):
LGTM

PR #2046, opened 2023-02-03
fix codeql workflow failure

rshraga:
The CodeQL workflow has started failing 3 weeks ago. I traced the issue to a change in how setuptools handles warnings in virtualenvs https://github.com/pypa/setuptools/issues/3772 Reverting to an older setuptools package fixes the issue, but I am not sure how to address properly

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -35,7 +35,7 @@ jobs:
           sudo ln -s /usr/bin/ninja /usr/bin/ninja-build
 
       - name: Build TorchText
-        run: python setup.py develop --user
+        run: python setup.py develop
 
       # If any code scanning alerts are found, they will be under Security -> CodeQL
       # Link: https://github.com/pytorch/text/security/code-scanning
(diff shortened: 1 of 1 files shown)


rshraga:
I traced this issue to the setuptools package: https://github.com/pypa/setuptools/issues/3772 Seems this warning has been thrown forever, but starting with setuptools 66 the warning starts throwing an actual exception. The back and forth on that decision is spicy and I am not sure it will get fixed since installation using the setup.py command is not recommended. Workaround is to revert setup tools to `65.7.0` but I am not sure this is wise long term.

joecummings:
If we just need to get this passing in time for the release, this seems fine as it only affects the setuptools used for running codeql. 

How important is codeql to our health signals? @osalpekar ?

joecummings (approved):

osalpekar:
> If we just need to get this passing in time for the release, this seems fine as it only affects the setuptools used for running codeql.
> 
> How important is codeql to our health signals? @osalpekar ?

@joecummings It's a good security check, but likely not the most critical since many other ecosystem libraries and pytorch core don't have this.

PR #2051, opened 2023-02-08
modify integration test workflow to use pytorch generic CI job

rshraga:
Changing this workflow to use the generic pytorch CI job like the unit tests https://github.com/pytorch/text/blob/main/.github/workflows/test-linux-cpu.yml


joecummings (approved):
Thx lgtm!

PR #2056, opened 2023-02-13
Install portalocker for testing

ejguan:
We are adding this [patch](https://github.com/pytorch/data/pull/1007) to minimize the hard dependency for TorchData. So, `portalocker` won't be automatically installed when installing with `torchdata`.

This PR would add `portalocker` to ensure CI green

### DIFF
--- a/.circleci/unittest/linux/scripts/install.sh
+++ b/.circleci/unittest/linux/scripts/install.sh
@@ -30,7 +30,8 @@ printf "* Installing PyTorch\n"
 )
 
 
-printf "Installing torchdata nightly\n"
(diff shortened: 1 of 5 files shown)


Nayef211 (approved):
LGTM thanks for adding this!

h-vetinari:
Shouldn't torch/text add an explicit dependency on portalocker now? Without it, I get:
```
= 979 failed, 204 passed, 12 skipped, 1 deselected, 6 warnings in 495.47s (0:08:15) =
```
that's >80% failed tests, and probably does not represent a functional torchtext?

PR #2064, opened 2023-02-17
Promote t5 and variants

joecummings:


torchtext/models/__init__.py
```diff
@@ -1 +1,2 @@
 from .roberta import *  # noqa: F401, F403
+from .t5 import *  # noqa: F401, F403
```
rshraga, line 2: t5 not added yet so this will cause errors

joecummings: It's added in this commit

joecummings: OH nvm u right

test/torchtext_unittest/models/gpu_tests/models_gpu_test.py
```diff
@@ -3,11 +3,12 @@
 import pytest
 import torch
 from torchtext_unittest.common.torchtext_test_case import TorchtextTestCase
-from torchtext_unittest.models.models_test_impl import BaseTestModels
+from torchtext_unittest.models.roberta_models_test_impl import RobertaBaseTestModels
```
rshraga, line 6: where is this added?

PR #2071, opened 2023-02-20
Fix validation workflow for test channel

atalman:
Fix validation workflow for test chan, fix conda installs for 3.11

### DIFF
--- a/.github/workflows/validate-nightly-binaries.yml
+++ b/.github/workflows/validate-nightly-binaries.yml
@@ -23,5 +23,5 @@ jobs:
   nightly:
     uses: ./.github/workflows/validate-binaries.yml
     with:
-      channel: test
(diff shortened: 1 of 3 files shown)


.github/workflows/validate-nightly-binaries.yml
```diff
@@ -23,5 +23,5 @@ jobs:
   nightly:
     uses: ./.github/workflows/validate-binaries.yml
     with:
-      channel: nightly
```
atalman, line 26: Please note will revert this change before pushing. This is to test, test chan

weiwangmeta (approved):
Please leave a note on why -c malfet is needed and merge when green. Looks like we have some conda failures.

osalpekar (approved):

PR #2076, opened 2023-02-24
Cuda 11.8 & 12 builds and Python 3.11

johnnynunez:
Changes:
Python base test: 3.10 (Release in march by Anaconda)
Future Python 3.11 support
Cuda 11.8 builds
Cuda 12.0 builds

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -17,7 +17,7 @@ jobs:
     uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
     with:
       job-name: Build doc
-      runner: amz2.linux.2xlarge
+      runner: linux.2xlarge
       repository: pytorch/text
       gpu-arch-type: cpu
       timeout: 120
--- a/.github/workflows/test-linux-cpu.yml
+++ b/.github/workflows/test-linux-cpu.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
     uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
     with:
-      runner: amz2.linux.12xlarge
+      runner: linux.12xlarge
       repository: pytorch/text
       script: |
         # Mark Build Directory Safe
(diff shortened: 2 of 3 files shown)


joecummings:
cc @atalman @osalpekar

osalpekar:
Looks like these are adding 3.11 support for the unittest jobs, but the jobs are failing due to version conflicts since some dependencies are not available for 3.11 in conda. We should perhaps wait for this until the deps are available and unittest jobs are able to run green reliably so the red signal doesn't affect torchtext dev's workflow - @joecummings lmk what you think about this.

johnnynunez:
> Looks like these are adding 3.11 support for the unittest jobs, but the jobs are failing due to version conflicts since some dependencies are not available for 3.11 in conda. We should perhaps wait for this until the deps are available and unittest jobs are able to run green reliably so the red signal doesn't affect torchtext dev's workflow - @joecummings lmk what you think about this.

it's similar problem with torchvision, but in theory anaconda will release the python binaries already in mid-March. So, I already prepare test...

joecummings:
> > Looks like these are adding 3.11 support for the unittest jobs, but the jobs are failing due to version conflicts since some dependencies are not available for 3.11 in conda. We should perhaps wait for this until the deps are available and unittest jobs are able to run green reliably so the red signal doesn't affect torchtext dev's workflow - @joecummings lmk what you think about this.
> 
> it's similar problem with torchvision, but in theory anaconda will release the python binaries already in mid-March. So, I already prepare test...

Got it. Thanks for the addition! I agree w/ @osalpekar that we'll wait until mid-March to merge, but I appreciate you being so on top of things here.

johnnynunez:
> > > Looks like these are adding 3.11 support for the unittest jobs, but the jobs are failing due to version conflicts since some dependencies are not available for 3.11 in conda. We should perhaps wait for this until the deps are available and unittest jobs are able to run green reliably so the red signal doesn't affect torchtext dev's workflow - @joecummings lmk what you think about this.
> > 
> > 
> > it's similar problem with torchvision, but in theory anaconda will release the python binaries already in mid-March. So, I already prepare test...
> 
> Got it. Thanks for the addition! I agree w/ @osalpekar that we'll wait until mid-March to merge, but I appreciate you being so on top of things here.

conda 3.11 binaries are avaible

johnnynunez:
Python 3.11 support added via https://github.com/conda/conda/pull/12256, this will be included in the May release of conda 23.5.0

PR #2091, opened 2023-03-01
Fix UTF8 decoding error in GPT2BPETokenizer `decode` method

Nayef211:
Summary:
- PyBind11 throws an error when decoding a C++ `std::string` which contains incomplete UTF8 byte sequences since the default UTF8 conversion uses `"strict"` error handling ([ref](https://docs.python.org/3/library/codecs.html#error-handlers))
- To resolve user issues (see [post](https://fb.workplace.com/groups/pytorchtext/permalink/899318121386487/)) we set the error handling to `"ignore"` which ignores the malformed data and continues decoding the string

Differential Revision: D43361716

fbshipit-source-id: 4ac488e4b4b894c8049728941a2ee36b1799258a


joecummings (approved):
lgtm

PR #2096, opened 2023-03-07
Include padding mask in generation

joecummings:
### Bug

Expect batched input to match single input e.g.

1. [seq1, ... seq_m] -> generate -> [output1, ...., output_m]
2. [seq1] -> generate -> [output1]

Before this would not create the same output1. The issue was that the src_key_padding_mask was not being propagated forward. 

### Fix

Create padding mask and add it to `model_kwargs` and pass it to the forward function.

### DIFF
--- a/test/integration_tests/test_generate.py
+++ b/test/integration_tests/test_generate.py
@@ -14,26 +14,25 @@ class TestGenerationUtil(TorchtextTestCase):
         self.model = t5_base.get_model()
         self.model.eval()
         # Examples taken from T5 Paper and Huggingface
-        self.inputs = self.transform(
-            [
-                "summarize: studies have shown that owning a dog is good for you",
-                "translate English to German: That is good.",
-                "cola sentence: The course is jumping well.",
-                "stsb sentence1: The rhino grazed on the grass. sentence2: A rhino is grazing in a field.",
-                "summarize: state authorities dispatched emergency crews tuesday to survey the damage after an onslaught of severe weather in mississippi...",
-            ]
-        )
+        self.inputs = [
+            "summarize: studies have shown that owning a dog is good for you",
+            "translate English to German: That is good.",
+            "cola sentence: The course is jumping well.",
+            "stsb sentence1: The rhino grazed on the grass. sentence2: A rhino is grazing in a field.",
+            "summarize: state authorities dispatched emergency crews tuesday to survey the damage after an onslaught of severe weather in mississippi...",
+        ]
+        self.transformed_inputs = self.transform(self.inputs)
         torch.manual_seed(0)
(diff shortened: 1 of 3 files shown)


torchtext/prototype/generate.py
```diff
@@ -48,7 +48,7 @@ def _prepare_decoder_ids_for_generation(
             return torch.ones((batch_size, 1), dtype=torch.long, device=device) * pad_idx
 
     def greedy_search(
-        self, input_ids: torch.Tensor, max_length: int, eos_idx: int, pad_idx: Optional[int] = None, **model_kwargs
+        self, input_ids: torch.Tensor, max_length: int, eos_idx: int, pad_idx: int, **model_kwargs
```
rshraga, line 51: Does changing pas_idx from Optional to required break any call sites?

torchtext/prototype/generate.py
```diff
@@ -77,18 +78,16 @@ def greedy_search(
...
 
             # For any finished sequences, padding idx should be the last token
-            if eos_idx is not None:
-                if pad_idx is not None:
-                    next_tokens = next_tokens * unfinished_sequences + pad_idx * (1 - unfinished_sequences)
+            next_tokens = next_tokens * unfinished_sequences + pad_idx * (1 - unfinished_sequences)
 
             # Append the next tokens to the previous tokens
-            input_ids = torch.cat([input_ids, next_tokens], dim=-1)
+            input_ids = torch.cat([input_ids, next_tokens[:, None]], dim=-1)
```
rshraga, line 87: what does the `[:, None]` do here?

torchtext/prototype/generate.py
joecummings: Nope. Only being called from the entry point method atm.

torchtext/prototype/generate.py
joecummings: Same thing as unsqueezing the last dim

test/integration_tests/test_generate.py
```diff
@@ -14,41 +14,47 @@ def setUp(self) -> None:
...
         ]
 
         self.assertEqual(generated_text, expected_generated_text)
 
+        inputs = self.transform([self.inputs[0]])
+
+        tokens_for_single_example = generation_model.generate(inputs, num_beams=1, max_length=30)
+        generated_text_for_single_example = self.transform.decode(tokens_for_single_example.tolist())
+
+        self.assertEqual(generated_text[0], generated_text_for_single_example[-1])
```
Nayef211, line 48: Why do we do `generated_text_for_single_example[-1]` instead of `generated_text_for_single_example[0]`?

PR #2120, opened 2023-03-17
Set stable to 0.15.0

Nayef211:
Set stable to 0.15.0


svekars (approved):

svekars:
Just adding a note here that the symlink needs to be created locally, not through the UI by running the following:
```
rm stable
ln -s 0.14.0 stable
```

PR #2121, opened 2023-03-18
Add padding direction

smsm8898:
give small flexibility to pad transformer


SM-Jang:
1. Can you refer to our style guide and make sure the code passes our linting?
Okay, I check flake8
and I follow change the name as begin:bool, default=False
To verify it, i modify the unittest and pass it all
2. Can you add a brief note on the motivation behind this addition?
When I work for modeling timeserise, I have to pad on begin or end.
The old PadTransform only give left pad... So I have to use torch.nn.functional.pad
That's why

PR #2128, opened 2023-03-27
Optionally ignore utf-8 decoding error for scripted C++ tokenizers.

shuminghu:
Summary: Binding and test to make sure we can use 'ignore' option for utf-8 decoding added to pytorch in D43970697( https://github.com/pytorch/pytorch/pull/97282).

Reviewed By: Nayef211

Differential Revision: D44315169


Nayef211:
Don't need to export diffs to GH

Issue #2127, opened 2023-03-24
Add _shift_right helper function to T5 model


joecummings:
See https://github.com/huggingface/transformers/blob/68287689f2f0d8b7063c400230b3766987abf18d/src/transformers/models/t5/modeling_t5.py#L835

This is needed for training.

PR #2131, opened 2023-03-28
Add shift-right method to T5 model

yohann-benchetrit:
We add a `_shift_right` method to the T5 Model. This is used for training (see [Issue 2127](https://github.com/pytorch/text/issues/2127)).

It shifts an input batch of sequences of token-indices, one index to the right.
Their first token is replaced with the start-of-sequence token, which is the padding-token index for T5 models


joecummings:
@yohann-benchetrit  Can you add a unittest for this?

Nayef211 (commented):
This looks fine to me. @joecummings do you want to take a look and stamp if everything looks okay?

joecummings:
@yohann-benchetrit  Looks like there's an error in how the config is being called within _shift_right

PR #2136, opened 2023-03-31
Allow overwriting of local SPM model path

joecummings:
Fixes issue in which some download managers (Manifold) do not successfully download the SPM model; however, because the file has been created, the model does not attempt another redownload and instead tries to load an empty file.

Bug was found and fix authored by @debowin

### DIFF
--- a/torchtext/models/t5/t5_transform.py
+++ b/torchtext/models/t5/t5_transform.py
@@ -39,9 +39,16 @@ class T5Transform(nn.Module):
         >>> transform(["hello world", "attention is all you need!"])
     """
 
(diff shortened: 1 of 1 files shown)


joecummings:
@osalpekar These Macos build errors don't seem to be coming from changes to this PR - can you take a look?

osalpekar:
@DanilBaibak @atalman Any ideas what these MacOS issues are (both in Unittests and Build workflows)? Anything changed in the configuration for those instances? Also with MacOS x86 support being deprecated, do we want to start phasing these jobs out from domains as well?

Nayef211 (approved):

PR #2137, opened 2023-04-03
Add top-p and top-k sampling to GenerationUtils

yohann-benchetrit:
1. Add the following token-decoding helpers to `GenerationUtils.generate`:
- `top-p`
- `top-k`
- `remove_invalid_values`
- `temperature` (see [Issue 2138](https://github.com/pytorch/text/issues/2138))

2. Add corresponding unit tests and docstrings.

### DIFF
--- a/torchtext/prototype/generate.py
+++ b/torchtext/prototype/generate.py
@@ -48,7 +48,12 @@ class GenerationUtils:
             return torch.ones((batch_size, 1), dtype=torch.long, device=device) * pad_idx
 
     def greedy_search(
-        self, input_ids: torch.Tensor, max_length: int, eos_idx: int, pad_idx: int, **model_kwargs
+        self,
+        input_ids: torch.Tensor,
+        max_length: int,
+        eos_idx: int,
+        pad_idx: int,
+        **model_kwargs,
     ) -> torch.Tensor:
         """Greedy search decoding for text generation. Takes the most likely next token every time.
 
@@ -117,7 +122,6 @@ class GenerationUtils:
             max_length (int): Max length to generate responses.
             pad_idx (int): Padding index. Defaults to 0.
             eos_idx (int): End of sequence index. Defaults to 1.
-
         Returns:
             Tensor of Tensors containing output sequences as ids.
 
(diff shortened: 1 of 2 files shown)


joecummings:
High level, I notice you implemented the sampling as part of the `greedy_search` function. Technically, with sampling this name is incorrect. I see a couple of ways of going about this. 

1. Rename the function to a more generic `sample` method and remove the `do_sample` parameter to avoid confusion. Then, if `top_k` or `top_p` was specified, a true sampling would be implemented. Otherwise, it would essentially be a greedy search. (in this case, as well, top_k should probably default to 1 instead of 0.) This would be the best way to reuse code IMO. 

2. Follow HF and have separate methods for sampling and greedy. From an adoption perspective, this might be beneficial as users of torchtext tend to also have experience with HF; however, there would certainly be some duplicated code.

Thoughts? @yohann-benchetrit

torchtext/prototype/generate.py
```diff
@@ -75,10 +88,21 @@ def greedy_search(
...
             output_key = "logits" if self.is_huggingface_model else "decoder_output"
             decoder_output = outputs[output_key]
-
-            # Calculate probabilities and take the most likely next token
-            probs = F.log_softmax(decoder_output[:, -1], dim=-1)
-            next_tokens = torch.argmax(probs, dim=-1)
+            log_probs = F.log_softmax(decoder_output[:, -1], dim=-1)
+
+            if do_sample:
+                probs = log_probs.softmax(dim=-1)
```
joecummings, line 94: Taking double softmax here? Probs are already softmax'd on L91

yohann-benchetrit: Thanks for spotting this ! It should be a `.exp` to retrieve the original probabilities indeed.

yohann-benchetrit:
> Thoughts? @yohann-benchetrit

Agreed, thanks @joecummings  and @Nayef211 for your comments ! 

So with this additional information my understanding is:
- `generate` is the "maximum generality" decoding-interface
- `greedy_search` and `beam_search` are (or will be) implemented on their own as public subroutines of `generate` but they are not necessarily intended to be directly called by the end-user.

My thoughts on this:
1. As @Nayef211 suggested and closer to your proposition 2 (HF), I would tend to also leave `greedy_search` sampling-free and delegate 'sampling with a single beam' (i.e what I implemented here) to a `beam_sample` method that will build on  `beam_search` once it will be in `main` (up to efficiency concerns, it could also make sense to have a single `beam_search` method, that can do sampling). 
However, since `beam_search` is not currently in `main`, does this mean that this PR should solely restrict to adding `_get_top_{k,p}_restriction` methods (with corresponding tests) ?

2. As a side note, although HF also recommends going solely for using `generate`, I like the idea of having the fundamentals such as `greedy_search` and `beam_search` on the top of the API, for "no-extra-thought" usage.

Nayef211:
> 1. As @Nayef211 suggested and closer to your proposition 2 (HF), I would tend to also leave `greedy_search` sampling-free and delegate 'sampling with a single beam' (i.e what I implemented here) to a `beam_sample` method that will build on  `beam_search` once it will be in `main`.
>    However, since `beam_search` is not currently in `main`, does this mean that this PR should solely restrict to adding `_get_top_{k,p}_restriction` methods (with corresponding tests) ?

I like this idea more and unless there are customers that are asking for sampling to be implemented in `greedy_search`, I would hold off on adding these params to the method and instead just add the helper methods like you mentioned. However if customers are asking for this, I don't see any harm in adding the functionality to `greedy_search` and removing it later down the line. Ultimately, I'll leave the decision to @joecummings.

yohann-benchetrit:
@joecummings @Nayef211 

Thanks again for your comments. I addressed them as follows:
- I kept only the helper methods `_get_top_{k,p}_restriction` and deleted their usage in decoding (except for unit tests).
- I added an `_apply_temperature` method (and a unit test), as Joe was suggesting, addressing issue [#2138](https://github.com/pytorch/text/issues/2138) .

joecummings (approved):
Looks good. Thanks @yohann-benchetrit !

PR #2142, opened 2023-04-05
Add file_lock when loading SentencePieceModel

joecummings:
Summary:
SentencePieceModel loading can cause a RuntimeError when concurrent threads try to load/download it (e.g. when using T5 tokenizer in a DDP model training). Adding a file lock ensures the first thread to acquire the lock will actually download the model and the other ones will just use the existing path (which will not contain a partially downloaded model).

This diff was inspired by D42686913 and reverts D44566854 behavior (there is no need to overwrite anymore).

It should also disable unit test flakiness such as https://www.internalfb.com/intern/test/281475067136403?ref_report_id=0 and solve https://fb.workplace.com/groups/pytorchtext/permalink/920234369294862/.

Reviewed By: joecummings

Differential Revision: D44604474

fbshipit-source-id: 1c117fb6d1e72cce31cbf30bf72d513ad535b0d4

### DIFF
--- a/torchtext/utils.py
+++ b/torchtext/utils.py
@@ -10,6 +10,7 @@ from torchtext import _CACHE_DIR
 
 from ._download_hooks import _DATASET_DOWNLOAD_MANAGER
 
+from iopath.common.file_io import file_lock
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


torchtext/utils.py
```diff
@@ -6,11 +6,11 @@
 import zipfile
 
 import torch
+from iopath.common.file_io import file_lock
```
mthrok, line 9: This is the third time that iopath is being introduced as a dependency.
Did they become available on all platforms that torchtext is distributed?

* 1st
    https://github.com/pytorch/text/pull/1319
    https://github.com/pytorch/text/pull/1381
* 2nd
    https://github.com/pytorch/text/pull/1905

joecummings: Nope - will need to rework this to not depend on `iopath`.

joecummings:
This seems to be a suitable stdlib fix: https://gist.github.com/jirihnidek/430d45c54311661b47fb45a3a7846537

mthrok:
> https://gist.github.com/jirihnidek/430d45c54311661b47fb45a3a7846537

IIRC `fcntl`-based approach is not portable to Windows, so need an alternative for that.

joecummings:
Looking into using `portalocker` as a soln.

PR #2148, opened 2023-04-13
[NovaX] Windows CPU Unittests

osalpekar:
Adds Windows CPU Unittests as a GHA Nova Workflow. Supports Python 3.8-3.10 like the current CircleCI job. If these work well for a week or so, we can deprecate the old CircleCI jobs.


atalman (approved):
LGTM, looks good!

PR #2158, opened 2023-04-21
File lock Linux/Unix using Python stdlib

Nayef211:
Summary:
Builds on D44604474 but uses Python stdlib so as to not as a new dependency on `iopath`.

Limitations:
- Does not work on Windows. Currently looking for a better solution; however, this should unblock bugs internally

Reviewed By: gl3lan

Differential Revision: D45048667

fbshipit-source-id: ef2f00d9c0ba99ca8b400206b8d84bad6c28f5c3


abhinavarora (approved):

Nayef211:
Closing in favor of https://github.com/pytorch/text

PR #2104, opened 2023-03-08
Add decorator to skip tests and log a warning if URL is unaccessible

joecummings:
Closes #2103

PR #2163, opened 2023-04-26
Add decorator to skip tests and log a warning if URL is unaccessible …

Nayef211:
…(#2104)

* Add decorator to skip tests and log a warning if URL is unaccessible

* More descriptive error message


atalman (approved):
LGTM

PR #2170, opened 2023-05-16
Build docs in GHA

atalman:


.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,102 @@
...
+      - v[0-9]+.[0-9]+.[0-9]
+      - v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+
+  workflow_dispatch:
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        python_version: ["3.8"]
+      fail-fast: false
```
malfet, line 20: Why do you need a matrix here at all?
```suggestion
```

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,102 @@
...
+      gpu-arch-type: cpu
+      timeout: 120
+      upload-artifact: docs
+
+      script: |
+        # Mark Build Directory Safe
+        git config --global --add safe.directory /__w/text/text
+
+        # Set up Environment Variables
+        export PYTHON_VERSION="${{ matrix.python_version }}"
```
malfet, line 35:
```suggestion
        export PYTHON_VERSION="3.8"
```

malfet (approved):

.github/workflows/build-docs.yml
atalman: Done!

.github/workflows/build-docs.yml
atalman: Done!

PR #2179, opened 2023-06-13
Add TorchFix linter

kit1980:
Add TorchFix linter and fix deprecated usages of `assert_allclose` to make it pass.

Also move the comments in the flake8 config to separate lines, as more modern flake8 versions do not support inline comments for any of the keys - https://flake8.pycqa.org/en/latest/user/configuration.html

The bandit failure and M1 failures are pre-existing.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,10 +1,13 @@
 [flake8]
 ignore =
     E401,E402,E501,E722,W503,W504,F821,B006,B007,B008,B009,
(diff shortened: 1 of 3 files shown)


huydhn (approved):

kit1980:
The tests failing are also per-existing, see the same errors on a dummy change https://github.com/pytorch/text/pull/2180

So merging this PR.

PR #2212, opened 2023-10-16
Drop RoCM Support for Text Linux Wheels

osalpekar:


seemethere (approved):
LGTM will want a second opinion as well from @kartikayk

abhinavarora (approved):
LGTM. Do you think this change will resolve the version mismatch issue as seen in installation of torchmultimodal? Also, could you list down any potential side-effects of this change?

PR #2225, opened 2024-01-25
Add 0.17.0 docs

atalman:
Same as: https://github.com/pytorch/text/pull/2207

Docs taken from: https://github.com/pytorch/text/actions/runs/7534160197/job/20508030622

cc @svekars @NicolasHug


NicolasHug:
Thanks for the PR Andrey. At a high-level this looks good. I'm still seeing instances of "Torchtext 0.17.0.dev20240115" though, this would need to be `sed`ed into just `Torchtext 0.17.0`.


In any case we should not merge this before the release of course :)

NicolasHug (approved):
I took a brief look at the rendered docs, this should be good! Thank you @atalman . Approving now, but of course we should wait for the release before merging.

PR #2229, opened 2024-02-07
Update README.rst

joecummings:
Make it clear that this repository is not actively maintained by PyTorch anymore.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,6 +12,8 @@
 torchtext
 +++++++++
 
+CAUTION: This repository is no longer actively maintained by PyTorch. Use at your own risk. 
+
 This repository consists of:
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -12,6 +12,8 @@
 torchtext
 +++++++++
 
+CAUTION: This repository is no longer actively maintained by PyTorch. Use at your own risk.
```
NicolasHug, line 15: Nit: maybe we want to communicate a slightly less scary message, like the one that was written on the [0.16 release notes](https://github.com/pytorch/text/releases/tag/v0.16.0) ?  But LGTM anyway, thanks @joecummings 

> As of September 2023 we have paused active development of TorchText because our focus has shifted away from building out this library offering. We will continue to release new versions but do not anticipate any new feature development as we figure out future investments in this space.

NicolasHug (approved):

PR #2240, opened 2024-03-21
Add version to validate from torchtext validate binaries call

PaliC:
Adds option to validate against a specific version of pytorch


atalman:
We would need a PR to:
```
pytorch/test-infra/.github/workflows/validate-domain-library.yml
```

atalman:
Also need to change /.github/scripts/validate_binaries.sh to install required version of text

atalman (approved):

